PR #1, opened 2024-12-28
feat(multi-chain): create application deployment setup

tk-o:
This PR enables running parallel application deployments. Each of the deployments is indexing a specific chain independently, while sharing the infrastructure deployments (db, cache) with other application deployments.

This change aims to enable running multiple indexing engines based of the ENS SDK (defining ponder config and schema per ENS deployment).
image

### Docker services

Testing application and infrastructure deployments locally is possible with Docker Compose setup. It includes all required dependencies and allows using a GQL API the same way for each chain-specific deployment:
image

To run the services, use your terminal to go to the root project dir and execute:
```
make start
```

To stop the services:
```
make stop
```


shrugs (commented):
hey tko! this is a great implementation of dockerizing the project, but i fear it may be a bit early, let me know what you think:

i was expecting this to look basically like the current ponder project but with some configuration to switch between which deployment we're targeting (mainnet, base, linea) rather than a multi-indexer docker compose setup. if someone would like to index all three deployments they could run/deploy multiple ponder instances (optionally targeting different schemas on the same database). i get that the compose makes running them all at once perhaps easier, but i'm not certain the added complexity is worth the squeeze right now.

it's true that in v2 we'll likely want to run some additional services (redis, namehash label healing service, user-facing api service, etc) but i don't expect we'll need that until we start working on the v2—for now i don't think we need more than what ponder gives us at the moment. 

i think the priority should be modifying the ponder project to support the base and linea deployments, effectively writing `ponder-plugin-ens` such that one can run something like `DATABASE_SCHEMA=base ENS_DEPLOYMENT=base ponder run start` (as well as the linea and mainnet versions). these should have identical schemas but will have different values for the ponder config as well as different abis and extra logic for the handlers like you mentioned. writing this in a type-safe and readable way will be a good challenge, i expect.

once we begin writing a true multichain indexer, merging all known ens deployments into a single ponder index, we'll likely want to come back to this PR in order to make deployment of the project easier, since it'll likely contain multiple services at that point. 

what do you think?

tk-o:
It makes sense to avoid the tooling overhead in the v1. I'm parking this PR for later.

PR #3, opened 2024-12-31
feat(ponder-ens): introduce plugin architecture

tk-o:
This change allows defining simple ponder plugins per supported subnames. Only one plugin can be active during indexing, and to select it, please set the `ACTIVE_PLUGIN` env var to the name of one of the available plugins (see `src/plugins` directory). For example:
```
ACTIVE_PLUGIN=eth
ACTIVE_PLUGIN=base.eth
```

Also, please [make sure to know](https://ponder.sh/docs/getting-started/database#database-schema) how to use the `DATABASE_SCHEMA` env var.

### Examples

Running indexer with the requested subname plugin for ponder:

```
ACTIVE_PLUGIN=base.eth DATABASE_SCHEMA=my_base_eth pnpm start 
```


lightwalker-eth (commented):
@tk-o Hey thanks for all the efforts on this 🙌 Excited to see the "plugins" start to come to life! Will defer to @shrugs on more detailed review for the proposed plugin architecture.

A more general feedback from my side. For each PR we make, let's please also consider what documentation updates might be relevant and include them in the PR. Appreciate this takes time, but good documentation for other devs in the ENS community will be key to our success.

For example, ideas that would be great to include as documentation in the README.md with this PR might include:

1. Overall introduction and framing for what we're working to do here. For example: enable any permutation of chain-specific ENS indexers to be dynamically activated via environment variables.
2. Communicate known oversimplifications in our current approach and identify that our roadmap will remove these oversimplifications in time. For example:
   1. At the moment we're making oversimplifications where indexing "mainnet" represents indexing the legacy ENS Subgraph, indexing "base" represents only indexing subnames of "base.eth" on Base, and indexing "linea" represents only indexing subnames of "linea.eth" on Linea. Etc.
   2. At the moment we're making oversimplifications where we assume that all the delegation relationships of subname issuance from an L1 resolver (using ENSIP-10 and EIP-3668) to an offchain gateway server that reads state from a L2 subname registry is immutable. In other words, we're currently assuming that the current L2 subname registry for subnames of "base.eth" on Base is guaranteed to continue serving in that role.
4. Perhaps an .env.example file that includes comments documenting the environment variables we use and different example values that might be set. Am I correct to assume that there's also some ponder-related environment variables that we should be documenting in an .env.example file too? For example: RPC endpoints for different chains? Let's write our documentation under the assumption that the reader has never used Ponder before and should not need to familiarize themselves with Ponder if they just want to get started and run their own instance.
5. A "getting started" section that explains step-by-step how to go from cloning the repo to having a running instance of our indexer.
6. Any other ideas you suggest will be useful :+1: 😄

src/ponder-ens-plugins/ethereum/ponder.config.ts
```diff
@@ -0,0 +1,100 @@
...
+      startBlock: Math.min(16925618, END_BLOCK),
+      endBlock: END_BLOCK,
+    },
+    NameWrapper: {
+      network: "mainnet",
+      abi: NameWrapper,
+      address: NAME_WRAPPER_ADDRESS,
+      startBlock: Math.min(16925608, END_BLOCK),
+      endBlock: END_BLOCK,
+    },
```
tk-o, line 98:
```suggestion
    Ethereum_RegistryOld: {
      network: "mainnet",
      abi: Registry,
      address: REGISTRY_OLD_ADDRESS,
      startBlock: 3327417,
      endBlock: END_BLOCK,
    },
    Ethereum_Registry: {
      network: "mainnet",
      abi: Registry,
      address: REGISTRY_ADDRESS,
      startBlock: 9380380,
      endBlock: END_BLOCK,
    },
    Ethereum_OldRegistryResolvers: {
      network: "mainnet",
      abi: RESOLVER_ABI,
      address: factory({
        address: REGISTRY_OLD_ADDRESS,
        event: getAbiItem({ abi: Registry, name: "NewResolver" }),
        parameter: "resolver",
      }),
      startBlock: 9380380,
      endBlock: END_BLOCK,
    },
    Ethereum_Resolver: {
      network: "mainnet",
      abi: RESOLVER_ABI,
      address: factory({
        address: REGISTRY_ADDRESS,
        event: getAbiItem({ abi: Registry, name: "NewResolver" }),
        parameter: "resolver",
      }),
      startBlock: 9380380,
      endBlock: END_BLOCK,
    },
    Ethereum_BaseRegistrar: {
      network: "mainnet",
      abi: BaseRegistrar,
      address: BASE_REGISTRAR_ADDRESS,
      startBlock: 9380410,
      endBlock: END_BLOCK,
    },
    Ethereum_EthRegistrarControllerOld: {
      network: "mainnet",
      abi: EthRegistrarControllerOld,
      address: ETH_REGISTRAR_CONTROLLER_OLD_ADDRESS,
      startBlock: 9380471,
      endBlock: END_BLOCK,
    },
    Ethereum_EthRegistrarController: {
      network: "mainnet",
      abi: EthRegistrarController,
      address: ETH_REGISTRAR_CONTROLLER_ADDRESS,
      startBlock: Math.min(16925618, END_BLOCK),
      endBlock: END_BLOCK,
    },
    Ethereum_NameWrapper: {
      network: "mainnet",
      abi: NameWrapper,
      address: NAME_WRAPPER_ADDRESS,
      startBlock: Math.min(16925608, END_BLOCK),
      endBlock: END_BLOCK,
    },
```

tk-o: If we don't introduce any namespace, we'll have issues around types if the same contract name would rely on different set of arguments. I'd prefer merging contract definitions across chains in the v2 release.

src/ponder-ens-plugins/main.ts
```diff
@@ -1,11 +1,12 @@
...
 /**
  * Main entry point for the Ponder ENS plugins.
  * It tries to activate all the enlisted plugins.
  */
 function main() {
-  const plugins = [ethereumPlugin];
+  const plugins = [basePlugin, ethereumPlugin];
 
   for (const plugin of plugins) {
     if (plugin.canActivate) {
```
lightwalker-eth, line 12: Small feedback: maybe we should rename `canActivate` to `shouldActivate`.

src/ponder-ens-plugins/base/ponder.config.ts
```diff
@@ -0,0 +1,91 @@
...
+        parameter: "resolver",
+      }),
+      startBlock: L2_RESOLVER_START_BLOCK,
+      endBlock: END_BLOCK,
+    },
+    [ns("BaseRegistrar")]: {
+      network: "base",
+      abi: BaseRegistrar,
+      address: BASE_REGISTRAR_ADDRESS,
+      startBlock: 9380410,
```
lightwalker-eth: Should we define a named constant for this?

src/ponder-ens-plugins/base/ponder.config.ts
```diff
@@ -0,0 +1,91 @@
...
+const L2_RESOLVER_ADDRESS = '0xC6d566A56A1aFf6508b41f6c90ff131615583BCD';
+const L2_RESOLVER_START_BLOCK = 17575714;
+
+export const config = Object.freeze({
+  networks: {
+    base: {
+      chainId: base.id,
+      transport: http(process.env[`PONDER_RPC_URL_${base.id}`]),
+    },
+    mainnet: {
```
lightwalker-eth, line 56: Interesting, so the base indexer also makes use of mainnet RPC calls?

src/ponder-ens-plugins/chain.ts
```diff
@@ -1,5 +1,21 @@
 export function isChainIndexingActive(chainId: number) {
-  return process.env.CHAINS_TO_INDEX?.split(",")
-    .map((maybeChainId) => parseInt(maybeChainId, 10))
-    .includes(chainId);
+  if (!process.env.ENS_DEPLOYMENT_CHAIN_ID) {
```
lightwalker-eth, line 2: Perhaps we should rename the `ENS_DEPLOYMENT_CHAIN_ID` variable so it is more self documenting?

For example, as I understand this environment variable is currently used to identify what chain id should have its indexing activated on. In my mind this is a different idea than an "ENS deployment chain id" because currently there's only an ENS deployment to mainnet and sepolia.

It can be helpful to separate the idea of ENS from any particular TLD such as ".eth" or any particular subname. ENS is rooted in overall ENS root which lives only on mainnet (and a testnet deployment on sepolia).

One quick alternative idea is to call it `INDEXED_CHAIN_ID` -- open to other ideas too 😄

src/ponder-ens-plugins/chain.ts
```diff
@@ -1,5 +1,21 @@
...
+    console.warn("ENS_DEPLOYMENT_CHAIN_ID is not set");
+    return false;
+  }
+
+  return parseInt(process.env.ENS_DEPLOYMENT_CHAIN_ID, 10) === chainId;
 }
+
+export function createNs<ChainId extends number>(chainId: ChainId) {
```
lightwalker-eth, line 10: My understanding is that this is a utility function to help with namespacing?

Suggest we define our namespacing based on ENS names, not on chain ids. It's true that we are applying a simplifying assumption right now that mainnet => direct subnames of .eth and base => direct subnames of base.eth. I'm happy with this simplifying assumption, but prefer where possible we try to avoid leaning into this too hard.

Would it add much scope to make it so that each of our plugins included not only a defined chain id, but also a defined ENS name of direct subnames the plugin takes responsibility for indexing? For example, our mainnet plugin would set this config to "eth" and our base plugin would set this config to "base.eth".

Please let me know if any concerns or questions :+1:

src/ponder-ens-plugins/base/README.md
```diff
@@ -0,0 +1,9 @@
+# Base ENS plugin for Ponder indexer
```
lightwalker-eth, line 1: Suggest we rename all references to this plugin from "base" to "base.eth".

Suggest we have a mindset that the fundamental way we refer to plugins is through ENS names and not chains.

The following is an open task for after we finish our V1. Suggest we only follow up on this later. Please also add a task to the root README as a pending action item for V2+:

* Separate the "ethereum" plugin into two separate items:
   * "Root" ENS contracts, which are not specific to the "eth" TLD. The ENS name associated with these should be "" (empty string, representing the ENS root).
   * "eth" TLD contracts, which should be separated into a plugin for the ENS name "eth".

src/ponder-ens-plugins/base/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+This plugin contains configuration required to run blockchain indexing with [the Ponder app](https://ponder.sh/). It includes relevant ABI files, contract addresses and the block numbers those contracts were deployed at on a selected network.
+
+## Architecture
+
+The Base implementation for ENS protocol has custom naming convention, which was described here:
```
lightwalker-eth, line 7: Very happy to see this documentation here! Here's some suggestions for details in the language we use:

1. We shouldn't refer to this as "The Base implementation for ENS protocol". This plugin is for indexing subnames of "base.eth". It just so happens that the related contracts are on Base today, but in theory that might change in the future. We should think of plugins based on ENS names, not on chains.
2. We shouldn't use the language "... has custom naming convention". In the ENS community, "naming convention" might mean different things to different people. Suggest instead we use more specific ENS terminology. For example: "The `base.eth` subname registrar / registry is similar to the `eth` subname registrar with a few customizations as described here:"

src/lib/ens-helpers.ts
```diff
@@ -3,6 +3,7 @@ import { type Hex, concat, keccak256, namehash, toHex } from "viem";
 // TODO: pull from ens utils lib or something
 export const NAMEHASH_ZERO = namehash("");
 export const NAMEHASH_ETH = namehash("eth");
+export const NAMEHASH_BASE_ETH = namehash("base.eth");
```
lightwalker-eth, line 6: In ENS terminology, "namehash" is a function and the value returned by that function is referred to as a "node". Therefore, suggest renaming each of these constants to be prefixed with "NODE_" instead of "NAMEHASH_"

src/lib/ens-helpers.ts
```diff
@@ -3,6 +3,7 @@ import { type Hex, concat, keccak256, namehash, toHex } from "viem";
 // TODO: pull from ens utils lib or something
 export const NAMEHASH_ZERO = namehash("");
```
lightwalker-eth, line 4: In ENS terminology, this might be referred to as "NODE_ROOT". It's true that this is a bunch of zeros, but more symbolically it represents the root node in ENS.

lightwalker-eth (changes requested):
@tk-o Hey nice updates! I like the direction you're moving in 🚀 Reviewed and shared feedback :+1:

src/ponder-ens-plugins/base/ponder.config.ts
```diff
@@ -0,0 +1,102 @@
...
+import { EarlyAccessRegistrarController } from "./abis/EARegistrarController";
+
+// just for testing...
+const END_BLOCK = undefined;
+
+export const ns = createNs(base.id);
+
+export type NsType<T extends string> = NsReturnType<T, typeof base.id>;
+
+const REGISTRY_ADDRESS = '0xb94704422c2a1e396835a571837aa5ae53285a95';
```
shrugs: i didn't do this in the original config but now i'd suggest just inlining all of these config variables in each of the plugins — defining variables for each of these seems verbose and unnecessary

shrugs (commented):
an abi diff between mainnet and base would be helpful for identifying any logic discrepancies

please configure your editor to use biome to format files — the columnWidth being changed from 80 to 100 is causing a lot of formatting-related changes

src/ponder-ens-plugins/base/ponder.config.ts
```diff
@@ -0,0 +1,102 @@
...
+const REVERSE_REGISTRAR_ADDRESS = '0x79ea96012eea67a83431f1701b3dff7e37f9e282';
+const REVERSE_REGISTRAR_START_BLOCK = 17571485;
+
+const L1_RESOLVER_ADDRESS = '0xde9049636F4a1dfE0a64d1bFe3155C0A14C54F31';
+const L1_RESOLVER_START_BLOCK = 20420641;
+
+const L2_RESOLVER_ADDRESS = '0xC6d566A56A1aFf6508b41f6c90ff131615583BCD';
+const L2_RESOLVER_START_BLOCK = 17575714;
+
+export const config = Object.freeze({
```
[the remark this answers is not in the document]
shrugs: actually i'm realize the endblock stuff doesn't make a lot of sense at the moment, as those are chain-specific and also not important to implement, so please ignore this comment in favor of what i've proposed in #4

src/ponder-ens-plugins/base/ponder.config.ts
tk-o: Makes sense, these are not used outside of the config object.

src/ponder-ens-plugins/eth/ponder.config.ts
```diff
@@ -30,20 +30,20 @@ export const config = createConfig({
...
       abi: Registry,
-      address: REGISTRY_OLD_ADDRESS,
+      address: "0x314159265dd8dbb310642f98f50c066173c1259b",
       startBlock: 3327417,
     },
     [ns("Registry")]: {
       network: "mainnet",
       abi: Registry,
-      address: REGISTRY_ADDRESS,
+      address: "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e",
```
shrugs, line 39: ah sorry — if it's an address that's re-used for a factory config or used more than once we should keep it in a constant so that the reader knows that it's guaranteed to be identical 😅  — sorry i was not specific originally

src/ponder-ens-plugins/base/ponder.config.ts
tk-o: I discussed with @shrugs to inline the config params that are not used outside the config file. Reading the config file should be a good start to understanding which contract is indexed with what params. What do you think?

src/ponder-ens-plugins/chain.ts
tk-o: Using namespaces is needed for correct ponder app build. Ponder flattens all contracts configured across our ponder ens plugins, which causes type conflicts. Namespaces resolve that issue.

src/lib/ens-helpers.ts
tk-o: Updated

src/ponder-ens-plugins/base/README.md
tk-o: Updated

src/lib/ens-helpers.ts
tk-o: Updated

src/ponder-ens-plugins/base/README.md
tk-o: Updated

src/ponder-ens-plugins/main.ts
tk-o: We dropped the activation getter and simply check the value of the `INDEX_BASENAME` env var

src/handlers/Registry.ts
```diff
@@ -1,15 +1,25 @@
...
+import { NsReturnType } from "../lib/plugins";
+import { upsertAccount } from "../lib/upserts";
 
-// a domain is migrated iff it exists and isMigrated is set to true, otherwise it is not
-async function isDomainMigrated(context: Context, node: Hex) {
-  const domain = await context.db.find(domains, { id: node });
-  return domain?.isMigrated ?? false;
+type NsType<T extends string> = NsReturnType<T, "/eth">;
+
+export async function setup({ context }: { context: Context }) {
```
lightwalker-eth, line 12: Can we please document the big idea behind this function? For example: What purpose does it solve? What is the bigger theory behind why it exists?

For example, I'm assuming this function has the responsibility to initialize the indexed state of the registry? Assuming so, can we please document this idea? If that's incorrect, no worries, whatever the correct idea is, let's please document it :+1:

src/handlers/Registry.ts
```diff
@@ -1,15 +1,25 @@
...
+import { ROOT_NODE, encodeLabelhash, makeSubnodeNamehash } from "../lib/ens-helpers";
+import { makeResolverId } from "../lib/ids";
+import { NsReturnType } from "../lib/plugins";
+import { upsertAccount } from "../lib/upserts";
 
-// a domain is migrated iff it exists and isMigrated is set to true, otherwise it is not
-async function isDomainMigrated(context: Context, node: Hex) {
-  const domain = await context.db.find(domains, { id: node });
-  return domain?.isMigrated ?? false;
+type NsType<T extends string> = NsReturnType<T, "/eth">;
```
lightwalker-eth, line 10: Let's please document the thinking here. Thanks

src/handlers/Registry.ts
```diff
@@ -1,15 +1,25 @@
...
+
+export async function setup({ context }: { context: Context }) {
+  // ensure we have an account for the zeroAddress
+  await upsertAccount(context, zeroAddress);
+
+  // ensure we have a root Domain, owned by the zeroAddress, that is default not migrated
```
lightwalker-eth, line 16:
```suggestion
  // initialize the ENS root to be owned by the zeroAddress and not migrated
```

src/handlers/Resolver.ts
```diff
@@ -0,0 +1,270 @@
+import { type Context, type Event } from "ponder:registry";
+import { domains, resolvers } from "ponder:schema";
+import { base, mainnet } from "viem/chains";
+import { hasNullByte, uniq } from "../lib/helpers";
+import { makeResolverId } from "../lib/ids";
+import { NsReturnType } from "../lib/plugins";
+import { upsertAccount, upsertResolver } from "../lib/upserts";
+
+type NsType<T extends string> = NsReturnType<T, "/eth">;
```
lightwalker-eth, line 9: Can we please add a comment here explaining the thinking?

src/lib/plugins.ts
```diff
@@ -0,0 +1,18 @@
+// TODO: change the chainId to the root node value (i.e. namehash("base.eth"))
```
lightwalker-eth, line 1: Should we update this comment? It looks like it might be old?

src/lib/plugins.ts
```diff
@@ -0,0 +1,18 @@
...
+export function createNs<Domain extends EnsRootDomain>(domain: Domain) {
+  /** Creates a name-spaced contract name */
+  return function ns<ContractName extends string>(
+    contractName: ContractName,
+  ): NsReturnType<ContractName, Domain> {
+    return `${domain}/${contractName}` as NsReturnType<ContractName, Domain>;
+  };
+}
+
+export type NsReturnType<
```
lightwalker-eth, line 11:
```suggestion
export type NsContractName<
```

Just a suggestion. Let me know what you think.

src/lib/plugins.ts
```diff
@@ -0,0 +1,18 @@
...
+  ContractName extends string,
+  Domain extends EnsRootDomain,
+> = `${Domain}/${ContractName}`;
+
+export const EnsRootDomains = ["/eth", "/eth/base"] as const;
```
lightwalker-eth, line 16:
```suggestion
export const EnsIndexedDomains = ["/eth", "/eth/base"] as const;
```

Just a suggestion. Let me know what you think.

src/lib/plugins.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type EnsRootDomain = (typeof EnsRootDomains)[number];
```
lightwalker-eth, line 18:
```suggestion
export type EnsIndexedDomain = (typeof EnsIndexedDomains)[number];
```

Just a suggestion. Let me know what you think.

.env.local.example
```diff
@@ -0,0 +1,15 @@
+# RPC urls, follow the format RPC_URL_{chainId}={rpcUrl}
+
+PONDER_RPC_URL_1=https://ethereum-rpc.publicnode.com
+PONDER_RPC_URL_8453=https://base-rpc.publicnode.com
+PONDER_RPC_URL_59144=https://linea-rpc.publicnode.com
+
+# ponder indexer ens deployment configuration
+
+INDEX_BASENAME=.base.eth
```
lightwalker-eth, line 9:
```suggestion
INDEX_SUBNAMES=base.eth
```

Two suggestions here:

1. Rename `INDEX_BASENAME` to `INDEX_SUBNAMES`: "subnames" is more common language in ENS and "base" already has multiple potential interpretations in our project.
2. Don't prefix the value with "." In other words, make it "base.eth" instead of ".base.eth". This way the value is a valid ENS name.

.env.local.example
```diff
@@ -0,0 +1,15 @@
+# RPC urls, follow the format RPC_URL_{chainId}={rpcUrl}
+
+PONDER_RPC_URL_1=https://ethereum-rpc.publicnode.com
+PONDER_RPC_URL_8453=https://base-rpc.publicnode.com
+PONDER_RPC_URL_59144=https://linea-rpc.publicnode.com
+
+# ponder indexer ens deployment configuration
```
lightwalker-eth, line 7:
```suggestion
# Identify which ENS indexer plugin to activate.
```

Can we please add more docs here? For example? Does this value always need to be a single value, or can it be multiple values? If it can be multiple values, how to format it? It can also be nice to document which values are supported here or perhaps in the overall README.md.

.env.local.example
```diff
@@ -0,0 +1,15 @@
...
+
+INDEX_BASENAME=.base.eth
+
+# ponder indexer database configuration
+
+## one schema name per chain ID, i.e. ponder_ens_${INDEX_BASENAME}
+DATABASE_SCHEMA=ponder_ens_base.eth
```
lightwalker-eth, line 14: Can you please document more details about how this `DATABASE_SCHEMA` should be set? Please also explain how this value might relate to the `INDEX_SUBNAMES` environment variable.

.env.local.example
```diff
@@ -0,0 +1,15 @@
...
+PONDER_RPC_URL_8453=https://base-rpc.publicnode.com
+PONDER_RPC_URL_59144=https://linea-rpc.publicnode.com
+
+# ponder indexer ens deployment configuration
+
+INDEX_BASENAME=.base.eth
+
+# ponder indexer database configuration
+
+## one schema name per chain ID, i.e. ponder_ens_${INDEX_BASENAME}
```
lightwalker-eth, line 13: This comment seems old and no longer relevant? Can we please review & update? Thanks!

ponder.config.ts
```diff
@@ -1,98 +1,29 @@
...
+export default ((): AllConfigs => {
+  switch (BASENAME) {
+    case ethBaseName:
+      activateEth();
+      return ethereumConfig as AllConfigs;
+    case baseBaseName:
+      activateBase();
+      return baseConfig as AllConfigs;
+    default:
+      throw new Error(`Unsupported base name: ${BASENAME}`);
```
lightwalker-eth, line 27:
```suggestion
      throw new Error(`Unsupported subname indexer: ${BASENAME}`);
```

ponder.config.ts
```diff
@@ -1,98 +1,29 @@
...
+import { BASENAME } from "./src/lib/ens-helpers";
+import {
+  activate as activateBase,
+  baseName as baseBaseName,
+  config as baseConfig
+} from "./src/ponder-ens-plugins/eth.base/ponder.config";
+import {
+  activate as activateEth,
+  baseName as ethBaseName,
+  config as ethereumConfig
```
lightwalker-eth, line 10:
```suggestion
  config as ethConfig
```

src/lib/ens-helpers.ts
```diff
@@ -1,8 +1,26 @@
 import { type Hex, concat, keccak256, namehash, toHex } from "viem";
 
+class BaseName {
```
lightwalker-eth, line 3: Can you please document what the goals are for this? It will help me to review and share advice :+1:

src/handlers/Registrar.ts
```diff
@@ -0,0 +1,134 @@
+import { type Context, type Event } from "ponder:registry";
+import { domains, registrations } from "ponder:schema";
+import { type Hex, namehash } from "viem";
+import { isLabelValid, makeSubnodeNamehash, tokenIdToLabel } from "../lib/ens-helpers";
+import { NsReturnType } from "../lib/plugins";
+import { upsertAccount, upsertRegistration } from "../lib/upserts";
+
+type NsType<T extends string> = NsReturnType<T, "/eth">;
```
lightwalker-eth, line 8: Can we please document the idea here?

src/handlers/Registrar.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const GRACE_PERIOD_SECONDS = 7776000n; // 90 days in seconds
+
+export const makeRegistryHandlers = (baseName: `.${`${string}`}eth`) => {
```
lightwalker-eth, line 12: Can we please document the idea here? I'm having difficulty reviewing and sharing feedback. Thanks

src/ponder-ens-plugins/eth/ponder.config.ts
```diff
@@ -0,0 +1,96 @@
...
+import { EthRegistrarController } from "./abis/EthRegistrarController";
+import { EthRegistrarControllerOld } from "./abis/EthRegistrarControllerOld";
+import { LegacyPublicResolver } from "./abis/LegacyPublicResolver";
+import { NameWrapper } from "./abis/NameWrapper";
+import { Registry } from "./abis/Registry";
+import { Resolver } from "./abis/Resolver";
+
+const RESOLVER_ABI = mergeAbis([LegacyPublicResolver, Resolver]);
+
+export const baseName = ".eth" as const;
```
lightwalker-eth, line 15: Please see my other feedback on this:

1. We shouldn't prefix with "."
2. The ENS community doesn't use language such as "baseName". We should discuss what language would be better.
3. In this context, I assume we should just be calling this "indexedSubnames"?

src/ponder-ens-plugins/eth/ponder.config.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const nestedNamespace = "/eth" as const;
+export const ns = createNs(nestedNamespace);
+export type NsType<T extends string> = NsReturnType<T, typeof nestedNamespace>;
```
lightwalker-eth, line 19: We need to document what each of these lines are doing. It's not clear when reviewing the code. Cheers thanks

src/ponder-ens-plugins/eth.base/ponder.config.ts
```diff
@@ -0,0 +1,79 @@
...
+import { base } from "viem/chains";
+
+import { NsReturnType, createNs } from "../../lib/plugins";
+import { BaseRegistrar } from "./abis/BaseRegistrar";
+import { EarlyAccessRegistrarController } from "./abis/EARegistrarController";
+import { L2Resolver } from "./abis/L2Resolver";
+import { RegistrarController } from "./abis/RegistrarController";
+import { Registry } from "./abis/Registry";
+
+export const baseName = ".base.eth" as const;
```
lightwalker-eth, line 12: Please see my comments on this from other files :+1:

lightwalker-eth (changes requested):
@tk-o Hey great to see the updates here. 👍  Reviewed and shared feedback 😄

shrugs:
i much prefer the `root` argument passed to `makeRegistryHandlers()` because it makes clear that the registry handlers are universal except for root that nodes in those handlers are relative to. it also means that in the future v2 this shared code should be relatively identical. using the constant is less clear and requires more mental overhead for the reader

src/handlers/Registrar.ts
```diff
@@ -1,51 +1,54 @@
...
-type NsType<T extends string> = NsReturnType<T, "/eth">;
-
 const GRACE_PERIOD_SECONDS = 7776000n; // 90 days in seconds
 
-export const makeRegistryHandlers = (baseName: `.${`${string}`}eth`) => {
-  const baseNameNode = namehash(baseName.slice(1));
+/**
+ * A factory function that returns Ponder indexing handlers for a specified index name/subname.
+ */
+export const makeRegistryHandlers = (indexedSubname: `${string}eth`) => {
```
shrugs, line 12: now that i understand the nomenclature better, let's call this `indexedSubname` variable `managedSubname` (and `managedSubnameNode`) so that people know that these Registrar handlers are in the context of a registrar that manages the provided subname

src/handlers/Registrar.ts
```diff
@@ -1,51 +1,54 @@
...
+        .set({ labelName: name, name: `${name}${indexedSubname}` });
     }
 
     await context.db.update(registrations, { id: label }).set({ labelName: name, cost });
   }
 
   return {
-    get baseNameNode() {
-      return baseNameNode;
+    get indexedSubnameNode() {
```
shrugs, line 32: why provide this back, it will be known to the caller because they provided it as an input agument

src/handlers/Registrar.ts
```diff
@@ -1,51 +1,54 @@
...
     },
 
     async handleNameRegistered({
       context,
       event,
     }: {
       context: Context;
-      event: Omit<Event<NsType<"BaseRegistrar:NameRegistered">>, "name">;
+      event: {
+        block: { timestamp: bigint };
```
shrugs, line 42: i agree on splitting out the args for clarity, but can you type the block argument as a ponder `Block`?

src/handlers/Registrar.ts
```diff
@@ -114,12 +119,16 @@ export const makeRegistryHandlers = (baseName: `.${`${string}`}eth`) => {
       args: { tokenId, from, to },
     }: {
       context: Context;
-      args: Event<NsType<"BaseRegistrar:Transfer">>["args"];
+      args: {
```
shrugs, line 122: and i think for each of these functions in `Registrar.ts` we should standardize the api to be like a ponder handler: it always accepts context and event, and `args` is within the event. we need not do this in this pr, i can manage it once we've merged if you'd like

src/handlers/Registry.ts
```diff
@@ -1,19 +1,19 @@
...
-import { NsReturnType } from "../lib/plugins";
 import { upsertAccount } from "../lib/upserts";
 
-type NsType<T extends string> = NsReturnType<T, "/eth">;
-
-export async function setup({ context }: { context: Context }) {
+/**
+ * Initialize the ENS root node with the zeroAddress as the owner.
+ */
+export async function handleRootNodeCreation({ context }: { context: Context }) {
```
shrugs, line 12: agreed with lightwalker's naming concern, and apologies for the nitpick, but how about calling this `setupRootNode`

src/handlers/Resolver.ts
```diff
@@ -1,37 +1,19 @@
 import { type Context, type Event } from "ponder:registry";
 import { domains, resolvers } from "ponder:schema";
-import { base, mainnet } from "viem/chains";
+import { Hex } from "viem";
 import { hasNullByte, uniq } from "../lib/helpers";
 import { makeResolverId } from "../lib/ids";
-import { NsReturnType } from "../lib/plugins";
 import { upsertAccount, upsertResolver } from "../lib/upserts";
 
-type NsType<T extends string> = NsReturnType<T, "/eth">;
```
shrugs: very nice

src/lib/ponder-plugin-utils.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/** The return type of the `ponderNamespace` function */
+export type PonderNamespaceReturnType<
+  ContractName extends string,
+  NsPath extends PonderNsPath,
+> = `${NsPath}/${ContractName}`;
+
+type PonderNsPath<T extends PonderNsPath = "/"> = `` | `/${string}` | `/${string}${T}`;
+
+type EthSubname = `${string}eth`;
```
shrugs, line 64: very nice type work in this file 🫡

shrugs (commented):
excellent documentation & typescript type work! i only have nitpicks in here, i think, so it's looking very close to merge

src/handlers/Registrar.ts
tk-o: We need this value in the caller for some inline indexing handlers:
https://github.com/namehash/ens-multichain-indexer/blob/aad651cf532f1d89b214695994bb2d8e2574b7e3/src/ponder-ens-plugins/eth.base/handlers/Registrar.ts#L8-L17
https://github.com/namehash/ens-multichain-indexer/blob/aad651cf532f1d89b214695994bb2d8e2574b7e3/src/ponder-ens-plugins/eth.base/handlers/Registrar.ts#L31

shrugs:
looks like biome got misconfigured or something — some line width formatting related changes in the recent commits

src/handlers/Registrar.ts
```diff
@@ -1,51 +1,55 @@
...
 import { upsertAccount, upsertRegistration } from "../lib/upserts";
 
-type NsType<T extends string> = NsReturnType<T, "/eth">;
-
 const GRACE_PERIOD_SECONDS = 7776000n; // 90 days in seconds
 
-export const makeRegistryHandlers = (baseName: `.${`${string}`}eth`) => {
-  const baseNameNode = namehash(baseName.slice(1));
+/**
+ * A factory function that returns Ponder indexing handlers for a specified index name/subname.
```
lightwalker-eth, line 11:
```suggestion
 * A factory function that returns Ponder indexing handlers for a specified subname.
```

ponder.config.ts
```diff
@@ -1,29 +1,29 @@
...
-    case baseBaseName:
-      activateBase();
-      return baseConfig as AllConfigs;
+      return ethConfig as AllConfigs;
+    case ethBaseManagedSubname:
+      activateEthBase();
+      return ethBaseConfig as AllConfigs;
     default:
-      throw new Error(`Unsupported base name: ${BASENAME}`);
+      throw new Error(`Unsupported base name: ${INDEXED_SUBNAME}`);
```
lightwalker-eth, line 27:
```suggestion
      throw new Error(`Unsupported INDEXED_SUBNAME: ${INDEXED_SUBNAME}`);
```

src/lib/ponder-plugin-utils.ts
```diff
@@ -0,0 +1,64 @@
...
+    return `${path}/${contractName}`;
+  };
+}
+
+type TransformDomain<T extends string> = T extends `${infer Sub}.${infer Rest}`
+  ? `/${TransformDomain<Rest>}/${Sub}`
+  : `/${T}`;
+
+/**
+ * Transforms a domain name into a path-like format
```
lightwalker-eth, line 42: Happy to see this comment here!

Can we please also document why we created this function? Anytime we write a function that has a non-obvious purpose we should be documenting why it exists.

lightwalker-eth (changes requested):
@tk-o Thanks for your updates here! Very happy to see docs improving. Let's continue to push hard on making our docs and use of language properly aligned.

Reviewed and shared feedback :+1:

src/ponder-ens-plugins/eth/ponder.config.ts
```diff
@@ -12,11 +12,9 @@ import { Resolver } from "./abis/Resolver";
 
 const RESOLVER_ABI = mergeAbis([LegacyPublicResolver, Resolver]);
 
-export const baseName = ".eth" as const;
+export const managedSubname = "eth";
```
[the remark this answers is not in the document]
tk-o: cc @shrugs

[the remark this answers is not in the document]
tk-o: I've reverted back to `indexedSubname`

src/ponder-ens-plugins/eth.base/ponder.config.ts
```diff
@@ -2,21 +2,16 @@ import { createConfig, factory } from "ponder";
...
-export const baseName = ".base.eth" as const;
+export const managedSubname = "base.eth" as const;
 
-const nestedNamespace = "/eth/base" as const;
-export const ns = createNs(nestedNamespace);
-export type NsType<T extends string> = NsReturnType<T, typeof nestedNamespace>;
-
-const START_BLOCK = undefined; // 17607350;
-const END_BLOCK = undefined; // 17607351;
+export const ponderNamespace = createPonderNamespace(managedSubname);
```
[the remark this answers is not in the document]
shrugs: I might suggest `contractNamespace`

src/ponder-ens-plugins/eth.base/ponder.config.ts
```diff
@@ -2,21 +2,16 @@ import { createConfig, factory } from "ponder";
...
-import { NsReturnType, createNs } from "../../lib/plugins";
+import { createPonderNamespace } from "../../lib/ponder-plugin-utils";
 import { BaseRegistrar } from "./abis/BaseRegistrar";
 import { EarlyAccessRegistrarController } from "./abis/EARegistrarController";
 import { L2Resolver } from "./abis/L2Resolver";
 import { RegistrarController } from "./abis/RegistrarController";
 import { Registry } from "./abis/Registry";
 
-export const baseName = ".base.eth" as const;
+export const managedSubname = "base.eth" as const;
```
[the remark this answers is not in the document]
shrugs: @lightwalker-eth it got lost in the github comments but: the registrar handlers are in the context of the subname that the registrar 'manages' — the ETH Registrar manages .eth, base's l2 registrar deployment manages .base.eth subnames, etc. perhaps there's a better word but i don't like 'indexed' in this case, since the argument is specifying which parent node all of the events in this registrar are relative to

src/handlers/Registrar.ts
shrugs: ah, hmm. i don't love this but it's ok for the moment 👍  we'll refactor this in v2

lightwalker-eth (approved):
@tk-o Hey super awesome updates! 🚀 🚀 🚀

PR #4, opened 2025-01-03
Refactor/plugin logic

shrugs:
this pr:

- pulls out the shared logic of Registry and Resolver into a shared `handlers/` directory, and each plugin's handlers reference that code
  - the Registrar logic still needs to be DRY'd up
  - eth mainnet-specific logic (old registry migration) is kept in eth-specific plugin
  - event types of the shared handlers should be improved, rn lucky that the event types are perfect matches between base and mainnet

- simplifies the plugin architecture
  - the root ponder.config.ts only exports a single ponder config instead of merging them
  - but types it as both, making the generated types happy
  - could create a `PonderPlugin` type that defines `{ config, activate }` rather than just importing the modules, but is ok


tk-o (approved):
I like how you exported a result of `createConfig` function from each plugin separately.

The top-level shared handlers look OK, even if we use the event names type from the `.eth` namespace.

Feel free to merge 🚀

PR #13, opened 2025-01-09
feat(plugin): index linea.eth

tk-o:
This PR builds on top of
- https://github.com/namehash/ens-multichain-indexer/pull/2
and needs to be only merged afterwards.

The PR introduces the `linea.eth` plugin for Ponder indexer. The `linea.eth` plugin is very similar to the `eth` plugin. We can say it extends it (with minor naming alterations in some interfaces).


src/lib/plugin-helpers.ts
```diff
@@ -48,7 +48,7 @@ export function createPluginNamespace<Subname extends string>(subname: Subname)
 }
 
 type TransformNameIntoPath<Name extends string> = Name extends `${infer Sub}.${infer Rest}`
-  ? `/${TransformNameIntoPath<Rest>}/${Sub}`
+  ? `${TransformNameIntoPath<Rest>}/${Sub}`
```
tk-o, line 51: This is a bugfix, as without it, the type would always start with a double slash. We need a single slash.

.github/workflows/static-analysis.yml
```diff
@@ -24,8 +24,14 @@ jobs:
...
+        run: pnpm audit --audit-level=low
+
       - name: Install dependencies
         run: pnpm install
 
       - name: Run Biome CI
         run: pnpm biome ci
+
+      - name: Run TypeScript type checking
+        run: pnpm run typecheck
```
tk-o, line 37: I accidentally broke the `eth` plugin while making changes to `handlers` files to enable `linea.eth` integration. This CI step will notify me if I do a similar mistake in the future.

src/lib/plugin-helpers.ts
lightwalker-eth: @tk-o Nice, thanks for catching this :+1:

What do you think about creating an issue for us to action in a future PR? The issue would be to add vitest to the project and to define some initial unit tests.

src/handlers/NameWrapper.ts
```diff
@@ -61,150 +61,154 @@ async function handleTransfer(
...
-  context,
-  event,
-}: {
-  context: Context;
-  event: {
-    args: {
-      node: Hex;
-      fuses: number;
-    };
+export const makeRegistryHandlers = (ownedName: `${string}eth`) => {
```
shrugs, line 65: `makeRegistrarHandlers`*

shrugs (approved):
only renaming factory function, otherwise LGTM,

src/lib/plugin-helpers.ts
tk-o: Recorded your request here:
- https://github.com/namehash/ens-multichain-indexer/issues/15

shrugs (approved):
LGTM nice work

PR #21, opened 2025-01-12
feat(cross-subname-indexing): allow activating multiple plugins together

tk-o:
This PR allows cross-subname indexing. For example. with this change, it's possible to get a list of all names a given address owns across multiple subnames (i.e. `base.eth`, `linea.eth`).

### Example query

```gql
query GetOwnedDomains($ownerAddreess: String!) {
  domainss(
    where: {OR: [{ownerId: $ownerAddreess}, {wrappedOwnerId: $ownerAddreess}]}
  ) {
    items {
      labelName
      name
      ownerId
      wrappedOwnerId
      createdAt
      expiryDate
    }
  }
}
```

### Preview

Domain names under `eth`, `base.eth`, and `linea.eth`:
image

### DIFF
--- a/ponder.config.ts
+++ b/ponder.config.ts
@@ -4,13 +4,10 @@ import * as baseEthPlugin from "./src/plugins/base.eth/ponder.config";
 import * as ethPlugin from "./src/plugins/eth/ponder.config";
 import * as lineaEthPlugin from "./src/plugins/linea.eth/ponder.config";
 
+/** list of all available plugins — any of them can be activated in the runtime */
 const plugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
 
-// The type of the exported default is the intersection of all plugin configs to
-// each plugin can be correctly typechecked
 type AllPluginsConfig = IntersectionOf<(typeof plugins)[number]["config"]>;
-
-export default loadActivePlugins();
 
 /**
  * Activates the indexing handlers included in selected active plugins and returns their combined config.
@@ -26,3 +23,7 @@ function loadActivePlugins(): AllPluginsConfig {
 
   return config as AllPluginsConfig;
 }
+
+// The type of the defuexporte is the intersection of all plugin configs to
+// each plugin can be correctly typechecked
+export default loadActivePlugins();
--- a/ponder.config.ts
+++ b/ponder.config.ts
@@ -4,17 +4,20 @@ import * as baseEthPlugin from "./src/plugins/base.eth/ponder.config";
 import * as ethPlugin from "./src/plugins/eth/ponder.config";
 import * as lineaEthPlugin from "./src/plugins/linea.eth/ponder.config";
 
-/** list of all available plugins — any of them can be activated in the runtime */
+// list of all available plugins
+// any of them can be activated in the runtime
 const plugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
 
+// intersection of all available plugin configs to support correct typechecking
+// of the indexing handlers
 type AllPluginsConfig = IntersectionOf<(typeof plugins)[number]["config"]>;
 
-/**
- * Activates the indexing handlers included in selected active plugins and returns their combined config.
- */
-function loadActivePlugins(): AllPluginsConfig {
+// Activates the indexing handlers included in selected active plugins and
+// returns and intersection of their combined config.
+function getActivePluginsConfig(): AllPluginsConfig {
   const activePlugins = getActivePlugins(plugins);
 
+  // load indexing handlers from the active plugins into the runtime
   activePlugins.forEach((plugin) => plugin.activate());
 
   const config = activePlugins
@@ -24,6 +27,6 @@ function loadActivePlugins(): AllPluginsConfig {
   return config as AllPluginsConfig;
 }
 
-// The type of the defuexporte is the intersection of all plugin configs to
-// each plugin can be correctly typechecked
-export default loadActivePlugins();
+// The type of the default export is the intersection of all available plugin
+// configs to each plugin can be correctly typechecked
+export default getActivePluginsConfig();
--- a/ponder.config.ts
+++ b/ponder.config.ts
@@ -1,29 +1,39 @@
-import { ACTIVE_PLUGIN } from "./src/lib/plugin-helpers";
+import { createConfig } from "ponder";
+import { deepMergeRecursive } from "./src/lib/helpers";
+import { type IntersectionOf, getActivePlugins } from "./src/lib/plugin-helpers";
 import * as baseEthPlugin from "./src/plugins/base.eth/ponder.config";
 import * as ethPlugin from "./src/plugins/eth/ponder.config";
 import * as lineaEthPlugin from "./src/plugins/linea.eth/ponder.config";
 
 const plugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
 
-// here we export only a single 'plugin's config, by type it as every config
-// this makes all of the mapping types happy at typecheck-time, but only the
-// relevant config is run at runtime
-export default ((): AllConfigs => {
-  const pluginToActivate = plugins.find((p) => p.ownedName === ACTIVE_PLUGIN);
-
-  if (!pluginToActivate) {
-    throw new Error(`Unsupported ACTIVE_PLUGIN: ${ACTIVE_PLUGIN}`);
-  }
-
-  pluginToActivate.activate();
-
-  return pluginToActivate.config as AllConfigs;
-})();
-
-// Helper type to get the intersection of all config types
-type IntersectionOf<T> = (T extends any ? (x: T) => void : never) extends (x: infer R) => void
-  ? R
-  : never;
 // The type of the exported default is the intersection of all plugin configs to
 // each plugin can be correctly typechecked
-type AllConfigs = IntersectionOf<(typeof plugins)[number]["config"]>;
+type AllPluginsConfig = IntersectionOf<(typeof plugins)[number]["config"]>;
+
+const activePlugins = loadActivePlugins();
+
+export default (() =>
+  createConfig({
+    contracts: {
+      ...activePlugins.contracts,
+    },
+    networks: {
+      ...activePlugins.networks,
+    },
+  }) as AllPluginsConfig)();
+
+/**
+ * Activates the indexing handlers included in selected active plugins and returns their combined config.
+ */
+function loadActivePlugins() {
+  const activePlugins = getActivePlugins(plugins);
+
+  activePlugins.forEach((plugin) => plugin.activate());
+
+  const config = activePlugins
+    .map((plugin) => plugin.config)
+    .reduce((acc, val) => deepMergeRecursive(acc, val), {} as AllPluginsConfig);
+
+  return config;
+}
--- a/ponder.config.ts
+++ b/ponder.config.ts
@@ -1,4 +1,3 @@
-import { createConfig } from "ponder";
 import { deepMergeRecursive } from "./src/lib/helpers";
 import { type IntersectionOf, getActivePlugins } from "./src/lib/plugin-helpers";
 import * as baseEthPlugin from "./src/plugins/base.eth/ponder.config";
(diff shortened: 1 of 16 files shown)


src/handlers/Registrar.ts
```diff
@@ -23,7 +23,7 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
     if (domain.labelName !== name) {
       await context.db
         .update(domains, { id: node })
-        .set({ labelName: name, name: `${name}${ownedName}` });
+        .set({ labelName: name, name: `${name}.${ownedName}` });
```
tk-o, line 26: This was a bug, that existed since we dropped `.` from the `ownedName` param on plugin's configuration object.

src/handlers/Registry.ts
```diff
@@ -15,12 +15,15 @@ export async function setupRootNode({ context }: { context: Context }) {
...
-  });
+  await context.db
+    .insert(domains)
+    .values({
+      id: ROOT_NODE,
+      ownerId: zeroAddress,
+      createdAt: 0n,
+      isMigrated: false,
+    })
+    .onConflictDoNothing();
```
tk-o, line 26: I had to introduce `.onConflictDoNothing()` as multiple plugins run the `setupRootNode()` function, so every call apart from the initial one would cause an error.

src/plugins/base.eth/ponder.config.ts
```diff
@@ -13,6 +14,11 @@ export const ownedName = "base.eth" as const;
 
 export const pluginNamespace = createPluginNamespace(ownedName);
 
+// constrain the ponder indexing between the following start/end blocks
+// https://ponder.sh/0_6/docs/contracts-and-networks#block-range
+const START_BLOCK: ContractConfig["startBlock"] = 24944146;
```
tk-o, line 19: Test value here, will replace it with `undefined` before merging.

ponder.config.ts
```diff
@@ -1,29 +1,39 @@
...
+// each plugin can be correctly typechecked
+type AllPluginsConfig = IntersectionOf<(typeof plugins)[number]["config"]>;
 
-  if (!pluginToActivate) {
-    throw new Error(`Unsupported ACTIVE_PLUGIN: ${ACTIVE_PLUGIN}`);
-  }
+const activePlugins = loadActivePlugins();
 
-  pluginToActivate.activate();
+export default (() =>
```
shrugs, line 16: can you just return `createConfig(loadActivePlugins()) as AllPluginsConfig`? feels like spreading `contracts: {...activePlugins.contracts}` isn't doing much?

tk-o: Thanks for this suggestion, now it looks way cleaner ;)

src/handlers/Registry.ts
```diff
@@ -130,6 +132,12 @@ export const handleNewOwner =
       await context.db.update(domains, { id: domain.id }).set({ name, labelName });
     }
 
+    if (await incrementSubdomainCountOnParent()) {
+      await context.db
+        .update(domains, { id: parent!.id })
+        .set((row) => ({ subdomainCount: row.subdomainCount + 1 }));
+    }
```
tk-o, line 139: @shrugs, I hope you'll like this change. It allows not updating the parent domain entry if it does not exist. Useful during partial block-range indexing, and neutral to full block-range indexing.

src/lib/helpers.ts
```diff
@@ -16,3 +16,55 @@ export const blockConfig = (
...
+ * @param source The source object to merge from.
+ * @returns The merged object.
+ * @see https://stackoverflow.com/a/48218209
+ * @example
+ * const obj1 = { a: 1, b: 2, c: { d: 3 } };
+ * const obj2 = { a: 4, c: { e: 5 } };
+ * const obj3 = deepMergeRecursive(obj1, obj2);
+ * // { a: 4, b: 2, c: { d: 3, e: 5 } }
+ */
+export function deepMergeRecursive<T extends AnyObject, U extends AnyObject>(
```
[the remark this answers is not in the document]
tk-o: We could use lib, I tried `ts-deepmerge` and `deepmerge` — couldn't get the types to work easily, so gave up.  On the other hand, this one works well and does its job.

src/lib/plugin-helpers.ts
```diff
@@ -81,3 +81,33 @@ type PluginNamespacePath<T extends PluginNamespacePath = "/"> =
 
 /** @var the requested active plugin name (see `src/plugins` for available plugins) */
 export const ACTIVE_PLUGIN = process.env.ACTIVE_PLUGIN;
```
lightwalker-eth, line 83: Suggest we rename this to `ACTIVE_PLUGINS` (plural) too. That includes the environment variable.

Please also carefully review all existing related comments / docs to ensure they are updated to reflect this change too.

src/lib/plugin-helpers.ts
```diff
@@ -81,3 +81,33 @@ type PluginNamespacePath<T extends PluginNamespacePath = "/"> =
...
+ *
+ * The `ACTIVE_PLUGIN` environment variable is a comma-separated list of plugin
+ * names. The function returns the plugins that are included in the list.
+ *
+ * @param plugins is a list of available plugins
+ * @returns the active plugins
+ */
+export function getActivePlugins<T extends { ownedName: string }>(plugins: readonly T[]): T[] {
+  const pluginsToActivateByOwnedName = ACTIVE_PLUGIN
+    ? ACTIVE_PLUGIN.split(",").map((p) => p.toLowerCase())
```
lightwalker-eth, line 96: Suggest we not make plugin names case insensitive.

src/lib/plugin-helpers.ts
```diff
@@ -81,3 +81,33 @@ type PluginNamespacePath<T extends PluginNamespacePath = "/"> =
...
+    : [];
+
+  if (!pluginsToActivateByOwnedName.length) {
+    throw new Error("No active plugins found. Please set the ACTIVE_PLUGIN environment variable.");
+  }
+
+  return plugins.filter((plugin) =>
+    pluginsToActivateByOwnedName.includes(plugin.ownedName.toLowerCase()),
```
lightwalker-eth, line 104: Suggest we not make plugin names case insensitive.

Also suggest we document that plugin names are case sensitive in all the relevant places :+1:

src/lib/plugin-helpers.ts
```diff
@@ -81,3 +81,33 @@ type PluginNamespacePath<T extends PluginNamespacePath = "/"> =
...
+export function getActivePlugins<T extends { ownedName: string }>(plugins: readonly T[]): T[] {
+  const pluginsToActivateByOwnedName = ACTIVE_PLUGIN
+    ? ACTIVE_PLUGIN.split(",").map((p) => p.toLowerCase())
+    : [];
+
+  if (!pluginsToActivateByOwnedName.length) {
+    throw new Error("No active plugins found. Please set the ACTIVE_PLUGIN environment variable.");
+  }
+
+  return plugins.filter((plugin) =>
```
lightwalker-eth, line 103: What do you think about throwing an error if any of the plugins listed in `ACTIVE_PLUGINS` is not recognized? I think it's good to proactively throw an error if even a single value there is unrecognized.

src/lib/helpers.ts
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Thanks for proposing this. Agreed with @shrugs that a well-tested library for this sounds attractive. My head hurts thinking for code review of testing on this.

Appreciate if it might take some time to figure out how to get one of those libraries to work. If we can find a solution to that in this PR, then great :+1: If it needs more time, then please feel welcome to create a new GitHub Issue for us to track this goal and we can come back to it in a separate PR.

src/handlers/Registry.ts
lightwalker-eth: @tk-o Hey really appreciate your PR comment on this! Could you please also document this idea on each place in the code where this idea applies? This way the relevant documentation will live on in our code in an easy to find place after this PR is merged.

Taking a step-back, I'm a bit confused on why there might be conflicts caused by other plugins here. I suppose the conflict might be caused by how we are enabling the base.eth or linea.eth plugins to be activated without the .eth plugin being activated. In other words, we don't have any concept of plugin dependencies implemented, such that the activation of plugin X also requires the activation of plugin Y.

Do I understand that correctly? Appreciate your advice.

I'm ok to not introduce plugin dependencies yet. I think we can come back to this as a separate goal in the future. It would add complexity and I'm not sure if its even the best idea. For now just trying to understand the current situation better. Thanks :+1:

ponder.config.ts
```diff
@@ -1,29 +1,28 @@
...
 import * as lineaEthPlugin from "./src/plugins/linea.eth/ponder.config";
 
 const plugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
 
-// here we export only a single 'plugin's config, by type it as every config
-// this makes all of the mapping types happy at typecheck-time, but only the
-// relevant config is run at runtime
-export default ((): AllConfigs => {
-  const pluginToActivate = plugins.find((p) => p.ownedName === ACTIVE_PLUGIN);
+// The type of the exported default is the intersection of all plugin configs to
```
lightwalker-eth, line 9: A few questions:

1. What is "the type of the exported default"?
2. When you write "all plugin configs" or `AllPluginsConfig`, is this all activated plugins, or all available plugins? Appreciate if you could clarify terminology.

ponder.config.ts
```diff
@@ -1,29 +1,28 @@
-import { ACTIVE_PLUGIN } from "./src/lib/plugin-helpers";
+import { deepMergeRecursive } from "./src/lib/helpers";
+import { type IntersectionOf, getActivePlugins } from "./src/lib/plugin-helpers";
 import * as baseEthPlugin from "./src/plugins/base.eth/ponder.config";
 import * as ethPlugin from "./src/plugins/eth/ponder.config";
 import * as lineaEthPlugin from "./src/plugins/linea.eth/ponder.config";
 
 const plugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
```
lightwalker-eth, line 7:
```suggestion
const availablePlugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
```

Suggesting this to help us give more distinct naming to all available plugins vs all activated plugins.

src/handlers/Registrar.ts
lightwalker-eth: Good catch, thanks.

src/handlers/Registry.ts
lightwalker-eth: @tk-o Hey thanks for your PR comment above. I see that comment is specific to lines 135-139. What about all the other changes being proposed to `handleNewOwner` in this PR? Can you please help us understand the bigger goal for these changes?

While we're working on our "V1" milestone I'm cautious about changes to indexing logic that are difficult for me to easily review. Open to making them if there's a good reason :+1:

lightwalker-eth (changes requested):
@tk-o Super exciting updates in this PR! 🚀 Reviewed and shared feedback.

src/plugins/linea.eth/handlers/EthRegistrar.ts
```diff
@@ -8,13 +11,34 @@ const {
...
 } = makeRegistrarHandlers(ownedName);
 
 export default function () {
   ponder.on(pluginNamespace("BaseRegistrar:NameRegistered"), handleNameRegistered);
   ponder.on(pluginNamespace("BaseRegistrar:NameRenewed"), handleNameRenewed);
 
   ponder.on(pluginNamespace("BaseRegistrar:Transfer"), async ({ context, event }) => {
+    if (event.args.from === zeroAddress) {
+      /**
+       * Address the issue where in the same transaction the Transfer event occurs before the NameRegistered event.
```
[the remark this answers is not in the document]
tk-o: 1. The contract emits the `Transfer` event before emitting the `NameRegistered` event. It's how the contract implementation is done. Our `BaseRegistrar:Transfer` handler previously assumed that the domain name would already be included in the `domains` table, but that's not always the case. Registering a new subname first transfers an NFT token from the zero address to the owner, and then emits the `NameRegistered` event.

2. Minting the subname NFT is effectively transferring it from the zero address to the owner.

src/lib/plugin-helpers.ts
```diff
@@ -81,3 +81,33 @@ type PluginNamespacePath<T extends PluginNamespacePath = "/"> =
...
+ * Returns the active plugins list based on the `ACTIVE_PLUGIN` environment variable.
+ *
+ * The `ACTIVE_PLUGIN` environment variable is a comma-separated list of plugin
+ * names. The function returns the plugins that are included in the list.
+ *
+ * @param plugins is a list of available plugins
+ * @returns the active plugins
+ */
+export function getActivePlugins<T extends { ownedName: string }>(plugins: readonly T[]): T[] {
+  const pluginsToActivateByOwnedName = ACTIVE_PLUGIN
```
shrugs: let's rename this to ACTIVE_PLUGINS now that it's a plural array, and make sure to update .env.local.example as well

ponder.config.ts
tk-o: 1. I should say "the type of the default export"
2. "all plugins config" mean an intersection type of the available plugins (regardless if active or not). This type is required to support all defined indexing handlers type-safety.

PR #23, opened 2025-01-13
singularize schema table names

shrugs:
- first commit moves the individual imports to a module import
  - from `import { domains } from "ponder:schema"` and `.insert(domains)`
  - to `import schema from "ponder:schema"` and `.insert(schema.domains)`
- second comment renames the tables so
  - from `.insert(schema.domains)`
  - to `.insert(schema.domain)`

this makes the autogenerated api correct

### DIFF
--- a/ponder.schema.ts
+++ b/ponder.schema.ts
@@ -1,7 +1,7 @@
 import { onchainTable, relations } from "ponder";
 import type { Address } from "viem";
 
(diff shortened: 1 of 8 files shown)


shrugs:
@tk-o i'm going to YOLO merge this into main cause we've already talked about it and i'd like to brand off of it for graphql work

PR #24, opened 2025-01-13
feat(cross-chain-indexing): enable multiple concurrent plugins

tk-o:
For example. with this change, it's possible to get a list of all names a given address owns across multiple chains (i.e. `base.eth` subnames on Base, and `linea.eth` subnames on Linea.

### Example query

```gql
query GetOwnedDomains($ownerAddreess: String!) {
  domains(
    where: {OR: [{ownerId: $ownerAddreess}, {wrappedOwnerId: $ownerAddreess}]}
  ) {
    items {
      labelName
      name
      ownerId
      wrappedOwnerId
      createdAt
      expiryDate
    }
  }
}
```

### Preview

Domain names under `eth`, `base.eth`, and `linea.eth`:

untitled (1)

---

This PR overrides:
- https://github.com/namehash/ens-multichain-indexer/pull/21


.env.local.example
```diff
@@ -7,13 +7,25 @@ RPC_URL_59144=https://linea-rpc.publicnode.com
...
+# 2) prevent data corruption
+#
+# No two indexer instances can use the same database schema at the same time.
+#
+# Read more about database schema rules here:
+# https://ponder.sh/docs/api-reference/database#database-schema-rules
+DATABASE_SCHEMA=ens
+
+# This is the connection string for the database that the indexer will use to store data.
+# It should be in the format of `postgresql://<username>:<password>@<host>:<port>/<database>`
```
tk-o, line 30: @lightwalker-eth I've added more context around what those env vars are meant for

ponder.config.ts
```diff
@@ -1,29 +1,32 @@
-import { ACTIVE_PLUGIN } from "./src/lib/plugin-helpers";
+import { deepMergeRecursive } from "./src/lib/helpers";
```
tk-o, line 1: @shrugs I will replace that tomorrow with a previously suggested library.

ponder.config.ts
```diff
@@ -1,29 +1,32 @@
...
+// intersection of all available plugin configs to support correct typechecking
+// of the indexing handlers
+type AllPluginsConfig = IntersectionOf<(typeof plugins)[number]["config"]>;
 
-  if (!pluginToActivate) {
-    throw new Error(`Unsupported ACTIVE_PLUGIN: ${ACTIVE_PLUGIN}`);
-  }
+// Activates the indexing handlers included in selected active plugins and
+// returns and intersection of their combined config.
+function getActivePluginsConfig(): AllPluginsConfig {
```
tk-o, line 17: @lightwalker-eth updated the comments to express the active plugins config idea more clearly.

src/handlers/Registrar.ts
```diff
@@ -23,7 +23,7 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
     if (domain.labelName !== name) {
       await context.db
         .update(schema.domain, { id: node })
-        .set({ labelName: name, name: `${name}${ownedName}` });
+        .set({ labelName: name, name: `${name}.${ownedName}` });
```
tk-o, line 26: Fixing the bug introduced when we dropped the starting dot from the `ownedName` param.

src/handlers/Registry.ts
```diff
@@ -15,12 +15,15 @@ export async function setupRootNode({ context }: { context: Context }) {
...
-  });
+  await context.db
+    .insert(schema.domain)
+    .values({
+      id: ROOT_NODE,
+      ownerId: zeroAddress,
+      createdAt: 0n,
+      isMigrated: false,
+    })
+    .onConflictDoNothing();
```
tk-o, line 26: There are no dependencies between plugins, but every plugin needs to make sure it has a domain for the ROOT_NODE created before starting the indexing. For example, if we activate three plugins, only one of them will create the domain for the root node, while remaining ones will just call a no-op.

src/lib/helpers.ts
```diff
@@ -16,3 +16,55 @@ export const blockConfig = (
...
+ * @param source The source object to merge from.
+ * @returns The merged object.
+ * @see https://stackoverflow.com/a/48218209
+ * @example
+ * const obj1 = { a: 1, b: 2, c: { d: 3 } };
+ * const obj2 = { a: 4, c: { e: 5 } };
+ * const obj3 = deepMergeRecursive(obj1, obj2);
+ * // { a: 4, b: 2, c: { d: 3, e: 5 } }
+ */
+export function deepMergeRecursive<T extends AnyObject, U extends AnyObject>(
```
tk-o, line 34: To be replaced with an NPM library.

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -14,48 +16,82 @@ const {
...
-    return await handleNameTransferred({
+    const { id, from, to } = event.args;
+
+    if (event.args.from === zeroAddress) {
+      // The ens-subgraph `handleNameTransferred` handler implementation
+      // assumes the domain record exists. However, when an NFT token is
+      // minted, there's no domain record yet created. The very first transfer
+      // event has to initialize the domain record. This is a workaround to
+      // meet the subgraph implementation expectations.
+      await handleDomainNameInitialized({
```
tk-o, line 75: @lightwalker-eth, I added context behind the need for initializing the domain name for the very first transfer event.

src/lib/helpers.ts
```diff
@@ -24,47 +25,10 @@ type AnyObject = { [key: string]: any };
...
-            : source[key];
-        }
-      } else {
-        Object.assign(output, { [key]: source[key] });
-      }
-    });
-  }
-
-  return output;
+  return tsDeepMerge(target, source) as T & U;
```
tk-o, line 33: @shrugs, here's my idea to make the `tsDeepMerge` work in runtime and not cause any issues with type-checks.

ponder.config.ts
tk-o: Done

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -14,48 +16,82 @@ const {
...
+  const label = tokenIdToLabel(id);
+  const node = makeSubnodeNamehash(ownedSubnameNode, label);
+  return context.db
+    .insert(schema.domain)
+    .values({
+      id: node,
+      ownerId: owner,
+      createdAt: event.block.timestamp,
+    })
+    .onConflictDoNothing();
```
lightwalker-eth, line 43: Someone reviewing our code could have concerns of potentially hidden issues when they view these `onConflictDoNothing` statements. I appreciate that there might be valid reasons for them. Could you please ensure any instance of these include a comment for why they exist? Thanks

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -14,48 +16,82 @@ const {
...
+}
+
 export default function () {
   // support NameRegisteredWithRecord for BaseRegistrar as it used by Base's RegistrarControllers
   ponder.on(pluginNamespace("BaseRegistrar:NameRegisteredWithRecord"), async ({ context, event }) =>
     handleNameRegistered({ context, event }),
   );
 
   ponder.on(pluginNamespace("BaseRegistrar:NameRegistered"), async ({ context, event }) => {
+    await upsertAccount(context, event.args.owner);
```
lightwalker-eth, line 53: Could you please document why we are calling `upsertAccount` here?

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -14,48 +16,82 @@ const {
...
 
-  // Base's BaseRegistrar uses `id` instead of `tokenId`
   ponder.on(pluginNamespace("BaseRegistrar:Transfer"), async ({ context, event }) => {
-    return await handleNameTransferred({
+    const { id, from, to } = event.args;
+
+    if (event.args.from === zeroAddress) {
+      // The ens-subgraph `handleNameTransferred` handler implementation
+      // assumes the domain record exists. However, when an NFT token is
+      // minted, there's no domain record yet created. The very first transfer
```
lightwalker-eth, line 72: These docs are great and super helpful!

One question:

"... there's no domain record yet created. ..."

Can you please add a bit more detail here on exactly where the domain record is not created yet? Is it in the subregistry or the shadow registry?

src/plugins/base.eth/handlers/Registrar.ts
lightwalker-eth: @tk-o Amazing! Super helpful 🚀

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -14,48 +16,82 @@ const {
   ownedSubnameNode,
 } = makeRegistrarHandlers(ownedName);
 
+/**
+ * Idempotent handler to insert a domain record when a new domain is
```
lightwalker-eth, line 20: Noting how you mention this function is "Idempotent". That's very helpful.

Perhaps we should rename this function then to better communicate what it is doing? For example: `ensureDomainExists`? Appreciate your suggestions and advice.

src/plugins/linea.eth/handlers/EthRegistrar.ts
```diff
@@ -8,21 +12,69 @@ const {
   handleNameRenewedByController,
   handleNameRenewed,
   handleNameTransferred,
+  ownedSubnameNode,
 } = makeRegistrarHandlers(ownedName);
```
lightwalker-eth, line 17: Please see all the related comments I shared in the base.eth version of this file :+1:

src/handlers/Registry.ts
lightwalker-eth: @tk-o Great comment. Can you please transfer this to be a comment in the code next to the `.onConflictDoNothing` perhaps? Or open to putting it somewhere else. The main point is we want this idea to live in our code rather than a PR comment.

lightwalker-eth (changes requested):
@tk-o Super updates :+1: Reviewed and shared feedback 🙌

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -14,48 +16,82 @@ const {
...
+ * initialized. For example, right after an NFT token for the domain
+ * is minted.
+ *
+ * @returns a newly created domain record
```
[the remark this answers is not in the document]
tk-o: Maybe I should say it's about database entity?

src/plugins/base.eth/handlers/Registrar.ts
tk-o: I will rename `handleDomainNameInitialized` into `ensureDomainExists` and will comment that calling `onConflictDoNothing()` allows the `ensureDomainExists` to be idempotent.

src/plugins/base.eth/handlers/Registrar.ts
tk-o: This is directly reflecting [the reference subgraph implementation](https://github.com/ensdomains/ens-subgraph/blob/master/src/ethRegistrar.ts#L41-L42). I've just moved it couple lines up, compared to the previous version of the `src/plugins/base.eth/handlers/Registrar.ts` file.

src/handlers/Registry.ts
tk-o: I think I'll use the `ensureDomainExists` function here to create a database entity for the ROOT_NODE idempotently. The `ensureDomainExists` will be documented to describe the idempotency mechanism (`onConflictDoNothing()` call)

.env.local.example
```diff
@@ -5,7 +5,8 @@ RPC_URL_1=https://ethereum-rpc.publicnode.com
 RPC_URL_8453=https://base-rpc.publicnode.com
 RPC_URL_59144=https://linea-rpc.publicnode.com
 
-# Identify which indexer plugin to activate (see `src/plugins` for available plugins)
+# Identify which indexer plugins to activate (see `src/plugins` for available plugins)
+# This is a comma separated list of one or more available plugin names.
 
 # ACTIVE_PLUGINS=eth,base.eth,linea.eth
 # ACTIVE_PLUGINS=base.eth,linea.eth
```
lightwalker-eth, line 12: Perhaps by default we should activate all our plugins in the .env.local.example file? Currently we're only activating `eth` by default.

src/lib/upserts.ts
```diff
@@ -19,3 +19,25 @@ export async function upsertRegistration(
...
+ * @returns domain database entity
+ */
+export async function ensureDomainExists(
+  context: Context,
+  values: typeof schema.domain.$inferInsert,
+): Promise<typeof schema.domain.$inferSelect> {
+  const domainEntity = await context.db.insert(schema.domain).values(values).onConflictDoNothing();
+
+  if (!domainEntity) {
+    throw new Error("domain expected");
```
lightwalker-eth, line 39: Could we please document what's going on here in more detail and how it relates to `onConflictDoNothing`?

src/lib/upserts.ts
```diff
@@ -19,3 +19,25 @@ export async function upsertRegistration(
 ) {
   return context.db.insert(schema.registration).values(values).onConflictDoUpdate(values);
 }
+
+/**
+ * Idempotent handler to ensure a domain entity for requested node exists in
+ * the database. It inserts a domain entity if it does not exist. Otherwise,
```
lightwalker-eth, line 25:
```suggestion
 * the database. It inserts the provided domain entity if it does not exist. Otherwise,
```

src/lib/upserts.ts
```diff
@@ -19,3 +19,25 @@ export async function upsertRegistration(
 ) {
   return context.db.insert(schema.registration).values(values).onConflictDoUpdate(values);
 }
+
+/**
+ * Idempotent handler to ensure a domain entity for requested node exists in
```
lightwalker-eth, line 24: I don't think it's right for us to call this function "Idempotent" since multiple calls to this function with the same input might produce different output if there is an existing record with state different than the provided value.

src/lib/upserts.ts
```diff
@@ -19,3 +19,25 @@ export async function upsertRegistration(
 ) {
   return context.db.insert(schema.registration).values(values).onConflictDoUpdate(values);
 }
+
```
lightwalker-eth, line 22: Suggest renaming this file to something more general since it contains operations that technically aren't upserts.

src/handlers/Registry.ts
```diff
@@ -15,15 +15,12 @@ export async function setupRootNode({ context }: { context: Context }) {
...
-  await context.db
-    .insert(schema.domain)
-    .values({
-      id: ROOT_NODE,
-      ownerId: zeroAddress,
-      createdAt: 0n,
-      isMigrated: false,
-    })
-    .onConflictDoNothing();
+  await ensureDomainExists(context, {
```
lightwalker-eth, line 18: For each of our plugins that calls `ensureDomainExists` in their respective Registry, can you please document why we are calling `ensureDomainExists`? Ex: any permutation of plugins might be activated and multiple plugins expect the ENS root to exist.

We need to clarify in detail why we are doing something special here that's different to the existing ENS subgraph code.

Also, is there something we should say here about the order that Ponder executes this function across plugins if we have multiple plugins activated? For example, if we activate all 3 of our plugins, in what order would Ponder call this function across all 3 plugins? Is it undefined or is it predictable?

Appreciate your advice.

src/handlers/Registry.ts
```diff
@@ -104,7 +101,7 @@ export const handleNewOwner =
       await context.db.update(schema.domain, { id: domain.id }).set({ ownerId: owner, isMigrated });
     } else {
       // otherwise create the domain
-      domain = await context.db.insert(schema.domain).values({
+      domain = await ensureDomainExists(context, {
```
lightwalker-eth, line 104: Not sure. Should this stay as an insert based on the overall context of this function? Appreciate your advice.

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -56,35 +56,39 @@ export default function () {
     // any Registrar:NameRegistered events. in the future we will likely happily upsert domains, but
     // in order to avoid prematurely drifting from subgraph equivalancy, we upsert the domain here,
```
lightwalker-eth, line 57: My understanding is that `ensureDomainExists` does not implement an "upsert"

lightwalker-eth (changes requested):
@tk-o Thanks for updates :+1: Reviewed and shared a few more points of feedback.

src/handlers/Registry.ts
tk-o: `ensureDomainExists` is effectively the insert operation with conflict management. The current `handleNewOwner` handler will only attempt inserting a new domain entity into database if the entity was not present. The conflict is impossible here. In this case, we can come back to the previous version of the code, if that seems clearer to read.

src/lib/upserts.ts
tk-o: I can extract the `ensureDomainExists` function to another file. All other handles are exactly upserts, IMO.

src/handlers/Registry.ts
tk-o: Updated the comment on top of the `setupRootNode ` function .

src/plugins/base.eth/handlers/Registrar.ts
tk-o: You are right, the `ensureDomainExists` do not allow updating the existing database entity. It only takes effect, by inserting a database row into the `domains` table,  when there is no database entity for the select `id` value (which is node value). Otherwise, it's a no-op.

shrugs:
just to chime in, i think the original inline `.onConflictDoNothing()` is acceptable without abstracting it into an `ensure` or `upsert` function — that's very normal/standard ponder behavior and is well documented within ponder. i agree that a comment is in order, to describe the fact that this `setupRootNode` is called multiple times and needs to be idempotent. it could be renamed `ensureRootNode` or `setup_ensureRootNode` to make that even clearer

src/lib/upserts.ts
lightwalker-eth: Not sure what the goal for this file is. Is the goal to collect various data access layer utility functions? If so maybe  rename the file to "dal-helpers.ts" or "db-helpers.ts" or something like that.

src/handlers/Registry.ts
```diff
@@ -9,6 +9,24 @@ import { ensureDomainExists, upsertAccount } from "../lib/upserts";
 
 /**
  * Initialize the ENS root node with the zeroAddress as the owner.
+ * Any permutation of plugins might be activated and multiple plugins expect
```
lightwalker-eth, line 12: @tk-o Thanks for expanding the docs here! 🙌

src/lib/upserts.ts
```diff
@@ -21,23 +21,20 @@ export async function upsertRegistration(
 }
 
 /**
- * Idempotent handler to ensure a domain entity for requested node exists in
- * the database. It inserts a domain entity if it does not exist. Otherwise,
- * just returns the existing domain entity from the db.
+ * A function to ensure a domain entity for requested node value exists in
```
lightwalker-eth, line 24:
```suggestion
 * Ensure that some domain entity value (not necessarily the provided value) for the requested node exists in
```

Carefully rewording to reduce potential confusion that the node value in the database will be the node value provided.

lightwalker-eth (changes requested):
@tk-o Thanks for updates :+1: Reviewed and shared feedback

.env.local.example
```diff
@@ -1,19 +1,35 @@
...
+RPC_REQUEST_RATE_LIMIT_1=50
+RPC_REQUEST_RATE_LIMIT_8453=20
+RPC_REQUEST_RATE_LIMIT_59144=20
 
-# Identify which indexer plugin to activate (see `src/plugins` for available plugins)
+# Identify which indexer plugins to activate (see `src/plugins` for available plugins)
+# This is a comma separated list of one or more available plugin names.
 
-ACTIVE_PLUGIN=base.eth
+ACTIVE_PLUGINS=base.eth,linea.eth
```
tk-o, line 15:
```suggestion
ACTIVE_PLUGINS=eth,base.eth,linea.eth
```

.env.local.example
```diff
@@ -1,19 +1,33 @@
...
-RPC_URL_8453=https://base-rpc.publicnode.com
-RPC_URL_59144=https://linea-rpc.publicnode.com
-
-# Identify which indexer plugin to activate (see `src/plugins` for available plugins)
-
-ACTIVE_PLUGIN=base.eth
+# For the RPC URL of each chain, follow the format: RPC_URL_{chainId}={rpcUrl}
+RPC_URL_1=https://eth.drpc.org
+RPC_URL_8453=https://base.drpc.org
+RPC_URL_59144=https://linea.drpc.org
```
tk-o, line 5: @shrugs @lightwalker-eth we now have the dRPC URLs as defaults for public RPC nodes

src/lib/db-helpers.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Otherwise, just returns the existing domain entity from the db.
+ *
+ * @param context ponder context to interact with database
+ * @param values domain properties, where `id` is the node value
+ * @returns domain database entity
+ */
+export async function ensureDomainExists(
+  context: Context,
+  values: typeof schema.domain.$inferInsert,
+): Promise<typeof schema.domain.$inferSelect> {
```
shrugs, line 35: return type can be inferred

src/lib/helpers.ts
```diff
@@ -16,3 +17,53 @@ export const blockConfig = (
...
+ * to the RPC endpoint. For public RPC endpoints, it is recommended to set
+ * a rate limit to low values (i.e. below 30 rps) to avoid being rate limited.
+ * For private RPC endpoints, the rate limit can be set to higher values,
+ * depending on the capacity of the endpoint. For example, 500 rps.
+ *
+ * @param chainId the chain ID to read the rate limit for from the environment variable
+ * @returns the rate limit in requests per second (rps)
+ */
+export const rpcRequestRateLimit = (chainId: number): number => {
+  if (typeof process.env[`RPC_REQUEST_RATE_LIMIT_${chainId}`] === "string") {
```
shrugs, line 43: nitpick: extract this to `varName` or something

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -21,41 +23,52 @@ export default function () {
...
-    await upsertAccount(context, owner);
-    await context.db
-      .insert(schema.domain)
-      .values({
-        id: node,
-        ownerId: owner,
-        createdAt: event.block.timestamp,
-      })
-      .onConflictDoNothing();
+    await ensureDomainExists(context, {
```
shrugs, line 32: ahhhh i'm SO sorry but i'd REALLY like to avoid refactoring all of these situations — i explicitly did not create an `upsertDomain` function because i wanted the logic in the handlers to be as explicit and subgraph-like as possible. can we please revert all of this refactor? this is also why i was against the `ensureDomainExists` helper originally.

@lightwalker-eth i know this likely derived from your suggestions, so FYI. i don't think this level of abstraction is necessary and makes the code more complicated to follow

shrugs: if you find it necessary you can use `ensureDomainExists` for the root node setup and the plugin-specific handlers, but please do not refactor any of the shared handler logic for the moment

shrugs (changes requested):

src/plugins/base.eth/handlers/Registrar.ts
lightwalker-eth: @shrugs Thanks for feedback 🙌  I missed that this was going to change anything related to our "eth" plugin, which includes the shared handler code. Agreed that we want to avoid making changes to the logic there.

@tk-o Sorry for any feedback from me that might have sent this in the wrong direction. My mistake.

I believe it can be ok to selectively add refined logic if it only impacts the base.eth or linea.eth plugins.

Also, adding additional comments to the "eth" plugin related logic is fantastic. So perhaps if there's relevant comments to retain there, please consider it.

Thanks guys :+1:

src/lib/helpers.ts
tk-o: Not sure what's expected. Could you post a suggestion, please?

src/lib/helpers.ts
```diff
@@ -16,3 +17,53 @@ export const blockConfig = (
   startBlock: Math.min(Math.max(start || 0, startBlock), end || Number.MAX_SAFE_INTEGER),
   endBlock: end,
 });
+
+// default rate limit for request per second to RPC endpoints
```
lightwalker-eth, line 21:
```suggestion
// default request per second rate limit for RPC endpoints
```

src/lib/helpers.ts
```diff
@@ -16,3 +17,53 @@ export const blockConfig = (
...
+const DEFAULT_RPC_RATE_LIMIT = 50;
+
+/**
+ * Reads the RPC request rate limit for a given chain ID from the environment
```
lightwalker-eth, line 25:
```suggestion
 * Gets the RPC request rate limit for a given chain ID.
```

Then suggest moving or removing the remainder of this first paragraph.

The comments here are for the overall function. Generally it's best not to document the implementation details of the function here. These comments on the outside of the function should just be for the callers of this function and those callers generally shouldn't be concerned with purely internal implementation details. Ex: that the value is read from environment variables.

You're welcome to move this paragraph of comments to the first line of code inside the function. These details are specific to the internal implementation.

The other paragraphs here are great to keep.

src/lib/helpers.ts
```diff
@@ -16,3 +17,53 @@ export const blockConfig = (
...
+ * variable is not set for the requested chain ID, use `DEFAULT_RPC_RATE_LIMIT`
+ * as the default value.
+ *
+ * The rate limit is the maximum number of requests per second that can be made
+ * to the RPC endpoint. For public RPC endpoints, it is recommended to set
+ * a rate limit to low values (i.e. below 30 rps) to avoid being rate limited.
+ * For private RPC endpoints, the rate limit can be set to higher values,
+ * depending on the capacity of the endpoint. For example, 500 rps.
+ *
+ * @param chainId the chain ID to read the rate limit for from the environment variable
```
lightwalker-eth, line 39:
```suggestion
 * @param chainId the chain ID to get the rate limit for
```

Please see comment above for why suggesting this.

src/lib/helpers.ts
```diff
@@ -16,3 +17,53 @@ export const blockConfig = (
...
+/**
+ * Reads the RPC request rate limit for a given chain ID from the environment
+ * variable: RPC_REQUEST_RATE_LIMIT_{chainId}.
+ * For example, for Ethereum mainnet the chainId is `1`, so the env variable
+ * can be set as `RPC_REQUEST_RATE_LIMIT_1=400`. This will set the rate limit
+ * for the mainnet (chainId=1) to 400 requests per second. If the environment
+ * variable is not set for the requested chain ID, use `DEFAULT_RPC_RATE_LIMIT`
+ * as the default value.
+ *
+ * The rate limit is the maximum number of requests per second that can be made
```
lightwalker-eth, line 33: Suggest moving the ideas in this paragraph into the relevant location of the .env.local.example file.

These are nice ideas and are best for people to read where they set those environment variables. Many more people will read that example .env file than will read this comment here.

src/plugins/README.md
```diff
@@ -1,8 +1,8 @@
 # Indexer plugins
 
-This directory contains plugins which allow defining subname-specific processing of blockchain events.
-Only one plugin can be active at a time. Use the `ACTIVE_PLUGIN` env variable to select the active plugin, for example:
+This directory contains plugins which define subname-specific processing of blockchain events.
```
lightwalker-eth, line 3:
```suggestion
This directory contains plugins that define subname-specific processing of blockchain events.
```

src/plugins/base.eth/ponder.config.ts
```diff
@@ -13,19 +14,25 @@ export const ownedName = "base.eth" as const;
 
 export const pluginNamespace = createPluginNamespace(ownedName);
 
+// constrain the ponder indexing between the following start/end blocks
```
lightwalker-eth, line 17:
```suggestion
// constrain indexing between the following start/end blocks
```

src/plugins/linea.eth/ponder.config.ts
```diff
@@ -24,6 +24,7 @@ export const config = createConfig({
     linea: {
       chainId: linea.id,
       transport: http(process.env[`RPC_URL_${linea.id}`]),
```
lightwalker-eth, line 26: Suggest adding a helper function for this that all our plugins could use.

The helper function would get the RPC URL for a given chainId and work similar to `rpcRequestRateLimit`.

Importantly: If no environment variable is set for a RPC URL that we expect this utility function should throw a helpful error message that the expected environment variable is missing.

lightwalker-eth (changes requested):
@tk-o Great updates. Made another careful review. Here's a few more suggestions :+1:

src/lib/helpers.ts
shrugs: ah sorry. it's really not a big deal. after looking at the code further i might write this function like

```ts
const DEFAULT_RPC_RATE_LIMIT = 50; // or "50" if parseInt strictly requires a string

export const rpcRequestRateLimit = (chainId: number): number => {
  const envVarName = `RPC_REQUEST_RATE_LIMIT_${chainId}`;
  try {
    return parseInt(process.env[envVarName] || DEFAULT_RPC_RATE_LIMIT, 10);
  } catch (e) {
    throw new Error(
      `Invalid ${envVarName} value: ${e}. Please provide a valid number.`,
    );
  }
};
```

this way if the var doesn't exist or is empty string the `||` operator uses the default instead. and if parsing fails we throw the error as before.

but like i said it's minor thing, don't feel that you must accept this suggestion — only if you prefer this as well

lightwalker-eth (changes requested):
@tk-o Hey nice updates, thanks. Just very minor nitpicks :+1:

src/handlers/Registry.ts
```diff
@@ -42,7 +42,7 @@ export async function setupRootNode({ context }: { context: Context }) {
       createdAt: 0n,
       isMigrated: false,
     })
-    // make sure to only insert the domain entity into database
+    // only insert the domain entity into the database if it doesn't already exist
     // only if it doesn't already exist
```
lightwalker-eth, line 46: This line should be removed

src/lib/helpers.ts
```diff
@@ -63,19 +68,30 @@ export const rpcRequestRateLimit = (chainId: number): number => {
...
     return DEFAULT_RPC_RATE_LIMIT;
   }
 
   // otherwise
   try {
     // parse the rate limit value from the environment variable
-    return parseInt(process.env[envVarName], 10);
+    const rpcRequestRateLimit = parseInt(envVarValue, 10);
+    
+    if (Number.isNaN(rpcRequestRateLimit)) {
```
lightwalker-eth, line 84: Nice thanks for adding the `isNan` check. Please also add a check for <= 0 since the value must be positive.

lightwalker-eth (changes requested):
@tk-o Thanks for updates. I made one last end-to-end review of the PR (rather than just reviewing the incremental changes since a prior review).

Let's do a great job on all the details. Thanks!

src/plugins/linea.eth/handlers/EthRegistrar.ts
```diff
@@ -8,21 +11,42 @@ const {
...
 
   ponder.on(pluginNamespace("BaseRegistrar:Transfer"), async ({ context, event }) => {
-    return await handleNameTransferred({ context, args: event.args });
+    const { tokenId, from, to } = event.args;
+
+    if (event.args.from === zeroAddress) {
+      // The ens-subgraph `handleNameTransferred` handler implementation
+      // assumes the domain record exists. However, when an NFT token is
+      // minted, there's no domain entity in the database yet. The very first
+      // transfer event has to ensure the domain entity for the requested
```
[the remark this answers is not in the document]
tk-o: A minting event is the transfer event with the from value set to `0x0`

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -21,41 +22,52 @@ export default function () {
   );
 
   ponder.on(pluginNamespace("BaseRegistrar:NameRegistered"), async ({ context, event }) => {
+    await upsertAccount(context, event.args.owner);
     // base has 'preminted' names via Registrar#registerOnly, which explicitly does not update Registry.
     // this breaks a subgraph assumption, as it expects a domain to exist (via Registry:NewOwner) before
     // any Registrar:NameRegistered events. in the future we will likely happily upsert domains, but
```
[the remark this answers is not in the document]
tk-o: Not sure about the exact wording here, but can just drop that extra sentence.

src/plugins/linea.eth/handlers/EthRegistrar.ts
```diff
@@ -23,12 +23,12 @@ export default function () {
 
     if (event.args.from === zeroAddress) {
       // The ens-subgraph `handleNameTransferred` handler implementation
-      // assumes the domain record exists. However, when an NFT token is
-      // minted, there's no domain entity in the database yet. The very first
-      // transfer event has to ensure the domain entity for the requested
-      // token ID has been inserted into the database. This is a workaround to
-      // meet expectations of the `handleNameTransferred` subgraph
-      // implementation.
+      // assumes an indexed record for the domain already exists. However,
```
lightwalker-eth, line 26: @tk-o Nice update here! Helps a lot!

src/plugins/base.eth/handlers/Registrar.ts
lightwalker-eth: @tk-o This is still an open question for me. Can you please clarify?

src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -23,10 +23,10 @@ export default function () {
 
   ponder.on(pluginNamespace("BaseRegistrar:NameRegistered"), async ({ context, event }) => {
     await upsertAccount(context, event.args.owner);
-    // base has 'preminted' names via Registrar#registerOnly, which explicitly does not update Registry.
-    // this breaks a subgraph assumption, as it expects a domain to exist (via Registry:NewOwner) before
-    // any Registrar:NameRegistered events. in the future we will likely happily upsert domains, but
-    // in order to avoid prematurely drifting from subgraph equivalancy, we insert the domain entity here,
+    // Base has 'preminted' names via Registrar#registerOnly, which explicitly
```
lightwalker-eth, line 26: Thanks for updating the docs here. Looks good :+1:

When base performs these 'premint' operations, are there any related events emitted? In other words, is this the best place to put this logic, or should we investigate moving it into a different event handler?

If so, assuming this needs more research, could you please create an Issue in GitHub and link back to this comment for it? Happy to not grow this PR more :+1:

src/lib/plugin-helpers.ts
```diff
@@ -79,28 +79,29 @@ type PluginNamespacePath<T extends PluginNamespacePath = "/"> =
...
-    throw new Error("No active plugins found. Please set the ACTIVE_PLUGINS environment variable.");
+  if (!requestedPlugins.length) {
+    throw new Error("Set the ACTIVE_PLUGINS environment variable to activate one or more plugins.");
   }
 
   // Check if the requested plugins are valid and can become active
-  const invalidPlugins = pluginsToActivateByOwnedName.filter(
-    (plugin) => !plugins.some((p) => p.ownedName === plugin),
+  const invalidPlugins = requestedPlugins.filter(
+    (plugin) => !availablePlugins.some((p) => p.ownedName === plugin),
```
lightwalker-eth, line 104:
```suggestion
    (requestedPlugin) => !availablePlugins.some((availablePlugin) => availablePlugin.ownedName === requestedPlugin),
```

ponder.config.ts
```diff
@@ -1,32 +1,32 @@
...
-  const config = activePlugins
+  const activePluginsConfig = activePlugins
     .map((plugin) => plugin.config)
     .reduce((acc, val) => deepMergeRecursive(acc, val), {} as AllPluginConfigs);
 
-  return config as AllPluginConfigs;
+  return activePluginsConfig as AllPluginConfigs;
 }
 
 // The type of the default export is the intersection of all active plugin configs
```
lightwalker-eth, line 30:
```suggestion
// The type of the default export is a merge of all active plugin configs
```

Please also see my comment above with feedback on the distinction of static vs dynamic types. Whatever the answer is there, appreciate if you can align this comment here as well.

ponder.config.ts
```diff
@@ -1,32 +1,32 @@
...
+// any available plugin can be activated at runtime
+const availablePlugins = [baseEthPlugin, ethPlugin, lineaEthPlugin] as const;
 
-// intersection of all available plugin configs to support correct typechecking
+// merge of all available plugin configs to support correct typechecking
 // of the indexing handlers
-type AllPluginConfigs = IntersectionOf<(typeof plugins)[number]["config"]>;
+type AllPluginConfigs = MergedTypes<(typeof availablePlugins)[number]["config"]>;
 
 // Activates the indexing handlers of activated plugins and
```
lightwalker-eth, line 15:
```suggestion
// Activates the indexing handlers of activated plugins.
```

ponder.config.ts
```diff
@@ -1,32 +1,32 @@
...
 // returns the intersection of their combined config.
```
lightwalker-eth, line 16:
```suggestion
// Statically typed as the merge of all available plugin configs. However at runtime returns the merge of all activated plugin configs.
```

I'm not sure if I'm understanding this correctly. Is what I wrote above correct? Appreciate if you could refine if needed :+1:

lightwalker-eth (changes requested):
@tk-o Very close. Getting there 😄 A few more feedbacks.

src/plugins/base.eth/handlers/Registrar.ts
tk-o: @shrugs, please confirm:

When [we insert a record into `domains` table](https://github.com/namehash/ensnode/pull/24/files#diff-5cc50dbf0d6ef2a9d7148562b56e141dc7440cddfe373882a411e9b41b023a1fR33), we set its `ownerId` to some value (as it is `notNull` column). Because of how we have the ponder schema relations defined, the `ownerId` must have a reference in the `accounts` table.

That's why we first insert a record into `accounts` table, followed by a record added into `domains` table. The later must point to the former.

lightwalker-eth (approved):
@tk-o All now looks good to me 🚀 @shrugs we created a separate follow up question for you here that we can follow up on outside of this PR if you like: https://github.com/namehash/ensnode/issues/34

@shrugs Please take the lead on reviewing this PR and merging it in when it looks good to you :+1:

src/plugins/base.eth/handlers/Registrar.ts
shrugs: correct.

side note: ponder doesn't enforce foreign key constraints in the postgres schema, so _technically_ either order works but we should prefer the order you've suggested here, since domain records _should_ depend on account record existing.

shrugs (approved):
LGTM, merging

PR #26, opened 2025-01-14
Subgraph-compatible Autogenerated API

shrugs:
1. copy/paste original ponder schema generation code & hono middleware
2. update that code to be subgraph compatible
3. hosts original ponder api at `/` and new subgraph api at `/subgraph`

also includes some fixes of the eth handlers based on snapshot tool discoveries — have not tested against base/linea

1. resolvedAddress is Account, not string literal
2. labelhash
4. lowercase address in resolverId
5. default null array types
6. (in)correctly implements root node migration state
7. `Resolver#VersionChanged` needs to use upsert pattern

if you're reviewing the code, i suggest commit-by-commit, as the commit history here is relatively orderly


ponder.schema.ts
```diff
@@ -45,7 +45,10 @@ export const domain = onchainTable("domains", (t) => ({
 }));
 
 export const domainRelations = relations(domain, ({ one, many }) => ({
-  // has one owner
+  resolvedAddress: one(account, {
+    fields: [domain.resolvedAddressId],
+    references: [account.id],
+  }),
```
tk-o, line 51: Nice catch

src/plugins/eth/ponder.config.ts
```diff
@@ -57,7 +61,7 @@ export const config = createConfig({
       network: "mainnet",
       abi: RESOLVER_ABI,
       address: factory({
-        address: "0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e",
+        address: REGISTRY_ADDRESS,
```
tk-o: I've checked the `linea.eth` and the `base.eth` plugins, and each of them uses the registry address for the resolver contract factory method. Which matches what you did here as well 👍

src/plugins/eth/handlers/Registry.ts
```diff
@@ -49,6 +47,12 @@ export default function () {
   });
 
   ponder.on(pluginNamespace("RegistryOld:Transfer"), async ({ context, event }) => {
+    // NOTE: this logic derived from the subgraph introduces a bug for queries with a blockheight
+    // below 9380380, when the new Registry was deployed, as it implicitly ignores Transfer events
+    // of the ROOT_NODE. as a result, the root node's owner is always zeroAddress until the new
+    // Registry events are picked up. for backwards compatibility this beahvior is re-implemented
+    // here.
```
tk-o, line 54: 👍

src/api/middleware.ts
```diff
@@ -45,7 +51,7 @@ export const graphql = (
   return createMiddleware(async (c) => {
     if (yoga === undefined) {
       const metadataStore = c.get("metadataStore");
-      const graphqlSchema = c.get("graphqlSchema");
+      const graphqlSchema = buildGraphQLSchema(schema);
```
tk-o, line 54: Very nice

src/handlers/Registry.ts
```diff
@@ -182,12 +184,14 @@ export async function handleNewResolver({
         domainId: event.args.node,
         address: event.args.resolver,
       })
-      .onConflictDoNothing();
+      .onConflictDoUpdate({}); // noop update to return the existing record
```
tk-o, line 187: I believe that `onConflictDoNothing()` also returns the existing record:
- https://ponder.sh/docs/indexing/write-to-the-database#upsert--conflict-resolution

Perhaps `onConflictDoNothing()` and `onConflictDoUpdate({})` have the same effect?

src/plugins/base.eth/ponder.config.ts
```diff
@@ -30,12 +30,24 @@ export const config = createConfig({
     [pluginNamespace("Resolver")]: {
       network: "base",
       abi: L2Resolver,
-      address: factory({
-        address: "0xb94704422c2a1e396835a571837aa5ae53285a95",
-        event: getAbiItem({ abi: Registry, name: "NewResolver" }),
-        parameter: "resolver",
-      }),
-      startBlock: 17575714,
+      // NOTE: this indexes every event ever emitted that looks like this
```
tk-o, line 33: 👍

tk-o:
We might to tweak the gql schema generation for sorting params. The subgraph use specific types for order definition:
```
domains(
  skip: Int = 0
  first: Int = 100
  orderBy: Domain_orderBy
  orderDirection: OrderDirection
  where: Domain_filter
  block: Block_height
  subgraphError: _SubgraphErrorPolicy_! = deny
): [Domain!]!
```

Our schema uses `String` type instead.

src/plugins/eth/handlers/Registry.ts
lightwalker-eth: @shrugs Great docs 👍

src/handlers/Registry.ts
```diff
@@ -19,7 +19,7 @@ export async function setupRootNode({ context }: { context: Context }) {
     id: ROOT_NODE,
     ownerId: zeroAddress,
     createdAt: 0n,
-    isMigrated: false,
+    isMigrated: true,
```
lightwalker-eth, line 22: Interesting. What about indexed state between the time of the old registry being deployed and the new registry being deployed?

lightwalker-eth (commented):
@shrugs Awesome update! 🚀 Direction looks great.

src/handlers/Registry.ts
tk-o: Ok, I've checked with the Ponder crew. If we need to:
- insert a DB entity if it does not exists, AND
- do nothing with it when it already exists, AND
- return the DB entity

Then we need `onConflictUpdate({})`. [More context on that issue](https://t.me/ponder_sh/9841):
> on conflict do nothing does not return anything when nothing is done. this is how Postgres does it. it is certainly a bit unintuitive. see this discussion for some further information on this functionality in postgres and workarounds https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql

src/handlers/Registry.ts
shrugs: no you're correct, it's a bug in the subgraph, but a relatively minor one

shrugs: i documented this behavior in the Registry handler function, but i'll add another comment here of the same nature

src/api/index.ts
```diff
@@ -1,6 +1,10 @@
...
+import { graphql as ponderGraphQL } from "ponder";
+import { graphql as subgraphGraphQL } from "./middleware";
 
-// use our custom graphql middleware
-ponder.use("/", graphql());
+// use ponder middleware at root
+ponder.use("/", ponderGraphQL());
+
+// use our custom graphql middleware at /subgraph
+ponder.use("/subgraph", subgraphGraphQL());
```
tk-o, line 10: Awesome plugin :)

src/handlers/Registrar.ts
```diff
@@ -18,7 +18,9 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
 
     const node = makeSubnodeNamehash(ownedSubnameNode, label);
     const domain = await context.db.find(schema.domain, { id: node });
-    if (!domain) throw new Error("domain expected");
+
+    // encode the runtime assertion here https://github.com/ensdomains/ens-subgraph/blob/master/src/ethRegistrar.ts#L101
+    if (!domain) throw new Error("domain expected in setNamePreimage but not found");
```
tk-o, line 23: 👍

docs/GRAPHQL.md
```diff
@@ -1,5 +1,14 @@
 # graphql info/spec
 
+## backwards-compatibility notes
+
+the following features of the subgraph graphql api are explicitly unsupported, because they are not used by ensjs or ens-app-v3
+
+- `_nocase` case-insensitive where filters
+- [fulltext search queries](https://thegraph.com/docs/en/subgraphs/querying/graphql-api/#fulltext-search-queries)
+- 1-level-nested Entity `_orderBy` param
+- subgraph `_Meta_` object (ponder's `_meta` is available)
```
tk-o, line 9: Thank you for an explicit list of API elements not utilised by the clients we seek to support.

src/api/graphql.ts
```diff
@@ -0,0 +1,747 @@
+/**
```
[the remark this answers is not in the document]
shrugs: oh this file is not being autogenerated, it's for the auto-generated graphql api that ponder serves... what wording do you think would be best?

docs/GRAPHQL.md
```diff
@@ -1,5 +1,18 @@
 # graphql info/spec
 
+## backwards-compatibility notes
```
lightwalker-eth, line 3: @shrugs Great job with the docs here 💪 

Should we also include time-travel queries? https://thegraph.com/docs/en/subgraphs/querying/graphql-api/#time-travel-queries

docs/V2.md
```diff
@@ -41,6 +41,8 @@ the 'empty' domains should be handled more accurately, depending on how importan
   - removes need to increment during domain creation
   - new impl likely needs to exclude 'empty' domains (see registry notes for context)
 
+domain createdAt should not update on re-registration, should be original createdAt
```
lightwalker-eth, line 44: So happy to see you documenting these opportunities :+1:

src/handlers/Resolver.ts
```diff
@@ -6,6 +6,12 @@ import { upsertAccount, upsertResolver } from "../lib/db-helpers";
 import { hasNullByte, uniq } from "../lib/helpers";
 import { makeResolverId } from "../lib/ids";
 
+// NOTE: both subgraph and this indexer us upserts in this file because a 'Resolver' is _any_
```
lightwalker-eth, line 9:
```suggestion
// NOTE: both subgraph and this indexer use upserts in this file because a 'Resolver' is _any_
```

src/handlers/Resolver.ts
```diff
@@ -6,6 +6,12 @@ import { upsertAccount, upsertResolver } from "../lib/db-helpers";
...
+// contract on the chain that emits an event with this signature, which may or may not actually be
+// an ENS-specific Resolver. because of this each even could theoretically be the first event the
```
lightwalker-eth, line 11: My understanding is that any contract with the signature of what we define as a "Resolver" should be considered to be an "ENS-specific Resolver".

My interpretation of the current sentence here is it seems to suggest that we might be indexing "Resolvers" but for a different name service that's not ENS? Suggest we find another way to write this. Happy to discuss on Slack :+1:

lightwalker-eth, line 11:
```suggestion
// an ENS-specific Resolver. Because of this each event could theoretically be the first event the
```

src/handlers/Resolver.ts
```diff
@@ -6,6 +6,12 @@ import { upsertAccount, upsertResolver } from "../lib/db-helpers";
...
+// indexer has seen for a given Resolver id and therefore needs to use an upsert and not assume
```
lightwalker-eth, line 12: Suggest to remove everything here after "... and therefore needs to use an upsert." It seems like too strong of a statement to say that an indexer needs to not assume anything else about this Resolver's state. For example: The indexer might already have other knowledge about this Resolver's state from previously indexed events, etc.

src/handlers/Resolver.ts
```diff
@@ -245,15 +251,18 @@ export async function handleVersionChanged({
   const { node } = event.args;
   const id = makeResolverId(event.log.address, node);
   const domain = await context.db.find(schema.domain, { id: node });
-  if (!domain) throw new Error("domain expected");
 
-  // materialize the Domain's resolvedAddress field
-  if (domain.resolverId === id) {
+  // materialize the Domain's resolvedAddress field iff exists and is set to this Resolver
```
lightwalker-eth, line 255:
```suggestion
  // clear the Domain's resolvedAddress field iff the domain is already indexed and the domain's active Resolver is the Resolver emitting this event
```

There's more to the logic here, but cautious if I'm interpreting the next ideas correctly:

The key issue for improvement here is how we're documenting "Domain" here and how it relates to the Resolver that is emitting this event.

For example, if we review how `id` and `domain` are determined as a function of the `event`, it looks like this domain has a relation to the resolver emitting the event.

Trying to get this PR feedback sent to you quickly so won't investigate further right now.

lightwalker-eth (approved):
@shrugs Hey super work here :+1: Shared a few small suggestions / comments with you. You're welcome to merge this in whenever you're ready 🚀

lightwalker-eth (commented):
@shrugs One more feedback based on your latest commits

PR #33, opened 2025-01-17
WIP Event Support

shrugs:
- schema
- handlers
- api


shrugs:
as part of this i would like to rename the indexer app to `indexer` or something, since it alone is no longer 'ensnode', wdyt?

PR #35, opened 2025-01-20
fix(plugin-helpers): plugin activation

tk-o:
Check all constraints before activiating a given requested plugin. Before, all requested plugins would get activated, which was an uninteded behaviour.


lightwalker-eth (approved):
@tk-o Looks good :+1: Thanks

Issue #15, opened 2025-01-09
Setup testing suite


tk-o:
This issue will be done once there's a testing suite allowing to test the libraries we create for subnames indexing. First, we'd need unit tests to ensure the helpers/utils functions keep working as intented.

lightwalker-eth:
@tk-o This seems like a high value priority :+1:

lightwalker-eth:
We should define some unit tests for `decodeDNSPacketBytes` inside src/lib/subname-helpers.ts.

PR #38, opened 2025-01-20
Subgraph Logic Fixes Post 21M Diff

### LINKED ISSUE #36: issues surfaced by diff tool at block 21M
- [x] resolver texts/coinTypes being empty array when they should be null
- [x] ponder's Registrations.cost for `0x3581397a478dcebdc1ee778deed625697f624c6f7dbed8bb7f780a6ac094b772` is `290823255294308` when the subgraph says it should be `null` (??)
- [x] ponder's `wrappedDomain.name` for `0x19a2c81ceaecdd40aa03e760f30aa3f087a534f87c68038cc518742e4949131d` is `8436.eth.eth` when it should be `null` — definitely something odd going on with dns decoding, why two .eth tlds...
- [ ] #39
- [x] ponder's Domain.expiryDate for `0x3dc9be0027174c741f04f692470874ec1ea26fc044af101cde7d4c6c350d799f` is null when it should be `1960355140`

shrugs:
closes #36

### DIFF
--- a/src/handlers/NameWrapper.ts
+++ b/src/handlers/NameWrapper.ts
@@ -7,7 +7,7 @@ import { bigintMax } from "../lib/helpers";
 import { makeEventId } from "../lib/ids";
 import { decodeDNSPacketBytes, tokenIdToLabel } from "../lib/subname-helpers";
 
-// if the wrappedDomain in question has pcc burned (?) and a higher (?) expiry date, update the domain's expiryDate
+// if the wrappedDomain has PCC set in fuses, set domain's expiryDate to the greatest of the two
 async function materializeDomainExpiryDate(context: Context, node: Hex) {
   const wrappedDomain = await context.db.find(schema.wrappedDomain, { id: node });
   if (!wrappedDomain) throw new Error(`Expected WrappedDomain(${node})`);
@@ -18,7 +18,7 @@ async function materializeDomainExpiryDate(context: Context, node: Hex) {
   // make sure to remember that if you compare the logic in this function to the original subgraph logic [here](https://github.com/ensdomains/ens-subgraph/blob/master/src/nameWrapper.ts#L87)
   // related GitHub issue: https://github.com/ensdomains/ens-subgraph/issues/88
 
-  // do not update expiry if PCC is burned
+  // if PCC is burned (not set), we do not update expiry
   if (checkPccBurned(BigInt(wrappedDomain.fuses))) return;
 
   // update the domain's expiry to the greater of the two
--- a/src/handlers/NameWrapper.ts
+++ b/src/handlers/NameWrapper.ts
@@ -153,12 +153,18 @@ export const makeNameWrapperHandlers = (ownedName: `${string}eth`) => {
     }) {
       const { node, fuses } = event.args;
 
-      // NOTE: subgraph does an implicit ignore if no WrappedDomain record.
-      // we will be more explicit and update logic if necessary
-      await context.db.update(schema.wrappedDomain, { id: node }).set({ fuses });
+      // NOTE: subgraph no-ops this event if there's not a wrappedDomain already in the db.
+      // https://github.com/ensdomains/ens-subgraph/blob/master/src/nameWrapper.ts#L144
+      const wrappedDomain = await context.db.find(schema.wrappedDomain, { id: node });
+      if (wrappedDomain) {
+        // set fuses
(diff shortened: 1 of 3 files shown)


src/handlers/NameWrapper.ts
```diff
@@ -156,6 +156,9 @@ export const makeNameWrapperHandlers = (ownedName: `${string}eth`) => {
       // NOTE: subgraph does an implicit ignore if no WrappedDomain record.
       // we will be more explicit and update logic if necessary
       await context.db.update(schema.wrappedDomain, { id: node }).set({ fuses });
+
+      // materialize the domain's expiryDate
+      await materializeDomainExpiryDate(context, node);
```
tk-o, line 161: 👍

src/handlers/Registrar.ts
```diff
@@ -14,6 +14,16 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
   const ownedSubnameNode = namehash(ownedName);
 
   async function setNamePreimage(context: Context, name: string, label: Hex, cost: bigint) {
+    // NOTE: ponder intentionally removes null bytes to spare users the footgun of
+    // inserting null bytes into postgres. We don't like this behavior, though, because it's
+    // important that 'vitalik\x00'.eth and vitalik.eth are differentiable.
+    // https://github.com/ponder-sh/ponder/issues/1456
+    // So here we use the labelhash fn to determine whether ponder modified our `name` argument,
+    // in which case we know that it used to have null bytes in it, and we should ignore it.
```
tk-o, line 22: Does that mean we don't index names that include null bytes, or we do it somewhere else?

src/lib/subname-helpers.ts
```diff
@@ -71,17 +70,55 @@ export const isLabelIndexable = (label: string) => {
...
+    if (offset > 1) {
+      list.push(".");
+    } else {
+      firstLabel = labelStr;
+    }
+    list.push(labelStr);
+    offset += len;
+    len = buf[offset++];
+  }
+  return [firstLabel, list.join("")];
```
tk-o, line 123: @lightwalker-eth this one looks like something you'd be the best person to review :) I can only suggest creating unit tests for. But I'm also aware I have to first integrate a testing suite into this project as per
- https://github.com/namehash/ensnode/issues/15

lightwalker-eth (approved):
@shrugs Looks good 🚀 Shared 1 comment seeking your confirmation.

src/handlers/Registrar.ts
shrugs: it means that we ignore events that have names with null bytes

src/lib/subname-helpers.ts
shrugs: we're copying and pasting subgraph code, testing the function is lower priority than adding tests to ensjs for this case, imo

shrugs:
subgraph excludes events that contain null byte names, we are doing the same. it specifically only excludes them when setting 'preimage' (whatever that is) for the name registered by controller and name renewed by controller events. this implementation does the same. only addition here is correctly identifying when ponder has already removed the null bytes that we were trying to detect with isLabelIndexable

src/handlers/Registrar.ts
lightwalker-eth: Noted. I've created an issue to track an enhancement opportunity: https://github.com/namehash/ensnode/issues/45

tk-o (approved):
Looks good to me 🚀

src/lib/subname-helpers.ts
lightwalker-eth: This is definitely a lower priority goal at the moment, but as I understand there's still value for us to define our own unit tests for `decodeDNSPacketBytes` under the assumption that ensjs defines `bytesToPacket` but not `decodeDNSPacketBytes`.

If my assumptions are off there please let me know. Cheers :+1:

lightwalker-eth (approved):
@shrugs Hey looks good :+1: Thanks!

src/lib/subname-helpers.ts
shrugs: those are intended to be the same function, although we do need to inject the subgraph-specific character checks (`isLabelIndexable`) into their logic as well (their util should probably handle that case tbh). but yeah some tests here are ok, i was cautioning against trying to exhaustively understand and test this function when its guaranteed to do what we desire for the moment (because it comes from subgraph and we're aiming for subgraph 1:1)

PR #43, opened 2025-01-21
feat(testing-suite): include basic unit tests

tk-o:
Use `vitest` runner to ensure unit tests are passing OK.

Resolves: #15


.github/workflows/runtime-tests.yml
```diff
@@ -0,0 +1,34 @@
+name: Runtime Tests
```
shrugs, line 1: 'Unit Tests' perhaps more standard

src/lib/subname-helpers.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it("should return [null, null] for labels with unindexable characters", () => {
+    const buf = new Uint8Array([116, 101, 115, 116, 0]); // 'test\0'
+    expect(decodeDNSPacketBytes(buf)).toEqual([null, null]);
+  });
+});
+
+describe("tokenIdToLabel", () => {
+  it("should convert bigint tokenId to hex string", () => {
+    const tokenId = BigInt("1234567890123456789012345678901234567890");
```
shrugs, line 41: perhaps better to use a real value here

src/lib/subname-helpers.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(tokenIdToLabel(tokenId)).toBe(
+      "0x00000000000000000000000000000003a0c92075c0dbf3b8acbc5f96ce3f0ad2",
+    );
+  });
+});
+
+describe("makeSubnodeNamehash", () => {
+  it("should return the correct namehash for a subnode", () => {
+    const node = makeSubnodeNamehash(namehash("base.eth"), tokenIdToLabel(123n));
```
shrugs, line 50: perhaps best to use real info and make it like

```
expect(makeSubnodeNameHash(namehash("base.eth"), labelhash("0xtko"))).toBe(namehash("0xtko.base.eth"))
```

src/lib/subname-helpers.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("should return false for labels containing unindexable characters", () => {
+    expect(isLabelIndexable("test\0")).toBe(false);
+    expect(isLabelIndexable("test.")).toBe(false);
+    expect(isLabelIndexable("test[")).toBe(false);
+    expect(isLabelIndexable("test]")).toBe(false);
+  });
+
+  it("should return true for labels without unindexable characters", () => {
+    expect(isLabelIndexable("test")).toBe(true);
+    expect(isLabelIndexable("example")).toBe(true);
```
shrugs, line 20: add a unicode emoji example like `test👍`

shrugs (changes requested):

src/lib/subname-helpers.spec.ts
shrugs: also makes me want to reorder the arguments to `makeSubnodeNameHash` to be `(label, node)` so it looks right.

src/lib/subname-helpers.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  it("should return [null, null] for labels with unindexable characters", () => {
+    const buf = new Uint8Array([116, 101, 115, 116, 0]); // 'test\0'
+    expect(decodeDNSPacketBytes(buf)).toEqual([null, null]);
+  });
+});
+
+describe("tokenIdToLabel", () => {
+  it("should convert bigint tokenId to hex string", () => {
+    const tokenId = BigInt("1234567890123456789012345678901234567890");
+    expect(tokenIdToLabel(tokenId)).toBe(
```
lightwalker-eth, line 42: Nice! This unit test finally makes the `tokenIdToLabel` function more clear for me. I saw it being used in strange ways in the indexing code.

I've created an issue to follow up in separate PR: https://github.com/namehash/ensnode/issues/44

.github/workflows/runtime-tests.yml
lightwalker-eth, line 1:
```suggestion
name: Unit Tests
```

And perhaps rename this workflow file as well?

src/lib/helpers.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    it("should handle undefined start and end", () => {
+      const config = blockConfig(undefined, 10, undefined);
+      expect(config).toEqual({ startBlock: 10, endBlock: undefined });
+    });
+  });
+
+  describe("rpcEndpointUrl", () => {
+    const originalEnv = process.env;
```
lightwalker-eth, line 48: Cool strategies here for building unit tests around `process.env` but suggest we take a different approach:

1. Refactor `rpcEndpointUrl` such that you split it into two separate functions:
   1. `rpcEndpointUrlImpl` (or something like that) that doesn't read anything from environment variables but instead just takes a `string | undefined` (or whatever the type of an environment variable is) as a param.
   2. `rpcEndpointUrl` that is a little convenience wrapper of `rpcEndpointUrlImpl` that reads from an env variable.
2. ... then you can just build the unit tests here around `rpcEndpointUrlImpl` and not need any special environment variable logic here.

lightwalker-eth (changes requested):
@tk-o Nice work 💪 Reviewed and shared some suggestions with feedback :+1:

src/lib/helpers.spec.ts
tk-o: If we read from the env, then I suggest our test also reads from the env. Of course, we can always add one layer of abstraction and mock an underlying getter/provider method. It does not change the accuracy of the test in anyway — just hides implementation details behind an extra abstraction. 

I recommend us making things simpler and easier to read, over adding abstractions when there's a little gain from that abstraction.

What do you think, @lightwalker-eth?

src/lib/helpers.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    it("should throw an error if the environment variable value is invalid", () => {
+      process.env.RPC_URL_1 = "invalid-url";
+      expect(() => rpcEndpointUrl(1)).toThrow(
+        "Invalid 'RPC_URL_1' environment variable value: 'invalid-url'. Please provide a valid RPC URL.",
+      );
+    });
+  });
+
+  describe("rpcMaxRequestsPerSecond", () => {
```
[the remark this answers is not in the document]
tk-o: I'd apply the same response to this comment as I did for the other one 🙃

src/lib/subname-helpers.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+});
+
+describe("decodeDNSPacketBytes", () => {
+  it('should return ["", "."] for empty buffer', () => {
+    expect(decodeDNSPacketBytes(new Uint8Array())).toEqual(["", "."]);
+  });
+
+  it("should return [null, null] for labels with unindexable characters", () => {
+    const buf = new Uint8Array([116, 101, 115, 116, 0]); // 'test\0'
```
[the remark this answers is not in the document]
shrugs: `viem#stringToBytes` should do the trick

src/lib/helpers.spec.ts
shrugs: agree less abstraction here. if it were me i'd even have less code that's more like `panic if rpc url is null when that plugin is enabled` and call it a day. not sure how abstraction helps us here, where would a non-env config come from?

src/lib/helpers.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+    it("should return the rate limit from the environment variable", () => {
+      process.env.RPC_REQUEST_RATE_LIMIT_1 = "400";
+      expect(rpcMaxRequestsPerSecond(1)).toBe(400);
+    });
+
+    it("should return the default rate limit if the environment variable is missing", () => {
+      expect(rpcMaxRequestsPerSecond(1)).toBe(50);
+    });
+
+    it("should throw an error if the environment variable value is invalid", () => {
```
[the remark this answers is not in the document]
tk-o: NaN is not possible, it's always a `string | undefined` when read from env vars.

.github/workflows/runtime-tests.yml
```diff
@@ -0,0 +1,34 @@
...
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  static-analysis:
+    runs-on: ubuntu-latest
+
+    steps:
```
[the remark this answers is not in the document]
tk-o: Why would we need `.nvmrc` file? is that for the `nvm install` command to know which node version to apply?

src/lib/helpers.spec.ts
lightwalker-eth: As I understand, these unit tests have the goal of validating our logic that takes the `string | unknown` (or whatever the type of an environment variable is) and handles it in a defined way.

I don't see value in testing how we read from an environment variable.

Let me know what you think :+1:

.github/workflows/runtime-tests.yml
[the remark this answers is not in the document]
tk-o: Perhaps we'll run the unit tests per package once we landed on the monorepo setup.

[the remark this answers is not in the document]
lightwalker-eth: @tk-o Yep, that's right. The .nvmrc file is a nice way to make the expected node version explicit.

src/lib/helpers.spec.ts
lightwalker-eth: Sure thanks for feedback guys. Happy to follow your suggestions :+1:

tk-o: Got your point. You actually want to test the rpc settings parsers, and don't want to test the way they are delivered from env vars. Consider it done.

src/lib/helpers.ts
```diff
@@ -33,25 +33,29 @@ export const rpcEndpointUrl = (chainId: number): string => {
...
+  try {
+    return parseRpcEndpointUrl(envVarValue);
+  } catch (e: any) {
+    throw new Error(`Environment variable parse error: '${envVarName}' was ${e.message}`);
+  }
+};
+
+export const parseRpcEndpointUrl = (rawValue?: string): string => {
+  // no RPC URL provided
+  if (!rawValue) {
```
tk-o, line 45: @lightwalker-eth here's the split of functionality creating RPC config values. One is to read raw value from env vars, the other is to have it parsed into expected value range. We unit test the later.

src/lib/helpers.ts
```diff
@@ -70,27 +74,33 @@ export const rpcMaxRequestsPerSecond = (chainId: number): number => {
...
-  // no rate limit provided in env var
-  if (!envVarValue) {
+  try {
+    return parseRpcMaxRequestsPerSecond(envVarValue);
+  } catch (e: any) {
+    throw new Error(`Environment variable parse error: '${envVarName}' was ${e.message}`);
+  }
+};
+
+export const parseRpcMaxRequestsPerSecond = (rawValue?: string): number => {
```
tk-o, line 84: @lightwalker-eth here's the split of functionality creating RPC config values. One is to read raw value from env vars, the other is to have it parsed into expected value range. We unit test the later.

src/lib/subname-helpers.ts
```diff
@@ -99,7 +99,7 @@ export function decodeDNSPacketBytes(buf: Uint8Array): [string | null, string |
   let hex = Buffer.from(buf).toString("hex");
   let firstLabel = "";
 
-  if (len === 0) {
+  if (!len) {
```
tk-o, line 102: FYI @shrugs, this change fixes this failing test case:
image

src/lib/helpers.spec.ts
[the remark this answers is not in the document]
tk-o: Ok, I did split the factory function from the parser, and now, the parser is covered with unit tests.

.github/workflows/runtime-tests.yml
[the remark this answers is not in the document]
tk-o: .nvmrc file got added

src/lib/subname-helpers.ts
tk-o: I've skipped the test so we could stick to the original `decodeDNSPacketBytes` implementation.

lightwalker-eth (approved):
@tk-o Looks good 🚀

PR #49, opened 2025-01-22
fix(ponder-build): move tests out the `src/` dir

tk-o:
This is needed to avoid issues with ponder build process. If tests are included in the `src/` directory, ponder won't be able to start indexing:

image

Also, this PR drops the end block limit on the `eth` plugin and adds new check to static analysis task on CI.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -35,3 +35,6 @@ jobs:
 
       - name: Run TypeScript type checks
         run: pnpm run typecheck
+
+      - name: Run Ponder build checks
+        run: pnpm codegen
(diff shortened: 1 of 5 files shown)


lightwalker-eth (approved):
@tk-o looks good :+1:

.github/workflows/static-analysis.yml
```diff
@@ -35,3 +35,6 @@ jobs:
 
       - name: Run TypeScript type checks
         run: pnpm run typecheck
+
+      - name: Run Ponder build checks
+        run: pnpm codegen
```
tk-o, line 40: @lightwalker-eth I've added a step that ensures the PR will fail if ponder build process can't get to successful completion.

lightwalker-eth: @tk-o Nice 🚀

PR #67, opened 2025-01-22
Align node version

lightwalker-eth:
I'm not an expert in this minutiae, but creating this PR just to flag how I noticed the changes merged in https://github.com/namehash/ensnode/pull/62/files are not in full alignment with the current package.json.

Open to whatever you guys suggest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "vitest": "^3.0.3"
   },
   "engines": {
-    "node": ">=18.14"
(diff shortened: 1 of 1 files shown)


shrugs:
this is fine, too. the `>=` that ponder suggests allowed for specifying 8.20.5, but we can be even more specific if we'd like

shrugs (approved):

Issue #52, opened 2025-01-22
Move www.ensnode.io to monorepo


lightwalker-eth:
Key goals include:

- [x] Transition to monorepo structure
  - https://github.com/namehash/ensnode/pull/68
- [ ] Move [the Next.js app for the ENSNode.io homepage](https://github.com/namehash/namekit/tree/main/apps/ensnode.io) out of NameKit into the ENSNode monorepo.

Here are some resources that might be helpful:

1. [Ponder's example monorepo configuration](https://github.com/ponder-sh/ponder/tree/main/examples/with-nextjs)
2. Review the NameKit monorepo configuration:https://github.com/namehash/namekit

Key subtasks include:

1. Update Railway config (for Ponder app builds)
2. Update Vercel config (for Next.js app builds of the ensnode.io homepage)

shrugs:
Will there be any shared code between the website and ensnode project? If not i'm not a _huge_ fan of the monorepo, since it adds complexity (deploys, merge-rules, etc) for very little organizational gain. Once the ensnode project goes multi-service in v2, though, a pnmp workspace makes sense, as it will likely share code amongst the various services.

If we just make the website it's own repo, configuring stuff like railway/vercel/etc becomes strictly happy-path, and we'll run into fewer of the obstacles that show up with monorepo deployments.

shrugs:
Ah, now that i say that, it does seem like the website might share code in the future, to power the dashboard, i see. in that case a monorepo could make sense — we'll just want to make sure things are configured correctly as we set it up.

lightwalker-eth:
@shrugs Good feedback, thanks.

In my understanding, there's two distinct web apps in our roadmap currently:
1. A "live dashboard" showing metrics and utilities for monitoring the status of an ENSNode instance. More details in the following Slack thread. I'll need more time to transition the ideas there into a task here in GitHub: https://namehash.slack.com/archives/C086Z6FNBHN/p1737206149803769. In theory each instance of ENSNode might run this. Therefore it might be accessible through domains such as https://alpha.ensnode.io.
[...]

PR #68, opened 2025-01-22
feat(monorepo): split indexer into apps and packages

tk-o:
Related to #52 

## Goals

- [x] easy exporting for ponder schema to support the upcoming [`@ponder/client` feature](https://ponder-docs-git-kjs-live-ponder-sh.vercel.app/docs/query/client)
  - ~ponder plugins to be split into packages~
    - cannot be done as ponder app requires all handlers to be defined within ponder's `rootDir`
    - `rootDir` is a single location, hence, having plugins in separate directories won't work
  - [x] `ponder-shema` package to host `ponder.schema.ts`
- [x] `ensnode-common` package to host all primitives that can be used outside of ponder app context, for example, on the client side
- [x] opt-in to using the Subgraph-compatible middleware
  - [x] the Subgraph middleware will have its own package
- [x] single directory `apps` to host all runnable modules
  - [x] `ensnode` ponder app will be the first module here
  - ~`ensnode` app will import ponder schema & config from the `ponder-bootstrap` package~
    - the ponder app expects these files to be loaded from a single `rootDir` — having ponder plugins in play, also place in the `rootDir`, prevents extracting `ponder.config.ts` & `ponder.schema.ts` into a separate package


lightwalker-eth (commented):
@tk-o Exciting to see this taking shape :+1: A few very small comments.

packages/ensnode-common/src/plugin-helpers.ts
lightwalker-eth, line 1: What's the use case you see for these utility functions to live outside the Ponder app?

It seems that at the minimum `getActivePlugins` and `MergedTypes` should be moved out of this library and into the Ponder app.

packages/ensnode-common/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "ensnode-common",
```
lightwalker-eth, line 2: What do you think about renaming this to something like `ensnode-utils`? Open to feedback :+1:

packages/ensnode-common/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "@biomejs/biome": "catalog:",
+    "@types/node": "catalog:",
+    "typescript": "catalog:",
+    "viem": "catalog:",
+    "vitest": "catalog:"
+  },
+  "engines": {
+    "node": ">=18.14"
+  }
+}
```
lightwalker-eth, line 30: Could you please create an initial basic README file for this package? Thanks. It's nice if each app / package has a README

packages/ensnode-subgraph-graphql/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "ensnode-subgraph-graphql",
```
lightwalker-eth, line 2: This package also needs a copy of the LICENSE

packages/ensnode-common/package.json
```diff
@@ -0,0 +1,30 @@
+{
```
lightwalker-eth, line 1: It think it's nice to add a few more fields to each of our `package.json` files. Specifically the following:

* description (please propose a 1 liner for each)
* license (always "MIT" for us)
* repository
* homepage

You can find a reference here: https://github.com/namehash/namekit/blob/main/packages/nameguard-sdk/package.json#L3-L11

lightwalker-eth (commented):
@tk-o Hey nice updates :+1: Shared a few more small ideas.

apps/ensnode/package.json
```diff
@@ -8,26 +8,24 @@
     "start": "ponder start",
     "db": "ponder db",
     "codegen": "ponder codegen",
-    "test": "echo 'no test to run'",
+    "test": "vitest",
     "lint": "biome check --write .",
+    "lint:ci": "biome ci",
     "typecheck": "tsc --noEmit"
   },
```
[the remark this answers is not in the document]
tk-o: I think it's the package manager job to ensure correct `node_modules` state across entire workspace. Clean task can be helpful with clearing build artifacts, though.

package.json
```diff
@@ -4,11 +4,10 @@
...
-    "typescript": "^5.2.2",
-    "tsup": "^8.3.5",
-    "vitest": "^3.0.3"
+    "@biomejs/biome": "^1.9.4"
+  },
+  "engines": {
+    "node": ">=18.14"
   },
   "resolutions": {
     "vite": "5.4.12"
```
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Ah. Is this related to the fix to that dependabot warning we had? I understand this is only temporary in our package.json file then until some of our dependencies update their downstream dependencies?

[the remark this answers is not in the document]
tk-o: Yeah, it's temporary, but from my experience, also recurring.

tk-o:
@shrugs, do you think, for this PR, that we should also work towards:

> easy exporting for ponder schema to support the upcoming [@ponder/client feature](https://ponder-docs-git-kjs-live-ponder-sh.vercel.app/docs/query/client)

?

I wanted to export `ponder.schema.ts` from one package and import it in another, but it gives me some drizzle type resolution errors, for example:
```
DTS Build start
../../apps/ensnode/ponder.schema.ts(4,14): error TS2742: The inferred type of 'domain' cannot be named without a reference to '.pnpm/drizzle-orm@0.36.4_@electric-sql+pglite@0.2.13_@opentelemetry+api@1.9.0_kysely@0.26.3_pg@8.13.1_react@18.3.1/node_modules/drizzle-orm/pg-core'. This is likely not portable. A type annotation is necessary.
```
I've found [an explanation to that drizzle problem](https://github.com/microsoft/TypeScript/pull/58176#issuecomment-2052698294), but didn't yet figure how to solve it. Not sure how to progress about that.

What do you think?

packages/ensnode-utils/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "outDir": "dist",
+    "strict": true,
+    "allowImportingTsExtensions": true,
+    "emitDeclarationOnly": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true
+  },
+  "include": [
+    "src/**/*",
+    "../../apps/ensnode/src/lib/plugin-helpers.ts",
```
shrugs, line 18: this seems like an anti-pattern, what is it accomplishing?

apps/ensnode/README.md
```diff
@@ -0,0 +1,213 @@
...
+- `pnpm ponder dev` for development mode,
+- `pnpm ponder start` for production mode.
+
+To learn more about those commands, go to https://ponder.sh/docs/api-reference/ponder-cli#dev
+
+### Query index
+
+The ENSNode exposes two GraphQL endpoints to query:
+- `/` uses a Ponder-native GraphQL schema
+- `/subgraph` uses a subgraph-native GraphQL schema
```
shrugs, line 63: nit: subgraph-compatible GraphQL schema

apps/ensnode/README.md
```diff
@@ -0,0 +1,213 @@
...
+  - ensjs equivalency confirmed via [ens-subgraph-transition-tools](https://github.com/namehash/ens-subgraph-transition-tools)
+- v2 — **optimized multichain indexer w/ unified namespace**
+  - true multichain indexing (mainnet, base, linea, etc)
+  - flattened, unified, multichain namespace
+  - support key ens-app-v3 and wallet ENS funtions via optimized resolvers & PRs
+  - high quality human-readable (healed) list of names by owner, necessary for many UX
+  - (possible) continued backwards compatibility with subgraph
+  - support indexing subset of data, i.e. only domains under parent node
+
+## next up
```
shrugs, line 184: i will move these to github issues, you can delete this section

packages/ensnode-utils/src/helpers.ts
```diff
@@ -0,0 +1,5 @@
+export const uniq = <T>(arr: T[]): T[] => [...new Set(arr)];
```
shrugs, line 1: this file can/should live in the indexer, it doesn't need to be shared. and the file name helpers really only makes sense in that context.

packages/ensnode-utils/src/ids.ts
```diff
@@ -7,7 +6,7 @@ export const makeResolverId = (address: Address, node: Hex) =>
 
 // https://github.com/ensdomains/ens-subgraph/blob/master/src/utils.ts#L5
 // produces `blocknumber-logIndex` or `blocknumber-logindex-transferindex`
-export const makeEventId = (event: Event, transferIndex?: number) =>
-  [event.block.number.toString(), event.log.logIndex.toString(), transferIndex?.toString()]
+export const makeEventId = (blockNumber: bigint, logIndex: number, transferIndex?: number) =>
```
shrugs, line 9: i would prefer this file live in the indexer as well — i'm not sure we'll need to generate any of these ids outside of that context, and these ids are indexer-specific. there was a mention of maybe resolverId being used outside, but let's move all of these back to indexer for now, and only extract that one iff we actually need to do so.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -5,7 +5,6 @@ import { type Hex, concat, keccak256, namehash, toHex } from "viem";
 // or exposed by ensjs or viem
```
shrugs, line 5: i support this file being moved to the utils package like we have here

packages/ensnode-subgraph-graphql/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 NameHash
```
shrugs, line 3: let's update these years to 2025 across the board

packages/ensnode-subgraph-graphql/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": "0.0.1",
+  "type": "module",
+  "description": "A Hono middleware for generating Subgraph-compatible GraphQL schema.",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/namehash/ensnode.git",
+    "directory": "packages/ensnode-utils"
```
shrugs, line 10: update directory to match

packages/ensnode-utils/tsconfig.json
```diff
@@ -13,6 +13,10 @@
     "esModuleInterop": true,
     "skipLibCheck": true
   },
-  "include": ["src/**/*"],
+  "include": [
```
lightwalker-eth, line 16: Hmm, not sure if I'm reading this right. Is this creating a dependency from the `ensnode-utils` package back into the `ensnode` app? If so I don't believe we should be doing that. `ensnode-utils` is a dependency of `ensnode`, not the other way around.

packages/ensnode-subgraph-graphql/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "ensnode-subgraph-graphql",
```
shrugs, line 2: let's name this package `ponder-subgraph-api` or `ponder-subgraph-compat` since that's all it will do, and we can eventually publish it as such or `@namehash/ponder-subgraph-api` instead of contributing it back to ponder directly

pnpm-workspace.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - apps/*
+  - packages/*
+
+catalog:
+  '@biomejs/biome': ^1.9.4
+  '@types/node': ^20.9.0
+  hono: ^4.6.14
+  typescript: ^5.7.3
+  viem: ^2.21.57
+  vitest: ^3.0.3
```
shrugs, line 11: editor EOF

packages/ensnode-utils/README.md
```diff
@@ -0,0 +1,3 @@
+# ENSNode Utils
+
+This package is a utility library for interacting with ENS (Ethereum Name Service) data. It contains various helper functions and tools to facilitate interactions with ENS nodes, such as name resolution, hashing, and other common operations related to ENS.
```
lightwalker-eth, line 3:
```suggestion
This package is a utility library for interacting with ENS (Ethereum Name Service) data. It contains various helper functions and tools to facilitate interactions with ENS and ENSNode instances.
```

packages/ensnode-subgraph-graphql/package.json
```diff
@@ -2,6 +2,14 @@
   "name": "ensnode-subgraph-graphql",
   "version": "0.0.1",
   "type": "module",
+  "description": "A Hono middleware for generating Subgraph-compatible GraphQL schema.",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/namehash/ensnode.git",
+    "directory": "packages/ensnode-utils"
```
lightwalker-eth, line 10: We should update this directory

packages/ensnode-utils/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "emitDeclarationOnly": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true
+  },
+  "include": [
+    "src/**/*",
+    "../../apps/ensnode/src/lib/plugin-helpers.ts",
+    "../../apps/ensnode/test/plugin-helpers.spec.ts"
+  ],
+  "exclude": ["*spec.ts"]
```
shrugs, line 21: necessary?

packages/ensnode-subgraph-graphql/package.json
```diff
@@ -2,6 +2,14 @@
...
+  },
+  "homepage": "https://github.com/namehash/ensnode/tree/main/packages/ensnode-utils",
```
lightwalker-eth, line 12: We should update this homepage

README.md
```diff
@@ -1,3 +1,9 @@
 # ENSNode
 
 ENSNode is an open-source ENS data provider for web applications.
```
lightwalker-eth, line 3:
```suggestion
ENSNode is a multichain indexer for ENS, powered by Ponder.
```

packages/ensnode-utils/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
```
shrugs, line 2: we should leave the indexer's tsconfig as ponder has generated it, but if we ever have more than one other package using a similar tsconfig we should do the monorepo thing where we add a `tsconfig` package that other projects can depend on an inherit from.

README.md
```diff
@@ -1,3 +1,9 @@
...
+
+This repository is built from modules falling into two categories:
```
lightwalker-eth, line 5:
```suggestion
The ENSNode monorepo contains multiple modules in the following subdirectories:
```

README.md
```diff
@@ -1,3 +1,9 @@
...
+- `apps` for runnable code,
```
lightwalker-eth, line 6:
```suggestion
- `apps` for executable applications.
```

Please also make `apps` a link to the `apps` subdirectory.

README.md
```diff
@@ -1,3 +1,9 @@
...
+- `packages` for library code (to be included in runnable code).
```
lightwalker-eth, line 7:
```suggestion
- `packages` for libraries you can embed into your own apps.
```

shrugs (changes requested):
great work on this with the catalog definitions and the readmes and whatnot. some minor stuff i've noted from afar, but i'll check this out and start running it now

README.md
```diff
@@ -1,3 +1,9 @@
...
+
+The main module of this repository is `apps/ensnode`. Read more about [`apps/ensnode` here](apps/ensnode/README.md).
```
lightwalker-eth, line 9:
```suggestion
The main module of this repository is the ENSNode app found in [`apps/ensnode`](apps/ensnode).
```

apps/ensnode/README.md
```diff
@@ -19,18 +19,26 @@
```
lightwalker-eth, line 19: All the content in the Quick Start section should move to the root README in the monorepo.

lightwalker-eth (changes requested):
@tk-o Nice updates :+1: Shared a few more ideas with feedback 😄

package.json
```diff
@@ -1,39 +1,10 @@
...
-    "ponder": "^0.8.26",
-    "ts-deepmerge": "^7.0.2",
-    "viem": "^2.21.57"
-  },
   "devDependencies": {
-    "@biomejs/biome": "^1.9.4",
-    "@types/node": "^20.9.0",
-    "typescript": "^5.2.2",
-    "vitest": "^3.0.3"
+    "@biomejs/biome": "^1.9.4"
```
shrugs, line 7: necessary?

shrugs (commented):
may not need to include vite resolution in root package.json?

README.md
shrugs: 'for libraries both published and used here internally'?

shrugs:
just noting for myself: i'd really like for env variable management to be exclusively at the root — can likely configure this with dotenv executing the script before running ponder if necessary

shrugs:
also note to self: i worry this is too package-happy but the plugin framework can be extracted into its own package, initially private but publishable in the future

shrugs:
ok i've moved the tsconfigs to shared-configs and fixed my editor config related to biome, tests and static is passing. @tk-o give it a look and let me know what you think

packages/ensnode-utils/tsconfig.json
tk-o: Yup, this wa not meant to land in there.

tk-o:
Thanks you so much for your support here, @shrugs 🚀

package.json
lightwalker-eth: I think so, based on the script reference in this package.json

README.md
```diff
@@ -1,9 +1,166 @@
 # ENSNode
 
-ENSNode is an open-source ENS data provider for web applications.
+ENSNode is a multichain indexer for ENS, powered by Ponder.
 
-This repository is built from modules falling into two categories:
-- `apps` for runnable code,
-- `packages` for library code (to be included in runnable code).
+The ENSNode monorepo contains multiple modules in the following subdirectories:
+- `apps`  executable applications.
```
lightwalker-eth, line 6: Please make "apps" a link to the apps directory. Same for "packages".

README.md
```diff
@@ -1,9 +1,166 @@
...
+- `packages` for libraries, both published and used here internally, can be embedded into your own apps.
```
lightwalker-eth, line 7:
```suggestion
- `packages` for libraries that can be embedded into apps.
```

README.md
```diff
@@ -1,9 +1,166 @@
...
+```
+
+Install workspace dependencies:
+```
+pnpm install
+```
+
+#### Prepare application environment
+
+Go into the application root directory:
```
lightwalker-eth, line 42:
```suggestion
Go into the main ENSNode application root directory:
```

README.md
```diff
@@ -1,9 +1,166 @@
...
+
+Once your `.env.local` is configured, launch the indexer by running:
+- `pnpm ponder dev` for development mode,
+- `pnpm ponder start` for production mode.
+
+To learn more about those commands, go to https://ponder.sh/docs/api-reference/ponder-cli#dev
+
+### Query index
+
+The ENSNode exposes two GraphQL endpoints to query:
```
lightwalker-eth, line 67:
```suggestion
ENSNode exposes two GraphQL endpoints to query:
```

README.md
```diff
@@ -1,9 +1,166 @@
...
+- `DATABASE_SCHEMA` is arbitrary, with the limitations mentioned in the linked documentation
+- `DATABASE_URL` is your postgres database connection string
+
+Once your `.env.local` is configured, launch the indexer by running:
+- `pnpm ponder dev` for development mode,
+- `pnpm ponder start` for production mode.
+
+To learn more about those commands, go to https://ponder.sh/docs/api-reference/ponder-cli#dev
+
+### Query index
```
lightwalker-eth, line 65:
```suggestion
### Query indexed data
```

README.md
```diff
@@ -1,9 +1,166 @@
...
+
+The ENSNode exposes two GraphQL endpoints to query:
+- `/` uses a Ponder-native GraphQL schema
+- `/subgraph` uses a subgraph-compatible GraphQL schema
+
+#### Examples
+
+Fetching data about most recently-created domains while skipping some initial records.
```
lightwalker-eth, line 73:
```suggestion
Fetch data about the 3 most recently-created domains.
```

Please see some related comments below.

README.md
```diff
@@ -1,9 +1,166 @@
...
+
+<details>
+  <summary>Ponder-native query</summary>
+
+  ```gql
+  {
+    domains(
+      orderBy: "createdAt"
+      orderDirection: "desc"
+      after: "eyJjcmVhdGVkQXQiOnsiX190eXBlIjoiYmlnaW50IiwidmFsdWUiOiIxNjM5ODk1NzYxIn0sImlkIjoiMHhkNTczOGJjNGMxYzdhZDYyYWM0N2IyMWNlYmU1ZGZjOWZkNjVkNTk4NTZmNmYyNDIxYjE5N2Q0ZjgxNmFkZTRjIn0"
```
lightwalker-eth, line 83: Suggest removing the complexity of this "after" idea from the example queries. We should make the first example queries as simple as possible for people who aren't familiar with GraphQL yet.

README.md
```diff
@@ -1,9 +1,166 @@
...
+    ```
+  </details>
+</details>
+
+<details>
+  <summary>Subgraph-native query</summary>
+
+  ```gql
+  {
+    domains(orderBy: createdAt, orderDirection: desc, skip: 40, first: 3) {
```
lightwalker-eth, line 135: Also suggest removing any "skip" here. Please see comment above.

README.md
```diff
@@ -1,9 +1,166 @@
...
+
+  <details>
+    <summary>Subgraph-native response</summary>
+
+    ```
+    {
+      "data": {
+        "domains": [
+          {
+            "name": "🐧🐧🐧🐧🐧🐧🐧🐧🐧.eth",
```
lightwalker-eth, line 150: Ideally the names and expiry dates shown for both example queries would be identical. If this is tricky to action within this PR then no worries. Please create a separate follow up issue for it and we can improve this later. Appreciate your advice.

README.md
```diff
@@ -1,9 +1,166 @@
...
+            "expiryDate": "1710785244"
+          },
+          {
+            "name": "rebelteenapeclub.eth",
+            "expiryDate": "1679228224"
+          },
+          {
+            "name": "[b4201276b6f7ffe5a50b0c3c1406c21295ab9f553107ddc9c715be2f9a6f6e90].[e5e14487b78f85faa6e1808e89246cf57dd34831548ff2e6097380d98db2504a].[dec08c9dbbdd0890e300eb5062089b2d4b1c40e3673bbccb5423f7b37dcf9a9c]",
```
lightwalker-eth, line 158: Let's please try to avoid an example response that includes unknown labels. We should pick more simple and beginner friendly examples.

apps/ensnode/README.md
```diff
@@ -181,18 +23,6 @@ Fetching data about most recently-created domains while skipping some initial re
   - (possible) continued backwards compatibility with subgraph
   - support indexing subset of data, i.e. only domains under parent node
 
-## next up
```
lightwalker-eth: @tk-o Hey if I remember correctly I saw an earlier request from Matt to remove the ideas in this "next up" section and transfer them into GitHub issues. I think it's best we action that outside of this PR and have Matt take the lead on that transition of ideas, since he's the original author of these and he'll understand them best.

Let's revert this change so that we retain these ideas as they are for now.

Sorry for the back and forth on this.

apps/ensnode/package.json
```diff
@@ -23,14 +23,16 @@
...
   "dependencies": {
     "@ensdomains/ensjs": "^4.0.2",
```
lightwalker-eth, line 25: Should we also add ensjs to the catalog? I assume it will be used by multiple modules soon if it isn't already.

packages/ponder-schema/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "ponder-schema",
+  "private": true,
+  "version": "0.0.1",
+  "type": "module",
+  "description": "a shared ponder schema for ensnode",
```
lightwalker-eth, line 6:
```suggestion
  "description": "A shared ponder schema for ENSNode",
```

packages/shared-configs/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@namehash/shared-configs",
+  "private": true,
+  "version": "0.0.1",
+  "type": "module",
+  "description": "shared configs for the project",
```
lightwalker-eth, line 6:
```suggestion
  "description": "Shared configs for the ENSNode project.",
```

apps/ensnode/package.json
tk-o: We'll do that once we get more modules using the dependency 👍

packages/ensnode-utils/tsconfig.json
tk-o: Not really if we don't plan to emit types and share them as an npm package.

package.json
```diff
@@ -1,43 +1,15 @@
...
-    "graphql-scalars": "^1.24.0",
-    "graphql-yoga": "^5.10.9",
-    "hono": "^4.6.14",
-    "ponder": "^0.8.26",
-    "ts-deepmerge": "^7.0.2",
-    "viem": "^2.21.57"
-  },
   "devDependencies": {
     "@biomejs/biome": "^1.9.4",
     "@types/node": "^20.9.0",
```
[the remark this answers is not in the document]
tk-o: Removed all of them apart from biome — I'd like the linter to control the all files in monorepo.

apps/ensnode/README.md
tk-o: Reverted. FYI @shrugs

.github/workflows/ci.yml
```diff
@@ -36,8 +36,17 @@ jobs:
       - name: Run TypeScript type checks
         run: pnpm -r typecheck
 
-      - name: Run Ponder build checks
-        run: pnpm --filter ensnode codegen
+      # This will run the dev command in the background, wait for 5 seconds,
+      # and then kill the process if it is still running. If the command does
+      # not throw an error within that time, the step will exit successfully.
+      # If it does throw an error, the step will exit and fail the CI pipeline.
+      - name: Run Ponder runtime integrity checks
```
lightwalker-eth, line 43: @tk-o Very interesting :) A few questions:

1. There's no Postgres environment variable here. As I understand that means this would make use of PGLite? Assuming so, could you please also document that here. I guess this would use PGLite running locally on the GitHub Actions Runner instance?
2. There's no RPC environment variables here. Therefore we would fallback to the defaults. Any issues with that? We could add some GitHub secrets for these. Perhaps that's getting more complicated for an already big PR though. If you think this is a nice idea please feel welcome to create a separate follow up for this. This can also be a nice one to share with Sebastian to advance next steps with.
3. There's a nice way to set environment variables in these scripts outside of the `run` command. You can see an example here: https://github.com/namehash/namekit/blob/main/.github/workflows/nameai-api-lambda-deploy.yml#L55-L64
4. Is there still value in keeping the old `pnpm --filter ensnode codegen` command for "Run Ponder build checks"? Or is that totally redundant?
5. Open to adding a bit more than 5 seconds if it helps.
6. Is there any special reason why this only runs the `eth` plugin and not all plugins?

lightwalker-eth (changes requested):
@tk-o Great updates! Just one more set of open questions based on your proposed CI updates.

.github/workflows/ci.yml
lightwalker-eth: @tk-o Also, instead of using `pnpm --filter`, what do you think about something like this where you just set the `working-directory`? https://github.com/namehash/namekit/blob/main/.github/workflows/nameai-api-lambda-deploy.yml#L79

tk-o: 1. Yes, sir. The aim of the check is to see if the service can start indexing. It will just write into a file on the CI environment. Will add a comment on that.
2. No issues in using public RPCs, the check would terminate after 5 seconds anyway.
3. Thanks, will apply this format 🤝 
4. No, not really. The codegen process gets run internally by the `dev` command as well.
5. Sure, I'll do that with a local env var set on the step.
6. Sure, let's run all plugins.

On the working directory, for monorepo, in my mind, the working directory is the root directory. What do you think?

lightwalker-eth: @tk-o Great updates!

I'm really not an expert on these devops things, but it might be the case that instead of using `pnpm --filter ensnode` you could just run `pnpm run dev` but set the `working-directory` to `apps/ensnode`.

Not sure :) Appreciate your advice :+1:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+      # will exit and fail the CI pipeline. This runtime check uses a pglite
+      # database that only lives in the CI environment. It will be discarded
+      # after the CI run.
+      - name: Run Ponder runtime integrity checks
+        env:
+          ACTIVE_PLUGINS: eth,base.eth,linea.eth
+          RPC_URL_1: https://eth.drpc.org
+          RPC_URL_8453: https://base.drpc.org
+          RPC_URL_59144: https://linea.drpc.org
+          RUNTIME_CHECK_TIMEOUT: 10
```
lightwalker-eth, line 52:
```suggestion
          RUNTIME_CHECK_TIMEOUT_SECONDS: 10
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
+name: Static Analysis
```
lightwalker-eth, line 1:
```suggestion
name: CI Checks
```

Or something like that. Open to other ideas, but this is no longer just static analysis.

lightwalker-eth (changes requested):
@tk-o Super updates. Replied to your question and small comments :+1:

.github/workflows/ci.yml
tk-o: We can use `working-directory` if you like

lightwalker-eth (approved):
@tk-o Awesome! PR now looks good to me 🚀 😄 

@shrugs Passing PR review over to you 🙌

shrugs:
if ensnode-utils is going to be internal, let's add `private: true` to the package.json and we can remove the extra stuff like the `repo` and `homepage` sections.

do we need the `typecheck` script in any of these packages?

we should standardize on whether we're going to org-prefix package names or not. i've done so for shared-configs (it's package name is `@namehash/shared-configs`, as an example. the benefit is that for internal packages where the name might be relatively generic ('shared-configs') when we import it we import it like `@namehash/shared-configs` which communicates that it's a namehash-specific module. if it's just named `shared-configs` a reader might assume that it's some npm package called `shared-configs`. up to you if we want to do this or not for the other packages, or whether we want to remove the prefix for shared-configs and go un-prefixed only.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: Run Ponder runtime integrity checks
+        working-directory: apps/ensnode
+        env:
+          ACTIVE_PLUGINS: eth,base.eth,linea.eth
+          RPC_URL_1: https://eth.drpc.org
+          RPC_URL_8453: https://base.drpc.org
+          RPC_URL_59144: https://linea.drpc.org
+          RUNTIME_CHECK_TIMEOUT_SECONDS: 10
+        run: |
+          pnpm dev -vv &
```
shrugs, line 55: start command here, yeah?

shrugs (approved):
ready to merge, up to you on package name stuff, or whether to tackle separately. well done!

lightwalker-eth:
@shrugs Agreed on your feedback about org prefixing all our package names here :+1:

Let's keep all our packages private for now, including `ensnode-utils`, but I like the idea of keeping fields there that will be relevant once we change that to `public` and setup everything for automatically publishing to NPM, etc.. But let's handle that in some future separate PR. This one is already big enough!

tk-o:
I'm going to update the Custom Start Command on Railway for the indexing service.

From:
```
pnpm start --schema $RAILWAY_DEPLOYMENT_ID
```
to:
```
pnpm --filter ensnode start --schema $RAILWAY_DEPLOYMENT_ID
```

lightwalker-eth:
@tk-o Sounds good! Thanks!

PR #72, opened 2025-01-24
Added Dockerfile and docker-compose for local setup

BanaSeba:


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:20-slim
```
lightwalker-eth, line 1: Perhaps this is all good, but not sure if we should change anything here based on our `.nvmrc` file: https://github.com/namehash/ensnode/blob/main/.nvmrc or the version of node referenced in our root package.json: https://github.com/namehash/ensnode/blob/main/package.json#L38-L40

Ideally these would all be in total alignment.

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+        dockerfile: Dockerfile
+      ports:
+        - "8080:42069"
+      env_file:
+      - .env.docker
+      depends_on:
+        - postgres
+
+  postgres:
+    image: postgres:15
```
lightwalker-eth, line 15: Currently in Railway I see we're using:

CleanShot 2025-01-24 at 15 49 09

Is there some reason why this is different?

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+services:
+
+  indexer:
```
lightwalker-eth, line 3:
```suggestion
  ensnode:
```

lightwalker-eth (changes requested):
@BanaSeba Thanks! I'm definitely not an expert on Docker / devops stuff. So appreciate the feedback and advice from you @tk-o and @shrugs on what our next steps should be here.

Shared a few small ideas for now.

Dockerfile
tk-o: I believe we should support the oldest version possible, but use [the active LTS version](https://nodejs.org/en/about/previous-releases#release-schedule). There's no point in voluntarily using outdated software.

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+      build:
+        context: .
+        dockerfile: Dockerfile
+      ports:
+        - "8080:42069"
+      env_file:
+      - .env.docker
```
tk-o, line 10: Would using the [`.env.local` file](https://github.com/namehash/ensnode/blob/a2f3ce22dc993dc45db5444538bb5cf1310a5aff/.env.local.example) work here? It was meant to be loaded in to the indexer runtime. Let me know what you think 🙃

lightwalker-eth:
@shrugs Also I remember you shared some concerns 1-2 weeks about about us introducing Docker into the project at that time. Do you still have a concern on that? If so, could you please share more details with your advice and suggestions? Much appreciated! Thanks!

Sebastian will be on holiday for the coming week. @shrugs @tk-o If either of you guys have suggestions for the general direction @BanaSeba takes here, please share advice in a comment.

Please also note the advice I already shared with @BanaSeba in this Slack thread: https://namehash.slack.com/archives/C086Z6FNBHN/p1737717125731929

Dockerfile
shrugs: agree this should be `18` for now, but truthfully we should be able to update everything to run on 20. if we do so we'll want to do that as a single pr, updating nvmrc, engines fields, and docker bases at once. i've added an issue for that in the prject.

docker-compose.yml
shrugs: def should be 16

docker-compose.yml
shrugs: agree, call it `ensnode` for now, but imo we'll rename the ensnode service to something like `indexer` or `onchain-indexer` in the future, since ensnode is now the collection of services. i've added an issue to track this as well.

shrugs:
project is now multi-service, so docker is fine solution. i still worry about logistics-creep but it's the right time to add

.dockerignore
```diff
@@ -0,0 +1 @@
+.env*
```
shrugs, line 1: editor EOF configuration, please automatically include newlines on save

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+      POSTGRES_PASSWORD: mypassword
+      POSTGRES_DB: mydatabase
+    ports:
+      - "5432:5432"
+    volumes:
+      - postgres_data:/var/lib/postgresql/data
+
+volumes:
+  postgres_data:
+    driver: local
```
shrugs, line 28: editor EOF

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+COPY . /app
+WORKDIR /app
+
+RUN pnpm install --prod --frozen-lockfile
+
+EXPOSE 8000
+CMD [ "pnpm", "start" ]
```
shrugs, line 11: editor EOF

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node:20-slim 
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+COPY . /app
+WORKDIR /app
+
+RUN pnpm install --prod --frozen-lockfile
+
+EXPOSE 8000
```
shrugs, line 10: expose 8000 here but mapping `8080:42069` in compose?

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+services:
+
+  indexer:
+      build:
+        context: .
+        dockerfile: Dockerfile
+      ports:
+        - "8080:42069"
```
shrugs, line 8: can we rely on the ponder defaults and instead of specifying a port to that container we just map `42069:42069`?

shrugs (changes requested):

shrugs:
I'm working on some stuff that will be useful for you in updating this PR, so let's collaborate on slack when you pick this up again

BanaSeba:
Everything related to Dockerfile and compose was settled in PR 186: https://github.com/namehash/ensnode/pull/186

Issue #81, opened 2025-01-26
Registration not found error


tk-o:
```
9:20:11 PM WARN  user       Failed 'registrations.update()' database method 
9:20:11 PM ERROR indexing   Error while processing '/eth/linea/EthRegistrarController:OwnerNameRegistered' event in 'linea' block 6684681
RecordNotFoundError: No existing record found in table 'registrations'
    at setNamePreimage (/app/apps/ensnode/src/handlers/Registrar.ts:46:5)
    at handleNameRegisteredByController (/app/apps/ensnode/src/handlers/Registrar.ts:99:7)
    at /app/apps/ensnode/src/plugins/linea.eth/handlers/EthRegistrar.ts:57:7
  44 |     }
  45 |
> 46 |     await context.db.update(schema.registration, { id: label }).set({ labelName: name, cost });
     |     ^
  47 |   }
  48 |
  49 |   return {
db.update arguments:
  id   0xa1772418ccf3e352d0eefaba4e867638d6a697be4ede2eafcda98f5f4f11c887
Event arguments:
  label     0xa1772418ccf3e352d0eefaba4e867638d6a697be4ede2eafcda98f5f4f11c887
  owner     0xeCa8A765ebc86C95DDF8F0aAc4838454EA60cbeF
  name      erigganewmoney
  expires   4842836578
Log:
  index     18
  address   0xDb75Db974B1F2bD3b5916d503036208064D18295
Transaction:
  hash   0x1ee0a3e2facc4bb2895a366a9a5fb38f25299b4ac4c12f9d79ea374593889c5a
  from   0x3357e73873944D885c5496718D0A1326f2Cab9B4
  to     0xDb75Db974B1F2bD3b5916d503036208064D18295
Block:
  hash        0x454cd7724ae363689d9995a1572b4d49479d718e68b24f19f702427b3f12bef8
  number      6684681
  timestamp   1720772578
9:20:11 PM INFO  app        Indexed 10000 events with 93.1% complete and 12s remaining
9:20:11 PM WARN  process    Encountered indexing error, starting shutdown sequence
9:20:12 PM FATAL process    Finished shutdown sequence, terminating (exit code 1)
```

shrugs:
closed in #82

Issue #70, opened 2025-01-23
Missing domain error


tk-o:
Transaction details: https://etherscan.io/tx/0x109e7530e5c1d5d4e45c2623df553901d470117a31704e5793efd3152b5c89b4

```
9:02:22 AM WARN user Failed 'domains.update()' database method (id=edc0065d)

RecordNotFoundError: No existing record found in table 'domains'
```

<details>
  <summary>Full error message</summary>

```
9:02:19 AM INFO  app        Indexed 3055 events with 93.4% complete and 20s remaining

9:02:22 AM WARN  user       Failed 'domains.update()' database method (id=edc0065d)

RecordNotFoundError: No existing record found in table 'domains'

    at file:///app/node_modules/.pnpm/ponder@0.8.28_@opentelemetry+api@1.9.0_@types+node@20.17.10_@types+pg@8.11.10_hono@4.6.14_typ_k2ne6a4izfbuk3jfnqcbtlq2cy/node_modules/ponder/src/indexing-store/historical.ts:658:35

    at processTicksAndRejections (node:internal/process/task_queues:105:5)

    at HeadlessKysely.wrap (file:///app/node_modules/.pnpm/ponder@0.8.28_@opentelemetry+api@1.9.0_@types+node@20.17.10_@types+pg@8.11.10_hono@4.6.14_typ_k2ne6a4izfbuk3jfnqcbtlq2cy/node_modules/ponder/src/database/kysely.ts:58:24)

    at handleNameTransferred (/app/src/handlers/Registrar.ts:151:7)

    at /app/src/plugins/eth/handlers/EthRegistrar.ts:18:5

    at executeEvent (file:///app/node_modules/.pnpm/ponder@0.8.28_@opentelemetry+api@1.9.0_@types+node@20.17.10_@types+pg@8.11.10_hono@4.6.14_typ_k2ne6a4izfbuk3jfnqcbtlq2cy/node_modules/ponder/src/indexing/service.ts:440:5)

    at processEvents (file:///app/node_modules/.pnpm/ponder@0.8.28_@opentelemetry+api@1.9.0_@types+node@20.17.10_@types+pg@8.11.10_hono@4.6.14_typ_k2ne6a4izfbuk3jfnqcbtlq2cy/node_modules/ponder/src/indexing/service.ts:255:20)

[...]

Issue #44, opened 2025-01-21
Refine `tokenIdToLabel`


lightwalker-eth:
Several issues with the current use of this function:

- [ ] It needs to be renamed for clarity and to be technically correct in what it does.
- [ ] The assumptions this function makes needs to be explicitly documented.
   - [ ] Ex: relationship between tokenId and labelhash (within BaseRegistrars).
   - [ ] Ex: relationship between tokenId and namehash (within NameWrapper).
- [ ] Probably needs to be split into different functions and moved to more logical places in our code to avoid creating future confusion.
- [ ] Refine related unit tests.

shrugs:
ERC721 needs a unit256 type as the id. like you mentioned, to create this id the contracts just re-format (re-encode?) the node's labelhash from a bytes32 to a uint256.

```sol
// NameWrapper.sol
uint256 tokenId = uint256(keccak256(bytes(label)));

// BaseRegistrar.sol
//     * @param id The token ID (keccak256 of the label).
```

this does the reverse, re-encoding a `uint256` into a 32-byte long hex string. it's pretty straightforward and am unsure what you'd prefer it be called or how it might be split up. the only improvement i could think of are:

- rename to `tokenIdToLabelhash()`
- add a comment specifically explaining the above relation (tokenId is just labelhash as uint256)

are there any situations where tokenId is not a uint256-encoded labelhash?

lightwalker-eth:
One more quick thought: We shouldn't define `tokenIdToLabelHash` at a "global" level. There is no universal algorithm for this. A distinct implementation of this idea needs to be defined inside each NFT issuing contract we index.

shrugs:
i see, it's not officially standardized, just a common practice. in that case i might suggest:

generic helper in the shared utils package:
- `uint256ToBytes32()` - the current function's implementation
- a comment like 're-encodes a uint256 as bigint into bytes32 as hex string'

in the handler file for a given nft-issuing contract
- `tokenIdToLabelhash` - the contract-specific logic, which for now is just
- a comment linking to the implementation of the contract where this is documented

```ts
const tokenIdToLabelhash = (bigint tokenId) => uint256ToBytes32(tokenId);
```

lightwalker-eth:
@shrugs Agreed 🙌

lightwalker-eth:
[...]

PR #82, opened 2025-01-26
fix(registration): use unique registration ID

tk-o:
This PR fixes a primary key constraint issue on `registrations` table in database. All names registered for plugins, except for the `eth` plugin, will use node value (namehash function output) for registration ID. The `eth` plugin will continue using registration IDs based of the label value only.

Before this change, all registrations would have their respective ID derived from the label only. Which means that there would be multiple names competing for the same registration (i.e. `www.vitalik.eth` & `www.vitalik.linea.eth`). In consequence, subgraph-compatible indexing logic would assume that for an existing (conflicting) registration entity, there must be a related domain entity.

That was not always the case, and the runtime error would occur. Not anymore.

Fixes:
- #70 
- #81


packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -11,6 +11,15 @@ export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([
 // converts uint256-encoded nodes to hex
 export const tokenIdToLabel = (tokenId: bigint) => toHex(tokenId, { size: 32 });
 
+// Makes a cross-chain unique registration ID for a given owned name.
+export const makeRegistrationId = (ownedName: string, labelhash: Hex, namehash: Hex) =>
+  ownedName === "eth"
+    ? // To keep subgraph compatibility, we still use labelhash for registrations
```
lightwalker-eth, line 17:
```suggestion
    ? // To keep subgraph compatibility, we still use labelhash for RegistrationIds
```

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -11,6 +11,15 @@ export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([
 // converts uint256-encoded nodes to hex
 export const tokenIdToLabel = (tokenId: bigint) => toHex(tokenId, { size: 32 });
 
+// Makes a cross-chain unique registration ID for a given owned name.
+export const makeRegistrationId = (ownedName: string, labelhash: Hex, namehash: Hex) =>
```
lightwalker-eth, line 15:
```suggestion
export const makeRegistrationId = (ownedName: string, labelhash: Hex, node: Hex) =>
```

1. "namehash" refers to the function, while "node" returns to the return value of that function. There should be detailed documentation for this param, ex: that this is the node of the name that was registered.
2. Please also refine the following:
   1. Please add detailed documentation for `ownedName` and what this represents. Ex: It is the `ownedName` of the ENSNode plugin that is processing the registration event.
   2. Please add detailed documentation for `labelhash` and what this represents. For example, it needs to be 100% clear what this is the labelhash of. Ex: Is it the labelhash of the direct subname of `ownedName`? What if the registration is more than 1 level of subname beneath `ownedName`? What if `labelhash` has no child relationship at all with `ownedName`? For example: Imagine the "eth" plugin where a name such as tko.box is explicitly added to the ENS Registry. Exactly what is labelhash the labelhash of in this case? The same question if the name abc.123.456.tko.box was explicitly added to the ENS Registry under the "eth" plugin. I'm assuming that this is the labelhash of the subname being registered, but that we also need to make it clear that the subname being registered could be any subname at any level within the scope of registrations indexed by the ENSNode plugin associated with `ownedName`. It should also be noted how the "eth" plugin can theoretically index all possible names given the NameWrapper.


Please also note how a source of some of the complexity here is how this `makeRegistrationId` utility function is in a totally generic "subname-helpers.ts" file, which suggests that the logic inside of it should apply universally in all cases. This is a key reason for some of the feedback above.

Alternatively, it could be considered to only put this function in `apps/ensnode/src/handlers/Registrar.ts` and NOT export it. It would still need careful documentation, but perhaps a bit less, since as I understand it would only apply to direct subnames of the `ownedName` of the associated plugin.

Advice appreciated :+1:

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -146,11 +155,16 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
 
       const label = tokenIdToLabel(tokenId);
```
lightwalker-eth, line 156: The concerns I raised in https://github.com/namehash/ensnode/issues/44 make me concerned here too. Suggest we give priority to fixing https://github.com/namehash/ensnode/issues/44 and then return back to this PR to help us ensure we're doing this correctly here.

Until issue 44 is fixed it's too confusing to review the code we have and ensure it's doing the right thing.

lightwalker-eth (changes requested):
@tk-o Hey super effort to advance this on the weekend 💪 Much appreciated! Reviewed and shared some ideas with feedback 😄

apps/ensnode/src/handlers/Registrar.ts
shrugs: i think it's ok to merge this and then fix 44 — since in this PR we're not specifically checking the logic of the function as a whole, only that each instance of id for a registration record is updated to be `makeRegistrationId` with the correct arguments. for now you can assume that the logic that `const label = tokenIdToLabel(tokenId);` implements is correct, even if it's not yet fully accurate to the ideal mental model.

packages/ensnode-utils/src/subname-helpers.ts
shrugs: i support this helper living in `ids.ts`

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -11,6 +11,15 @@ export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([
...
+  ownedName === "eth"
+    ? // To keep subgraph compatibility, we still use labelhash for registrations
+      // handled by `eth` plugin
+      labelhash
+    : // For other plugins, we use the namehash to ensure uniqueness
+      namehash;
```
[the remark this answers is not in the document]
shrugs: personally i think collisions can be safely ignored given that they're keccak hashes. no known collision has ever been discovered and if such an event occurs i don't think anyone would say ensnode is at fault for not handling that case.

re: validation — it seems safe to throw that validation logic in, but i question the need for it, since we do have verified contract source. _perhaps_ that may not be true in the future, if we integrate a malicious registrar? but it doesn't seem like a likely threat at the moment. lmk if i'm off here. regardless should definitely be a separate issue

ps. sadly a concatenation also isn't as trivial because of the `Hex` type of the field, it may break some assumptions within ponder if we commit non-hex data, even if it is represented as TEXT in postgres. sticking with label or node is efficient, clear, and causes 0 likely problems, so i think it's a perfect solution.

packages/ensnode-utils/src/subname-helpers.ts
shrugs: i also think the included documentation is mostly acceptable, perhaps some additional context at the top is warranted like:

> a Registration record's `id` is historically the labelhash of the label directly under the name the Registrar manages (i.e. for a registration of `test.eth`, a Registration's id is `labelhash('test')`). Because the original subgraph implementation didn't consider indexing any Registrars besides the .eth Registrar, this leaves no room in the namespace for Registration records from other Registrars (like the .base.eth and .linea.eth Registrars). To account for this, while preserving backwards compatibility, Registration records created for the .eth Registrar use `label` as `id` and those created by any other Registrar use `node` (i.e. namehash(`test.base.eth`) to avoid collisions that would otherwise occur.

i think the inline comments are excellent, though perhaps unnecessary if the larger comment on the helper looks like my suggestion above

shrugs (approved):

apps/ensnode/src/lib/primitives.ts
```diff
@@ -0,0 +1,31 @@
+import type { Hex } from "viem";
```
shrugs, line 1: this is nice actually, good idea! a nitpick on filename: maybe `ens-types.ts`? or i've just used just `types.ts` in the past but maybe too generic.

shrugs (approved):
nice

apps/ensnode/src/lib/primitives.ts
tk-o: I've had it as `types`, but `primitives` also worked in my mind. However, let's stick to common naming standards for TS, and go with just `types` for the time being.

lightwalker-eth (changes requested):
@tk-o Hey nice work here. Reviewed and shared feedback :+1:

apps/ensnode/test/ids.spec.ts
```diff
@@ -17,4 +17,18 @@ describe("ids", () => {
       expect(makeEventId(123n, 456)).toEqual("123-456");
     });
   });
+
+  describe("makeRegistrationId", () => {
+    it("should use labelhash when ownedName is `eth` to ensure subgraph compatibility", () => {
```
lightwalker-eth, line 22: Suggest we change the text here now that the ownedName param is renamed

apps/ensnode/test/ids.spec.ts
```diff
@@ -17,4 +17,18 @@ describe("ids", () => {
...
+      expect(makeRegistrationId("eth", labelhash("vitalik"), namehash("vitalik.eth"))).toEqual(
+        labelhash("vitalik"),
+      );
+    });
+
+    it("should use node when ownedName is not `eth`", () => {
```
lightwalker-eth, line 28: Please see related comment above

apps/ensnode/src/lib/ids.ts
```diff
@@ -10,3 +11,24 @@ export const makeEventId = (blockNumber: bigint, logIndex: number, transferIndex
...
+ * Makes a cross-registrar unique registration ID.
+ *
+ * A Registration record's id is historically the labelhash of the label
+ * directly under the name the Registrar manages (i.e. for a registration
+ * of test.eth, a Registration's id is labelhash('test')). Because the original
+ * subgraph implementation didn't consider indexing any Registrars besides
+ * the .eth Registrar, this leaves no room in the namespace for Registration
+ * records from other Registrars (like the base.eth and linea.eth Registrars).
+ * To account for this, while preserving backwards compatibility, Registration
+ * records created for the .eth Registrar use label as id and those created by
```
lightwalker-eth, line 25: labelhash

apps/ensnode/src/lib/ids.ts
```diff
@@ -10,3 +11,24 @@ export const makeEventId = (blockNumber: bigint, logIndex: number, transferIndex
...
+ * any other Registrar use node (i.e. namehash(test.base.eth) to avoid
+ * collisions that would otherwise occur.
+ *
+ * @param registrarName the name of the registrar issuing the registration
+ * @param labelHash the labelHash of the direct subname of `registrarName` that was registered
+ * @param node the node of the full name that was registered
+ */
+export const makeRegistrationId = (registrarName: string, labelHash: Labelhash, node: Node) =>
+  registrarName === "eth" ? labelHash : node;
```
lightwalker-eth, line 34: Prefer to keep the inline comments I proposed in the separate PR. Ex: It matters that we explicitly defend details of how this function is implemented.

lightwalker-eth (changes requested):
@tk-o Thanks for updates. A few follow ups.

apps/ensnode/src/lib/ids.ts
```diff
@@ -10,3 +11,24 @@ export const makeEventId = (blockNumber: bigint, logIndex: number, transferIndex
...
+ * To account for this, while preserving backwards compatibility, Registration
+ * records created for the .eth Registrar use label as id and those created by
+ * any other Registrar use node (i.e. namehash(test.base.eth) to avoid
+ * collisions that would otherwise occur.
+ *
+ * @param ownedName the ownedName of the ENSNode plugin that is processing the registration event
+ * @param label registration's label
+ * @param node registration's node
+ */
+export const makeRegistrationId = (ownedName: OwnedName, label: Labelhash, node: Node) =>
```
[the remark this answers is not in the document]
shrugs: based on my understanding of the terminology, a nitpick: `labelhash` over `labelHash` for these variables and typenames

lightwalker-eth (approved):
@tk-o Looks great! Approved 🚀

apps/ensnode/src/lib/ids.ts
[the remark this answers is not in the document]
lightwalker-eth: @shrugs Worry here isn't for capitalization. Instead the issue here is I've noticed several places in the code where we are using "label" where we should instead be using "labelHash". There's an important distinction there.

Issue #86, opened 2025-01-28
Use more specific, updated terminology for types & variable names


lightwalker-eth:
- [x] use `Node` and `Labelhash` type where possible to better communicate semantic type
- [x] rename `ownedSubnameNode` to `ownedNameNode`
- [x] rename makeSubnodeNamehash to just makeSubnode and reverse the argument order (so it reads like `[label].[node]`)
- [x] rename `label` to `lablehash` where appropritate
  - keep event name `label` as-is, but re-name when expanding args
- [x] validate namehash/labelhash parameters
  - https://github.com/namehash/ensnode/pull/94#discussion_r1933007622
- [x] general terminology pass

shrugs:
also: renaming `label` to `labelhash` where appropriate.

notably the events emitted by these contracts frequently (exclusively?) call this value `label` (whereas we're standardizing on `labelhash`), and I _don't_ think that those instances should be updated. in that case, the event should be typed like `EventWithArgs<{ label: Hex }>` and in the handler we can rename the variable like `const { label: labelhash } = event.args`. this creates a nice boundary between the misaligned terminologies.

lightwalker-eth:
@shrugs Yes it drives me crazy how ENS contracts use incorrect / imprecise terminology. This issue also spills into the ENS Subgraph. It makes developing on ENS harder than it needs to be.

Really appreciate efforts that help us do better when ENSNode! One of our exciting opportunities is to build a more precise and expressive data model in our V2.

shrugs:
i've consolidated many specific terminology update tasks into this task, as i believe the changes are best done in a single PR, with commits splitting individual concerns for easier review.

PR #94, opened 2025-01-28
fix: refactor and document tokenId decoding logic

### LINKED ISSUE #44: Refine `tokenIdToLabel`

shrugs:
closes #44

### DIFF
--- a/apps/ensnode/src/handlers/NameWrapper.ts
+++ b/apps/ensnode/src/handlers/NameWrapper.ts
@@ -11,7 +11,9 @@ import { EventWithArgs } from "../lib/ponder-helpers";
 import type { OwnedName } from "../lib/types";
 
 /**
- * The NameWrapper contract encodes a domain's `node` value as uint256 for use as the tokenId
+ * When a name is wrapped in the NameWrapper contract, an ERC1155 token is minted that tokenizes
+ * ownership of the name. The minted token will be assigned a unique tokenId represented as
+ * uint256(namehash(name)) where name is the fully qualified ENS name being wrapped.
  * https://github.com/ensdomains/ens-contracts/blob/mainnet/contracts/wrapper/ERC1155Fuse.sol#L262
  */
 const tokenIdToNode = (tokenId: bigint): Node => uint256ToHex32(tokenId);
--- a/apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts
+++ b/apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts
@@ -8,7 +8,10 @@ import { upsertAccount } from "../../../lib/db-helpers";
 import { ownedName, pluginNamespace } from "../ponder.config";
 
 /**
- * Base RegistrarController contract's tokenId is uint256(labelhash)
+ * When direct subnames of base.eth are registered through the base.eth RegistrarController contract
+ * on Base a NFT is minted that tokenizes ownership of the registration. The minted NFT will be
+ * assigned a unique tokenId represented as uint256(labelhash(label)) where label is the direct
+ * subname of base.eth that was registered.
  * https://github.com/base-org/basenames/blob/main/src/L2/RegistrarController.sol#L488
  */
 const tokenIdToLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);
--- a/apps/ensnode/src/plugins/eth/handlers/EthRegistrar.ts
+++ b/apps/ensnode/src/plugins/eth/handlers/EthRegistrar.ts
@@ -5,7 +5,10 @@ import { makeRegistrarHandlers } from "../../../handlers/Registrar";
 import { ownedName, pluginNamespace } from "../ponder.config";
 
 /**
- * ETHRegistrarController contract's tokenId is uint256(labelhash)
+ * When direct subnames of .eth are registered through the ETHRegistrarController contract on
+ * Ethereum mainnet, a NFT is minted that tokenizes ownership of the registration. The minted NFT
+ * will be assigned a unique tokenId represented as uint256(labelhash(label)) where label is the
+ * direct subname of .eth that was registered.
  * https://github.com/ensdomains/ens-contracts/blob/mainnet/contracts/ethregistrar/ETHRegistrarController.sol#L215
  */
 const tokenIdToLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);
--- a/apps/ensnode/src/plugins/linea.eth/handlers/EthRegistrar.ts
+++ b/apps/ensnode/src/plugins/linea.eth/handlers/EthRegistrar.ts
@@ -8,7 +8,10 @@ import { upsertAccount } from "../../../lib/db-helpers";
 import { ownedName, pluginNamespace } from "../ponder.config";
 
 /**
- * Linea's ETHRegistrarController contract's tokenId is uint256(labelhash)
+ * When direct subnames of linea.eth are registered through the linea.eth ETHRegistrarController
+ * contract on Linea a NFT is minted that tokenizes ownership of the registration. The minted NFT
+ * will be assigned a unique tokenId represented as uint256(labelhash(label)) where label is the
+ * direct subname of linea.eth that was registered.
  * https://github.com/Consensys/linea-ens/blob/main/packages/linea-ens-contracts/contracts/ethregistrar/ETHRegistrarController.sol#L447
  */
 const tokenIdToLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);


apps/ensnode/src/handlers/NameWrapper.ts
```diff
@@ -36,7 +37,7 @@ async function handleTransfer(
   to: Address,
 ) {
   await upsertAccount(context, to);
-  const node = tokenIdToLabel(tokenId);
+  const node = decodeTokenIdToNode(tokenId);
```
tk-o, line 40: `map` or `covert` might be a better choice, as `decode` suggests there's some `encode`, but hashing is 
a one-way operation.

apps/ensnode/src/lib/ens-helpers.ts
```diff
@@ -0,0 +1,17 @@
+import { uint256ToHex32 } from "ensnode-utils/subname-helpers";
+
+/**
+ * Registrar contracts operate exclusively on names that are a direct subdomain of an OwnedName.
```
lightwalker-eth, line 4: It's not necessarily true that a Registrar contract has to exclusively operate only on names that are a direct subname of the Registrar's name. A Registrar contract could theoretically operate on an unconstrained set of names at different levels of the ENS hierarchy.

Its difficult for me to think of any constraints ENS places on Registrar contracts. They are essentially completely unconstrained in their behaviour.

It might be true that all 3 the Registrar contracts we're currently indexing only operate on their direct subnames. But there's no rule causing this to be universally true.

apps/ensnode/src/lib/ens-helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Because of this, tokenIds that they emit are uint256-encoded `labelhash` values.
```
lightwalker-eth, line 5: Several details to note:

1. We should separate the ideas of a Registrar from a "NFT Issuer". Not all Registrars are "NFT Issuers" and not all "NFT Issuers" are Registrars. For example, you could implement a contract to serve as a Registrar for subnames of example.eth. This contract can truly serve as a Registrar. Ex: it might be an onchain Registrar that writes data about registered subnames to the ENS Registry. Therefore, that contract meets the definition of a Registrar. However, nothing I just described necessarily needs to involve the issuance of NFTs. NFT Issuers operate at a separate layer from ENS.
2. There's no universal relationship between tokenIds and Labelhash values. Please see my other comments related to this.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -8,8 +8,11 @@ export const ROOT_NODE = namehash("");
 
 export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([node, label]));
 
-// converts uint256-encoded nodes to hex
-export const tokenIdToLabel = (tokenId: bigint) => toHex(tokenId, { size: 32 });
+/**
+ * Encodes a uint256 bigint as hex string (sized to 32 bytes), useful for NFTs in the ENS
```
lightwalker-eth, line 12: "... useful for ..." the way this reads suggests that this is the only use of this utility function. Suggest to frame this in a more general way, ex: "... uses include ...".

Please also note the other comments I've shared about how it isn't correct to assume any universal relationship between tokenIds and labelhash values.

apps/ensnode/src/handlers/NameWrapper.ts
lightwalker-eth: I'd be happy to remove the "decode" prefix. Ex:
* `tokenIdToLabelHash` (note "Hash" suffix as this doesn't return a label)
* `tokenIdToNode`

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -122,7 +119,7 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
     }) {
       const { id, expires } = event.args;
 
-      const label = tokenIdToLabel(id);
+      const label = decodeTokenIdToLabelhash(id);
```
lightwalker-eth, line 122: Let's please do a search through all our code and look for opportunities to change instances of "label" to "labelhash". The capitalization of "labelhash" doesn't matter to me, but it does matter to keep the ideas of "label" and "labelhash" distinct.

This will need to be done with some care, as perhaps there are some cases where "label" truly is a label and not a "labelhash".

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -61,7 +58,7 @@ export const makeRegistrarHandlers = (ownedName: `${string}eth`) => {
 
       await upsertAccount(context, owner);
 
-      const label = tokenIdToLabel(id);
+      const label = decodeTokenIdToLabelhash(id);
```
lightwalker-eth, line 61: This comment may not make sense until reading through all the others.

I note how:
1. This file contains generic "Registrar" logic to be shared across all of our Registrar contracts.
2. This file seems to need to convert tokenIds into labelhash values.
3. There is no universal strategy to convert from tokenIds into labelhash values.

Therefore, suggest we introduce some mechanism for this logic to be dynamic. Ex: Each specific Registrar implementation passes in some callback function that implements its own strategy for converting from tokenId to labelhash.

lightwalker-eth (changes requested):
@shrugs Hey great to see this PR 😄 Reviewed and shared feedback :+1:

apps/ensnode/src/handlers/NameWrapper.ts
shrugs: well it is an encode/decode -> they're encoding a bytes32 hash as a uint256 and we're re-encoding/decoding it back to bytes32. also happy to remove the decode prefix, so i'll do that

apps/ensnode/src/handlers/Registrar.ts
shrugs: yup — saw this but i think best left to #86

apps/ensnode/src/handlers/Registrar.ts
shrugs: i see — will think about how best to implement this. is leaving the shared handler code as-is and documenting this fact for v2 a valid option?

apps/ensnode/src/handlers/Registrar.ts
shrugs: i've updated these references in Registrar.ts for this PR but we should do another pass as part of #86

apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -1,11 +1,18 @@
...
-import { makeSubnodeNamehash, tokenIdToLabel } from "ensnode-utils/subname-helpers";
+import { makeSubnodeNamehash, uint256ToHex32 } from "ensnode-utils/subname-helpers";
+import type { Labelhash } from "ensnode-utils/types";
 import { zeroAddress } from "viem";
 import { makeRegistrarHandlers } from "../../../handlers/Registrar";
 import { upsertAccount } from "../../../lib/db-helpers";
 import { ownedName, pluginNamespace } from "../ponder.config";
 
+/**
+ * Base RegistrarController contract's tokenId is uint256(labelhash)
```
lightwalker-eth, line 11:
```suggestion
 * When direct subnames of base.eth are registered through the base.eth RegistrarController contract on Base a NFT is minted that tokenizes ownership of the registration. The minted NFT will be assigned a unique tokenId represented as uint256(labelhash(label)) where label is the direct subname of base.eth that was registered.
```

apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -1,11 +1,18 @@
...
+ * https://github.com/base-org/basenames/blob/main/src/L2/RegistrarController.sol#L488
+ */
+export const tokenIdToLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);
```
lightwalker-eth, line 14: Suggest we not export this.

lightwalker-eth, line 14:
```suggestion
export const tokenIdToDirectSubnameLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);
```

Appreciate that's a mouthful.. But putting priority on clarity and avoiding misinterpretations.

apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts
```diff
@@ -51,7 +96,7 @@ export default function () {
 
     await handleNameTransferred({
       context,
-      event: { ...event, args: { from, to, tokenId } },
+      event: { ...event, args: { from, to, labelhash: labelhash } },
```
lightwalker-eth, line 99:
```suggestion
      event: { ...event, args: { from, to, labelhash } },
```

apps/ensnode/src/plugins/eth/handlers/EthRegistrar.ts
```diff
@@ -1,7 +1,15 @@
...
+import { uint256ToHex32 } from "ensnode-utils/subname-helpers";
+import type { Labelhash } from "ensnode-utils/types";
 import { makeRegistrarHandlers } from "../../../handlers/Registrar";
 import { ownedName, pluginNamespace } from "../ponder.config";
 
+/**
+ * ETHRegistrarController contract's tokenId is uint256(labelhash)
+ * https://github.com/ensdomains/ens-contracts/blob/mainnet/contracts/ethregistrar/ETHRegistrarController.sol#L215
+ */
+export const tokenIdToLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);
```
lightwalker-eth, line 11: Please see the feedback I shared about this inside apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts

This includes the suggestion not to export this and to refine the comments about the function.

apps/ensnode/src/plugins/linea.eth/handlers/EthRegistrar.ts
```diff
@@ -1,11 +1,18 @@
...
 import { zeroAddress } from "viem";
 import { makeRegistrarHandlers } from "../../../handlers/Registrar";
 import { upsertAccount } from "../../../lib/db-helpers";
 import { ownedName, pluginNamespace } from "../ponder.config";
 
+/**
+ * Linea's ETHRegistrarController contract's tokenId is uint256(labelhash)
+ * https://github.com/Consensys/linea-ens/blob/main/packages/linea-ens-contracts/contracts/ethregistrar/ETHRegistrarController.sol#L447
+ */
+export const tokenIdToLabelhash = (tokenId: bigint): Labelhash => uint256ToHex32(tokenId);
```
lightwalker-eth, line 14: Please see the feedback I shared about this inside apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts

This includes the suggestion not to export this and to refine the comments about the function.

apps/ensnode/src/plugins/linea.eth/handlers/EthRegistrar.ts
```diff
@@ -46,7 +78,7 @@ export default function () {
 
     await handleNameTransferred({
       context,
-      event: { ...event, args: { from, to, tokenId } },
+      event: { ...event, args: { from, to, labelhash: labelhash } },
```
lightwalker-eth, line 81:
```suggestion
      event: { ...event, args: { from, to, labelhash } },
```

apps/ensnode/src/handlers/NameWrapper.ts
```diff
@@ -1,15 +1,21 @@
...
 import { makeEventId } from "../lib/ids";
+import { bigintMax } from "../lib/lib-helpers";
 import { EventWithArgs } from "../lib/ponder-helpers";
 import type { OwnedName } from "../lib/types";
 
+/**
+ * The NameWrapper contract encodes a domain's `node` value as uint256 for use as the tokenId
+ * https://github.com/ensdomains/ens-contracts/blob/mainnet/contracts/wrapper/ERC1155Fuse.sol#L262
+ */
+export const tokenIdToNode = (tokenId: bigint): Node => uint256ToHex32(tokenId);
```
lightwalker-eth, line 17: Please see the feedback I shared about this inside apps/ensnode/src/plugins/base.eth/handlers/Registrar.ts

This includes the suggestion not to export this and to refine the comments about the function.

lightwalker-eth (changes requested):
@shrugs Nice! Thank you! This is definitely moving in the right direction now. Reviewed and shared feedback :+1:

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -47,35 +48,34 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
...
     await context.db
       .update(schema.registration, {
-        id: makeRegistrationId(ownedName, label, node),
+        id: makeRegistrationId(ownedName, labelhash, node),
       })
       .set({ labelName: name, cost });
   }
 
   return {
     get ownedSubnameNode() {
```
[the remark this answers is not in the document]
shrugs: created issue #99

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -92,7 +92,7 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
       // log RegistrationEvent
       await context.db.insert(schema.nameRegistered).values({
         ...sharedEventValues(event),
-        registrationId: label,
+        registrationId: labelhash,
```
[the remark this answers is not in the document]
shrugs: ah no that should be `makeRegistrationId`

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -18,22 +14,27 @@ const GRACE_PERIOD_SECONDS = 7776000n; // 90 days in seconds
...
+    cost: bigint,
+  ) {
     // NOTE: ponder intentionally removes null bytes to spare users the footgun of
     // inserting null bytes into postgres. We don't like this behavior, though, because it's
     // important that 'vitalik\x00'.eth and vitalik.eth are differentiable.
     // https://github.com/ponder-sh/ponder/issues/1456
     // So here we use the labelhash fn to determine whether ponder modified our `name` argument,
     // in which case we know that it used to have null bytes in it, and we should ignore it.
-    const didHaveNullBytes = labelhash(name) !== label;
+    const didHaveNullBytes = _labelhash(name) !== labelhash;
```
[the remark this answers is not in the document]
shrugs: in this context the events (`NameRegistered`, `NameRenewed`) emit what we now call a 'label' called `name` and what we're calling a 'labelhash' called `label`. the variable names here stem from those events and the subgraph, which names them the same thing.

i've separated these concerns into #100

PR #95, opened 2025-01-28
fix: only use nvm for node version management

### LINKED ISSUE #71: remove .tool-versions in favor of nvm
having a second way to manage node version seems like a bad idea

shrugs:
closes #71


tk-o (approved):
LGTM

PR #97, opened 2025-01-28
fix: reference correct table for relation plural fields

shrugs:


lightwalker-eth (approved):
@shrugs Thanks! LGTM :+1:

Issue #84, opened 2025-01-27
ENSRainbow (Label Healing Sidecar)


shrugs:
implement ENSRainbow (the label healing internal api sidecar), as discussed in #55 

- [ ] create another project in `ensnode/apps/ensrainbow`
   - Add initial README.md with simple message that includes sharing thanks to https://github.com/graphprotocol/ens-rainbow
   - Add MIT license
- [x] @lightwalker-eth to setup R2
- [ ] use git annex to track/download the rainbowtables from R2
- [ ] write an `ingest.ts` file that:
  - instantiates pglite w/ datadir
  - unzips and ingests [ens_names.sql.gz](https://storage.cloud.google.com/ens-files/ens_names.sql.gz)
  -  saves the output to a datadir
- [ ] write an `index.ts` Hono server
  - [load pglite with datadir](https://pglite.dev/docs/filesystems#node-fs)
  - a single api endpoint `/v1/heal/:labelhash` that returns something like `{ healed: string | null }`
- [ ] write a Dockerfile that:
  - copies `ens_names.sql.gz` into the container
  - runs `ingest.ts`
  - deletes the copied `ens_names.sql.gz` file
  - sets `index.ts` as its entrypoint. 
- [ ] add a package.json script to run the correct docker build & push commands
- [ ] update the docker-compose
- [ ] update the railway template to pull this container and set the ens var reference correctly

this should produce a container with an embedded datadir that server.ts can serve requests from.

lightwalker-eth:
R2 should now be setup for our goals here.

The exact same Rainbow Table snapshot currently linked from https://github.com/graphprotocol/ens-rainbow can now be found here:
* https://bucket.ensrainbow.io/ens_names.sql.gz (6.37 GB)
* https://bucket.ensrainbow.io/ens_names.sql.gz.sha256sum

Please let me know if there's anything else I can help with on this :+1:

lightwalker-eth:
Hey guys, what do you think about deploying ENSRainbow independently of ENSNode?

In other words, what if we hosted a deployment of ENSRainbow at api.ensrainbow.io and gave it an independent deployment lifecycle from ENSNode?

I suppose one factor there might be latency changes depending on where the ENSNode instance is hosted in relation to the ENSRainbow instance.. but then couldn't we just make the endpoint for ENSRainbow an environment variable in ENSNode? If someone wanted to host their own instance of ENSRainbow, they could?

Please attack this. Thanks :+1:

[...]

PR #101, opened 2025-01-29
Ensrainbow init

djstrong:
https://github.com/namehash/ensnode/issues/84


apps/ensrainbow/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+COPY --from=ingestor /app/data ./data
+RUN npm install --production
+
+# Set environment variables
+ENV NODE_ENV=production
+ENV PORT=3001
+
+EXPOSE 3001
+
+CMD ["node", "dist/index.js"]
```
shrugs, line 45: can we move this to a start script and use `CMD ["pnpm", "start"]`?

apps/ensrainbow/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "classic-level": "^1.4.1",
+    "hono": "^4.0.5",
+    "progress": "^2.0.3"
+  },
+  "devDependencies": {
+    "@types/node": "^20.11.19",
+    "@types/progress": "^2.0.7",
+    "typescript": "^5.3.3"
+  }
+}
```
shrugs, line 23: apologies for the nitpick, but can you configure your editor to always add newlines at end of files?

apps/ensrainbow/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "build": "tsc",
+    "start": "node dist/index.js",
+    "ingest": "node dist/ingest.js",
+    "docker:build": "docker build -t ensnode/ensrainbow .",
+    "docker:push": "docker push ensnode/ensrainbow"
+  },
+  "dependencies": {
+    "@hono/node-server": "^1.4.1",
+    "classic-level": "^1.4.1",
+    "hono": "^4.0.5",
```
shrugs, line 15: use the `catalog:` spec for hono, since it's a shared dep

apps/ensrainbow/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "target": "ES2020",
+    "module": "commonjs",
+    "lib": ["ES2020"],
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "outDir": "dist",
+    "rootDir": "src"
+  },
```
shrugs, line 12: can you make this tsconfig extend from the shared lib tsconfig? if anything specific needs to be changed, you can override it

```json
{
  "extends": "@namehash/shared-configs/tsconfig.lib.json",
  "include": ["src/**/*"],
  "exclude": ["node_modules"]
}
```

shrugs (commented):
looks awesome

apps/ensrainbow/package.json
djstrong: I have added `"files.insertFinalNewline": true` in vscode settings.

.vscode/settings.json
```diff
@@ -11,5 +11,6 @@
   "[jsonc]": {
     "editor.defaultFormatter": "biomejs.biome"
   },
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "files.insertFinalNewline": true
```
shrugs, line 15: an nice, ty 👍

shrugs (approved):

apps/ensrainbow/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@ensnode/ensrainbow",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "build": "tsc",
+    "start": "node dist/index.js",
+    "ingest": "node dist/ingest.js",
```
lightwalker-eth, line 8: I'm really not an expert here. Relatively new to JS / TS land 😅  Perhaps @shrugs or @tk-o could share their advice? Should we be writing these `start` and `ingest` commands differently such that they:
1. Don't execute directly through `node` but instead perhaps through `pnpm` (or whatever else).
2. Reference .ts files (that will be present even if a build hasn't been completed yet) rather than .js files in the `dist` subdirectory (which I understand are only present after a build is completed).

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
+# ENSRainbow
+
+ENSRainbow is a label healing sidecar service for ENS names. It provides a simple API endpoint to heal ENS labelhashes back to their original labels.
```
lightwalker-eth, line 3:
```suggestion
ENSRainbow is an ENSNode sidecar service for healing ENS labels. It provides a simple API endpoint to heal ENS labelhashes back to their original labels.
```

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+curl http://localhost:3001/v1/heal/0x[labelhash]
+```
+Example:
+```bash
+curl http://localhost:3001/v1/heal/0x78441d2a930a233460507c2f25aea5ec5dc278db6aeef9e6ee8b930ccc150e58
+```
+Response: `{"healed":"sprayfoamroofingaugusta"}`
+
+If the label is not found: `{"healed":null}`
+
```
lightwalker-eth, line 59: It would also be nice to document the schema that's returned with any errors to this endpoint.

apps/ensrainbow/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+    const labelhash = c.req.param('labelhash');
+    
+    try {
+        // Convert hex string to Buffer, stripping '0x' prefix if present
+        const hashBytes = Buffer.from(labelhash.replace(/^0x/, ''), 'hex');
+        const label = await db.get(hashBytes);
+        return c.json({ healed: label });
+    } catch (error) {
+        if ((error as any).code === 'LEVEL_NOT_FOUND') {
+            return c.json({ healed: null });
```
lightwalker-eth, line 26: A few ideas here:

1. Should we return a HTTP 404 code for this case?
2. What do you think about logging some info to the console about the labelhash that was requested but not healable?

apps/ensrainbow/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+});
+
+app.get('/v1/heal/:labelhash', async (c: Context) => {
+    const labelhash = c.req.param('labelhash');
+    
+    try {
+        // Convert hex string to Buffer, stripping '0x' prefix if present
+        const hashBytes = Buffer.from(labelhash.replace(/^0x/, ''), 'hex');
+        const label = await db.get(hashBytes);
+        return c.json({ healed: label });
```
lightwalker-eth, line 23: What do you think about logging some info to the console about the labelhash / label that was healed?

I'd find it kind of interesting to watch the logs 🤓 😅

apps/ensrainbow/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+        console.error('Error healing label:', error);
+        return c.json({ error: 'Internal server error' }, 500);
+    }
+});
+
+// Health check endpoint
+app.get('/health', (c: Context) => c.json({ status: 'ok' }));
+
+const port = parseInt(process.env.PORT || '3001', 10);
+console.log(`Server starting on port ${port}...`);
```
lightwalker-eth, line 37:
```suggestion
console.log(`ensrainbow server starting on port ${port}...`);
```

apps/ensrainbow/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+        if ((error as any).code === 'LEVEL_NOT_FOUND') {
+            return c.json({ healed: null });
+        }
+        console.error('Error healing label:', error);
+        return c.json({ error: 'Internal server error' }, 500);
+    }
+});
+
+// Health check endpoint
+app.get('/health', (c: Context) => c.json({ status: 'ok' }));
```
lightwalker-eth, line 34: What do you think about validating that the number of key / value pairs in the attached LevelDB is what we expect? Not sure if this is the right place to do that, but I think it would be good to have some validation of this.

Perhaps this isn't something we implement inside of `/health`. Perhaps instead we introduce another endpoint such a `/labels/count` or something like that which returns the count of healable labels?

Please attack this. Very open to suggestions 👍

apps/ensrainbow/src/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import { serve } from '@hono/node-server';
+import { ClassicLevel } from 'classic-level';
+import { join } from 'path';
+import type { Context } from 'hono';
+
+const app = new Hono();
+const DATA_DIR = process.env.DATA_DIR || join(process.cwd(), 'data');
+
+// Initialize LevelDB with proper types for key and value
+const db = new ClassicLevel<Buffer, string>(DATA_DIR, {
```
lightwalker-eth, line 11: A few questions here:

1. What happens if this fails? Ex: `DATA_DIR` is set to something silly. Will a helpful error message be provided to help the user debug the issue?
2. What do you think about logging to the console the `DATA_DIR` that is being loaded?

lightwalker-eth (approved):
@djstrong Really amazing work here 🚀 😄 🙌 !! I've added some suggestions here, but all are pretty low priority and can be done in separate future PRs. Please feel welcome to merge this anytime!

lightwalker-eth:
@djstrong Ah I see the CI is complaining about `ERR_PNPM_OUTDATED_LOCKFILE`.

In this commit: https://github.com/namehash/ensnode/pull/101/commits/314f320cded6946229b0ffdfd30f1060d642f94a I see we made a change to apps/ensrainbow/package.json but there was no accompanying change to apps/ensrainbow/package-lock.json. 

Easiest fix is to delete `package-lock.json` and then rebuild it automatically from `package.json` by running `pnpm install` from the overall monorepo root.

lightwalker-eth (approved):

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+Special thanks to [The Graph Protocol](https://github.com/graphprotocol/ens-rainbow) for their work on the original ENS Rainbow project and rainbow table generation.
+
+## Prerequisites
+
+- Docker installed on your system
+- Node.js v20 or later (for local development)
+
+## Getting the Rainbow Tables
+
+The rainbow tables (6.37 GB) are stored in a public bucket. To download them:
```
lightwalker-eth, line 14:
```suggestion
For backwards compatibility with the ENS Subgraph, the current rainbow tables (6.37 GB) are exactly the same as those published by [The Graph Protocol](https://github.com/graphprotocol/ens-rainbow). Our copies are stored in a public bucket. To download them:
```

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. Download the rainbow table and verify checksum:
```
lightwalker-eth, line 16:
```suggestion
1. Download the rainbow tables and verify checksum:
```

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+```
+Response: `{"healed":"sprayfoamroofingaugusta"}`
+
+If the label is not found: `{"healed":null}`
+
+## Local Development
+
+1. Install dependencies:
+```bash
+npm install
```
lightwalker-eth, line 64: There's several commands documented here that use `npm`. Suggest we change them all to `pnpm`.

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
+# ENSRainbow
+
+ENSRainbow is a label healing sidecar service for ENS names. It provides a simple API endpoint to heal ENS labelhashes back to their original labels.
+
+Special thanks to [The Graph Protocol](https://github.com/graphprotocol/ens-rainbow) for their work on the original ENS Rainbow project and rainbow table generation.
```
lightwalker-eth, line 5:
```suggestion
Special thanks to [The Graph Protocol](https://github.com/graphprotocol/ens-rainbow) for their work on the original ENS rainbow table generation used in the ENS Subgraph.
```

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+
```
lightwalker-eth, line 6: There's a nice opportunity for us to document some key ideas for people who don't know what the ENSRainbow project is all about. For example:

1. The problem: Unknown labels. What are they? How are they formatted? What causes them? Why are they forever a consideration in the ENS protocol? How do they influence indexing ENS data? How do they influence apps (and user interfaces) building on ENS?
2. Related background information: Ex: labels, labelhashes, keccak256, rainbow tables
3. The solution: How ENSRainbow works to mitigate the problem.

I appreciate that this is probably a pretty big task that probably deserves a PR all on its own. I also appreciate that some of these questions might be tricky to answer, so please don't hesitate at all to ask me or anyone else on the team to take the lead on writing about some of these ideas.

Suggest moving the ideas in this comment into its own independent issue and following up on it separately later.

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
+# ENSRainbow
```
lightwalker-eth, line 1: What do you think about updating the `README.md` file in the root of the monorepo so that we have a little directory of links to the apps in the ENSNode project? I think its nice to make `ensrainbow` more discoverable from the root readme.

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+ENSRainbow is a label healing sidecar service for ENS names. It provides a simple API endpoint to heal ENS labelhashes back to their original labels.
+
+Special thanks to [The Graph Protocol](https://github.com/graphprotocol/ens-rainbow) for their work on the original ENS Rainbow project and rainbow table generation.
+
+## Prerequisites
+
+- Docker installed on your system
+- Node.js v20 or later (for local development)
+
+## Getting the Rainbow Tables
```
lightwalker-eth, line 12: Very excited for ENSRainbow!

Perhaps we won't release it this Thursday, but rather the next coming Thursday (ex: Feb 6)? Let's see.

One key idea for us to document here would be to note how this initial release of ENSRainbow is focused on backwards compatibility with the ENS Subgraph, but that we'll be adding enhancements to ENSRainbow soon that will include the ability to heal many more unknown labels than the ENS Subgraph.

I think it's important and valuable for us to plant this seed of an idea in people's minds about what we're working to do it. It will shape their whole perception of the value proposition that we're working towards.

Open to different ideas for where we write these ideas in the readme. Perhaps we add them to the "Architecture Overview" section?

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+4. Start the service:
+```bash
+npm start
+```
+
+## Environment Variables
+
+- `PORT`: Server port (default: 3001)
+- `DATA_DIR`: Directory for LevelDB data (default: './data')
+- `NODE_ENV`: Node environment (default: 'production' in Docker)
```
lightwalker-eth, line 86:
```suggestion
- `NODE_ENV`: Node.js environment (default: 'production' in Docker)
```

Since we're in "ENS Land" there's multiple interpretations for "Node", therefore suggesting the modification above.

Also, could you please include more info on what the purpose of this environment variable is? I understand it's something for Node.js, but is there a special reason we're mentioning it in our readme file here?

apps/ensrainbow/README.md
lightwalker-eth, line 6: There's a nice opportunity for us to add an "Architecture Overview" section to the readme. Happy to start with something simple for this that might at least identify the big picture of how ENSRainbow is implemented.

For example:
* ENSRainbow uses LevelDB as an embedded key-value store to map from labelhash -> label.
* ENSRainbow exposes an API endpoint for healing labelhashes back to their label preimage.

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
+# ENSRainbow
+
+ENSRainbow is a label healing sidecar service for ENS names. It provides a simple API endpoint to heal ENS labelhashes back to their original labels.
+
+Special thanks to [The Graph Protocol](https://github.com/graphprotocol/ens-rainbow) for their work on the original ENS Rainbow project and rainbow table generation.
+
+## Prerequisites
```
lightwalker-eth, line 7: Suggest we also include a section in the readme about how we're operating an instance of ENSRainbow for the ENS community to freely make use of at https://api.ensrainbow.io.

I'm happy to add this to the readme even if this cloud instance hasn't gone live yet. I'm sure we'll get it live quickly!

Here's some inspiration for how we might communicate these ideas: https://github.com/namehash/ensnode/wiki/ENS-Subgraph#namehash-labs-hosted-graph-node-instance

lightwalker-eth (commented):
Sharing a few more suggestions. These are more lower priority ideas and suggest merging this PR in first before actioning these ideas. They can be done in one or more separate future PRs.

apps/ensrainbow/src/index.ts
tk-o, line 23: Perhaps we can just return the text value, as it's the only possible value to be returned:
```suggestion
        return c.text(label);
```

apps/ensrainbow/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Heal Label
+```bash
+curl http://localhost:3001/v1/heal/0x[labelhash]
+```
+Example:
+```bash
+curl http://localhost:3001/v1/heal/0x78441d2a930a233460507c2f25aea5ec5dc278db6aeef9e6ee8b930ccc150e58
+```
+Response: `{"healed":"sprayfoamroofingaugusta"}`
```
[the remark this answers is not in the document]
djstrong: It is the first entry (with lowest node) in the rainbow tables.

apps/ensrainbow/src/index.ts
djstrong: In LevelDB we need to iterate over all entries in DB to count them.

djstrong: And it takes a lot of time to count.

djstrong: In the sql file we have:
```
     37 COPY public.ens_names (hash, name) FROM stdin;
     38 0x78441d2a930a233460507c2f25aea5ec5dc278db6aeef9e6ee8b930ccc150e58      sprayfoamroofingaugusta
     39 0x72b754c974b7b4d51add4abcd1bf97895a4090a2fe32691560b56cc9c5733bdd      flawlessathens


133856931 0x44a48ba7d2c0d67962a8cccf149d6c364c879a21ddc044fbafe461b34fb095c3      minderasercorporation
133856932 0xbaa6df85830434ad8f5ee50734f3a620a73d9dca099dee38c612add2336f9dd6      tanyatest1
133856933 \.
133856934
```
So it is 133856932-38+1-1=133856894

apps/ensrainbow/README.md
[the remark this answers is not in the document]
djstrong: They might be unnormalized and contain anything. Would you like to add some restrictions?

apps/ensrainbow/package.json
djstrong: I have made some changes.

apps/ensrainbow/README.md
djstrong: New issue: https://github.com/namehash/ensnode/issues/117

apps/ensrainbow/README.md
djstrong: Actually, it does nothing now, but we can use it e.g. to show more debug logs. I am removing it from the readme.

apps/ensrainbow/src/index.ts
lightwalker-eth: @djstrong Thanks for investigating this. I created a separate story for us to optimize this later: https://github.com/namehash/ensnode/issues/121

apps/ensrainbow/src/index.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    });
+
+    afterAll(async () => {
+        // Cleanup
+        await server.close();
+    });
+
+    describe('GET /v1/heal/:labelhash', () => {
+        it('should return the label for a valid labelhash', async () => {
+            const validLabelhash = '0x7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d';
```
lightwalker-eth, line 31: It might be confusing for someone reading the code here how `validLabelHash` is not actually the "valid" labelHash for `validLabel`.

Suggest you import viem and use the `labelhash` function for the real labelhash of `validLabel`.

apps/ensrainbow/src/index.test.ts
```diff
@@ -0,0 +1,104 @@
...
+            expect(response.status).toBe(200);
+            const text = await response.text();
+            expect(text).toBe(validLabel);
+        });
+
+        it('should handle missing labelhash parameter', async () => {
+            const response = await fetch('http://localhost:3002/v1/heal/');
+            expect(response.status).toBe(404); // Hono returns 404 for missing parameters
+            const text = await response.text();
+            expect(text).toBe('404 Not Found'); // Hono's default 404 response
```
lightwalker-eth, line 48: Interesting. Perhaps we should add some comment about thisinside the implementation of this API?

For example, I see this code in the implementation:

```
    if (!labelhash) {
        return c.json({ error: 'Missing labelhash parameter' }, 400);
    }
```

That suggested to me that Hono would return exactly that, but according to this test, it doesn't. Not sure if we can configure Hono to handle that differently or if we should just leave Hono the way it is but add some comments in the implementation identifying that the handler won't even be called by Hono if the param is missing.

Advice appreciated.

apps/ensrainbow/src/index.test.ts
```diff
@@ -0,0 +1,104 @@
+/// <reference types="vitest" />
+import { describe, it, expect, beforeAll, afterAll, beforeEach } from 'vitest';
+import { app, db } from './index';
+import { serve } from '@hono/node-server';
+
```
lightwalker-eth, line 5: I created a low priority story with general ideas on the testing strategy here that we can follow up on later: https://github.com/namehash/ensnode/issues/123

lightwalker-eth (approved):
@djstrong Hey super updates! Nice work 🙌 Reviewed and shared feedback. Overall I suggest merging this PR and then following up on the comments in separate future PRs.

Thanks!

apps/ensrainbow/vitest.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
```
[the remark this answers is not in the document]
tk-o: This is configuration required for generating test coverage reports. We don't have that on the ensnode app, but we have the `test:coverage` task defined in `package.json` for the ensrainbow app.

tk-o:
Please let's merge #116 before going live with this PR.

PR #103, opened 2025-01-29
fix: add indexes to events for speedier reference lookups

shrugs:
adds index on `(parent_id, id)` for each type of event. the diff is more verbose than it perhaps needed to be but that's the only change here


lightwalker-eth (approved):
@shrugs Looks good to me!

Issue #79, opened 2025-01-24
do another 21M diff


shrugs:
depends on 16M diff
- [x] non-addresses in accounts #39 
  - no permutations of those addresses show up in the subgraph — is it silently ignoring those events?

depends on 21M diff
  - [x] confirm NameWrapped `name` is set correctly
  - [x] different `coinType` for Resolver `0x84c5adb77dd9f362a1a3480009992d8d47325dc3-0x9c8b7ac505c9f0161bbbd04437fce8c630a0886e1ffea00078e298f063a8a5df` in [this tx](https://etherscan.io/tx/0x1504d282bc89a407b029f9367774e2249f07edc8447db32a64970cebc44bb008)
- [x] Resolver `0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41-0xe6ae31d630cc7a8279c0f1c7cbe6e7064814c47d1785fa2703d9ae511ee2be0c` is missing events after the 50th in ponder (there are 99 in subgraph, 50 in ponder) — likely a default LIMIT or something? also occurs to `"0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41-0x63e78f16a69301cd0843c9e20992e81ac222f19acf67da797a7caf4029ce6b9a`

```gql
query WrappedDomainNoName {
  wrappedDomain(
    id: "0x9e0b4ce6b6f4afd31c698ce0465026f38b90f580a1853253aa6ce44a9e6a6b8f"
  ) {
    id
    name
  }
  
  nameWrappeds(
    where: {domainId: "0x9e0b4ce6b6f4afd31c698ce0465026f38b90f580a1853253aa6ce44a9e6a6b8f"}
  ) {
    id
  }
  
  domain(id: "0x9e0b4ce6b6f4afd31c698ce0465026f38b90f580a1853253aa6ce44a9e6a6b8f") {
    id
    name
    labelName
    events {
      ... on NameWrapped {
        id
        transactionID
      }
    }
  }
}
```

```gql
query Resolvers {
  resolver(id:"0x84c5adb77dd9f362a1a3480009992d8d47325dc3-0x9c8b7ac505c9f0161bbbd04437fce8c630a0886e1ffea00078e298f063a8a5df") {
    id
    coinTypes
  }
}
```

shrugs:
[...]

PR #104, opened 2025-01-29
fix: default to 100 for plurals without limit specified

shrugs:
self explanatory, fixes an issue tracked in #79


lightwalker-eth (approved):
LGTM

PR #105, opened 2025-01-29
fix: dont upsertAccount for multicoin address newAddress

### LINKED ISSUE #39: 'missing' some accounts in 21MM snapshot
seems to be an out-of-order response for the Accounts query, that's a lot of missing accounts for likely no reason. should be lexigraphically ordered, however, so unclear.

shrugs:
fixes #39 

i naively called `upsertAccount` for account values in MulticoinAddrChanged events, but that is both incorrect and unnecessary.

note the lack of Account access in this handler: https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/resolver.ts#L59


lightwalker-eth (approved):
Good fix :+1:

PR #116, opened 2025-01-30
fix(docker): use monorepo context for docker build

tk-o:
This PR fixes docker build by using optimised layers and building apps from directly monorepo root directory.
Also, automated code style fixes were applied on the existing codebase.

## Run it yourself

https://github.com/namehash/ensnode/blob/91c670ff4963a604627bbf0510c312fdd1fea09f/apps/ensrainbow/README.md?plain=1#L51-L64

## Docker image build

Stats for docker build on MBP M1 Max

image


.github/workflows/ci.yml
```diff
@@ -83,3 +86,47 @@ jobs:
...
+  ensraninbow-docker-build:
+    runs-on: ubuntu-latest
+    needs: [static-analysis, unit-tests]
+    if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/alpha' || github.ref == 'refs/heads/subgraph')
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Build and push Docker image
```
lightwalker-eth, line 99:
```suggestion
      - name: Build and push Docker image to GitHub
```

Is that right? Goal is to make it more explicit where it is being pushed to

.github/workflows/ci.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+      - alpha
+      - subgraph
```
lightwalker-eth, line 8: Hmm. Do I understand this right? Would this create publish docker builds to GitHub each time we even make even make a PR to alpha or subgraph?

As I understand, we only want to publish docker builds to GitHub when we merge into main.

Advice appreciated. I'm pretty weak on devops.

.github/workflows/ci.yml
```diff
@@ -83,3 +86,47 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [ensraninbow-docker-build]
+    
+    steps:
+      - name: Download image artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: docker-image
+          
+      - name: Load and push Docker image
```
lightwalker-eth, line 124:
```suggestion
      - name: Load and push Docker image to GitHub
```

Is that right? Goal is to make it more explicit where it is being pushed to

It's also not so clear to me what's the difference between "Load and push Docker image" and "Build and push Docker image" -- but I'm really weak at DevOps. Just sharing my thoughts.

apps/ensrainbow/Dockerfile
```diff
@@ -1,54 +1,58 @@
...
-RUN pnpm install --prod
-
+COPY --from=ensrainbow.runtime_deps /app/apps/ensrainbow/package*.json .
+COPY --from=ensrainbow.runtime_deps /app/apps/ensrainbow/node_modules node_modules
+COPY --from=ensrainbow.runtime_deps /app/apps/ensrainbow/dist dist
+
+# Ingestor stage
+FROM ensrainbow.runtime AS ingestor
+# Copy input data
+COPY --from=ensrainbow.data /app/ens_names.sql.gz .
```
lightwalker-eth, line 44: I don't know anything about Docker really, but would this be making multiple copies of the ~7GB file? Could we avoid making copies? Not sure if relevant. Please feel welcome to ignore.

lightwalker-eth (commented):
@tk-o Nice to see this 😄 I'm not good with Docker / Devops, so will leave it to @shrugs and/or @djstrong to give the approval on this PR.

I've shared some small comments.

apps/ensrainbow/Dockerfile
tk-o: What gets exported as the final docker image is defined below the last `FROM` directive in a docker file. So in our case, it'd be just the Level DB store (`COPY --from=ingestor /app/apps/ensrainbow/data ./data`).

tk-o: All other copies remain in the context of the host which was running `docker build` command.

shrugs (approved):
love it, LGTM

PR #131, opened 2025-01-30
feat: add slow query logging in subgraph-compat graphql

shrugs:


lightwalker-eth (approved):
@shrugs Looks good, nice idea :+1:

PR #132, opened 2025-01-30
fix: correctly implement dns packet decoding from subgraph

shrugs:
this fixes one of the issues in #79 regarding `WrappedDomain.name` being null — my original re-implementation of the dns packet decoding helper was subtly incorrect. this version goes line-by-line and mirrors the subgraph implementation more closely. i've included a test case to verify against the specific instance that surfaced this issue.

https://github.com/ensdomains/ens-subgraph/blob/master/src/nameWrapper.ts#L30


tk-o (approved):
LGTM

lightwalker-eth (approved):
Looks good :+1:

PR #133, opened 2025-01-30
ENSRainbow: More fixes

djstrong:
For some comments from https://github.com/namehash/ensnode/pull/101

### DIFF
--- a/apps/ensrainbow/package.json
+++ b/apps/ensrainbow/package.json
@@ -28,7 +28,8 @@
     "classic-level": "^1.4.1",
     "hono": "catalog:",
     "progress": "^2.0.3",
-    "tsx": "^4.7.1"
+    "tsx": "^4.7.1",
+    "viem": "catalog:"
   },
   "devDependencies": {
     "@namehash/shared-configs": "workspace:*",
--- a/apps/ensrainbow/package.json
+++ b/apps/ensrainbow/package.json
@@ -19,9 +19,9 @@
     "ingest:prod": "node dist/ingest.js",
     "docker:build": "docker build -t ensnode/ensrainbow .",
     "docker:push": "docker push ensnode/ensrainbow",
-    "test": "vitest run",
-    "test:watch": "vitest",
-    "test:coverage": "vitest run --coverage"
+    "test": "DATA_DIR=test-data vitest run",
+    "test:watch": "DATA_DIR=test-data vitest",
+    "test:coverage": "DATA_DIR=test-data vitest run --coverage"
   },
   "dependencies": {
     "@hono/node-server": "^1.4.1",
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -1,4 +1,3 @@
-import { join } from "path";
 import { serve } from "@hono/node-server";
 /// <reference types="vitest" />
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -1,6 +1,7 @@
 import { serve } from "@hono/node-server";
 /// <reference types="vitest" />
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { labelhash } from "viem";
 import { app, db } from "./index";
 
 describe("ENS Rainbow API", () => {
@@ -28,8 +29,8 @@ describe("ENS Rainbow API", () => {
 
   describe("GET /v1/heal/:labelhash", () => {
     it("should return the label for a valid labelhash", async () => {
-      const validLabelhash = "0x7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d7d";
       const validLabel = "test-label";
+      const validLabelhash = labelhash(validLabel);
 
       // Add test data
       const hashBytes = Buffer.from(validLabelhash.replace(/^0x/, ""), "hex");
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -103,7 +103,7 @@ describe("ENS Rainbow API", () => {
   });
 
   describe("LevelDB operations", () => {
-    it("should handle values containing null bytes", async () => {
+    it("should store labels containing null bytes", async () => {
       const labelWithNull = "test\0label";
       const labelWithNullLabelhash = labelhash(labelWithNull);
       const labelHashBytes = labelHashToBytes(labelWithNullLabelhash);
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -110,4 +110,16 @@ describe("ENS Rainbow API", () => {
       expect(data.count).toBe(3);
     });
   });
+
+  describe("LevelDB operations", () => {
+    it("should handle values containing null bytes", async () => {
+      const hash = "0x1111111111111111111111111111111111111111111111111111111111111111";
+      const hashBytes = Buffer.from(hash.replace(/^0x/, ""), "hex");
+      const labelWithNull = "test\0label";
+
+      await db.put(hashBytes, labelWithNull);
+      const retrieved = await db.get(hashBytes);
+      expect(retrieved).toBe(labelWithNull);
+    });
+  });
 });
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -85,7 +85,7 @@ describe("ENS Rainbow API", () => {
     it("should return correct count of entries", async () => {
       // Add some test data
       const testData = ["test1", "test2", "test3"].map(label => ({
-        hash: labelhash(label),
+        labelHashBytes: labelHashBytes(label),
         label: label
       }));
 
@@ -104,7 +104,7 @@ describe("ENS Rainbow API", () => {
   describe("LevelDB operations", () => {
     it("should handle values containing null bytes", async () => {
       const labelWithNull = "test\0label";
-      const labelWithNullLabelhash = labelhash(labelWithNull);
+      const labelHashBytes = labelHashBytes(labelWithNull);
       const labelHashBytes = Buffer.from(hexToBytes(labelWithNullLabelhash));
 
       await db.put(labelHashBytes, labelWithNull);
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -1,7 +1,7 @@
 import { serve } from "@hono/node-server";
 /// <reference types="vitest" />
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
-import { labelhash } from "viem";
+import { labelhash, hexToBytes } from "viem";
 import { app, db } from "./index";
 
 describe("ENS Rainbow API", () => {
@@ -33,8 +33,8 @@ describe("ENS Rainbow API", () => {
       const validLabelhash = labelhash(validLabel);
 
       // Add test data
-      const hashBytes = Buffer.from(validLabelhash.replace(/^0x/, ""), "hex");
-      await db.put(hashBytes, validLabel);
+      const labelHashBytes = Buffer.from(hexToBytes(validLabelhash));
+      await db.put(labelHashBytes, validLabel);
 
       const response = await fetch(`http://localhost:3002/v1/heal/${validLabelhash}`);
       expect(response.status).toBe(200);
@@ -53,7 +53,7 @@ describe("ENS Rainbow API", () => {
       const response = await fetch("http://localhost:3002/v1/heal/invalid-hash");
       expect(response.status).toBe(400);
       const data = await response.json();
-      expect(data).toEqual({ error: "Invalid labelhash - must be a 32 byte hex string" });
+      expect(data).toEqual({ error: "Invalid labelhash - must be a valid hex string" });
     });
 
     it("should handle non-existent labelhash", async () => {
@@ -84,24 +84,14 @@ describe("ENS Rainbow API", () => {
 
     it("should return correct count of entries", async () => {
       // Add some test data
-      const testData = [
-        {
-          hash: "0x1111111111111111111111111111111111111111111111111111111111111111",
-          label: "test1",
-        },
-        {
-          hash: "0x2222222222222222222222222222222222222222222222222222222222222222",
-          label: "test2",
-        },
-        {
-          hash: "0x3333333333333333333333333333333333333333333333333333333333333333",
-          label: "test3",
-        },
-      ];
+      const testData = ["test1", "test2", "test3"].map(label => ({
+        hash: labelhash(label),
+        label: label
+      }));
 
       for (const entry of testData) {
-        const hashBytes = Buffer.from(entry.hash.replace(/^0x/, ""), "hex");
-        await db.put(hashBytes, entry.label);
+        const labelHashBytes = Buffer.from(hexToBytes(entry.hash));
+        await db.put(labelHashBytes, entry.label);
       }
 
       const response = await fetch("http://localhost:3002/v1/labels/count");
@@ -113,12 +103,12 @@ describe("ENS Rainbow API", () => {
 
   describe("LevelDB operations", () => {
     it("should handle values containing null bytes", async () => {
-      const hash = "0x1111111111111111111111111111111111111111111111111111111111111111";
-      const hashBytes = Buffer.from(hash.replace(/^0x/, ""), "hex");
       const labelWithNull = "test\0label";
+      const labelWithNullLabelhash = labelhash(labelWithNull);
+      const labelHashBytes = Buffer.from(hexToBytes(labelWithNullLabelhash));
 
-      await db.put(hashBytes, labelWithNull);
-      const retrieved = await db.get(hashBytes);
+      await db.put(labelHashBytes, labelWithNull);
+      const retrieved = await db.get(labelHashBytes);
       expect(retrieved).toBe(labelWithNull);
     });
   });
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -1,3 +1,4 @@
+import { join } from "path";
 import { serve } from "@hono/node-server";
 /// <reference types="vitest" />
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
@@ -54,7 +55,7 @@ describe("ENS Rainbow API", () => {
       const response = await fetch("http://localhost:3002/v1/heal/invalid-hash");
       expect(response.status).toBe(400);
       const data = await response.json();
-      expect(data).toEqual({ error: "Labelhash must be 0x-prefixed" });
+      expect(data).toEqual({ error: "Invalid labelhash length 12 characters (expected 66)" });
     });
 
     it("should handle non-existent labelhash", async () => {
@@ -85,7 +86,7 @@ describe("ENS Rainbow API", () => {
 
     it("should return correct count of entries", async () => {
       // Add some test data
-      const testData = ["test1", "test2", "test3"].map(label => ({
+      const testData = ["test1", "test2", "test2", "test3"].map(label => ({
         hash: labelhash(label),
         label: label
       }));
--- a/apps/ensrainbow/src/index.test.ts
+++ b/apps/ensrainbow/src/index.test.ts
@@ -1,7 +1,8 @@
 import { serve } from "@hono/node-server";
 /// <reference types="vitest" />
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "vitest";
-import { labelhash, hexToBytes } from "viem";
+import { labelhash } from "viem";
+import { labelHashToBytes } from "./utils/label-utils";
 import { app, db } from "./index";
 
 describe("ENS Rainbow API", () => {
@@ -33,7 +34,7 @@ describe("ENS Rainbow API", () => {
       const validLabelhash = labelhash(validLabel);
 
       // Add test data
-      const labelHashBytes = Buffer.from(hexToBytes(validLabelhash));
+      const labelHashBytes = labelHashToBytes(validLabelhash);
       await db.put(labelHashBytes, validLabel);
 
       const response = await fetch(`http://localhost:3002/v1/heal/${validLabelhash}`);
@@ -53,7 +54,7 @@ describe("ENS Rainbow API", () => {
       const response = await fetch("http://localhost:3002/v1/heal/invalid-hash");
       expect(response.status).toBe(400);
       const data = await response.json();
-      expect(data).toEqual({ error: "Invalid labelhash - must be a valid hex string" });
+      expect(data).toEqual({ error: "Labelhash must be 0x-prefixed" });
     });
 
     it("should handle non-existent labelhash", async () => {
@@ -85,12 +86,12 @@ describe("ENS Rainbow API", () => {
     it("should return correct count of entries", async () => {
       // Add some test data
       const testData = ["test1", "test2", "test3"].map(label => ({
-        labelHashBytes: labelHashBytes(label),
+        hash: labelhash(label),
         label: label
       }));
 
       for (const entry of testData) {
-        const labelHashBytes = Buffer.from(hexToBytes(entry.hash));
+        const labelHashBytes = labelHashToBytes(entry.hash);
         await db.put(labelHashBytes, entry.label);
       }
 
@@ -104,8 +105,8 @@ describe("ENS Rainbow API", () => {
   describe("LevelDB operations", () => {
     it("should handle values containing null bytes", async () => {
       const labelWithNull = "test\0label";
-      const labelHashBytes = labelHashBytes(labelWithNull);
-      const labelHashBytes = Buffer.from(hexToBytes(labelWithNullLabelhash));
+      const labelWithNullLabelhash = labelhash(labelWithNull);
+      const labelHashBytes = labelHashToBytes(labelWithNullLabelhash);
 
       await db.put(labelHashBytes, labelWithNull);
       const retrieved = await db.get(labelHashBytes);
--- a/apps/ensrainbow/src/index.ts
+++ b/apps/ensrainbow/src/index.ts
@@ -27,15 +27,21 @@ try {
 }
 
 app.get("/v1/heal/:labelhash", async (c: Context) => {
-  const labelhash = c.req.param("labelhash").toLowerCase();
+  const labelhash = c.req.param("labelhash");
+  const prefixedLabelHash = (labelhash.startsWith('0x') ? labelhash : `0x${labelhash}`) as `0x${string}`;
 
-  if (!isHex(labelhash) || size(labelhash) !== 32) {
-    return c.json({ error: "Invalid labelhash - must be a 32 byte hex string" }, 400);
+  let labelHashBytes: Uint8Array;
+  try {
+    labelHashBytes = hexToBytes(prefixedLabelHash);
+    if (labelHashBytes.length !== 32) {
+      return c.json({ error: "Invalid labelhash - must be a 32 byte hex string" }, 400);
+    }
+  } catch (error) {
+    return c.json({ error: "Invalid labelhash - must be a valid hex string" }, 400);
   }
 
   try {
-    const hashBytes = Buffer.from(hexToBytes(labelhash));
-    const label = await db.get(hashBytes);
+    const label = await db.get(Buffer.from(labelHashBytes));
     console.info(`Successfully healed labelhash ${labelhash} to label "${label}"`);
     return c.text(label);
   } catch (error) {
--- a/apps/ensrainbow/src/index.ts
+++ b/apps/ensrainbow/src/index.ts
@@ -3,6 +3,7 @@ import { serve } from "@hono/node-server";
 import { ClassicLevel } from "classic-level";
 import { Hono } from "hono";
 import type { Context } from "hono";
+import { isHex, size, hexToBytes } from 'viem'
 
 export const app = new Hono();
 export const DATA_DIR = process.env.VITEST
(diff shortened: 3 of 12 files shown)


package.json
```diff
@@ -16,5 +16,8 @@
   },
   "resolutions": {
     "vite": "5.4.12"
+  },
+  "dependencies": {
+    "viem": "catalog:"
```
djstrong, line 21: is this correct?

shrugs: probably not at the root package.json — you likely want this in the apps/ensrainbow/package.json, where i see you've got it already. so we can remove this change entirely

apps/ensrainbow/src/index.ts
```diff
@@ -24,11 +26,7 @@ try {
...
 
 app.get("/v1/heal/:labelhash", async (c: Context) => {
-  const labelhash = c.req.param("labelhash")?.toLowerCase();
-
-  if (!labelhash) {
-    return c.json({ error: "Missing labelhash parameter" }, 400);
-  }
+  const labelhash = c.req.param("labelhash").toLowerCase();
 
   if (!/^(0x)?[0-9a-f]{64}$/.test(labelhash)) {
```
shrugs, line 31: let's replace this fn with `isHex(labelhash) && size(labelhash) === 32` to remove our custom regex

(`import { isHex, size } from 'viem'`)

shrugs: likewise below, can you see if `viem#hexToBytes` can replace the line like

```
const hashBytes = Buffer.from(labelhash.replace(/^0x/, ""), "hex");
```

shrugs (commented):
let's wrap the console.info on L43 about the unhealable labelhash request with an `if (process.env.NODE_ENV === 'development')` because i think that'll create a LOT of logs in production

apps/ensrainbow/src/index.ts
djstrong: @lightwalker-eth What do you think? Do you want logs about unhealable labelhashes on production?

djstrong:
@lightwalker-eth What do you think? Do you want logs about unhealable labelhashes on production?

apps/ensrainbow/src/index.ts
```diff
@@ -24,19 +27,14 @@ try {
...
 app.get("/v1/heal/:labelhash", async (c: Context) => {
-  const labelhash = c.req.param("labelhash")?.toLowerCase();
+  const labelhash = c.req.param("labelhash").toLowerCase();
 
-  if (!labelhash) {
-    return c.json({ error: "Missing labelhash parameter" }, 400);
-  }
-
-  if (!/^(0x)?[0-9a-f]{64}$/.test(labelhash)) {
+  if (!isHex(labelhash) || size(labelhash) !== 32) {
```
tk-o, line 32: Very nice

apps/ensrainbow/src/ingest.ts
```diff
@@ -57,8 +63,7 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
         const [hashVal, name] = parts;
         if (hashVal && name) {
```
tk-o, line 64: Following `isHex` usage from the other file on this PR:
```suggestion
        if (isHex(hashVal) && name) {
```

apps/ensrainbow/src/ingest.ts
```diff
@@ -73,8 +78,6 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
           }
         }
       }
-    } else {
-      bar.tick();
     }
```
tk-o, line 81: Now, that we can the `else` clause gone, we could flatten these nested ifs a bit. What do you think, @djstrong?

From:

https://github.com/namehash/ensnode/blob/1f63b6f2694bfa488efe2a93677acfae9d5962d5/apps/ensrainbow/src/ingest.ts#L60-L82

to:

```ts
if (!isCopySection) {
  continue;
}

const parts = line.trim().split("\t");
if (parts.length !== 2) {
  continue;
}

const [hashVal, name] = parts;

if (!isHex(hashVal) || !name) {
  continue;
}

try {
  const hashBytes = Buffer.from(hexToBytes(hashVal));
  batch.put(hashBytes, name);
  batchSize++;

  if (batchSize >= MAX_BATCH_SIZE) {
    await batch.write();
    batch = db.batch();
    batchSize = 0;
  }
  bar.tick();
} catch (e) {
  console.error(`Error processing hash: ${e} '${hashVal}'`);
}
```

lightwalker-eth:
@djstrong Agree with suggestion from @shrugs

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +57,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
-            console.error(`Error processing hash: ${e} '${hashVal}'`);
-          }
-        }
+    if (!isCopySection) {
+      continue;
+    }
+
+    const parts = line.trim().split("\t");
+    if (parts.length !== 2) {
+      continue;
```
lightwalker-eth, line 66: @djstrong Not sure if my memory is correct about this? I seem to remember that you posted an example in Slack of an invalid value in this input file?

What do you think about:
1. Adding some warning log messages if any line doesn't conform to the expected formatting (and therefore is discarded).
2. Running ingestion to make a list of any lines with bad formatting.
3. If there are any lines with bad formatting: Creating a GitHub issue that we can follow up on later with a list of all lines discovered to have bad formatting. If no lines have bad formatting, that's great. I still like to keep the code that would log warnings though.

apps/ensrainbow/src/ingest.ts
```diff
@@ -15,6 +16,11 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
     keyEncoding: "binary",
   });
 
+  // Total number of lines in the ENS rainbow table SQL dump
+  // This number represents the count of unique label-labelhash pairs
+  // as of January 30, 2024 from the Graph Protocol's ENS rainbow tables
+  // Source file: ens_names.sql.gz
+  // SHA256: a6316b1e7770b1f3142f1f21d4248b849a5c6eb998e3e66336912c9750c41f31
   const TOTAL_LINES = 133_856_894;
```
lightwalker-eth, line 24: Thanks for the docs here! Very nice!

Is this actually the number of lines in the .sql file, or is it the number of records in the .sql file? I understand these would be different, as not all lines have records.

Additionally, after performing the full ingestion, seems it would be nice to validate that the number of records ingested matched the expected number. What do you think?

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +57,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+    }
+
+    const parts = line.trim().split("\t");
+    if (parts.length !== 2) {
+      continue;
+    }
+
+    const [hashVal, name] = parts;
+    if (!isHex(hashVal) || size(hashVal) !== 32 || !name) {
+      continue;
```
lightwalker-eth, line 71: The ideas shared in my comment above apply here too.

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +57,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+    if (!isCopySection) {
+      continue;
+    }
+
+    const parts = line.trim().split("\t");
+    if (parts.length !== 2) {
+      continue;
+    }
+
+    const [hashVal, name] = parts;
```
lightwalker-eth, line 69:
```suggestion
    const [labelHash, label] = parts;
```

I assume the above is more accurate?

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +57,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+    if (!isHex(hashVal) || size(hashVal) !== 32 || !name) {
```
lightwalker-eth, line 70: Instead of calling `isHex` and `size` here, it might be better to just call `hexToBytes` in a try / catch block. If that fails then you know it wasn't hex. Then after the try / catch you can have a check if the returned bytes are the expected length.

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +57,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+      continue;
+    }
+
+    try {
+      const hashBytes = Buffer.from(hexToBytes(hashVal as Hex));
```
lightwalker-eth, line 75:
```suggestion
      const labelHashBytes = Buffer.from(hexToBytes(hashVal as Hex));
```

apps/ensrainbow/src/index.test.ts
```diff
@@ -109,4 +110,16 @@ describe("ENS Rainbow API", () => {
       expect(data.count).toBe(3);
     });
   });
+
+  describe("LevelDB operations", () => {
+    it("should handle values containing null bytes", async () => {
+      const hash = "0x1111111111111111111111111111111111111111111111111111111111111111";
+      const hashBytes = Buffer.from(hash.replace(/^0x/, ""), "hex");
+      const labelWithNull = "test\0label";
```
lightwalker-eth, line 118: This is a nice test case, but suggest we always stay pure to the label -> labelhash relationships in every line we write.

Therefore, let's not define `hash` as it is defined here. Instead it should be calculated dynamically from `labelhash(labelWithNull)`.

apps/ensrainbow/src/index.ts
```diff
@@ -3,9 +3,12 @@ import { serve } from "@hono/node-server";
 import { ClassicLevel } from "classic-level";
 import { Hono } from "hono";
 import type { Context } from "hono";
+import { isHex, size, hexToBytes } from 'viem'
 
 export const app = new Hono();
-export const DATA_DIR = process.env.DATA_DIR || join(process.cwd(), "data");
+export const DATA_DIR = process.env.VITEST
```
lightwalker-eth, line 9: What's the `VITEST` environment variable? If you want to customize the `DATA_DIR` during testing, why not just customize that environment variable during testing? Why introduce `VITEST`?

lightwalker-eth (changes requested):
@djstrong Thanks for these updates. Reviewed and shared feedback :+1:

apps/ensrainbow/src/index.ts
djstrong: This is automatic env variable during testing.

apps/ensrainbow/src/ingest.ts
djstrong: The log looks now:
```
Loading data into LevelDB...
Processing [                                        ] 10000/133856894 lines (0%) - 84034 lines/sec - 1592.8s remaining
Invalid hex format for hash: "hash"
Processing [========================================] 133856894/133856894 lines (100%) - 110292 lines/sec - 0.0s remaining

Data loading complete!
Successfully processed all 133856894 records
Done!
```
So there is only one problematic line.

Do you want to save this log entry in a list and print once again at the end of ingestion?

apps/ensrainbow/src/index.test.ts
```diff
@@ -84,24 +84,14 @@ describe("ENS Rainbow API", () => {
...
-          hash: "0x2222222222222222222222222222222222222222222222222222222222222222",
-          label: "test2",
-        },
-        {
-          hash: "0x3333333333333333333333333333333333333333333333333333333333333333",
-          label: "test3",
-        },
-      ];
+      const testData = ["test1", "test2", "test3"].map(label => ({
+        hash: labelhash(label),
```
lightwalker-eth, line 88: Nice improvements here! Please see my comment above for more details.

```suggestion
        labelHashBytes: labelHashBytes(label),
```

apps/ensrainbow/src/ingest.ts
```diff
@@ -63,18 +64,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
     const parts = line.trim().split("\t");
     if (parts.length !== 2) {
+      console.warn(`Invalid line format - expected 2 columns but got ${parts.length}: "${line.slice(0, 100)}"`);
       continue;
     }
 
-    const [hashVal, name] = parts;
-    if (!isHex(hashVal) || size(hashVal) !== 32 || !name) {
+    const [labelHash, label] = parts;
+    if (!label) {
```
lightwalker-eth, line 72: I think this check here is impossible to ever be true based on the `if (parts.length !== 2)` check up above?

apps/ensrainbow/src/ingest.ts
```diff
@@ -63,18 +64,33 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+      console.warn(`Missing label for hash ${labelHash}`);
       continue;
     }
 
+    let labelHashBytes: Buffer;
```
lightwalker-eth, line 77: The logic here is great! A few suggestions, this relates to the other comment I shared suggesting a `labelHashBytes` utility function.

Suggest transferring the logic you wrote here into a new utility function called something like `labelHashToBytes` (please note the "To" in that name, which makes it distinct from `labelHashBytes`).

`labelHashToBytes` would take as input a `Hex` (that hopefully should be properly formatted).

Then, it should use all the validation logic you implemented here to ensure its a valid hex format and exactly 32 bytes in length.

After you do that, suggest updating the `labelHashBytes` function such that it calls `labelHashToBytes` internally.

Then.. you could make use of the `labelHashToBytes` function here 😄 

Oh yeah, one more thing.. you might wonder how to handle different error messages for different issues in these helper functions. A pattern for this is to remove any `console.warn` inside the helper functions, and instead just throw an error with a message on it. Then, in a context like here inside ingest.ts, you can wrap the call to the utility functions in a try / catch, and then put the message from the error that you catch into console.warn.

Please feel welcome to ask questions if that doesn't make sense 😅

apps/ensrainbow/src/ingest.ts
lightwalker-eth: @djstrong Awesome! Great check here! Will follow up separately inside issue 140. I think no need to print again at the end of ingestion.

apps/ensrainbow/src/ingest.ts
```diff
@@ -16,19 +16,19 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
     keyEncoding: "binary",
   });
 
-  // Total number of lines in the ENS rainbow table SQL dump
+  // Total number of valid records in the ENS rainbow table SQL dump
```
lightwalker-eth, line 19:
```suggestion
  // Total number of expected records in the ENS rainbow table SQL dump
```

apps/ensrainbow/src/ingest.ts
```diff
@@ -16,19 +16,19 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
   // This number represents the count of unique label-labelhash pairs
   // as of January 30, 2024 from the Graph Protocol's ENS rainbow tables
   // Source file: ens_names.sql.gz
   // SHA256: a6316b1e7770b1f3142f1f21d4248b849a5c6eb998e3e66336912c9750c41f31
-  const TOTAL_LINES = 133_856_894;
+  const TOTAL_VALID_RECORDS = 133_856_894;
```
lightwalker-eth, line 24:
```suggestion
  const TOTAL_EXPECTED_RECORDS = 133_856_894;
```

apps/ensrainbow/src/ingest.ts
```diff
@@ -94,6 +110,13 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
 
   await db.close();
   console.log("\nData loading complete!");
+  
+  // Validate the number of processed records
+  if (processedRecords !== TOTAL_VALID_RECORDS) {
+    console.warn(`Warning: Expected ${TOTAL_VALID_RECORDS} records but processed ${processedRecords}`);
+  } else {
+    console.log(`Successfully processed all ${processedRecords} records`);
```
lightwalker-eth, line 118:
```suggestion
    console.log(`Successfully ingested all ${processedRecords} records`);
```

apps/ensrainbow/src/ingest.ts
```diff
@@ -43,6 +43,7 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
   let isCopySection = false;
   let batch = db.batch();
   let batchSize = 0;
+  let processedRecords = 0;
```
lightwalker-eth, line 46:
```suggestion
  let ingestedRecords = 0;
```

apps/ensrainbow/src/ingest.ts
```diff
@@ -94,6 +110,13 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
 
   await db.close();
   console.log("\nData loading complete!");
```
lightwalker-eth, line 112:
```suggestion
  console.log("\nData ingestion complete!");
```

apps/ensrainbow/src/index.ts
```diff
@@ -24,20 +27,21 @@ try {
 }
 
 app.get("/v1/heal/:labelhash", async (c: Context) => {
-  const labelhash = c.req.param("labelhash")?.toLowerCase();
+  const labelhash = c.req.param("labelhash");
```
shrugs, line 30: personally I'm ok with requiring that the input be 0x-prefixed and not allowing for any other format. i.e. required. this tightens the api surface in a very reasonable way. i.e. this section can simply be:

```ts
import { isHex, size } from 'viem';

// ...

const labelhash = c.req.param("labelhash");
if (!isHex(labelhash) || size(labelhash) !== 32) {
  return c.json({ error: "labelhash must be a 32-byte hex string" }, 400);
}
```

apps/ensrainbow/src/index.ts
```diff
@@ -24,20 +27,21 @@ try {
...
+    }
+  } catch (error) {
+    return c.json({ error: "Invalid labelhash - must be a valid hex string" }, 400);
   }
 
   try {
-    // Convert hex string to Buffer, stripping '0x' prefix if present
-    const hashBytes = Buffer.from(labelhash.replace(/^0x/, ""), "hex");
-    const label = await db.get(hashBytes);
+    const label = await db.get(Buffer.from(labelHashBytes));
```
shrugs, line 44: now we can do `const label = await db.get(Buffer.from(hexToBytes(labelhash)))` without additional checks

apps/ensrainbow/src/index.ts
```diff
@@ -27,15 +27,21 @@ try {
...
 app.get("/v1/heal/:labelhash", async (c: Context) => {
-  const labelhash = c.req.param("labelhash").toLowerCase();
+  const labelhash = c.req.param("labelhash");
+  const prefixedLabelHash = (labelhash.startsWith('0x') ? labelhash : `0x${labelhash}`) as `0x${string}`;
 
-  if (!isHex(labelhash) || size(labelhash) !== 32) {
-    return c.json({ error: "Invalid labelhash - must be a 32 byte hex string" }, 400);
+  let labelHashBytes: Uint8Array;
+  try {
+    labelHashBytes = hexToBytes(prefixedLabelHash);
```
lightwalker-eth, line 35: Please see my other messages about helper functions. They will be useful here too 😄

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +58,48 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+      continue;
+    }
+
+    const [labelHash, label] = parts;
+    if (!label) {
+      console.warn(`Missing label for hash ${labelHash}`);
+      continue;
+    }
+
+    let labelHashBytes: Buffer;
```
shrugs, line 77: these checks can/should use similar logic to the recommendation above — perhaps we can extract that into an

```ts
const isValidLabelhash = (labelhash: string): labelhash is Hex => isHex(labelhash) && size(labelhash) === 32;
```

shrugs (changes requested):
sorry, requesting a minor change on how we do the label hash validity logic, just to tidy things up and standardize on the viem helper fns

apps/ensrainbow/src/index.ts
lightwalker-eth: Agreed. I'm happy to tighten this up too.

apps/ensrainbow/src/ingest.ts
```diff
@@ -4,25 +4,33 @@ import { createInterface } from "readline";
...
 
 const DATA_DIR = process.env.DATA_DIR || join(process.cwd(), "data");
 const INPUT_FILE = process.env.INPUT_FILE || join(process.cwd(), "ens_names.sql.gz");
 
+// Total number of expected records in the ENS rainbow table SQL dump
+// This number represents the count of unique label-labelhash pairs
+// as of January 30, 2024 from the Graph Protocol's ENS rainbow tables
+// Source file: ens_names.sql.gz
+// SHA256: a6316b1e7770b1f3142f1f21d4248b849a5c6eb998e3e66336912c9750c41f31
+const TOTAL_EXPECTED_RECORDS = 133_856_894;
```
lightwalker-eth, line 18: What about the record that is invalid? How does it relate to this count? Suggest making a note about this detail here and including a link to https://github.com/namehash/ensnode/issues/140

apps/ensrainbow/src/index.test.ts
```diff
@@ -109,4 +101,16 @@ describe("ENS Rainbow API", () => {
       expect(data.count).toBe(3);
     });
   });
+
+  describe("LevelDB operations", () => {
+    it("should handle values containing null bytes", async () => {
```
lightwalker-eth, line 106:
```suggestion
    it("should store labels containing null bytes", async () => {
```

apps/ensrainbow/src/index.test.ts
```diff
@@ -83,24 +85,14 @@ describe("ENS Rainbow API", () => {
...
-        {
-          hash: "0x2222222222222222222222222222222222222222222222222222222222222222",
-          label: "test2",
-        },
-        {
-          hash: "0x3333333333333333333333333333333333333333333333333333333333333333",
-          label: "test3",
-        },
-      ];
+      const testData = ["test1", "test2", "test3"].map(label => ({
```
lightwalker-eth, line 88: Perhaps include "test2" twice, the final count should still be 3.

apps/ensrainbow/src/index.ts
lightwalker-eth: It might be convenient (for the person who wrote this code) while testing, but I don't find it convenient for myself or others who come along later and need to review this code to try to figure out what is happening here. For example: there's some environment variable being used here that's not documented in our README. Factors like this reduce confidence in our work. Why are we mixing up the ideas of our testing framework dependencies directly into our application code?

We need to ensure all our code is clear and easy to read. Whatever the goal is here, let's please look for a different strategy to achieve it.

lightwalker-eth (changes requested):
@djstrong Thanks for your updates here. Reviewed and shared feedback :+1:

apps/ensrainbow/src/ingest.ts
lightwalker-eth: Ah I see this was resolved already in PR 148 :+1:

apps/ensrainbow/src/index.ts
djstrong: You are right. Moved test data directory into tests.

djstrong: But it doesn't work:
```
process.env.DATA_DIR = join(process.cwd(), "test-data");
```

So, now it is in package.json:
```
    "test": "DATA_DIR=test-data vitest run",
    "test:watch": "DATA_DIR=test-data vitest",
    "test:coverage": "DATA_DIR=test-data vitest run --coverage"
```

apps/ensrainbow/src/utils/label-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Converts a label hash to bytes, with validation
+ * @param labelHash The label hash to convert
+ * @returns A Buffer containing the bytes
+ * @throws Error if the input is not a valid 32-byte hex string
+ */
+export function labelHashToBytes(labelHash: Hex): Buffer {
+  try {
+    const bytes = hexToBytes(labelHash);
```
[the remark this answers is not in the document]
djstrong: Ok!
The check:
```
    if (bytes.length !== 32) { // should be redundant but keeping it for the principle of defensive programming
      throw new Error(`Invalid labelhash length ${bytes.length} bytes (expected 32)`);
    }
```
seems to be redundant but keeping it for now.

apps/ensrainbow/src/index.ts
```diff
@@ -23,26 +27,33 @@ try {
...
-    // Convert hex string to Buffer, stripping '0x' prefix if present
-    const hashBytes = Buffer.from(labelhash.replace(/^0x/, ""), "hex");
-    const label = await db.get(hashBytes);
+    const label = await db.get(labelHashBytes);
     console.info(`Successfully healed labelhash ${labelhash} to label "${label}"`);
     return c.text(label);
   } catch (error) {
     if ((error as any).code === "LEVEL_NOT_FOUND") {
-      console.info(`Unhealable labelhash request: ${labelhash}`);
+      if (process.env.NODE_ENV === 'development') {
```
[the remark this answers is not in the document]
djstrong: Created issue: https://github.com/namehash/ensnode/issues/152

apps/ensrainbow/src/ingest.ts
```diff
@@ -51,30 +60,37 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
-            bar.tick();
-          } catch (e) {
-            console.error(`Error processing hash: ${e} '${hashVal}'`);
-          }
-        }
+    if (!isCopySection) {
+      continue;
+    }
+
+    const parts = line.trim().split("\t");
```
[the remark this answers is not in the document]
djstrong: Created issue for enabling this mode: https://github.com/namehash/ensnode/issues/152

apps/ensrainbow/src/ingest.ts
```diff
@@ -84,7 +100,14 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+  
+  // Validate the number of processed records
+  if (processedRecords !== TOTAL_EXPECTED_RECORDS) {
+    console.warn(`Warning: Expected ${TOTAL_EXPECTED_RECORDS} records but processed ${processedRecords}`);
+  } else {
+    console.log(`Successfully ingested all ${processedRecords} records`);
+  }
 }
 
 // Check if this module is being run directly
```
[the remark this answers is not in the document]
djstrong: It is just for running the ingest script and not running the ingestion when the file is imported (e.g. in tests).

I will implement CLI app and move the code so it can imported without running the ingestion: https://github.com/namehash/ensnode/issues/152

apps/ensrainbow/src/utils/rainbow-record.ts
```diff
@@ -0,0 +1,54 @@
+import { ByteArray, labelhash } from 'viem';
+import { Labelhash } from '../../../../packages/ensnode-utils/src/types';
```
lightwalker-eth, line 2: Suggest we ask @tk-o or @shrugs to help optimize how we import internally from other packages in our monorepo.

apps/ensrainbow/src/index.ts
```diff
@@ -3,17 +3,19 @@ import { serve } from "@hono/node-server";
...
 console.log(`Initializing ENS Rainbow with data directory: ${DATA_DIR}`);
 
-export let db: ClassicLevel<Buffer, string>;
+export let db: ClassicLevel<ByteArray, string>;
 
 // Initialize database with error handling
 try {
-  db = new ClassicLevel<Buffer, string>(DATA_DIR, {
+  db = new ClassicLevel<ByteArray, string>(DATA_DIR, {
     valueEncoding: "utf8",
```
lightwalker-eth, line 19: Suggest to flip the order that you wrote these fields.

In other words, put the `keyEncoding` before the `valueEncoding`. People normally think key -> value in that order.

apps/ensrainbow/package.json
```diff
@@ -19,16 +19,17 @@
     "ingest:prod": "node dist/ingest.js",
     "docker:build": "docker build -t ensnode/ensrainbow .",
     "docker:push": "docker push ensnode/ensrainbow",
-    "test": "vitest run",
-    "test:watch": "vitest",
-    "test:coverage": "vitest run --coverage"
+    "test": "DATA_DIR=test-data vitest run",
+    "test:watch": "DATA_DIR=test-data vitest",
+    "test:coverage": "DATA_DIR=test-data vitest run --coverage"
```
lightwalker-eth, line 24: Nice!

apps/ensrainbow/src/utils/label-utils.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  // Valid case for reference
+  it('should convert a valid 32-byte hex string', () => {
+    const validHash = '0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef' as Labelhash;
+    expect(() => labelHashToBytes(validHash)).not.toThrow();
+  });
+
+  // Test case: labelhash without 0x prefix
+  it('should throw error when labelhash does not begin with 0x', () => {
+    const noPrefix = '0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef';
+    // @ts-expect-error - Testing invalid input
```
lightwalker-eth, line 15: We should remove any directives such as @ts-expect-error

apps/ensrainbow/src/utils/label-utils.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(() => labelHashToBytes(noPrefix)).toThrow('Invalid labelhash length 64 characters (expected 66)');
+  });
+
+  // Test case: labelhash with 1x prefix
+  it('should throw error when labelhash begins with 1x', () => {
+    const onePrefix = '1x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef';
+    // @ts-expect-error - Testing invalid input
```
lightwalker-eth, line 22: We should remove any directives such as @ts-expect-error

lightwalker-eth (approved):
@djstrong Awesome! Thank you! I left a few other small suggestions but a lot of work into this PR already. Approved 🚀 We can merge this in and then follow up on the suggestions in another future PR.

apps/ensrainbow/src/utils/rainbow-record.ts
```diff
@@ -0,0 +1,54 @@
...
+    throw new Error(`Invalid line format - expected 2 columns but got ${parts.length}: "${line.slice(0, 100)}"`);
+  }
+
+  const [maybeLabelHash, label] = parts;
+  const labelHash = labelHashToBytes(maybeLabelHash as Labelhash);
+
+  if (options.validateLabelHash) {
+    const computedLabelHash = labelHashToBytes(labelhash(label) as Labelhash);
+
+    if (computedLabelHash.length !== labelHash.length) {
```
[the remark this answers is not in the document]
djstrong: Resolved in https://github.com/namehash/ensnode/pull/150

apps/ensrainbow/src/index.ts
djstrong: Resolved in https://github.com/namehash/ensnode/pull/150

apps/ensrainbow/src/utils/rainbow-record.ts
djstrong: I have created an issue: https://github.com/namehash/ensnode/issues/154

apps/ensrainbow/src/utils/label-utils.test.ts
djstrong: Created issue: https://github.com/namehash/ensnode/issues/155

apps/ensrainbow/src/utils/label-utils.test.ts
djstrong: Created issue: https://github.com/namehash/ensnode/issues/155

djstrong:
All comments from this PR should be fixed or addressed in new issues.

Issue #85, opened 2025-01-27
ENSNode: Integrate ENSRainbow (Label Healing Sidecar)


shrugs:
depends on #84 

- [ ] create helper in indexer app to query label healing service
- [ ] in two locations in ponder indexer handlers, call that helper
- [ ] that's it

lightwalker-eth:
@tk-o Hey, I saw your question in Slack about the goals of this issue (85) vs the goals of issue 122: https://github.com/namehash/ensnode/issues/122.

I'm happy to separate these goals, such that in this issue (85) we're first focused on "just make it work", and then in issue 122 we build a more mature solution.

A key goal that I do see as part of this issue (85) is the creation of a new `ensrainbow-sdk` package that at a minimum would contain the type definitions associated with ENSRainbow responses. These type definitions need to be shared between both ENSRainbow and ENSNode.

lightwalker-eth:
Solved by https://github.com/namehash/ensnode/pull/170

PR #134, opened 2025-01-31
feat(ensrainbow): introduce label healing

tk-o:
Fixes https://github.com/namehash/ensnode/issues/85

This PR introduces a label healing feature, where each unknown label that's detected onchain is first checked in the ENSRainbow API for its healed counterpart. If one is found, it is used as a replacement within `labelName` & `name` properties on the `domain` entity.

To configure a specific instance of ENSRainbow API, please use `ENSRAINBOW_API_BASE_URL` env variable. If not set, or incorrect URL was provided, no healing will be performed.

#### Label healing off

image

#### Label healing on

image

### DIFF
--- a/apps/ensnode/src/handlers/NameWrapper.ts
+++ b/apps/ensnode/src/handlers/NameWrapper.ts
@@ -6,6 +6,7 @@ import type { Node } from "ensnode-utils/types";
 import { type Address, type Hex, hexToBytes, namehash } from "viem";
 import { createSharedEventValues, upsertAccount } from "../lib/db-helpers";
 import { makeEventId } from "../lib/ids";
+import { labelHealing } from "../lib/label-healing";
 import { bigintMax } from "../lib/lib-helpers";
 import { EventWithArgs } from "../lib/ponder-helpers";
 import type { OwnedName } from "../lib/types";
@@ -105,10 +106,16 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
       await upsertAccount(context, owner);
 
       // decode the name emitted by NameWrapper
-      const [label, name] = decodeDNSPacketBytes(hexToBytes(event.args.name));
+      let [label, name] = decodeDNSPacketBytes(hexToBytes(event.args.name));
 
       // upsert the healed name iff valid
       if (label) {
+        // try to heal the label (and name) if possible
+        const labelHealingResult = await labelHealing.heal(label, name);
+        if (labelHealingResult) {
+          [label, name] = labelHealingResult;
+        }
+
         await context.db.update(schema.domain, { id: node }).set({ labelName: label, name });
       }
 
--- a/apps/ensnode/src/handlers/Registrar.ts
+++ b/apps/ensnode/src/handlers/Registrar.ts
@@ -5,6 +5,7 @@ import type { Labelhash } from "ensnode-utils/types";
 import { type Hex, labelhash as _labelhash, namehash } from "viem";
 import { createSharedEventValues, upsertAccount, upsertRegistration } from "../lib/db-helpers";
 import { makeRegistrationId } from "../lib/ids";
+import { labelHealing } from "../lib/label-healing";
 import { EventWithArgs } from "../lib/ponder-helpers";
 import type { OwnedName } from "../lib/types";
 
@@ -42,9 +43,18 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
     if (!domain) throw new Error("domain expected in setNamePreimage but not found");
 
     if (domain.labelName !== name) {
+      let domainLabel = name;
+      let domainName = `${name}.${ownedName}`;
+
+      // try to heal the label (and name) if possible
+      const labelHealingResult = await labelHealing.heal(domainLabel, domainName);
+      if (labelHealingResult) {
+        [domainLabel, domainName] = labelHealingResult;
+      }
+
       await context.db
         .update(schema.domain, { id: node })
-        .set({ labelName: name, name: `${name}.${ownedName}` });
+        .set({ labelName: domainLabel, name: domainName });
     }
 
     await context.db
--- a/apps/ensnode/src/handlers/Registry.ts
+++ b/apps/ensnode/src/handlers/Registry.ts
@@ -6,6 +6,7 @@ import type { Labelhash, Node } from "ensnode-utils/types";
 import { type Hex, zeroAddress } from "viem";
 import { createSharedEventValues, upsertAccount, upsertResolver } from "../lib/db-helpers";
 import { makeResolverId } from "../lib/ids";
+import { labelHealing } from "../lib/label-healing";
 import { EventWithArgs } from "../lib/ponder-helpers";
 import { OwnedName } from "../lib/types";
 
@@ -128,8 +129,14 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
 
           // TODO: implement sync rainbow table lookups
           // https://github.com/ensdomains/ens-subgraph/blob/master/src/ensRegistry.ts#L111
-          const labelName = encodeLabelhash(label);
-          const name = parent?.name ? `${labelName}.${parent.name}` : labelName;
+          let labelName = encodeLabelhash(label);
+          let name = parent?.name ? `${labelName}.${parent.name}` : labelName;
+
+          // try to heal the label (and name) if possible
+          const labelHealingResult = await labelHealing.heal(labelName, name);
+          if (labelHealingResult) {
+            [labelName, name] = labelHealingResult;
+          }
 
           await context.db.update(schema.domain, { id: domain.id }).set({ name, labelName });
         }
--- a/apps/ensnode/src/lib/label-healing.ts
+++ b/apps/ensnode/src/lib/label-healing.ts
@@ -1,6 +1,6 @@
 import { isUnknownLabel, unknownLabelAsHex } from "ensnode-utils/subname-helpers";
 
-interface LabelHealing {
+interface ENSRainbow {
   /**
    * Attempts to heal the given label. For healed labels,
    * it also heals the name if provided.
@@ -12,75 +12,61 @@ interface LabelHealing {
 }
 
 /**
- * A no-op implementation of `LabelHealing` that does nothing.
- * Useful when no valid ENSRainbow API URL is provided.
- */
-const labelHealingNoOp: LabelHealing = Object.freeze({
-  async heal() {
-    return null;
-  },
-});
-
-/**
- * A factory function to
+ * A factory function to create an instance of ENSRainbow.
  * @param ensRainbowApiUrl
  * @returns
  */
-function createLabelHealing(ensRainbowApiUrl: string | undefined): LabelHealing {
-  try {
-    if (!ensRainbowApiUrl) {
-      console.warn("LabelHealing: no ENSRainbow API URL was provided.");
-      return labelHealingNoOp;
-    }
+function createLabelHealing(ensRainbowApiUrl: string | undefined): ENSRainbow {
+  /**
+   * Fetches the healed label for the given unknown label.
+   * @param label any label
+   * @returns healed label if available
+   */
+  async function fetchHealedLabel(label: string) {
+    try {
+      const ensRainbowApiUrlParsed = ensRainbowApiUrl ? new URL(ensRainbowApiUrl) : null;
 
-    const ensRainbowApiUrlParsed = new URL(ensRainbowApiUrl);
-
-    /**
-     * Fetches the healed label for the given unknown label.
-     * @param label any label
-     * @returns healed label if available
-     */
-    async function fetchHealedLabel(label: string) {
-      try {
-        // first, let's make sure this is an encoded unknown label
-        if (!isUnknownLabel(label)) {
-          return null;
-        }
-
-        const response = await fetch(
-          new URL(`/v1/heal/${unknownLabelAsHex(label)}`, ensRainbowApiUrlParsed).toString(),
-        );
-
-        if (!response.ok) {
-          console.error("Failed to heal label", label, response.statusText);
-          return null;
-        }
-
-        return response.text();
-      } catch (error) {
-        console.error("Failed to heal label", label, error);
+      // If the ENSRainbow API URL is not provided, we can't heal the label
+      if (!ensRainbowApiUrlParsed) {
+        console.warn("LabelHealing: no ENSRainbow API URL was provided.");
         return null;
       }
+
+      // first, let's make sure this is an encoded unknown label
+      if (!isUnknownLabel(label)) {
+        return null;
+      }
+
+      const response = await fetch(
+        new URL(`/v1/heal/${unknownLabelAsHex(label)}`, ensRainbowApiUrlParsed).toString(),
+      );
+
+      if (!response.ok) {
+        console.error("Failed to heal label", label, response.statusText);
+        return null;
+      }
+
+      return response.text();
+    } catch (error) {
+      console.error("Failed to heal label", label, error);
+      return null;
     }
-
-    return {
-      async heal(label, name) {
-        const healedLabel = await fetchHealedLabel(label);
-
-        if (!healedLabel) {
-          return null;
-        }
-
-        // If the name is provided, try to heal it as well
-        const healedName = name ? name.replaceAll(label, healedLabel) : null;
-
-        return [healedLabel, healedName] as [typeof label, typeof name];
-      },
-    };
-  } catch (error: any) {
-    console.error(`LabelHealing: failed to created instance. ${error.message}`);
-    return labelHealingNoOp;
   }
+
+  return {
+    async heal(label, name) {
+      const healedLabel = await fetchHealedLabel(label);
+
+      if (!healedLabel) {
+        return null;
+      }
+
+      // If the name is provided, try to heal it as well
+      const healedName = name ? name.replaceAll(label, healedLabel) : null;
+
+      return [healedLabel, healedName] as [typeof label, typeof name];
+    },
+  } satisfies ENSRainbow;
 }
 
 export const labelHealing = createLabelHealing(process.env.ENSRAINBOW_API_URL);
--- a/apps/ensnode/src/lib/label-healing.ts
+++ b/apps/ensnode/src/lib/label-healing.ts
@@ -1,4 +1,4 @@
-import { isUnknownLabel, unknownLabelAsHex } from "ensnode-utils/subname-helpers";
+import { decodeLabelhash } from "@ensdomains/ensjs/utils";
 import { Hex } from "viem";
 
 interface ENSRainbow {
@@ -35,13 +35,8 @@ function createLabelHealing(ensRainbowApiUrl: URL): ENSRainbow {
 
   return {
     async heal(label, name) {
-      // first, let's make sure this is an encoded unknown label
-      if (!isUnknownLabel(label)) {
-        return null;
-      }
-
       try {
-        const healedLabel = await fetchHealedLabel(unknownLabelAsHex(label));
+        const healedLabel = await fetchHealedLabel(decodeLabelhash(label));
 
         if (!healedLabel) {
           return null;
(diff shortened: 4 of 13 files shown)


packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -50,14 +50,24 @@ export const uint256ToHex32 = (num: bigint) => toHex(num, { size: 32 });
 const UNINDEXABLE_LABEL_CHARACTERS = [
   "\0", // null byte: PostgreSQL does not allow storing this character in text fields.
   ".", // conflicts with ENS label separator logic.
-  "[", // conflicts with "unknown label" representations.
-  "]", // conflicts with "unknown label" representations.
 ];
```
djstrong, line 53: this may break `isLabelIndexable`

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -58,6 +58,18 @@ const UNINDEXABLE_LABEL_CHARACTER_CODES = new Set(
...
 
+const UNKNOWN_LABEL_PREFIX = "[" as const;
+const UNKNOWN_LABEL_SUFFIX = "]" as const;
+
+type UnknownLabel = `${typeof UNKNOWN_LABEL_PREFIX}${string}${typeof UNKNOWN_LABEL_SUFFIX}`;
+
+export const isUnknownLabel = (label: string): label is UnknownLabel =>
+  label.startsWith(UNKNOWN_LABEL_PREFIX) &&
+  label.endsWith(UNKNOWN_LABEL_SUFFIX) &&
+  label.length === 66;
```
tk-o, line 69: @djstrong did you mean something like:
```suggestion
  label.startsWith(UNKNOWN_LABEL_PREFIX) &&
  label.endsWith(UNKNOWN_LABEL_SUFFIX) &&
  label.length === 66 &&
  isHex(unknownLabelAsHex(label));
```

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -50,14 +50,24 @@ export const uint256ToHex32 = (num: bigint) => toHex(num, { size: 32 });
...
```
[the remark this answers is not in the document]
tk-o: Cheers, updated ;)

packages/ensnode-utils/src/subname-helpers.ts
djstrong: yes

djstrong:
One name can have more than one unknown label in itself - may it cause some problems here?

packages/ensnode-utils/src/subname-helpers.ts
tk-o: Reverted

packages/ensnode-utils/src/subname-helpers.ts
tk-o: Done

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        // If the name is provided, try to heal it as well
+        const healedName = name ? name.replaceAll(label, healedLabel) : null;
+
+        return [healedLabel, healedName] as [typeof label, typeof name];
+      },
+    };
+  } catch (error: any) {
+    console.error(`LabelHealing: failed to created instance. ${error.message}`);
+    return labelHealingNoOp;
```
lightwalker-eth, line 82: Please see my other comments about the importance of us giving successful defaults.

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    return {
+      async heal(label, name) {
+        const healedLabel = await fetchHealedLabel(label);
+
+        if (!healedLabel) {
+          return null;
+        }
+
+        // If the name is provided, try to heal it as well
```
lightwalker-eth, line 74: Better to move such logic into a different layer.

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -128,8 +129,14 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
 
           // TODO: implement sync rainbow table lookups
           // https://github.com/ensdomains/ens-subgraph/blob/master/src/ensRegistry.ts#L111
-          const labelName = encodeLabelhash(label);
-          const name = parent?.name ? `${labelName}.${parent.name}` : labelName;
+          let labelName = encodeLabelhash(label);
```
lightwalker-eth, line 132:
```suggestion
          let unknownLabel = encodeLabelhash(label);
```

When we encode a labelhash, it generates an "unknown label".

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -128,8 +129,14 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
+          let name = parent?.name ? `${labelName}.${parent.name}` : labelName;
+
+          // try to heal the label (and name) if possible
```
lightwalker-eth, line 135: Is there a special reason why we are building `name` before we make the call to ENSRainbow?

Why not first figure out what the childmost label should be represented as, and only after that build `name`?

Additionally, I'm pretty sure it would be a bug if you healed more than just the childmost label within name. This could create a situation where a given labelhash has both known and unknown representations at the same time within our indexed data.

We should take special care to ensure that's impossible. All representations of a given labelhash should always be equivalent within our indexed state.

Please ask me if any questions.

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -15,6 +16,12 @@ describe("isLabelIndexable", () => {
     expect(isLabelIndexable("test]")).toBe(false);
   });
 
+  it("should return true for unknown labels", () => {
+    expect(
+      isLabelIndexable("[72c29f4186361a46935e4e9c3af71d1cf73cac00186fceb1cd1945ed9ed3dec1]"),
```
lightwalker-eth, line 21: This breaks ENS Subgraph backwards compatibility.

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -26,6 +33,30 @@ describe("isLabelIndexable", () => {
...
 });
 
+describe("isUnknownLabel", () => {
+  it("should return true for unknown labels", () => {
+    expect(
+      isUnknownLabel("[72c29f4186361a46935e4e9c3af71d1cf73cac00186fceb1cd1945ed9ed3dec1]"),
+    ).toBe(true);
+  });
+
+  it("should return false for non-unknown labels", () => {
```
lightwalker-eth, line 43: Just because the length of the string between [ and ] is the expected length doesn't mean it's an unknown label. The value between those characters must also contain only hex digits.

apps/ensnode/src/handlers/NameWrapper.ts
```diff
@@ -105,10 +106,16 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
...
       // decode the name emitted by NameWrapper
-      const [label, name] = decodeDNSPacketBytes(hexToBytes(event.args.name));
+      let [label, name] = decodeDNSPacketBytes(hexToBytes(event.args.name));
 
       // upsert the healed name iff valid
       if (label) {
+        // try to heal the label (and name) if possible
+        const labelHealingResult = await labelHealing.heal(label, name);
+        if (labelHealingResult) {
+          [label, name] = labelHealingResult;
```
lightwalker-eth, line 116: I believe we should only be healing the childmost label for now.

Please consider how our indexing logic works:

Events are guaranteed to start from the overall ENSRoot (with name "") and then proceed down the name hierarchy from there.

In other words, consider the name abc.123.456.eth.

It should be guaranteed that if we have an event for abc.123.456.eth that we have already indexed 123.456.eth, which includes healing all of its labels to the furthest extent possible. Therefore, when we are inserting abc.123.456.eth into our indexed data for the first time, we should only be concerned with healing the childmost label (if necessary).

apps/ensnode/src/handlers/NameWrapper.ts
```diff
@@ -105,10 +106,16 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
       await upsertAccount(context, owner);
 
       // decode the name emitted by NameWrapper
-      const [label, name] = decodeDNSPacketBytes(hexToBytes(event.args.name));
+      let [label, name] = decodeDNSPacketBytes(hexToBytes(event.args.name));
 
       // upsert the healed name iff valid
       if (label) {
+        // try to heal the label (and name) if possible
+        const labelHealingResult = await labelHealing.heal(label, name);
```
lightwalker-eth, line 114: I'm really confused here. Why are we passing a "label" to `heal`? Only labelhash values should be passed. Is `label` actually a `labelhash`?

We really need to fix this stuff! It creates endless problems and confusion!

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -42,9 +43,18 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
     if (!domain) throw new Error("domain expected in setNamePreimage but not found");
 
     if (domain.labelName !== name) {
+      let domainLabel = name;
+      let domainName = `${name}.${ownedName}`;
+
+      // try to heal the label (and name) if possible
```
lightwalker-eth, line 49: Please see my suggestions about similar code in other files.

lightwalker-eth (changes requested):
Please review feedback. Thanks.

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,69 @@
...
+  ): Promise<[Label, Name] | null>;
+}
+
+const labelHealingNoOp: LabelHealing = Object.freeze({
+  async heal() {
+    return null;
+  },
+});
+
+function createLabelHealing(ensRainbowBaseUrl: string | undefined): LabelHealing {
```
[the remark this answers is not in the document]
tk-o: This is how I name service origin URL (protocol+hostname+port). I can drop `Base` from it to avoid creating confusion. Would `ensRainbowApiUrl work?

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,69 @@
...
+  try {
+    if (!ensRainbowBaseUrl) {
```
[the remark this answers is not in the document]
tk-o: We define a default value, but also write code defensively so it's easier to debug if something is not configured well. Of course, I can make the code to assume the URL is always there. But will it be there for sure?

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const ensRainbowBaseUrlParsed = new URL(ensRainbowBaseUrl);
+
+    async function fetchHealedLabel(label: string) {
+      try {
+        if (!isUnknownLabel(label)) {
+          return null;
+        }
+
+        const labelHex = unknownLabelAsHex(label);
```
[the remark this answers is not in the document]
tk-o: So how do we call values like `[e5e14487b78f85faa6e1808e89246cf57dd34831548ff2e6097380d98db2504a]`? Would it be an `encodedLabelhash`?

apps/ensnode/src/lib/label-healing.ts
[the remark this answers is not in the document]
tk-o: Fixed that with creating a const with default service URL.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -67,6 +80,8 @@ const UNINDEXABLE_LABEL_CHARACTER_CODES = new Set(
 export const isLabelIndexable = (label: string) => {
   if (!label) return false;
 
+  if (isUnknownLabel(label)) return true;
```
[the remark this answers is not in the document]
tk-o: Yeah, maybe we should have a feature flag to require Subgraph compatibility.

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,86 @@
...
+ * Useful when no valid ENSRainbow API URL is provided.
+ */
+const labelHealingNoOp: LabelHealing = Object.freeze({
+  async heal() {
+    return null;
+  },
+});
+
+/**
+ * A factory function to
```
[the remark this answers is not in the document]
tk-o: I think classes vs factory functions is a personal preference. Can make it a class, if you find it more straightforward.

apps/ensnode/src/lib/label-healing.ts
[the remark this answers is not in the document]
lightwalker-eth: @tk-o We can call it either an "unknown label" or an "encoded labelhash".

An "unknown label" is represented as an "encoded labelhash", since it is unknown.

packages/ensnode-utils/src/subname-helpers.ts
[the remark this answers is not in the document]
lightwalker-eth: Yes agreed. I think we should introduce such a feature flag as an environment variable. Perhaps create a separate issue for this though, because when we add that new environment variable there will be some other factors to consider to do it right. For example: If you set the subgraph compatibility feature flag then we should also validate the expected plugins are activated, etc.

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,86 @@
+import { isUnknownLabel, unknownLabelAsHex } from "ensnode-utils/subname-helpers";
+
+interface LabelHealing {
+  /**
+   * Attempts to heal the given label. For healed labels,
+   * it also heals the name if provided.
+   */
+  heal<Label extends string, Name extends string | null>(
```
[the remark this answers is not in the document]
lightwalker-eth: Sorry I don't understand your question.

I'm suggesting that the `heal` function here should maximally align with the "heal" API in ENSRainbow. Therefore there should be no `name` param at all here.

apps/ensnode/src/lib/label-healing.ts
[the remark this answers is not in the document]
lightwalker-eth: Yes please.

packages/ensnode-utils/src/subname-helpers.ts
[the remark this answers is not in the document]
lightwalker-eth: Created an issue for this opportunity: https://github.com/namehash/ensnode/issues/137

apps/ensnode/src/lib/label-healing.ts
```diff
@@ -0,0 +1,61 @@
...
+      throw new Error(`Failed to fetch healed label: ${response.statusText}`);
+    }
+
+    return response.text();
+  }
+
+  return {
+    async heal(label, name) {
+      try {
+        const healedLabel = await fetchHealedLabel(decodeLabelhash(label));
```
shrugs, line 39: does this need to decodeLabelHash? we should only ever pass in `heal(labelhash: Labelhash)` where Labelhash is a 32byte hex string. i don't think there's a case in which we pass this function an encoded labelhash

```
const label = await labelHealing.heal(labelhash); // Promise<string | null>
```

apps/ensnode/src/lib/label-healing.ts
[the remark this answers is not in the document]
shrugs: agree, the name parameter and whatever logic is associated with that should not be in this service and should instead be implemented at the callsite, like the subgraph does (for now, at least).

apps/ensnode/src/lib/label-healing.ts
shrugs: let's also include our `isLabelIndexable` check here, and this service should return null if the label is not indexable.

packages/ensnode-utils/src/subname-helpers.spec.ts
shrugs: correct, that label should not be considered indexable — it is an encoded label not the label value itself

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -42,9 +43,18 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
     if (!domain) throw new Error("domain expected in setNamePreimage but not found");
 
     if (domain.labelName !== name) {
+      let domainLabel = name;
```
shrugs, line 46: lmk if i'm missing something but this diff should literally just be, [based on the subgraph code](https://github.com/ensdomains/ens-subgraph/blob/master/src/ethRegistrar.ts#L56)

```ts
const labelName = await labelHealing.heal(labelhash);
const name = labelName ? `${labelName}.${ownedName}` : undefined;

// ...

await context.db.update(schema.domain, { id: node }).set({
  registrantId: owner,
  labelName,
  name, // added
});
```

apps/ensnode/src/handlers/Registry.ts
shrugs: this code should minimc the subgraph and the diff should look like:

```ts
const healedLabel = await labelHealing.heal(label);
const labelName = healedLabel || encodeLabelhash(label);
const name = parent?.name ? `${labelName}.${parent.name}` : labelName;

await context.db.update(schema.domain, { id: domain.id }).set({
  name,
  // only set domain.labelName iff label is healed and indexable
  // https://github.com/ensdomains/ens-subgraph/blob/master/src/ensRegistry.ts#L113
  labelName: healedLabel,
});
```

shrugs (changes requested):
I don't love that we're defaulting to people using the public instance (and encouraging that behavior with the .env.example) — that will generate a _ton_ of traffic every time someone runs the project, not to mention that because it's a remote api indexing will be notably slower for everyone involved.

If that's something we want to handle (in which case we should really set up cloudflare's cdn & cache control headers with cache immutability) then that's fine, but my initial expectation was that we would require users run their own in the local network (therefore _much_ faster queries) in order to run the ensnode indexer. I also think that the indexer shouldn't default to the public instance and should instead fail hard and fast on startup if the label healing service isn't reachable. And it should crash at runtime if it ever becomes unreachable. Anything else sacrifices subgraph compatibility, which is our number one goal for v1.

cc @lightwalker-eth for thoughts on the above ☝️ 

---

it's been addresses in the reviews, but the label healing api service should have an api surface of:

```ts
// NOTE: undefined return type is helpful for passing to drizzle, which will ignore the input rather that set the value in the db to `NULL` if we were to return `null`
function heal(labelhash: Labelhash): Promise<string | undefined>
```

and that function should broadly just look like

```ts
async function getHealedLabel(labelhash: Labelhash) Promise<string | undefined> {
  // await fetch etc
  // if timeout or error code >= 500, throw Error
  // if any other error code, return undefined
  // otherwise return response.healed (or whatever it is)
}

async function heal(labelhash: Labelhash) {
  // optional runtime check, ENS rainbow encodes this logic already
  if (!isHex(labelhash) || size(labelhash) !== 32) return undefined;
  const label = await getHealedLabel(labelhash);

  // if the indexer can't store the label yet, consider it unhealable
  if (!isLabelIndexable(label)) return undefined;

  // if the label isn't normalized, consider it unhealable
  if (label !== normalize(label)) return undefined;

  // this is a healed, normalized label that the indexer can store
  return label;
}
```

and then at the callsite we can write code that matches the subgraph like mentioned in the review comments

packages/ensnode-utils/src/subname-helpers.spec.ts
shrugs: does this test pass?? it shouldn't

lightwalker-eth:
@tk-o Hey, over the weekend Krzysztof and I have been making some advancements with ENSRainbow across several PRs. Highly suggest you wait on any further action on this PR until after all of the following PRs are merged first:

1. https://github.com/namehash/ensnode/pull/153
2. https://github.com/namehash/ensnode/pull/150
3. https://github.com/namehash/ensnode/pull/149

lightwalker-eth:
All of the above linked items that Krzysztof was working on have now been merged :+1: Continued progress on this PR is unblocked.

tk-o:
Replaced with
- #170

PR #142, opened 2025-01-31
fix: remove event index name

### LINKED ISSUE #126: event indexes not showing up in final schema?
could just be that i did my index without the events in the shared schema

shrugs:
closes #126 

giving the index a unique name meant that drizzle only applied the index to the first table that used the helper — letting drizzle autogenerate the idnex name ensures that each table gets a composite index


lightwalker-eth (approved):
@shrugs Great fix :+1:

PR #145, opened 2025-01-31
fix: temporarily store coinTypes as string array to avoid parsing bugs

shrugs:
this provides incorrect graphql typings but correct graphql responses (since bigints are represented as string in json anyway). so it's effectively subgraph compatible. when the bugs are fixed we can revert to correct behavior


lightwalker-eth (approved):
@shrugs Nice work tracking this down and reporting the bug / unit test to Ponder :+1:

PR #147, opened 2025-01-31
fix: stop upserting addresses in multicoinaddresschanged

shrugs:
probably some merge conflict when we wrapped `makeResolverHandlers` but the bugfix in #105 got overwritten so this is that again


lightwalker-eth (approved):
@shrugs Thanks. Approved x2 😄

Issue #140, opened 2025-01-31
One line in The Graph's ens_names.sql is wrong


djstrong:
Related discussion: https://github.com/namehash/ensnode/pull/133/files/1a7cc053d94c69a63e7e3150f5243a09e3c27bdf..9dd66d4861352e1f4a6f81a8f8e783d854b777ee#r1937584451

The ingestion script reports: `Invalid hex format for hash: "hash"`
This is on line 10878:
```
...
  10877 0x76c4c7c44afc5e13caa6822c74ea0416e54bcea48425174abcb1529615deab98      eedwardsgroup
  10878 hash    name
  10879 0x7b142872438233babf2827c95242d40a3670f35c3314aff073fd86b5c95fb0dc      syncktools
...
```

lightwalker-eth:
@djstrong Hahah, great catch!

# ENS Subgraph Source Code Research

Inside the ENS Subgraph code, the function `ens.nameByHash` is used. Usage can be seen here: https://github.com/search?q=repo%3Aensdomains%2Fens-subgraph%20ens.nameByHash&type=code

The implementation of this function is not in the ENS Subgraph, but instead is inside GraphNode: https://github.com/graphprotocol/graph-node

GraphNode implements this function as `ens_name_by_hash`: https://github.com/graphprotocol/graph-node/blob/070072be30b16718691d208435ac8fa360563a0f/runtime/wasm/src/host_exports.rs#L1048

This in turn calls `ens_lookup.find_name`. Which later calls into: https://github.com/graphprotocol/graph-node/blob/master/store/postgres/src/subgraph_store.rs#L1277-L1280

Then.. finally we get to the sql they generate through an ORM here: https://github.com/graphprotocol/graph-node/blob/070072be30b16718691d208435ac8fa360563a0f/store/postgres/src/primary.rs#L1779-L1788

![Image](https://github.com/user-attachments/assets/ceff6535-79b3-48a3-b9b6-23b9d4dcaa33)

The database schema used by GraphNode is: https://github.com/graphprotocol/graph-node/blob/070072be30b16718691d208435ac8fa360563a0f/store/postgres/migrations/2019-05-09-232642_ens_names/up.sql

# Conclusions

1. The GraphNode's database schema treats both `hash` and `name` fields as simple `varchar` with no validation other than `hash` is unique and `name` is not null.
2. The GraphNode implementation of `ens.nameByHash` is ultimately just a simple select statement. They don't perform any special preprocessing or validation.
[...]

PR #148, opened 2025-02-01
ENSRainbow: Add comments about Invalid record in ens rainbow tables

djstrong:
https://github.com/namehash/ensnode/issues/140


lightwalker-eth (approved):
@djstrong Thanks for the upgrade :+1:

Issue #121, opened 2025-01-30
ENSRainbow: Optimize Label Count API


lightwalker-eth:
# Goals

This story is about the `/v1/labels/count` API.

Suggest that we not count all records in LabelDB with each request to this endpoint.

Instead, let's make this an O(1) operation.

# Background

1. LevelDB provides a single keyspace.
2. (As I understand) our keys are all fixed width byte values currently? (Not strings)

# Suggestions

1. Introduce a special key that symbolically represents "label count" that is guaranteed to not collide with labelhash keys. There could be many strategies for this. Ex: use a byte value that's guaranteed to not collide, etc.. Open to different ideas.
2. Update the ingest step to also write the total count to this special "label count" key.
3. Make use of this "label count" key in the `/v1/labels/count` API.

djstrong:
We can do this. However, it may be problematic, because while putting an entry to database we don't know if the key exists. So if in the ENS rainbow tables are duplicated labelhashes then we will count each of them and in database will be less actually. Also rerunning the ingest will make it problematic - we can clear the database before each ingest.

lightwalker-eth:
@djstrong Great point!

One idea then is that before we write a batch, we try to read that batch and then identify if any of those keys already exist? If they do, we could remove them from the batch that would be written and modify how much we increment the record count accordingly.

And I like the idea of clearing the db before each ingest. :+1:

djstrong:
It will significantly degrade performance of writing.

I suggest clearing the db before each ingest and keep in memory all labelhashes to ensure there is no duplicates (it will increase RAM requirements).

lightwalker-eth:
@djstrong Good feedback, thanks :+1:

What do you think about a different strategy?

What if we implement the counting logic as "step 2" of completing an ingest? Then we take the result of that and write it back into a special key -> value record which then provides a super fast O(1) solution for returning the count of records when API calls are made (post ingest).

djstrong:
@lightwalker-eth 
I have tried this approach and it works good:
```
$ time pnpm count-keys

> @ensnode/ensrainbow@0.1.0 count-keys ensnode/apps/ensrainbow
> tsx src/count-keys.ts

Existing count in database: 133856894
Counting keys in database...
Total number of keys (excluding count key): 133856894
Updated count in database under LABELHASH_COUNT_KEY
Done!

real	7m21.411s
user	7m34.605s
sys	0m13.771s
```
So there is no duplicates.

PR #149, opened 2025-02-01
ENSRainbow: Optimize Label Count API

djstrong:
Fixes https://github.com/namehash/ensnode/issues/121

### DIFF
--- a/apps/ensrainbow/src/ingest.ts
+++ b/apps/ensrainbow/src/ingest.ts
@@ -4,8 +4,8 @@ import { createInterface } from "readline";
 import { createGunzip } from "zlib";
 import { ClassicLevel } from "classic-level";
 import ProgressBar from "progress";
-import { Hex } from "viem";
-import { labelHashToBytes } from "./utils/label-utils";
+import { ByteArray } from "viem";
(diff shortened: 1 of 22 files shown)


apps/ensrainbow/src/ingest.ts
```diff
@@ -24,6 +24,11 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
     keyEncoding: "binary",
   });
 
+  // Clear existing database before starting
+  console.log("Clearing existing database...");
+  await db.clear();
+  console.log("Database cleared.");
```
djstrong, line 30: It may take a lot of time to clear this way. The alternative is to to delete db directory.
The other way os to assert that the db is empty.

lightwalker-eth (approved):
@djstrong Overall everything in this PR looks ok to merge :+1: Just 1 priority change request:

1. Could you please resolve the merge conflicts with main? Not sure but maybe I created that issue when I merged PR 133 into main? If I created that issue, sorry!

As soon as the merge conflicts are fixed, suggest you merge this PR. I think we're starting to get into some change management challenges 😅

PR #156, opened 2025-02-03
Remove ts expect error directives

djstrong:
Is this correct way to solve this problem?


tk-o:
Yes, type-coercion seems to be the right tool to apply here if we want to test runtime and need to pass type checks in the build time.

tk-o (approved):

tk-o:
Now I noticed that I also solved this issue on my PR here:
- https://github.com/namehash/ensnode/pull/157/commits/20346ec819f91a6e999de2583dc374b30bf006b9

We could close this PR without merging, if that's alright?

Issue #163, opened 2025-02-03
ENSRainbow: Check if all labels in rainbow tables are normalized


djstrong:
Then we can be sure that all returned labels from ENSRainbow are normalized. But in the future versions it may change if we decide to include unnormalized labels.

lightwalker-eth:
@djstrong Thanks for recording this issue :+1: It's a good opportunity.

For the possibility of deciding to support unnormalized labels, I hope that we can postpone that for now. Have some ideas for how we might implement it. Ex: add optional param to heal API that allows unnormalized labels to be returned (off by default). While ingesting, store data for each RainbowRecord if the label is normalized or not. If an unnormalized label shouldn't be returned, even if an unnormalized label is found, it wouldn't be returned. Thereby simulating an unknown label from the perspective of ENSNode.

djstrong:
In `ens_names.sql` are 1026226 unnormalized names:
* 25200 with capital letters (e.g. `Babelish`)
* 1000832 punycode encoded (e.g. `xn--j7q08kz62bndm`)
* 194 others  (e.g. `437937076855494182_f416625f5125c74c16168d944ab1d6b45b60597b`, `b---h`, `bq--744a`, `betfirst_templates`, `fасebook`)

djstrong:
I have downloaded all domains from ENS Subgraph:
```
domains {
    id
    name
    labelName
    labelhash
    subdomainCount
    createdAt
}
```
sorted by `createdAt` and  organized the domains into monthly bins/groups for analysis.

### Unknown names per month

![Image](https://github.com/user-attachments/assets/98968c7f-f832-442e-9a1c-eda8c5be3c8c)
The chart shows total number of names per month and number of unknown names.

![Image](https://github.com/user-attachments/assets/e62687de-c0ae-4bca-bd6a-b800381ef211)
This shows percentage of unknown names in each month.

### Check unknown names against ENS rainbow tables

There is 17 unknown labels present in ENS rainbow tables:
```
0x000dc67ad99514dea87e9c662219d1237b6d5d9b988cc412d324a29992d4b1a5 four7coin.world
0xa0c6fe01c3f73632492e926800fe8af144b673740cdc361253aa0d6043572db8 jessica.gimmeeth
0xafe31d8787a658a66a6a50f4b91f07bd6627b4f3af5288b4f63f6316d1077569 new7777.shop
[...]

PR #170, opened 2025-02-04
feat(ensnode): integrate ensrainbow sdk

tk-o:
_Feel free to review on a comment-by-comment basis._

This PR aims to replace:
- #134 

Main reason for replacement is keeping things simple and clear to improve PR review experience, and also reduce risk of introducing bugs.

### Checklist
- [x] Carefully review the ENS Subgraph's logic for how it works with healed label values.
  - [x] Are there any checks or operations on the healed label values before the indexed state updates?

### Post-merge checklist
- [ ] Create an issue for refactoring `ensrainbow-sdk` as per [this comment](https://github.com/namehash/ensnode/pull/170/files#r1941706099).
- [ ] Create `normalizeLabelhash` function as per [this comment](https://github.com/namehash/ensnode/pull/170/files#r1941734181).
- [ ] Create an issue for handling ENSRainbow API Client errors as [per this comment](https://github.com/namehash/ensnode/pull/170/files#r1941696260).
- [ ] Create an issue to introduce short-term in-memory caching layer to avoid multiple network calls while healing the same labelhash.
- [ ] Create an issue about [improving how the ENSRainbow client handles network errors](https://github.com/namehash/ensnode/pull/170/files#r1943599996)
- [ ] Create an issue to [add a more graceful retry mechanism for the case that there's a client network error](https://github.com/namehash/ensnode/pull/170/files#r1943599996)

Resolves:
- #85


apps/ensrainbow/package.json
```diff
@@ -31,7 +31,7 @@
   "dependencies": {
     "@hono/node-server": "^1.4.1",
     "classic-level": "^1.4.1",
-    "ensnode-utils": "workspace:*",
+    "ensrainbow-sdk": "workspace:*",
```
tk-o, line 35: I included `ensnode-utils` in later commits.

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    return undefined;
+  }
+
+  const { label } = healResponse;
+
+  // if the indexer can't store the label yet, consider it unhealable
+  if (!isLabelIndexable(label)) {
+    return undefined;
+  }
```
tk-o, line 44: @lightwalker-eth this follows the Subgraph reference implementation:

https://github.com/ensdomains/ens-subgraph/blob/master/src/utils.ts#L68-L96

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  // if the label isn't normalized, consider it unhealable
+  if (label !== normalize(label)) {
+    return undefined;
+  }
```
tk-o, line 49: This was an extra step that @shrugs proposed in my previous PR.

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Heal a labelhash to its original label.
+ *
+ * @returns a healed label for a given labelhash, if possible.
+ *
+ * NOTE: undefined return type is helpful for passing to drizzle, which
+ * will ignore the input rather that set the value in the db to `NULL`
+ * if we were to return `null`.
+ **/
+export async function heal(labelhash: Labelhash) {
```
lightwalker-eth, line 21:
```suggestion
export async function labelByHash(labelhash: Labelhash) {
```

Suggesting this renaming to better align with ENS Subgraph code that uses `nameByHash` (except fixing their misuse of the word "name").

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+    labelHashToBytes(labelhash);
+  } catch (error) {
+    console.error(`Invalid labelhash - must be a valid hex string: ${error}`);
+
+    return undefined;
+  }
+
+  const healResponse = await ensRainbowApiClient.heal(labelhash);
+
+  if (healResponse.status === "error") {
```
lightwalker-eth, line 33: We shouldn't treat all errors the same.

We need to examine the error code. If the error code is 404 then its ok to return undefined. However if the error code is anything else then I believe this function should throw an exception.

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+    console.error(`Error healing labelhash ${labelhash}: ${healResponse.error}`);
+
+    return undefined;
+  }
+
+  const { label } = healResponse;
+
+  // if the indexer can't store the label yet, consider it unhealable
```
lightwalker-eth, line 41: Suggest moving this logic outside of the `labelByHash` helper function. The reason to move it out is just to help make code review of our indexing logic vs code review of the ENS Subgraph's code as easy to mentally diff and compare as possible.

The ENS Subgraph had to do these `isLabelIndexable` checks outside of the `nameByHash` function.

apps/ensnode/src/lib/label-helpers.ts
lightwalker-eth: I don't believe we should add any normalize check on healed labels right now. That would break backwards compatibility with the ENS Subgraph.

For the normalization question, we're going to attack it through this goal first: https://github.com/namehash/ensnode/issues/163

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+import { DEFAULT_ENSRAINBOW_URL } from "ensrainbow-sdk/consts";
+import { labelHashToBytes } from "ensrainbow-sdk/label-utils";
+import { normalize } from "viem/ens";
+
+const ensRainbowApiClient = new EnsRainbowApiClient({
+  endpointUrl: new URL(process.env.ENSRAINBOW_URL || DEFAULT_ENSRAINBOW_URL),
+});
+
+/**
+ * Heal a labelhash to its original label.
```
lightwalker-eth, line 13: Suggest we document how this function mirrors the `nameByHash` function that is implemented inside GraphNode. It would also be nice to include some more specific code references if possible. Maybe consider taking one / some (I don't think we need all of them) of the ideas from the notes I captured in: https://github.com/namehash/ensnode/issues/140#issuecomment-2628326254

The goal of this function is to mirror the logic of `nameByHash`.

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @returns a healed label for a given labelhash, if possible.
+ *
+ * NOTE: undefined return type is helpful for passing to drizzle, which
```
lightwalker-eth, line 17: Thanks for documenting this. Is there an alternative strategy that would decouple the definition of this function from any drizzle concerns? We want this function to maximally align with the `nameByHash` function implemented in GraphNode.

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
+import { isLabelIndexable } from "ensnode-utils/subname-helpers";
+import type { Labelhash } from "ensnode-utils/types";
+import { EnsRainbowApiClient } from "ensrainbow-sdk/client";
+import { DEFAULT_ENSRAINBOW_URL } from "ensrainbow-sdk/consts";
+import { labelHashToBytes } from "ensrainbow-sdk/label-utils";
+import { normalize } from "viem/ens";
+
+const ensRainbowApiClient = new EnsRainbowApiClient({
+  endpointUrl: new URL(process.env.ENSRAINBOW_URL || DEFAULT_ENSRAINBOW_URL),
```
lightwalker-eth, line 9: Suggest a few items here:

1. Update docs for additional environment variable: https://github.com/namehash/ensnode?tab=readme-ov-file#prepare-application-environment
2. Update: https://github.com/namehash/ensnode/blob/main/apps/ensnode/.env.local.example
3. Any other items you suggest we update for the introduction of a new environment variable.

Also, what do you think about creating a utility function such `rpcEndpointUrl` that currently lives inside `ponder-helpers.ts`? The main goal I'm worried about is: Please set the `ENSRAINBOW_URL` environment variable to some strange invalid value. This should cause a helpful error message to be logged that would help an operator of ENSNode track down and fix the issue.

apps/ensnode/test/label-helpers.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe("heal", () => {
+  it("heals a labelhash", async () => {
+    const labelhash = "0xaf2caa1c2ca1d027f1ac823b529d0a67cd144264b2789fa2ea4d63a67c7103cc";
+    const label = await heal(labelhash);
+
+    expect(label).toEqual("vitalik");
+  });
+
+  it("returns undefined for an invalid labelhash", async () => {
```
lightwalker-eth, line 13: This case should actually throw an exception. Please see my related comments in `label-helpers.ts`.

apps/ensnode/test/label-helpers.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, it } from "vitest";
+
+import { heal } from "../src/lib/label-helpers";
+
+describe("heal", () => {
+  it("heals a labelhash", async () => {
```
lightwalker-eth, line 6:
```suggestion
  it("heals a known labelhash", async () => {
```

And then suggest creating a separate unit test for the case of a labelhash that is properly formatted but that is not known / not healable.

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -76,8 +77,9 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
 
       const node = makeSubnodeNamehash(ownedNameNode, labelhash);
 
-      // TODO: materialze labelName via rainbow tables ala Registry.ts
-      const labelName = undefined;
+      // materialze labelName via rainbow tables ala Registry.ts
+      // https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/ethRegistrar.ts#L56-L61
```
lightwalker-eth, line 81: Very happy to see you include the link to the related code in the ENS Subgraph.

Suggest we refactor how we write the code here so that it appears as similar as possible to the logic in the ENS Subgraph. Please see my related comments inside `label-helpers.ts`.

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -126,9 +127,9 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
         if (!domain.name) {
           const parent = await context.db.find(schema.domain, { id: node });
 
-          // TODO: implement sync rainbow table lookups
-          // https://github.com/ensdomains/ens-subgraph/blob/master/src/ensRegistry.ts#L111
-          const labelName = encodeLabelhash(label);
+          // sync rainbow table lookups
+          // https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L112-L116
```
lightwalker-eth, line 131: Very happy to see you include the link to the related code in the ENS Subgraph.

Suggest we refactor how we write the code here so that it appears as similar as possible to the logic in the ENS Subgraph. Please see my related comments inside `label-helpers.ts`.

apps/ensnode/src/lib/label-helpers.ts
```diff
@@ -0,0 +1,53 @@
+import { isLabelIndexable } from "ensnode-utils/subname-helpers";
+import type { Labelhash } from "ensnode-utils/types";
+import { EnsRainbowApiClient } from "ensrainbow-sdk/client";
+import { DEFAULT_ENSRAINBOW_URL } from "ensrainbow-sdk/consts";
+import { labelHashToBytes } from "ensrainbow-sdk/label-utils";
+import { normalize } from "viem/ens";
+
```
lightwalker-eth, line 7: Suggest renaming this file to something like "graphnode-helpers" since the utility functions being defined here are actually simulations of functions defined inside GraphNode. More background info here: https://github.com/namehash/ensnode/issues/140#issuecomment-2628326254

lightwalker-eth (changes requested):
@tk-o Hey some great progress on this PR. It's moving in the right overall direction. Reviewed and shared feedback :+1:

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @example
+ * ```typescript
+ * const client = new EnsRainbowApiClient({
+ *  endpointUrl: new URL("https://api.ensrainbow.io"),
+ * });
+ * ```
+ */
+export class EnsRainbowApiClient {
+  constructor(private readonly options: EnsRainbowApiClientOptions) {}
```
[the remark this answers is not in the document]
tk-o: This is a syntactic sugar called [parameter properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties).

apps/ensnode/src/lib/label-helpers.ts
tk-o: Ok, will drop the normalization check.

cc @shrugs

apps/ensrainbow/src/operations.ts
```diff
@@ -1,16 +1,16 @@
...
 import type {
   CountError,
   CountResponse,
   CountSuccess,
   HealError,
   HealResponse,
   HealSuccess,
-} from "./utils/response-types.js";
-import { ErrorCode, StatusCode } from "./utils/response-types.js";
+} from "ensrainbow-sdk/types";
```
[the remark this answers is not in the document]
tk-o: It's trivial, will do it.

packages/ensrainbow-sdk/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "scripts": {
+    "test": "vitest",
+    "typecheck": "tsc --noEmit",
+    "lint": "biome check --write",
+    "lint:ci": "biome ci"
+  },
+  "devDependencies": {
+    "@namehash/shared-configs": "workspace:",
+    "@biomejs/biome": "catalog:",
+    "ensnode-utils": "workspace:",
```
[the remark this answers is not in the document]
tk-o: For `ensrainbow-sdk` package, we only import types from `ensnode-utils` package, so this would be OK as a dev dependency 👍

apps/ensnode/src/lib/label-helpers.ts
tk-o: I will change this function signature to match the original type signature, and will take care of drizzle concerns on the caller level.

The original `nameByHash` function has the following signature:
```ts
function nameByHash(hash: string): string | null;
```

Source:
https://github.com/graphprotocol/graph-tooling/blob/ee3a20985909de49ec25f378db80cd1c3e19b610/packages/ts/index.ts#L59

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -76,8 +77,15 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
...
+      const label = await labelByHash(labelhash);
+      const canIndexLabel = isLabelIndexable(label);
+      // NOTE: undefined type is helpful for passing to drizzle, which
+      // will ignore the input rather that set the value in the db to `NULL`
+      // if we were to return `null`.
+      const labelName = canIndexLabel ? label : undefined;
+      const name = canIndexLabel ? `${label}.${ownedName}` : undefined;
```
tk-o, line 88: @shrugs please share your thoughts 🚀

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -126,9 +127,10 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
+          const label = await labelByHash(event.args.label);
+          const labelName = isLabelIndexable(label) ? label : encodeLabelhash(event.args.label);
```
tk-o, line 133: @shrugs please share your thoughts 🚀

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Heal a labelhash to its original label. It mirrors `ens.nameByHash` function:
+ * https://github.com/graphprotocol/graph-node/blob/master/runtime/test/wasm_test/api_version_0_0_4/ens_name_by_hash.ts#L9-L11
+ *
+ * @returns a healed label for a given labelhash, if possible.
+ *
+ **/
+export async function labelByHash(labelhash: Labelhash): Promise<string | null> {
+  // runtime check, ENS rainbow enforces this validation as well
+  labelHashToBytes(labelhash);
```
tk-o, line 19: Any errors thrown by this runtime check will be propagated to the caller.

apps/ensrainbow/src/operations.ts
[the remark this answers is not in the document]
tk-o: However, there's a member name conflict between consts and types for the following module members:
- StatusCode
- ErrorCode

image

I'll have to do some refactorings.

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -94,6 +102,7 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
         registrantId: owner,
         expiryDate: expires + GRACE_PERIOD_SECONDS,
         labelName,
+        name,
```
tk-o, line 105: @shrugs & I have discussed adding this property to follow the ENS Subgraph logic more accurately:
https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/ethRegistrar.ts#L59

apps/ensnode/src/handlers/Registrar.ts
tk-o: Replicated the same behaviour in a slightly different syntax. The bottom line is that we don't want to update `labelName` & `name` on an indexed `domain` entity (or `labelName` on an indexed `registration` entity) if the healed label is not indexable:

```ts
// extracted from 
// https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/ethRegistrar.ts#L56-L61
let labelName = ens.nameByHash(label.toHexString());
if (checkValidLabel(labelName)) {
  domain.labelName = labelName;
  domain.name = labelName! + ".eth";
  registration.labelName = labelName;
}
```

apps/ensnode/src/handlers/Registry.ts
tk-o: The original ENS Subgraph implementation:
https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L112-L116
```ts
let label = ens.nameByHash(event.params.label.toHexString());
if (checkValidLabel(label)) {
  domain.labelName = label;
} else {
  label = "[" + event.params.label.toHexString().slice(2) + "]";
}
```

apps/ensnode/src/handlers/Registrar.ts
tk-o: Replicated the same behaviour in a slightly different syntax. The bottom line is that we don't want to update `labelName` & `name` on an indexed `domain` entity (or `labelName` on an indexed `registration` entity) if the healed label is not indexable.

The original ENS Subgraph implementation:
https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/ethRegistrar.ts#L56-L61
```ts
let labelName = ens.nameByHash(label.toHexString());
if (checkValidLabel(labelName)) {
  domain.labelName = labelName;
  domain.name = labelName! + ".eth";
  registration.labelName = labelName;
}
```

apps/ensrainbow/src/operations.ts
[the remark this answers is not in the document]
tk-o: @lightwalker-eth I'll create an issue for that to be solved in a follow-up PR.

packages/ensrainbow-sdk/src/client.ts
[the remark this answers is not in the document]
tk-o: However, with `options` optional, we'll apply regular class filed, and then have it initialised in consturtor.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,64 @@
...
+   * Note on returned labels: The service returns labels exactly as they appear in the source data. This means:
+   *
+   * - Labels may or may not be ENS-normalized
+   * - Labels can contain any valid string, including dots, null bytes, or be empty
+   * - Clients should handle all possible string values appropriately
+   *
+   * @param labelhash
+   * @returns
+   *
+   * @link https://github.com/namehash/ensnode/tree/effc77d/apps/ensrainbow#heal-label
```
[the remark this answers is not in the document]
tk-o: Sure, will drop the link.

My hope was if I left particular commit reference, someone could track the updates with Git Blame.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -73,7 +73,7 @@ const UNINDEXABLE_LABEL_CHARACTER_CODES = new Set(
  * Related logic in ENS Subgraph:
  * https://github.com/ensdomains/ens-subgraph/blob/master/src/utils.ts#L68
  */
-export const isLabelIndexable = (label: string) => {
+export const isLabelIndexable = (label: string | null): boolean => {
```
tk-o, line 76: The original ENS Subgraph implementation:
https://github.com/ensdomains/ens-subgraph/blob/master/src/utils.ts#L68
```ts
export function checkValidLabel(name: string | null): boolean {
```

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * // Output:
+   * // {
+   * //   status: "success",
+   * //   label: "reverse",
+   * // }
+   * ```
+   */
+  async heal(labelhash: Labelhash): Promise<HealResponse> {
+    try {
```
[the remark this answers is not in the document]
tk-o: I'll create an issue to capture the `normalizeLabelhash` requirement 👍

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,64 @@
...
+      const response = await fetch(new URL(`/v1/heal/${labelhash}`, this.options.endpointUrl));
```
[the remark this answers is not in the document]
tk-o: I think we should let all packages to use Web Fetch API. It's [the application's runtime job to provide right implementation](https://www.builder.io/blog/stop-polyfilling-fetch-in-your-npm-package).

On the other hand, we support Node.js >= v18, and [Node.js v18 supports Web Fetch API](https://nodejs.org/en/blog/announcements/v18-release-announce/#fetch-experimental) out of the box.

With all that in mind, I propose not doing any changes regarding how we use Web Fetch API in our stack. Please share your thoughts, @lightwalker-eth.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,64 @@
...
+
+      return response.json() as Promise<HealResponse>;
+    } catch (error) {
```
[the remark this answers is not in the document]
tk-o: In that case, I'll move try / catch to the `heal` method caller, which is `labelByHash` function located in `graphnode-helpers` of ENSNode application.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,64 @@
...
+      const errorMessage = error instanceof Error ? error.message : "Unknown error";
+      return {
+        error: errorMessage,
+        errorCode: 500,
+        status: "error",
+      } satisfies HealError;
```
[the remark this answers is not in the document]
tk-o: Quick comparison:

- `as` = Type assertion (_"trust me, I know this is type X"_)
  - tells TypeScript to treat something as a specific type without checking
- `satisfies` = Type validation (_"check if this matches type X, please"_)
  - verifies the value actually matches the type specification
  - provides better type inference and safer type checking

The `satisfies` operator is generally safer because it validates rather than asserting.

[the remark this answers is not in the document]
tk-o: I use `satisfies` operator when I'm creating inline objects and need TypeScript to keep me in check. If I used `as` operator, TypeScript would think I know what I'm doing and wouldn't help me find errors for incorrectly defined object value.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -31,8 +53,6 @@ export class EnsRainbowApiClient {
    * @param labelhash all lowercase hex string with 0x prefix with length of 66 characters in total
    * @returns
    *
-   * @link https://github.com/namehash/ensnode/tree/effc77d/apps/ensrainbow#heal-label
```
tk-o: @lightwalker-eth, I've dropped the link as requested.

pnpm-workspace.yaml
```diff
@@ -10,5 +10,5 @@ catalog:
   hono: ^4.6.14
   typescript: ^5.7.3
   viem: ^2.22.13
-  vitest: ^3.0.3
+  vitest: ^3.0.5
```
tk-o, line 13: Addressing security alert:
- https://github.com/namehash/ensnode/security/dependabot/6

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -15,10 +16,19 @@ const ensRainbowApiClient = new EnsRainbowApiClient({
...
+  let healResponse: HealResponse;
+
+  try {
+    healResponse = await ensRainbowApiClient.heal(labelhash);
+  } catch (error) {
+    // If the client fails to fetch data, we log the error
+    // and return null to the caller.
+    console.error(`ENSRainbow Client error ${labelhash}: ${error}`);
+    return null;
+  }
```
tk-o, line 32: `ensRainbowApiClient.heal` is a thin wrapper over Web Fetch API. It would only throw an error when:
- Network failures occur
- DNS failures
- Invalid URL format
- CORS violations
- Request timeout
- Server is unreachable

I assume, for all of issues listed above, the indexing process must continue, and the label should be considered as not present (hence, returning `null`).

Thoughts on that approach, @lightwalker-eth?

apps/ensnode/test/ponder-helpers.spec.ts
```diff
@@ -57,6 +59,28 @@ describe("ponder helpers", () => {
...
+    it("should parse a valid ENSRainbow endpoint URL", () => {
+      expect(parseEnsRainbowEndpointUrl("https://api.ensrainbow.io")).toBe(
+        "https://api.ensrainbow.io/",
+      );
+    });
+
+    it("should throw an error if the URL is invalid", () => {
+      expect(() => parseEnsRainbowEndpointUrl("almost_an_URL")).toThrowError(
+        "'almost_an_URL' is not a valid URL",
+      );
```
tk-o, line 72: @lightwalker-eth this test case addresses your request about testing invalid URL value provided as `ENSRAINBOW_URL` env var.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -73,7 +73,7 @@ const UNINDEXABLE_LABEL_CHARACTER_CODES = new Set(
...
   if (!label) return false;
```
djstrong, line 77: @lightwalker-eth empty label should be indexable, right?

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+  labelHashToBytes(labelhash);
+
+  let healResponse: HealResponse;
+
+  try {
+    healResponse = await ensRainbowApiClient.heal(labelhash);
+  } catch (error) {
+    // If the client fails to fetch data, we log the error
+    // and return null to the caller.
+    console.error(`ENSRainbow Client error ${labelhash}: ${error}`);
```
lightwalker-eth, line 30:
```suggestion
    console.error(`ENSRainbow client network error when healing ${labelhash}: ${error}`);
```

apps/ensnode/src/lib/graphnode-helpers.ts
lightwalker-eth: @tk-o Thanks for asking. No I don't think the indexing process can continue if there's an error that isn't `NotFound`. We could theoretically add some retry mechanism but don't suggest we do that in this PR.

For now, if this fails, we must throw an error to crash the ENSNode instance. This is because otherwise we could not properly guarantee consistent state if we just failed silently. Logging an error message is too silent. We must fail loud.

Please create 2 issue for us to follow up on in separate PRs for the following goals:
1. Improving how the ENSRainbow client handles network errors. Please include all the possible examples you documented in your comment there. I think it would be nice to update the `HealResponse` data model so that these network errors could also be represented there and not just throw an exception. This is nice for people building UIs. For example it makes it easier to build a nice UI where you show an specialized UIs for errors such as "Connection Timeout" or "Server is unreachable".
2. Add a more graceful retry mechanism for the case that there's a client network error here. Perhaps we would retry a few times with some backoff before finally giving up? Each failure should write a warning to the logs.

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  if (healResponse.status === StatusCode.Success) {
+    return healResponse.label;
+  }
+
+  if (healResponse.errorCode === ErrorCode.NotFound) {
+    // This is a warning because it's possible that the labelhash is not
+    // recorded in the ENSRainbow database.
+    console.warn(`Healing labelhash error: '${labelhash}' not found`);
```
lightwalker-eth, line 41: Sorry I don't follow? If you get a not found error, it definitely means the labelhash is not recorded in the ENSRainbow database. In other words, it's not right to say "it's possible". Instead, it is "certain".

I also don't understand why we would want to `console.warn` here? Why not just return `null`?

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    return null;
+  }
+
+  throw new Error(
```
lightwalker-eth, line 46: We need to add a `@throws ...` to the JSDoc for any function that can throw with details about that situation. Please be specific in how you document this.

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+    `Healing labelhash error ${labelhash}: ${healResponse.error}; error code: ${healResponse.errorCode}`,
```
lightwalker-eth, line 47:
```suggestion
    `Error healing labelhash: "${labelhash}". Error (${healResponse.errorCode}): ${healResponse.error}.`,
```

apps/ensnode/test/graphnode-helpers.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it("returns null for a valid unknown labelhash", async () => {
+    // labelhash comes from the ENSRainbow API logs:
+    // "Unhealable labelhash request: 0x00ca5d0b4ef1129e04bfe7d35ac9def2f4f91daeb202cbe6e613f1dd17b2da06"
+    expect(
+      await labelByHash("0x00ca5d0b4ef1129e04bfe7d35ac9def2f4f91daeb202cbe6e613f1dd17b2da06"),
+    ).toBeNull();
+  });
+
+  it("throws an error for an invalid too short labelhash", async () => {
```
lightwalker-eth, line 21: Nice work on these unit tests :+1:

I think all 4 of the "throws an error" test here are good :+1: I shared some feedback in a related file that might change the exact error message we get on these, but I agree it is correct that all should still throw.

packages/ensnode-utils/src/subname-helpers.ts
lightwalker-eth: @tk-o Thanks! Could you add a unit test for the `null` case? Let's make sure it returns false.

Can you please update the JSDoc for `isLabelIndexable` to identify that passing a value of `null` to this function is symbolic for the idea that the label is unknown.

Whenever there's a null somewhere, we should carefully document what it symbolically represents.

packages/ensnode-utils/src/subname-helpers.ts
lightwalker-eth: @djstrong Good question. For now I don't think so because then it would result in the way the `name` fields are generated to get more complicated to parse.

I have some code I wrote over the summer that has really strong ENS name processing logic and internal data models that would allow us to represent and handle these cases better. I went far with it (multiple weeks of hard work) but wasn't able to finish it 100%. I plan to bring this effort back to life through ENSNode after we finish shipping our V1.

In the end, we will build a solution to index all labels, no matter what 😄

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -76,8 +77,15 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
 
       const node = makeSubnodeNamehash(ownedNameNode, labelhash);
 
-      // TODO: materialze labelName via rainbow tables ala Registry.ts
-      const labelName = undefined;
+      // materialze labelName via rainbow tables ala Registry.ts
+      // https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/ethRegistrar.ts#L56-L61
+      const label = await labelByHash(labelhash);
+      const canIndexLabel = isLabelIndexable(label);
```
lightwalker-eth, line 83: Could we please rewrite this to maximally align with the way the related logic is written in the ENS subgraph? Right now I have to think to much to verify our logic is 100% aligned.

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -76,8 +77,15 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
 
       const node = makeSubnodeNamehash(ownedNameNode, labelhash);
 
-      // TODO: materialze labelName via rainbow tables ala Registry.ts
-      const labelName = undefined;
+      // materialze labelName via rainbow tables ala Registry.ts
```
lightwalker-eth, line 80:
```suggestion
      // attempt to heal the label associated with labelhash via ENSRainbow
```

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -126,9 +127,10 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
         if (!domain.name) {
           const parent = await context.db.find(schema.domain, { id: node });
 
-          // TODO: implement sync rainbow table lookups
-          // https://github.com/ensdomains/ens-subgraph/blob/master/src/ensRegistry.ts#L111
-          const labelName = encodeLabelhash(label);
+          // sync rainbow table lookups
```
lightwalker-eth, line 130:
```suggestion
          // attempt to heal the label associated with labelhash via ENSRainbow
```

apps/ensnode/src/handlers/Registry.ts
lightwalker-eth: @tk-o Please rewrite our logic here so that it is maximally similar to the related ENS Subgraph logic. It makes me think too much to verify it is the same. Thanks.

lightwalker-eth (changes requested):
@tk-o Hey some good updates here :+1: Reviewed in detail and shared feedback.

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Heal a labelhash to its original label. It mirrors `ens.nameByHash` function:
+ * https://github.com/graphprotocol/graph-node/blob/master/runtime/test/wasm_test/api_version_0_0_4/ens_name_by_hash.ts#L9-L11
+ *
+ * @returns a healed label for a given labelhash, if possible.
+ *
+ **/
+export async function labelByHash(labelhash: Labelhash): Promise<string | null> {
+  // runtime check, ENSRainbow API enforces this validation as well
+  labelHashToBytes(labelhash);
```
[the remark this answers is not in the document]
tk-o: Sure, we can do that and let the unhandled errors to appear 👍 We could learn from those errors (unless there are no errors, which would be also good to know!).

apps/ensnode/src/lib/graphnode-helpers.ts
tk-o: Understood. I added a task for creating those two issues to the _Post-merge checklist_ in this PR description. Will go through that list once the PR gets merged 👍

apps/ensrainbow/src/operations.ts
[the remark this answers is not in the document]
tk-o: @lightwalker-eth, I have a post-merge checklist attached to this PR description:
image

Will go through it after all requested changes are done and the code in its final shape will land in the `main` branch. It will be easier for me to add references in each issue to some actual code base. Hope that's alright 👍

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,86 @@
...
+      endpointUrl: new URL(DEFAULT_ENSRAINBOW_URL),
+    };
+  }
+
+  constructor(options: Partial<EnsRainbowApiClientOptions> = {}) {
+    this.options = {
+      ...EnsRainbowApiClient.defaultOptions(),
+      ...options,
+    };
+  }
```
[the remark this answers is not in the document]
tk-o: Great idea, increasing awareness about importance of using co-located services for production use cases.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,86 @@
...
+   * Note on returned labels: ENSRainbow returns labels exactly as they are
+   * represented in source rainbow table data. This means:
+   *
+   * - Labels may or may not be ENS-normalized
+   * - Labels can contain any valid string, including dots, null bytes, or be empty
+   * - Clients should handle all possible string values appropriately
+   *
+   * @param labelhash all lowercase hex string with 0x prefix with length of 66 characters in total
+   * @returns a response with the status and the label if successful
+   * @throws an error if the request fails or the labelhash is not found or invalid
```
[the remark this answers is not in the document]
tk-o: You are right, the code has changed since this comment was first added. Will alter it.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * // Output:
+   * // {
+   * //   status: "success",
+   * //   label: "reverse",
+   * // }
+   * ```
+   */
+  async heal(labelhash: Labelhash): Promise<HealResponse> {
+    const response = await fetch(new URL(`/v1/heal/${labelhash}`, this.options.endpointUrl));
```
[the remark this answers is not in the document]
tk-o: Yes, I will log this request once this PR is merged, as per _the Post-merge checklist_ attached to this PR description:
image

apps/ensnode/src/lib/graphnode-helpers.ts
tk-o: I'm dropping the try/catch here, as for now, we decided to let the errors stop indexing. In the very near future, when we add a retries mechanism, we will use it to wrap `ensRainbowApiClient.heal(labelhash)` call and handle network errors internally.

Does that sound OK to you, @lightwalker-eth?

apps/ensnode/src/lib/graphnode-helpers.ts
tk-o: Sure, that's a wrong word. Anyway, dropping the whole `console.warn` statement, as we will have a similar log to appear on the ENSRainbow deployment anyway.

packages/ensnode-utils/src/subname-helpers.ts
tk-o: As for:

> the idea that the label is unknown

I reckon saying:

> the idea that the label is unindexable

would be more precise.

What do you think, @lightwalker-eth?

apps/ensnode/.env.local.example
```diff
@@ -37,4 +37,11 @@ DATABASE_URL=postgresql://dbuser:abcd1234@localhost:5432/my_database
...
 # Identify which indexer plugins to activate (see `src/plugins` for available plugins)
 # This is a comma separated list of one or more available plugin names (case-sensitive).
-ACTIVE_PLUGINS=eth,base.eth,linea.eth
\ No newline at end of file
+ACTIVE_PLUGINS=eth,base.eth,linea.eth
+
+# Unknown label healing
+# This is the URL of the ENSRainbow API that the indexer will use to fetch label healing data.
+# If this is not set, DEFAULT_ENSRAINBOW_URL will be used.
+# Read more about ENSRainbow here:
```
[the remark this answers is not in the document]
tk-o: Done ✅

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,86 @@
...
+    return response.json() as Promise<HealResponse>;
+  }
+
+  /**
+   * Get current client options.
+   *
+   * @returns the current client options
+   */
+  getOptions(): EnsRainbowApiClientOptions {
+    return this.options;
```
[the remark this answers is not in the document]
tk-o: Done ✅

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -77,38 +77,45 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
...
+      // prepare domain entity object
+      const domain = {
         registrantId: owner,
         expiryDate: expires + GRACE_PERIOD_SECONDS,
-        labelName,
-        name,
-      });
+        labelName: undefined as string | undefined,
+        name: undefined as string | undefined,
+      } satisfies Partial<typeof schema.domain.$inferInsert>;
```
tk-o, line 97: These two object definitions are required to follow the reference implementation code style in the further part of this handler function.

shrugs (approved):
I preferred the more modern & ponder-esq versions of these functions, but must admit that these versions that closely match the subgraph implementation are easier to compare for equivalency, and they look right to me.

shrugs:
these tests for `graphnode-helpers` create a hard dependency on ensrainbow running — we should probably mock that client so that the ensnode tests can be run independently of ensrainbow.

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -100,40 +101,63 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
-        if (!domain.name) {
-          const parent = await context.db.find(schema.domain, { id: node });
-
-          // TODO: implement sync rainbow table lookups
-          // https://github.com/ensdomains/ens-subgraph/blob/master/src/ensRegistry.ts#L111
-          const labelName = encodeLabelhash(label);
-          const name = parent?.name ? `${labelName}.${parent.name}` : labelName;
-
-          await context.db.update(schema.domain, { id: domain.id }).set({ name, labelName });
+        // setting the domain's name and labelName, akin to
```
lightwalker-eth, line 128: @tk-o Appreciate all your efforts here 🙌

lightwalker-eth (changes requested):
@tk-o Hey really appreciate the big effort here. We're very close. A few small suggestions and questions.

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -76,30 +77,45 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
...
+      const registration = {
+        id: makeRegistrationId(ownedName, labelhash, node),
         domainId: node,
         registrationDate: event.block.timestamp,
         expiryDate: expires,
         registrantId: owner,
-        labelName,
-      });
+        // labelName is set later if the label is indexable
+        labelName: undefined as string | undefined,
```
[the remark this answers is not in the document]
tk-o: We can use either operator: `as` or `satisfies` to define initial object value.

#### `as` operator

When we use `as`, we can skip the fields initialisation:
```ts
undefined as string | undefined
```
because TypeScript would trust us that we would remember to set the values ourselves later. So we could initialise `registration` object this way:
```ts
const registration = {
  id: makeRegistrationId(ownedName, labelhash, node),
  domainId: node,
  registrationDate: event.block.timestamp,
  expiryDate: expires,
  registrantId: owner,
} as typeof schema.registration.$inferInsert;
```

#### `satisfies` operator

When we use `satisfies`, we have to be explicit to help TypeScript help us. That's why we mention all fields in the object, and those that we don't have value yet, we assign as `T | undefined` where `T` is a concrete value type, i.e. `string`:
```ts
undefined as string | undefined
```
TypeScript needs to know the whole object:
```ts
const registration = {
  id: makeRegistrationId(ownedName, labelhash, node),
  domainId: node,
  registrationDate: event.block.timestamp,
  expiryDate: expires,
  registrantId: owner,
  // labelName is set later if the label is indexable
  labelName: undefined as string | undefined,
} satisfies typeof schema.registration.$inferInsert;
```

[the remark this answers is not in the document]
tk-o: Now, that I think about it, we possibly want to go with `as` operator` and let TypeScript know that we manage the type.

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -76,30 +77,45 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
...
 
-      await context.db.update(schema.domain, { id: node }).set({
+      // prepare domain entity object
+      const domain = {
         registrantId: owner,
         expiryDate: expires + GRACE_PERIOD_SECONDS,
-        labelName,
-      });
+        labelName: undefined as string | undefined,
+        name: undefined as string | undefined,
```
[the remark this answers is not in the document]
tk-o: After some reflection, I'd go with alternative syntax using `as` operator. Explained difference in [another comment](https://github.com/namehash/ensnode/pull/170/files#r1946430144) on this PR.

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -100,40 +101,63 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
+        if (domain === null) {
           domain = await context.db.insert(schema.domain).values({
             id: subnode,
-            ownerId: owner,
-            parentId: node,
             createdAt: event.block.timestamp,
             labelhash: event.args.label,
+            subdomainCount: 0,
+            ownerId: owner,
+            parentId: null,
```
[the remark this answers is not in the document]
tk-o: We have to set `parent` for a newly inserted domain entity to `null` so that the subgraph-compatible code works as in the reference implementation:
```ts
// increment subdomainCount of parent domain, akin to
// https://github.com/ensdomains/ens-subgraph/blob/c68a889e0bcdc6d45033778faef19b3efe3d15fe/src/ensRegistry.ts#L104-L107
if (domain.parentId === null && parent !== null) {
```

This is just a constraint we inherited from the reference implementation.

I can drop explicit initialisation to avoid confusion.
```suggestion
```

Will also drop initialisation of `subdomainCount` field as it's has its default value set on schema definition.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  constructor(options: Partial<EnsRainbowApiClientOptions> = {}) {
+    this.options = {
+      ...EnsRainbowApiClient.defaultOptions(),
+      ...options,
+    };
+
+    if (this.options.endpointUrl === EnsRainbowApiClient.defaultOptions().endpointUrl) {
+      console.warn(
```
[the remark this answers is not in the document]
tk-o: Good idea, will fix it right away.

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -100,40 +101,63 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
             isMigrated,
```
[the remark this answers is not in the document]
tk-o: Fixed ✅

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -23,6 +23,7 @@ describe("isLabelIndexable", () => {
 
   it("should return false for empty labels", () => {
     expect(isLabelIndexable("")).toBe(false);
+    expect(isLabelIndexable(null)).toBe(false);
```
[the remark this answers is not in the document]
tk-o: Done ✅

packages/ensrainbow-sdk/src/client.ts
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Thanks for the follow up comments above!

apps/ensnode/src/handlers/Registry.ts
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Thanks for your follow up.

Can you please help me understand how the following subgraph code:

CleanShot 2025-02-07 at 17 54 05

Relates to our logic here:

CleanShot 2025-02-07 at 17 55 16

1. Why are we making a db insert here?
2. Why aren't we setting `subdomainCount` to 0?
3. Why are we setting `labelhash`, `ownerId`, and `isMigrated`?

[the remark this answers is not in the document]
tk-o: I've further modified the code to look as if it was ENS Subgraph code 👍 Introduced couple of helper methods to achieve that. All of them have their original ENS Subgraph counterparts.

apps/ensrainbow/src/index.test.ts
```diff
@@ -9,13 +9,14 @@ import { app, db } from "./index";
 import { LABELHASH_COUNT_KEY } from "./utils/constants";
 
 describe("ENS Rainbow API", () => {
+  const port = 3223;
```
tk-o, line 12: Had to pick other port that would be less likely to conflict with other local services 🚀

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -78,6 +80,60 @@ async function recursivelyRemoveEmptyDomainFromParentSubdomainCount(context: Con
...
+  await context.db
+    .insert(schema.domain)
+    .values(domain)
+    .onConflictDoUpdate(() => domain);
+
+  // garbage collect newly 'empty' domain iff necessary
+  if (domain.ownerId === zeroAddress) {
+    await recursivelyRemoveEmptyDomainFromParentSubdomainCount(context, domain.id);
+  }
+}
```
tk-o, line 135: @lightwalker-eth, all these helpers were created to mimic form of the original event handler. I hope the comments help highlighting the mapping accuracy.

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -102,10 +102,10 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
       }
 
       // akin to domain.save()
-      // https://github.com/ensdomains/ens-subgraph/blob/c68a889e0bcdc6d45033778faef19b3efe3d15fe/src/ethRegistrar.ts#L63
+      // https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ethRegistrar.ts#L63
```
tk-o, line 105: Just aligning the github commit refs to point to a single commit shortened ID.

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -78,6 +80,60 @@ async function recursivelyRemoveEmptyDomainFromParentSubdomainCount(context: Con
...
+  if (domain == null && node == ROOT_NODE) {
+    return createDomain(node, timestamp);
+  } else {
+    return domain;
+  }
+}
+
+// upsert the domain database entity object
+// akin to https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L84-L87
+async function saveDomain(context: Context, domain: Domain): Promise<void> {
```
lightwalker-eth, line 124: I hate to say it, but is there a special reason why the logic in this function differs from the ENS Subgraph? 😅

For example:
1. Why are we inserting here before the recursive remove? Why not switch the order?
2. Why are we only conditionally performing the recursive remove?

apps/ensnode/src/handlers/Registry.ts
lightwalker-eth: @tk-o Really appreciate this! Thank you 🙌 🙌 🙌

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -62,7 +64,7 @@ function isDomainEmpty(domain: typeof schema.domain.$inferSelect) {
 }
```
lightwalker-eth, line 64: This comment is about our `isDomainEmpty` function.

Is there a special reason we changed `(domain.resolver == null ||
      domain.resolver!.split("-")[0] == EMPTY_ADDRESS)` in the original subgraph logic to just `domain.resolverId === null`?

lightwalker-eth (commented):
@tk-o Hey really appreciate these updates. They help a lot. Not just for me, but for others who will be doing the same thing I'm doing now to gain confidence in our implementation.

There's 2 more small differences I detected. Appreciate advice on how we should approach those. Thank you!

shrugs:
just a heads up @lightwalker-eth @tk-o I'm becoming less confident in all of these logical changes of the handlers, even if they do more closely map to the subgraph line-by-line. I designed the original implementation to match the intended logic while staying modern, streamlined, and ponder-y, and the diff too has confirmed identical behavior.

The proposed changes here are giving me less, not more, confidence in the quality of our implementation. I think that the diff on these functions could be as simple as the ~8 lines i suggested in the original PR, and personally feel very confident in those line's accuracy to the subgraph, which we can trivial confirm using the diff tool.

first change: https://github.com/namehash/ensnode/pull/134#discussion_r1937728746

second change: https://github.com/namehash/ensnode/pull/134#discussion_r1937748693

Anything beyond these changes seems entirely unnecessary and allows a larger surface area for bugs to appear.

I'm happy to relent and keep all of these changes, but I struggle to see the necessity when the diff here could be trivial and testable against the diff tool.

apps/ensnode/src/handlers/Registry.ts
tk-o: Not sure about the answers, but I can make changes to follow the exact order of the reference implementation and we will see the output.

lightwalker-eth (approved):
@tk-o Approving this PR to unblock from my side. Appreciate if you could update this PR as you think would align with the vision of @shrugs. Once finished, please pass this PR to @shrugs for final reviews.

Thank you!

More details of my thoughts in Slack here: https://namehash.slack.com/archives/C086Z6FNBHN/p1739193537008069?thread_ts=1738859584.260149&cid=C086Z6FNBHN

shrugs (changes requested):
Single logical change necessary, but otherwise LGTM!

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -124,17 +128,25 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
+          const label = isLabelIndexable(healedLabel) ? healedLabel : encodeLabelhash(labelhash);
+          // update the name using the parent's name and the label value
+          // (either healed label or encoded labelhash)
+          // NOTE: for the root node, the parent is null, so we just use the label value as is
+          const name = parent?.name ? `${label}.${parent.name}` : label;
 
-          await context.db.update(schema.domain, { id: domain.id }).set({ name, labelName });
+          // akin to domain.save() c68a889
+          // https://github.com/ensdomains/ens-subgraph/blob/c68a889e0bcdc6d45033778faef19b3efe3d15fe/src/ensRegistry.ts#L86
+          await context.db.update(schema.domain, { id: domain.id }).set({ name, labelName: label });
```
shrugs, line 145: blah, lost my previous comment somehow. tldr, `Domain.labelName` needs to be undefined if label not indexable. current proposed change results in encoded label being stored there. in my suggested change in the other pr i assumed indexable check was done in the healing service client. so updated logic looks something like:

```ts
const healedLabel = await labelByHash(labelhash);
const validLabel = isLabelIndexable(healedLabel) ? healedLabel : undefined;
const label = validLabel || encodeLabelhash(labelhash);
const name = parent?.name ? `${label}.${parent.name}` : label;
await context.db.update(schema.domain, { id: domain.id }).set({ name, labelName: label });
```

shrugs: additional context: [this line](https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L113) says "set Domain.labelName to label if it is valid". it doesn't want that label to ever be encoded, which is only used for the domain.name setting logic [here](https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L117)

shrugs (approved):

PR #171, opened 2025-02-04
fix: use jsonb for texts to fix malformed string serialization bug

shrugs:
a workaround for the malformed `//U_W_U\\` text key identified in #79 

the issue is being tracked with ponder team in https://github.com/ponder-sh/ponder/issues/1484 but this workaround is acceptable for now. it produces the incorrect graphql type, similar to the bigint array stopgap, but the json data response itself continues to be subgraph-compatible


lightwalker-eth (approved):
@shrugs Nice work here!

PR #172, opened 2025-02-04
fix: add back tool versions for asdf users

shrugs:
I'm unable to get nvm and asdf to play nicely together on my system for the moment, and i'd prefer to keep asdf as i use it for basically everything — the simple fix to ensure i'm always using the correct node version for ensnode is to commit the .tool-versions file again


lightwalker-eth (approved):
LGTM

PR #186, opened 2025-02-07
ENSNode dockerfile

BanaSeba:


docker-compose.yml
```diff
@@ -12,5 +38,5 @@ services:
     restart: unless-stopped
 
 volumes:
-  ensrainbow_data:
```
BanaSeba, line 15: This volume is not being used anywhere. Removing

docker-compose.yml
```diff
@@ -1,6 +1,32 @@
...
 services:
+
+  ensnode:
+      build:
+        context: .
+        dockerfile: ./apps/ensnode/Dockerfile
+      ports:
+        - "42069:42069"
+      env_file:
+      - ./apps/ensnode/.env.local
```
BanaSeba, line 12: Reuse any variables defined in .env.local file. I don't know if this is the best approach, but I am open to any other suggestions. This way we can skip all variable definitions inside compose and share them between local non-docker/docker setup

docker-compose.yml
```diff
@@ -1,6 +1,33 @@
 version: '3.8'
 
 services:
+
+  ensnode:
+      build:
+        context: .
```
shrugs, line 7: don't need `context: .`, it's the default default

docker-compose.yml
shrugs: i like this!

docker-compose.yml
```diff
@@ -1,6 +1,33 @@
...
+      env_file:
+      - ./apps/ensnode/.env.local
+      environment:
+        # Ovveride DATABASE_URL from .env.local Refer to docker compose postgres instance
+        DATABASE_URL: postgresql://postgres:password@postgres:5432/ponder
+      depends_on:
+        - postgres
+
+  postgres:
+    image: postgres:16
```
shrugs, line 20: postgres:17

docker-compose.yml
```diff
@@ -1,6 +1,33 @@
 version: '3.8'
 
 services:
+
+  ensnode:
+      build:
+        context: .
+        dockerfile: ./apps/ensnode/Dockerfile
```
shrugs, line 8: if you add `image: namehash/ensnode` when we do `docker compose build` it will build/tag the image correctly without any other arguments

shrugs: likewise for ensrainbow

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
   - also helps indexer discovery
+
+## Dockerfile
+
+> Instructions for building the Docker image for ENSNode.
+
+To build the Docker image, navigate to the top of the monorepo and run the following command:
+
+```bash
+docker build -f apps/ensnode/Dockerfile -t ensnode .
```
shrugs, line 52: i think the easier approach is to add the `image` field in the docker compose and have people run `docker compose build` from the root of the monorepo. they can also specify a container like `docker compose build ensnode`

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+The Dockerfile expects the build context to be at the top of the monorepo. This means that all the necessary files and directories it needs to access during the build process should be available from that location. If you run the build command from a different directory, the Dockerfile may not find the required files, leading to errors. Keeping the context at the top ensures everything is organized and accessible for a smooth build process.
+
+### Team and Docker Image Repository
+The team responsible for this repository is publishing its own Docker image under the following Docker image repository: [ENSNode image repository](https://github.com/namehash/ensnode/pkgs/container/ensnode%2Fensnode). This allows users to easily pull the pre-built image without needing to build it themselves.
+
+### Running the Application
+To run the application using Docker, you can use the following command:
+
+```bash
+docker run -p 42069:42069 ensnode
```
shrugs, line 67: likewise, i think it would be easiest to standardize on docker-compose, to avoid having extra config flags floating around. i think the suggestion should be `docker compose up ensnode`

shrugs: and to that end, i'd actually prefer to remove the `docker:build` and `docker:push` scripts in the individual package folders and perhaps instead add a script at the root package.json like

```
"docker:build": "docker compose build"
"docker:push": "docker compose push"
```

and in the ci we can use it like `pnpm run docker:build ensnode && pnpm run docker:push ensnode` to build/push a specific image

docker-compose.yml
BanaSeba: Removed

docker-compose.yml
BanaSeba: Upgraded to 17

docker-compose.yml
BanaSeba: I've never used this parameter this way. Looking at docs I've found this https://docs.docker.com/reference/compose-file/build/#using-build-and-image 
So basically when you specify the build and image you will want to download the image first. If it fails then build using the local Dockerfile. So when we put `image: namehash/ensnode` there we just specify a local image name, when missing just build it from scratch. By default compose would name this image after the logical component name.
I've updated this approach as you suggested. But pay attention that this will force new rainbow image to be build at the first `compose up`

apps/ensnode/README.md
BanaSeba: I would believe that we are talking about two separate ideas here (Docker and Docker compose). Here we are just showing how to set up a single docker container as this is the context of this README (it's only describing ENSNode not a whole system).
Docker compose description at the top of the monorepo is a guide on how to set up "all" components to run on a single machine in a more holistic way. I can think about some devs or clients who are not willing to do that and would only pay attention to the ENSNode component

apps/ensnode/README.md
BanaSeba: Look at my comment above about Docker/Docker-compose. 
About `docker:build` and `docker:push`. I would also dispose of those scripts.
I'm sceptical however about relying on `docker compose build` to actually build the image before pushing. That would mean we want to build all images at once which might not be desired.
I would completely remove any Docker or docker-compose related scripts from `package.json` and rely on scripts provided in CI workflow (use CLI for Docker related tasks).

docker-compose.yml
shrugs: ah, was unaware of that specifics. seems acceptable, though? once we publish the ensrainbow image then every developer using this docker compose will pull first before attempting to build, right?

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+
+1. **Open a Terminal**: Navigate to the root directory of the ENSNode repository where the `docker-compose.yml` file is located.
+
+2. **Run Docker Compose**: Execute the following command to start the application:
+   ```bash
+   docker-compose up
+   ```
+
+   This command will:
+   - Build the Docker images for the ENSNode and ENSRainbow applications.
```
lightwalker-eth, line 241: I understand we're exploring strategies to accelerate the building of a docker image for ENSRainbow. Not sure of the status on that while I'm reviewing this line of code.

Should we make any note in the docs here (at least for now) with a notice that building ENSRainbow may take a lot of time? Advice appreciated!

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+
+3. **Access the Applications**: Once the containers are running, you can access the applications in your web browser:
+   - **ENSNode**: Open [http://localhost:42069](http://localhost:42069) to access the ENSNode indexer.
+   - **ENSRainbow**: Open [http://localhost:3001](http://localhost:3001) to access the ENSRainbow service.
+
+### Expected Outcome
+
+After running `docker-compose up`, you should see logs in your terminal indicating that the services are starting. Once everything is up and running, you can interact with the ENSNode application through the provided URLs.
+
+### Stopping the Application
```
lightwalker-eth, line 254:
```suggestion
### Stopping the Applications
```

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+
+To stop the running application, you can press `Ctrl + C` in the terminal where Docker Compose is running. If you want to remove the containers and networks created by Docker Compose, you can run:
```
lightwalker-eth, line 256:
```suggestion
To stop the running applications, you can press `Ctrl + C` in the terminal where Docker Compose is running. If you want to remove the containers and networks created by Docker Compose, you can run:
```

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+```bash
+docker-compose down
+```
+
+This command will stop and remove all containers defined in the `docker-compose.yml` file.
```
lightwalker-eth, line 261: I assume then that this also means this will remove all of the state in Postgres that might have been indexed? If so, could you please make that more explicit?

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+
+### Summary
+
+Using Docker Compose simplifies the process of setting up and running the ENSNode application along with its dependencies. By following the steps above, you can quickly get the application running on your local machine without needing to manually configure each component.
```
lightwalker-eth, line 265:
```suggestion
Using Docker Compose simplifies the process of setting up and running ENSNode along with its dependencies. By following the steps above, you can quickly get the application running on your local machine without needing to manually configure each component.
```

README.md
lightwalker-eth, line 254: Sorry if this is a dumb question.

Let's say I follow all the steps under "Running the Application" and then the steps for "Stopping the Application".

What do I do now if I want to run the application again? I assume I shouldn't need to fully rebuild all the docker containers again? Advice appreciated for how we communicate these ideas :+1:

docker-compose.yml
```diff
@@ -1,16 +1,43 @@
...
+  ensnode:
+      build:
+        dockerfile: ./apps/ensnode/Dockerfile
+      image: namehash/ensnode
+      ports:
+        - "42069:42069"
+      env_file:
+      - ./apps/ensnode/.env.local
+      environment:
+        # Ovveride DATABASE_URL from .env.local Refer to docker compose postgres instance
```
lightwalker-eth, line 14:
```suggestion
        # Override DATABASE_URL from .env.local to refer to docker compose postgres instance
```

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+```bash
+docker build -f apps/ensnode/Dockerfile -t namehash/ensnode .
+```
+
+This command will use the Dockerfile located in the `apps/ensnode` directory to create the image.
+
+### Important Note
+The Dockerfile expects the build context to be at the top of the monorepo. This means that all the necessary files and directories it needs to access during the build process should be available from that location. If you run the build command from a different directory, the Dockerfile may not find the required files, leading to errors. Keeping the context at the top ensures everything is organized and accessible for a smooth build process.
+
+### Team and Docker Image Repository
```
lightwalker-eth, line 60:
```suggestion
### Public Docker Image Repository
```

Is this change fair?

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+The team responsible for this repository is publishing its own Docker image under the following Docker image repository: [ENSNode image repository](https://github.com/namehash/ensnode/pkgs/container/ensnode%2Fensnode). This allows users to easily pull the pre-built image without needing to build it themselves.
```
lightwalker-eth, line 61:
```suggestion
NameHash Labs publishes prebuilt ENSNode Docker images for the community under a [Docker image repository](https://github.com/namehash/ensnode/pkgs/container/ensnode%2Fensnode). You can use this to easily pull a prebuilt image without needing to build it yourself.
```

A few questions:
1. Are the above suggestions fair?
2. If someone from the community wanted to use one of these prebuilt images, how would they do it? Sorry if this is a stupid question.

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+
+### Running the Application
+To run the application using Docker, you can use the following command:
```
lightwalker-eth, line 64:
```suggestion
After building the Docker image, you can run the application with the following command:
```

Is that change fair? The main goal in this proposed change is to make sure its very explicit that you need to finish building the image first before this command will work.

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+
+This command will use the Dockerfile located in the `apps/ensnode` directory to create the image.
+
+### Important Note
+The Dockerfile expects the build context to be at the top of the monorepo. This means that all the necessary files and directories it needs to access during the build process should be available from that location. If you run the build command from a different directory, the Dockerfile may not find the required files, leading to errors. Keeping the context at the top ensures everything is organized and accessible for a smooth build process.
+
+### Team and Docker Image Repository
+The team responsible for this repository is publishing its own Docker image under the following Docker image repository: [ENSNode image repository](https://github.com/namehash/ensnode/pkgs/container/ensnode%2Fensnode). This allows users to easily pull the pre-built image without needing to build it themselves.
+
+### Running the Application
```
lightwalker-eth, line 63: Sorry, but a few stupid questions:

1. I've noted the instructions for how to customize the env vars in the overall `docker-compose.yml` file. But what if I'm just building the Docker container exclusively for ENSNode using the Dockerfile in this `apps/ensnode` directory? How do I customize my env vars here? Do I do it when building the container and/or when I run the container?
2. How does it work if someone downloads the prebuilt Docker image from the image repository on GitHub? How do they customize the env vars there?

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+To run the application using Docker, you can use the following command:
+
+```bash
+docker run -p 42069:42069 namehash/ensnode
+```
+
+This command will start the container and map port `42069` from the container to your local machine, allowing you to access the application.
```
lightwalker-eth, line 70:
```suggestion
This command will start the container for ENSNode and map port `42069` from the container to your local machine, allowing you to access ENSNode.
```

Is that fair? The goal in this suggestion is to be more explicit on exactly what application this starts, since there's more than one application.

apps/ensnode/README.md
```diff
@@ -41,3 +41,33 @@ estimated mainnet-only backfill time @ <=500rps = **~13 hours** on M1 Macbook (>
...
+
+### Docker Compose
+For a more convenient setup, you can also use Docker Compose. Refer to the Docker Compose usage instructions at the top of the monorepo for details on how to run the application with Docker Compose, which simplifies the process of managing multi-container applications.
```
lightwalker-eth, line 73: A few ideas here:

1. Suggest changing "Refer to the Docker Compose usage instructions at the top of the monorepo" to an link to the associated README.md file that is an anchor link to the specific section you're referring to here.
2. As I understand, if someone just builds and runs the ENSNode docker image by itself then we should be telling them that they need to be sure to manually configure the env vars (please see my questions above related to how we communicate this instruction) for dependent services: Postgres and ENSRainbow. I'm worried about people running ENSNode through any strategy before they manually configure env vars. For each of the journeys we document for people to get it running, please ensure we always explicitly encourage configuring env vars before they actually run it.

lightwalker-eth (approved):
@BanaSeba Very happy to see this! Great work on these docs! I have a number of questions / suggestions but approving this PR since I assume it would be fine to merge this PR and work on the feedback I shared in a separate future PR.

Appreciate advice on the path forward! Thank you!

docker-compose.yml
BanaSeba: Yes, this should be fine

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
 
   </details>
 </details>
+
+## Using Docker Compose
+
+Docker Compose is a tool that allows you to define and run multi-container Docker applications. In this monorepo, we use Docker Compose to set up the ENSNode application along with its dependencies, such as a PostgreSQL database.
```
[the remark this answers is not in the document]
BanaSeba: Yes, adjusted

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+   ```
+
+2. **Prepare the Environment**: Ensure you have a `.env.local` file in the `apps/ensnode` directory. This file contains environment variables needed for the application. You can create it by copying the example file:
+   ```bash
+   cp apps/ensnode/.env.local.example apps/ensnode/.env.local
+   ```
+
+   Then, edit the `.env.local` file to configure your local settings as needed.
+
+### Running the Application
```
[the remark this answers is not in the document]
BanaSeba: Adjusted

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+
+To start the ENSNode application and its dependencies using Docker Compose, follow these steps:
+
+1. **Open a Terminal**: Navigate to the root directory of the ENSNode repository where the `docker-compose.yml` file is located.
```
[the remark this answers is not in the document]
BanaSeba: Adjusted

README.md
BanaSeba: I would not pay attention to that here. We already stated that in ENSRainbow docker image build README. Will follow up on this issue as fast as I can

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+
+2. **Run Docker Compose**: Execute the following command to start the application:
+   ```bash
+   docker-compose up
+   ```
+
+   This command will:
+   - Build the Docker images for the ENSNode and ENSRainbow applications.
+   - Start the PostgreSQL database container.
```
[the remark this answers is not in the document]
BanaSeba: All makes sense. One note here is that this port is not "public" for the Internet. This is only "public" for localhost that will run compose. If you would want to expose this port for traffic coming externally from "localhost" additional work is needed.
I would allow this because of the "development" nature of the docker compose. Clients will probably want to experiment and see by their own eyes what's going on inside of the database (same for the developers that can be using this). Need of clearing schemas from time to time is not crucial in this thinking.

README.md
BanaSeba: Adjusted

README.md
BanaSeba: Adjusted

README.md
BanaSeba: No it will not. For local and docker-compose setup we are not ovveriding `DATABASE_SCHEMA`. This means that DB that indexer is writing to is controlled inside ENV variable not a runtime argument. Stopping application is not forcing any reindexing procedure

README.md
BanaSeba: Adjusted

README.md
BanaSeba: There is no additional work neede between stopping and running applications. You just follow start/stop procedures. No rebuild needed as those images are already in you local Docker image registry. Would not extend this section with any additional details

README.md
```diff
@@ -199,3 +199,67 @@ Fetch data about the three most recently-created domains.
...
+   - Start the ENSNode application, which will be accessible on port `42069`.
+   - Start the ENSRainbow application, which will be accessible on port `3001`.
+
+3. **Access the Applications**: Once the containers are running, you can access the applications in your web browser:
+   - **ENSNode**: Open [http://localhost:42069](http://localhost:42069) to access the ENSNode indexer.
+   - **ENSRainbow**: Open [http://localhost:3001](http://localhost:3001) to access the ENSRainbow service.
+
+### Expected Outcome
+
+After running `docker-compose up`, you should see logs in your terminal indicating that the services are starting. Once everything is up and running, you can interact with the ENSNode application through the provided URLs.
```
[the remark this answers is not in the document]
BanaSeba: Yep, makes sense. Adjusted

apps/ensnode/README.md
BanaSeba: Yep, all good. Adjusted

apps/ensnode/README.md
BanaSeba: Yep, adjusted

apps/ensnode/README.md
BanaSeba: Adjusted

apps/ensnode/README.md
BanaSeba: 1. I've added mention about setting up env variables for container run. They are defined during contianer creation, no need to provide them during image build.
2. The same way. You are providing them when runnning `docker run` see an example provided.

apps/ensnode/README.md
BanaSeba: 1. Added links to correct files.
2. Added instructions for setting up env vars and creating their own DB instance for indexer

PR #206, opened 2025-02-10
ENSRainbow: Handle command line arguments

djstrong:


apps/ensrainbow/src/lib/operations.ts
```diff
@@ -75,7 +77,7 @@ export const countLabels = async (context: ENSRainbowContext): Promise<CountResp
     };
```
lightwalker-eth, line 77: GitHub won't let me put this comment several lines above where it belongs.

I'm looking at the call to `parseInt`.

This function is not quite what you want it to be.

Here's a quick suggestion that I generated with ChatGPT:

```
function parseNonNegativeInteger(input: string): number | null {
    const num = Number(input);

    // Ensure it's a finite number, an integer, and non-negative
    if (Number.isFinite(num) && Number.isInteger(num) && num >= 0) {
        return num;
    }
    
    return null; // Return null if invalid
}

// Examples:
console.log(parseNonNegativeInteger("42"));    // 42
console.log(parseNonNegativeInteger("3.14"));  // null (not an integer)
console.log(parseNonNegativeInteger("-5"));    // null (negative)
console.log(parseNonNegativeInteger("abc"));   // null (not a number)
```

A few suggestions:
1. Define a utility function such as `parseNonNegativeInteger`.
2. Include unit tests for that utility function.
3. Whenever reading the `LABELHASH_COUNT_KEY`, always run `parseNonNegativeInteger` on the response to ensure it is in the expected format. If it returns `null` then we should still throw the internal server error code.
4. In the definition of `CountSuccess`, add a comment on the `count` field that it is always a non-negative integer.

apps/ensrainbow/src/lib/operations.ts
```diff
@@ -35,25 +39,23 @@ export const heal = async (
...
     const label = await context.db.get(labelHashBytes);
-    console.info(`Successfully healed labelhash ${labelhash} to label "${label}"`);
+    logger.info(`Successfully healed labelhash ${labelhash} to label "${label}"`);
     const result: HealSuccess = {
       status: StatusCode.Success,
       label,
     };
     return result;
   } catch (error) {
     if ((error as any).code === "LEVEL_NOT_FOUND") {
```
lightwalker-eth, line 49: A few ideas here. Please feel welcome to follow up in a separate future PR if you prefer.

1. I notice we aren't doing a similar check on the `countLabels` operation. It seems maybe we should so that we can return a more helpful server error message if someone has somehow messed up the leveldb attached to the ENSRainbow server?
2. Would it make sense to build a little helper function in database.ts for how we lookup a key / value pair in LevelDB? Please feel welcome to attack that idea! I was thinking maybe we would have it always return `string | null` (where `null` represents key not found) and then also document how it will throw an error if any error occurs other than "LEVEL_NOT_FOUND".... (or something like that...). This might help to keep ideas that are internal to how LevelDB works inside `database.ts`.

apps/ensrainbow/src/lib/operations.ts
```diff
@@ -1,16 +1,20 @@
...
   HealSuccess,
-} from "./utils/response-types.js";
-import { ErrorCode, StatusCode } from "./utils/response-types.js";
+  StatusCode,
+} from "../utils/response-types.js";
+
+const logger = createLogger();
 
 export interface ENSRainbowContext {
   db: ClassicLevel<ByteArray, string>;
```
lightwalker-eth, line 20: Suggest defining something like `type ENSRainbowDB: ClassicLevel<ByteArray, string>;` inside database.ts.

This would help to further consolidate all LevelDB specific ideas into database.ts.

It also could be a nice place to document the schema (including our special label count key) used in the ENSRainbowDB schema.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -0,0 +1,23 @@
+import { join } from "path";
+import { ClassicLevel } from "classic-level";
+import { ByteArray } from "viem";
+import { createLogger } from "../utils/logger";
+
+const logger = createLogger();
+
+export const getDataDir = () => process.env.DATA_DIR || join(process.cwd(), "data");
+
+export const initializeDatabase = (dataDir: string): ClassicLevel<ByteArray, string> => {
```
lightwalker-eth, line 10: Please see my other comment suggesting to create an `ENSRainbowDB` type. We could use that here.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -0,0 +1,23 @@
...
+  logger.info(`Initializing database with data directory: ${dataDir}`);
+
+  try {
+    return new ClassicLevel<ByteArray, string>(dataDir, {
```
lightwalker-eth, line 14: Please see my other comment suggesting to create an `ENSRainbowDB` type. We could use that here.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -0,0 +1,23 @@
...
+      keyEncoding: "binary",
+      valueEncoding: "utf8",
+    });
+  } catch (error) {
```
lightwalker-eth, line 18: Suggest adding a specialized error message for the case the opening fails with error code `[LEVEL_LOCKED](https://github.com/Level/abstract-level#errors)`

If this error happens, suggest we give a nice error message about the database already being in use. Appreciate your advice.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -0,0 +1,23 @@
...
+const logger = createLogger();
+
+export const getDataDir = () => process.env.DATA_DIR || join(process.cwd(), "data");
+
+export const initializeDatabase = (dataDir: string): ClassicLevel<ByteArray, string> => {
+  logger.info(`Initializing database with data directory: ${dataDir}`);
+
+  try {
+    return new ClassicLevel<ByteArray, string>(dataDir, {
+      keyEncoding: "binary",
```
lightwalker-eth, line 15: I'm not sure if I'm reading these docs correctly: https://github.com/Level/transcoder#built-in-encodings

Should we change `keyEncoding` to `view`?  The goal in suggesting this so that if we ask LevelDB for a key it would be returned back to us as a `Uint8Array` (a byte array) too, rather than a `Buffer`?

Not sure, just thought it would be good to share :+1:

apps/ensrainbow/src/lib/database.ts
lightwalker-eth, line 14: I'm studying the docs here: https://github.com/Level/abstract-level?tab=readme-ov-file#dbopenoptions

It seems to me that we should split `initializeDatabase` into two separate functions:
1. createDatabase
2. openDatabase

The ingest command would use `createDatabase`, while the serve command would use `openDatabase`.

`createDatabase` should set `errorIfExists` to `true`.

`openDatabase` should set `createIfMissing` to `false`.

We should then have helpful error messages if `createDatabase` fails because it already exists.

Likewise, we should have helpful error messages if `openDatabase` fails because it doesn't already exist.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -0,0 +1,23 @@
...
+      valueEncoding: "utf8",
+    });
+  } catch (error) {
+    logger.error("Failed to initialize database:", error);
+    logger.error(`Please ensure the directory ${dataDir} exists and is writable`);
```
lightwalker-eth, line 20: Does the directory need to be writable even if we're just in serve mode?

apps/ensrainbow/src/index.test.ts
```diff
@@ -1,17 +1,28 @@
...
 
 describe("ENS Rainbow API", () => {
   let server: ReturnType<typeof serve>;
+  let db: ClassicLevel<ByteArray, string>;
+  let app: ReturnType<typeof createServer>;
 
   beforeAll(async () => {
+    db = initializeDatabase("test-data");
+    app = createServer(db, console);
     // Start the server on a different port than the main app
```
lightwalker-eth, line 26:
```suggestion
    // Start the server on a different port than what ENSRainbow defaults to
```

shrugs (approved):
LGTM

apps/ensrainbow/src/commands/ingest-command.ts
```diff
@@ -76,14 +73,27 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
+          );
+          invalidRecords++;
+          continue;
+        }
+      }
     } catch (e) {
       if (e instanceof Error) {
-        console.warn(
+        log.warn(
           `Skipping invalid record: ${e.message} - this record would be unreachable via ENS Subgraph`,
```
lightwalker-eth, line 91:
```suggestion
          `Skipping invalid record: ${e.message} - this record is safe to skip as it would be unreachable by the ENS Subgraph`,
```

apps/ensrainbow/src/commands/ingest-command.ts
```diff
@@ -2,36 +2,32 @@ import { createReadStream } from "fs";
...
-    valueEncoding: "utf8",
-  });
+export async function ingestCommand(options: IngestCommandOptions): Promise<void> {
+  const log = createLogger(options.logLevel);
+  const db = initializeDatabase(options.dataDir);
 
   // Clear existing database before starting
-  console.log("Clearing existing database...");
+  log.info("Clearing existing database...");
   await db.clear();
```
lightwalker-eth, line 29: I shared another related comment about the `initializeDatabase` function where I suggested splitting it into `openDatabase` and `createDatabase`.

Perhaps `createDatabase` should have a param to clear an existing database if and only if it is found? If so, this whole db.clear logic and its associated log messages might be moved there.

apps/ensrainbow/src/commands/ingest-command.ts
```diff
@@ -105,21 +115,18 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
   }
 
   await db.close();
-  console.log("\nData ingestion complete!");
+  log.info("\nData ingestion complete!");
 
   // Validate the number of processed records
   if (processedRecords !== TOTAL_EXPECTED_RECORDS) {
-    console.warn(
+    log.warn(
```
lightwalker-eth, line 122: Maybe this should change from a warning into an error? Both for the log statement and the prefix of the log message?

apps/ensrainbow/src/commands/ingest-command.ts
```diff
@@ -105,21 +115,18 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
   }
-}
 
-// Check if this module is being run directly
-if (import.meta.url === `file://${process.argv[1]}`) {
-  loadEnsNamesToLevelDB()
-    .then(() => console.log("Done!"))
-    .catch(console.error);
+  if (invalidRecords > 0) {
+    log.warn(`Found ${invalidRecords} records with invalid hashes during processing`);
```
lightwalker-eth, line 130: Maybe this should change to an error?

apps/ensrainbow/src/commands/ingest-command.ts
```diff
@@ -105,21 +115,18 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
...
 
   // Validate the number of processed records
   if (processedRecords !== TOTAL_EXPECTED_RECORDS) {
-    console.warn(
+    log.warn(
       `Warning: Expected ${TOTAL_EXPECTED_RECORDS} records but processed ${processedRecords}`,
     );
   } else {
-    console.log(`Successfully ingested all ${processedRecords} records`);
+    log.info(`Successfully ingested all ${processedRecords} records`);
```
lightwalker-eth, line 126: Maybe we should always write some messages like this one to the console, no matter what the log level is? Appreciate your advice.

apps/ensrainbow/src/commands/ingest-command.ts
```diff
@@ -76,14 +73,27 @@ async function loadEnsNamesToLevelDB(): Promise<void> {
     let record;
     try {
       record = buildRainbowRecord(line);
+
+      if (options.validateHashes) {
```
lightwalker-eth, line 77: This `validateHashes` check should be moved outside of this try / catch block. Only `buildRainbowRecord` should be in this try catch block to ensure the way we handle an error there is purely due to `buildRainbowRecord` failing and not something else.

lightwalker-eth (changes requested):
@djstrong This is a huge upgrade! 🚀 🚀 Reviewed and shared some suggestions. Sorry that many of these suggestions go outside the scope of this PR. Sometimes its just a lot easier to give feedback on a line of code as I'm reviewing it and not to create a bunch of separate issues to track.

apps/ensrainbow/src/cli.ts
```diff
@@ -0,0 +1,123 @@
...
+          type: "string",
+          description: "Path to the gzipped SQL dump file",
+          default: join(process.cwd(), "ens_names.sql.gz"),
+        })
+        .option("data-dir", {
+          type: "string",
+          description: "Directory to store LevelDB data",
+          default: getDataDir(),
+        })
+        .option("validate-hashes", {
```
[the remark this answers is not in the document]
djstrong: Created new issue: https://github.com/namehash/ensnode/issues/212

apps/ensrainbow/src/commands/ingest-command.ts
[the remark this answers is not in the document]
djstrong: Sure, we can, it is based on: https://github.com/namehash/ens-subgraph-transition-tools/tree/main/tools/snapshot-eq/commands

apps/ensrainbow/src/lib/database.ts
djstrong: I think you are right.

apps/ensrainbow/package.json
```diff
@@ -35,16 +35,21 @@
...
     "@types/node": "^20.17.14",
     "@types/progress": "^2.0.7",
     "@types/supertest": "^6.0.2",
+    "@types/yargs": "^17.0.32",
     "@vitest/coverage-v8": "^1.6.0",
     "supertest": "^6.3.4",
     "typescript": "^5.3.3",
     "vitest": "^1.6.0"
+  },
+  "bin": {
```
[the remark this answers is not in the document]
djstrong: It allows for installing an app globally in the operating system and using this command. I am removing it.

apps/ensrainbow/src/commands/ingest-command.ts
lightwalker-eth: @djstrong Suggest to either:
1. Remove the `db.clear()` here, or;
2. Add a param to `createDatabase` for optionally clearing an existing database. If you follow that path then I suppose the ingest command should also have an optional flag for clearing an existing database?

Appreciate your advice.

apps/ensrainbow/src/commands/count-command.ts
```diff
@@ -0,0 +1,44 @@
...
+  logLevel?: LogLevel;
+}
+
+export async function countCommand(options: CountCommandOptions): Promise<void> {
+  const log = createLogger(options.logLevel);
+  const db = openDatabase(options.dataDir, options.logLevel);
+
+  // Try to read existing count
+  try {
+    const existingCount = await db.get(LABELHASH_COUNT_KEY);
```
lightwalker-eth, line 17: Seems we could use the utility function here.

lightwalker-eth (approved):
@djstrong Hey the updates look great :+1: Reviewed and shared some small suggestions. Please feel welcome to merge this PR whenever you're ready.

apps/ensrainbow/src/commands/ingest-command.ts
djstrong: I have removed it.

PR #207, opened 2025-02-10
fix: add ensrainbow dependency to docker-compose

shrugs:
- version is deprecated
- adds ensrainbow dependency to enforce start order


tk-o (approved):
Looks good to me!

lightwalker-eth (approved):
@shrugs Also looks good to me :+1:

Issue #119, opened 2025-01-30
ENSRainbow: Handle interrupted ingestion


lightwalker-eth:
What happens if the ingest is interrupted for some reason before it finishes?

Goal for this story is to ensure we can gracefully recover from this issue. For example: either logging a message to the console with instructions for manual recovery or automatically recovering. Very open to either.

lightwalker-eth:
A quick idea:

- [ ] Introduce a special key to represent the idea that an Ingestion has started.
- [ ] When an ingestion is started, check if ingestion key is set. If so...?
   - [ ] ... Maybe just exit with error message for now with instructions to manually clear all data and start again?
- [ ] Set ingestion key when starting ingestion
- [ ] Upon completion of ingestion, clear ingestion key
- [ ] Upon startup of ENSRainbow, check if ingestion key is set, if it is exit with error message that ingestion started but not completed with instructions to run ingestion again.

lightwalker-eth:
@djstrong Appreciate that, but I still prefer the idea of adding a distinct key just for the purpose of identifying that an ingestion was interrupted and not completed successfully.

An example for why this is helpful: When someone runs the new `serve` command you built, ENSRainbow should check if this special key is found in its attached LevelDB. If it is, it can immediately report an error before it ever starts accepting API requests.

djstrong:
We are not iterating through all the keys in the `serve` command - we are reading `LABELHASH_COUNT_KEY`.
 
But if there is no key identifying that an ingestion was interrupted it needs to check also `LABELHASH_COUNT_KEY` if `count` command was successful.

lightwalker-eth:
@djstrong Yes, appreciate that we are not iterating through all the keys in the `serve` command :+1:

Ah I think I see what you're saying. I think we're generally talking about similar goals here. If I understand you right, you're suggesting to use the `LABELHASH_COUNT_KEY` for the goals we're discussing. On the other hand, I'm suggesting to add a new distinct key for this.

An example for why we should have a distinct key for this, please consider other goals we're discussing such as https://github.com/namehash/ensnode/issues/162

Please also see feedback I shared in PR 206 suggesting to merge the `count` command into "phase 2" of the `ingest` command: https://github.com/namehash/ensnode/pull/206

Here's an example scenario:

* Imagine that you successfully complete an ingest of label set "1".
   * This sets `LABELHASH_COUNT_KEY`.
* Then imagine you begin an ingest of label set "2", but for some reason this is interrupted.
[...]

PR #223, opened 2025-02-11
ENSRainbow: Handle interrupted ingestion

djstrong:
https://github.com/namehash/ensnode/issues/119

### DIFF
--- a/apps/ensrainbow/src/commands/count-command.ts
+++ b/apps/ensrainbow/src/commands/count-command.ts
@@ -1,5 +1,9 @@
-import { ENSRainbowDB, safeGet } from "../lib/database.js";
-import { LABELHASH_COUNT_KEY } from "../lib/database.js";
+import {
+  ENSRainbowDB,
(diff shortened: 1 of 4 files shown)


lightwalker-eth (approved):
@djstrong Hey thanks for this upgrade :+1:

apps/ensrainbow/src/commands/count-command.ts
```diff
@@ -39,6 +38,4 @@ export async function countCommand(options: CountCommandOptions): Promise<void>
```
lightwalker-eth, line 38: @djstrong Hey, this comment is for several lines above. Suggest adding one more check that the key we're counting isn't this `INGESTION_IN_PROGRESS_KEY`.

I like that because the count command should always return the correct count, even if run when the `INGESTION_IN_PROGRESS_KEY` key is set.

lightwalker-eth (approved):
@djstrong One enhancement suggestion.

PR #192, opened 2025-02-08
Generalize ens-deployments package

lightwalker-eth:
Fixes https://github.com/namehash/ensnode/issues/190

Related ideas: https://github.com/namehash/ensnode/issues/87

# Goals

* Prepare the `ens-deployments` package for publishing on NPM (in some future PR) for broader use by the ENS community.
* Carefully refine our documentation of important ENS nuances and subtleties.
   * Mitigate the risks that expert ENS devs might find disagreement with how this package models the details of ENS deployments.
   * Help non-expert ENS devs achieve more accurate mental models of how ENS actually works (such that they are less likely to make bad assumptions about the true implementation of ENS).

# Subtasks

@lightwalker-eth has proposed a starting point for this PR. Additional work remains, including:

- [x] Refine `ens-deployments` package
   - [x] Remove any remaining use of "plugin" language within `ens-deployments`.
   - [x] Remove any remaining references or dependencies to Ponder within `ens-deployments`
      - [x] Refine comments / definition of `SubregistryContractConfig`.
      - [x] Refine any references to `ContractConfig#filter`
   - [x] Add root `README.md` file that provides a simple introduce to the package. Suggest lifting content from `src/index.ts` as relevant.
      - [x] Identify that we make use of terminology as defined in https://www.ensnode.io/reference/terminology.
   - [x] Refine `package.json`
      - [x] Refine `description` field. Ex: "Catalog of ENS deployments, including key contract addresses and other attributes."
      - [x] Refine description of package in monorepo root README.md
      - [x] Remove `ponder` dependency.
      - [x] Change `viem` dependency so that it falls under `peerDependencies`.
   - [x] Question: Perhaps we should change the `SubregistryName` enum values as follows:
      - [x] `base` -> `base.eth`
      - [x] `linea` -> `linea.eth`
   - [x] Question: Would it make sense to move our abis into `ens-deployments`?
- [x] Refine `ensnode` app to make use of the new `ens-deployments` package structure
   - NOTE: Initial commits on this PR from @lightwalker-eth that made changes to the `ensnode` app are all side effects of proposed renamings within `ens-deployments`. Perhaps `ensnode` retains the concept of a "plugin" but it will need to "translate" from the generic "subregistry" definitions in `ens-deployments` as necessary.
- [x] Suggest next steps on the ideas discussed in the following:
   - [x] https://github.com/namehash/ensnode/pull/192/commits/0b939f449b862375e43712b6b5a53e3a541921ee
[...]

PR #225, opened 2025-02-11
212 ensrainbow implement validate command

djstrong:


apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,87 @@
...
+  let hashMismatches = 0;
+
+  log.info("Starting database validation...");
+
+  // Validate each key-value pair
+  for await (const [key, value] of db.iterator()) {
+    totalKeys++;
+
+    // Skip count key
+    if (byteArraysEqual(key, LABELHASH_COUNT_KEY)) {
```
lightwalker-eth, line 29: We should do something similar for `INGESTION_IN_PROGRESS_KEY`.

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,87 @@
...
+  let invalidHashes = 0;
+  let hashMismatches = 0;
+
+  log.info("Starting database validation...");
+
+  // Validate each key-value pair
+  for await (const [key, value] of db.iterator()) {
+    totalKeys++;
+
+    // Skip count key
```
lightwalker-eth, line 28:
```suggestion
    // Skip keys not associated with rainbow records
```

Suggesting to write this in a more general way

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,87 @@
...
+    log.error(`Count mismatch: stored=${parsedCount}, actual=${actualCount}`);
+    throw new Error(`Count mismatch: stored=${parsedCount}, actual=${actualCount}`);
+  } else {
+    log.info(`Count verified: ${actualCount} records`);
+  }
+
+  // Report results
+  log.info("\nValidation Results:");
+  log.info(`Total keys: ${totalKeys}`);
+  log.info(`Valid hashes: ${validHashes}`);
```
lightwalker-eth, line 74:
```suggestion
  log.info(`Valid rainbow records: ${validHashes}`);
```

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,87 @@
...
+  log.info(`Invalid hashes: ${invalidHashes}`);
```
lightwalker-eth, line 75:
```suggestion
  log.info(`Invalid rainbow records: ${invalidHashes}`);
```

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,87 @@
...
+  log.info(`Hash mismatches: ${hashMismatches}`);
```
lightwalker-eth, line 76:
```suggestion
  log.info(`labelhash mismatches: ${hashMismatches}`);
```

Is that fair?

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  // Verify count
+  const storedCount = await safeGet(db, LABELHASH_COUNT_KEY);
+  const actualCount = totalKeys - 1; // Subtract 1 for count key
+
+  if (!storedCount) {
+    log.error("Count key missing from database");
+    throw new Error("Count key missing from database");
+  }
+
```
lightwalker-eth, line 62: We should also have a check here if `INGESTION_IN_PROGRESS_KEY` is set. If it is, it should be a validation error.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -72,17 +72,22 @@ export const createDatabase = async (
...
 
   try {
-    return new ClassicLevel<ByteArray, string>(dataDir, {
+    const db = new ClassicLevel<ByteArray, string>(dataDir, {
       keyEncoding: "view",
       valueEncoding: "utf8",
       createIfMissing: false,
       errorIfExists: false,
     });
+    await db.open();
```
lightwalker-eth, line 89: @djstrong The code change you added here looks right. Thank you :+1:

Things are getting a bit complicated with change management at the moment. Please note the comment I shared on PR 192 from @shrugs: https://github.com/namehash/ensnode/pull/192#discussion_r1952287184

@djstrong I'm not sure what to suggest at the moment for resolving this, but appreciate if you could please take the lead on ensuring that eventually after all these PRs are merged we have the `db.open()` logic correct. Thank you!

apps/ensrainbow/README.md
```diff
@@ -251,6 +251,24 @@ The service handles graceful shutdown on SIGTERM and SIGINT signals (e.g., when
...
+
+```bash
+# For development
+pnpm validate
+
+# For production build
+pnpm validate:prod
+```
+
+The validator performs the following checks:
```
lightwalker-eth, line 266: Could we please also add a check for the `INGESTION_IN_PROGRESS_KEY`

apps/ensrainbow/README.md
```diff
@@ -251,6 +251,24 @@ The service handles graceful shutdown on SIGTERM and SIGINT signals (e.g., when
...
+- Verifies all keys are valid labelhashes
+- Ensures stored labels match their corresponding labelhashes
+- Validates the total record count
+- Reports detailed statistics including valid hashes, invalid hashes, and any mismatches
```
lightwalker-eth, line 270:
```suggestion
- Reports detailed statistics including valid rainbow records, invalid labelhashes, and any labelhash mismatches
```

apps/ensrainbow/README.md
```diff
@@ -251,6 +251,24 @@ The service handles graceful shutdown on SIGTERM and SIGINT signals (e.g., when
...
+pnpm validate
+
+# For production build
+pnpm validate:prod
+```
+
+The validator performs the following checks:
+- Verifies all keys are valid labelhashes
+- Ensures stored labels match their corresponding labelhashes
+- Validates the total record count
```
lightwalker-eth, line 269:
```suggestion
- Validates the total rainbow record count
```

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -0,0 +1,102 @@
...
+import { labelhash } from "viem";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { LABELHASH_COUNT_KEY, createDatabase } from "../lib/database.js";
+import { validateCommand } from "./validate-command.js";
+
+describe("Validate Command", () => {
+  let tempDir: string;
+
+  beforeEach(async () => {
+    tempDir = await mkdtemp(join(tmpdir(), "ensrainbow-test-"));
```
lightwalker-eth, line 14:
```suggestion
    tempDir = await mkdtemp(join(tmpdir(), "ensrainbow-test-validate"));
```

What do you think?

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  });
+
+  afterEach(async () => {
+    await rm(tempDir, { recursive: true, force: true });
+  });
+
+  it("should validate an empty database", async () => {
```
lightwalker-eth, line 21: Shouldn't an empty database fail because the count key isn't present?

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    const db = await createDatabase(tempDir);
+    await db.close();
+
+    await expect(validateCommand({ dataDir: tempDir })).rejects.toThrow();
+  });
```
lightwalker-eth, line 26: Could we please also add a check for the `INGESTION_IN_PROGRESS_KEY` being present?

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    // Add count
+    await db.put(LABELHASH_COUNT_KEY, testData.length.toString());
+
+    await db.close();
+
+    await expect(validateCommand({ dataDir: tempDir })).resolves.not.toThrow();
+  });
+
+  it("should detect invalid labelhash format", async () => {
```
lightwalker-eth, line 49: This unit test should be refined so that everything is valid with the exception of the one invalid labelhash format.

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    // Add record with invalid labelhash format
+    const invalidLabelhash = new Uint8Array([1, 2, 3]); // Too short
+    await db.put(invalidLabelhash, "test");
+
+    await db.close();
+
+    await expect(validateCommand({ dataDir: tempDir })).rejects.toThrow();
+  });
+
+  it("should detect labelhash mismatch", async () => {
```
lightwalker-eth, line 61: This unit test should be refined so that everything is valid with the exception of the one labelhash mismatch

lightwalker-eth (changes requested):
@djstrong Hey, as part of this PR, suggest we also remove the following. What do you think?

CleanShot 2025-02-12 at 20 12 57

CleanShot 2025-02-12 at 20 14 15

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,104 @@
...
+  const storedCount = await safeGet(db, LABELHASH_COUNT_KEY);
+
+  if (!storedCount) {
+    log.error("Count key missing from database");
+    throw new Error("Count key missing from database");
+  } else {
+    rainbowRecordCount = rainbowRecordCount - 1;
+  }
+
+  // Check if ingestion is in progress
```
lightwalker-eth, line 73: @djstrong Suggest moving this check before beginning to iterate through all the keys. Its nice to fail validation fast if this key is set.

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,104 @@
...
+    log.error("Database is in an invalid state: ingestion in progress flag is set");
+    throw new Error("Database is in an invalid state: ingestion in progress flag is set");
+  }
+
+  const parsedCount = parseNonNegativeInteger(storedCount);
+  if (parsedCount !== rainbowRecordCount) {
+    log.error(`Count mismatch: stored=${parsedCount}, actual=${rainbowRecordCount}`);
+    throw new Error(`Count mismatch: stored=${parsedCount}, actual=${rainbowRecordCount}`);
+  } else {
+    log.info(`Count verified: ${rainbowRecordCount} records`);
```
lightwalker-eth, line 85:
```suggestion
    log.info(`Count verified: ${rainbowRecordCount} rainbow records`);
```

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -0,0 +1,104 @@
...
+  const ingestionInProgress = await safeGet(db, INGESTION_IN_PROGRESS_KEY);
+  if (ingestionInProgress) {
+    log.error("Database is in an invalid state: ingestion in progress flag is set");
+    throw new Error("Database is in an invalid state: ingestion in progress flag is set");
+  }
+
+  const parsedCount = parseNonNegativeInteger(storedCount);
+  if (parsedCount !== rainbowRecordCount) {
+    log.error(`Count mismatch: stored=${parsedCount}, actual=${rainbowRecordCount}`);
+    throw new Error(`Count mismatch: stored=${parsedCount}, actual=${rainbowRecordCount}`);
```
lightwalker-eth, line 83: Should we really throw here? If we do, the way the logic is now, we never call `db.close`.

Right now we're mixing up responsibilities too much across different layers.

Maybe:

1. Refactor so that you move the majority of the validation logic here into a `validate` function in `database.ts`.
   2. This function should assume an already opened database is passed to it.
   3. This function should return a bool to indicate if validation passed or failed.
   4. This function should try to maximally complete full validation if possible (and not throw / exit early).
2. Keep a `validateCommand` function in this file.
   1. This should have responsibility for opening / closing the database.
   2. This should call `validate`.
   3. It should catch any exceptions that might be thrown from `validate` to ensure the db always ends up being closed.
3. Update unit tests for the validate command
   1. Change all to call `validate` rather than `validateCommand`.
   2. Move into the unit tests for database.ts.

lightwalker-eth (changes requested):
@djstrong Hey thanks for updates :+1: A few more suggestions :+1:

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -19,86 +10,12 @@ export async function validateCommand(options: ValidateCommandOptions): Promise<
...
-      labelHashToBytes(keyHex);
-      validHashes++;
-    } catch (e) {
-      log.error(`Invalid labelhash key format: ${keyHex}`);
-      invalidHashes++;
-      continue;
+  try {
+    const isValid = await validate(db, log);
+    if (!isValid) {
+      throw new Error("Validation failed");
```
lightwalker-eth, line 16: Maybe no need to throw any error here since this is in the context of running the validate command? Or is the goal to have the process exit with an error code? If that's the goal, it makes sense to me :+1: Suggest to add a little comment here then explaining why we're throwing.

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -18,112 +18,29 @@ describe("Validate Command", () => {
...
-
-    // Add count
-    await db.put(LABELHASH_COUNT_KEY, testData.length.toString());
-
-    await db.close();
-
-    await expect(validateCommand({ dataDir: tempDir })).resolves.not.toThrow();
-  });
-
-  it("should detect invalid labelhash format", async () => {
```
lightwalker-eth: @djstrong I think we should still keep all the validation tests. Those were nice! They should just move into `database.test.ts`.

apps/ensrainbow/src/commands/validate-command.test.ts
```diff
@@ -18,112 +18,29 @@ describe("Validate Command", () => {
     await rm(tempDir, { recursive: true, force: true });
   });
 
-  it("should detect an empty database", async () => {
-    const db = await createDatabase(tempDir);
-    await db.close();
-
+  it("should throw for invalid database", async () => {
     await expect(validateCommand({ dataDir: tempDir })).rejects.toThrow();
```
lightwalker-eth, line 22: Not sure if I understand this unit test? I understand that it should fail for an empty database. Suggest we bring back all the more specific unit tests for the validate operation, they should just live in `database.test.ts` instead of here.

lightwalker-eth (changes requested):
@djstrong Hey nice updates! Just 1 key issue remaining I think. Reviewed and shared feedback :+1:

apps/ensrainbow/src/commands/validate-command.test.ts
djstrong: Yes, they are moved to `database.test.ts`.

apps/ensrainbow/src/commands/validate-command.test.ts
djstrong: Changed description - it is for invalid `dataDir`

apps/ensrainbow/src/commands/validate-command.ts
djstrong: Yes, this if for the CLI to exit with error code.

apps/ensrainbow/src/lib/database.test.ts
```diff
@@ -0,0 +1,173 @@
...
+        await db.put(LABELHASH_COUNT_KEY, "2");
+
+        const isValid = await validate(db, log);
+        expect(isValid).toBe(false);
+      } finally {
+        await db.close();
+      }
+    });
+
+    it("should throw when ingestion is in progress", async () => {
```
lightwalker-eth, line 134:
```suggestion
    it("should detect when ingestion is in progress", async () => {
```

lightwalker-eth (approved):
@djstrong Awesome 🚀 Thank you! Happy to merge this PR! 1 tiny suggestion. Please take the lead on merging this PR :+1:

Issue #111, opened 2025-01-30
Transition deployments to GitHub Actions


lightwalker-eth:
# Background

* Current deployments are made through Railway using a "pull" strategy and limit our flexibility.

# Goals

1. Provide more control over the details of ENSNode deployments.
2. Optimize the deployment narrative for others who are interested to run their own instance of ENSNode.
3. Increase control over the schema name that an ENSNode instance uses.
   1. Some code changes don't need to trigger a full historical reindexing.
5. Optimize db cleanup of schemas that ENSNode built in the past, but that are no longer needed.

# Ideas

1. We're experiencing a lot of issues with indexing speed on Railway and are exploring switching away. If we do continue using Railway for some reason there are related docs here: https://blog.railway.com/p/github-actions

lightwalker-eth:
@BanaSeba I note how we're currently using a container-based deployment strategy for ENSRainbow -> Railway. However we're using a GitHub-based "pull" deployment strategy for ENSNode -> Railway.

Could you please review and critique how all is currently configured here and suggest some proposed next steps for us here?

lightwalker-eth:
@BanaSeba When you go to action this, suggest also taking into consideration the ideas being explored in some semi-related issues:
* https://github.com/namehash/ensnode/issues/22
* https://github.com/namehash/ensnode/issues/135

shrugs:
this is completed, i believe — closing for now

PR #241, opened 2025-02-13
Feat/railway ensnode deployments

BanaSeba:
Related to: https://github.com/namehash/ensnode/issues/111

- Added workflow for deploying ENSNode instances in Railway 
- Created a new Railway environment for Docker image-based ENSNodes
- Created ENSNode Docker images and pushed them to gcr
- Populated Railway related variables in Github


.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,66 @@
...
+    container: ghcr.io/railwayapp/cli:latest
+    needs: image-build-and-push
+    if:  ${{ needs.image-build-and-push.outputs.svc_id }}
+    env:
+        RAILWAY_TOKEN: ${{ secrets.RAILWAY_TOKEN }}
+        SVC_ID: ${{ needs.image-build-and-push.outputs.svc_id }}
+    steps:
+      - name: Deploy ENSNode on Railway
+        run: |
+         railway redeploy --service ${{ env.SVC_ID }} --yes
```
tk-o, line 65: I wonder if we can define a set of env vars for the deployment over here?

Found [a thread](https://help.railway.com/questions/setting-environment-variables-from-cli-3bf27092) on railway help forum. I can see there's no CLI command. However, some user posted a GQL mutation achieving exactly that. 

[The reason](https://t.me/ponder_sh/10282) I'm asking about setting env vars from gh action is:
> Sure! I'd like to calculate one hash from env vars (config hash) and the other one from codebase (build hash) to use them together for schema name.   I need that to handle feature flags where some of the event handlers are loaded conditionally. 
>
> Also, sometimes I'm doing just renames or refactorings that don't change the logic of event handlers, and then I'd like to tell ponder runtime that even though there's new build id, it should just carry on with the selected schema name.

BanaSeba: I don't believe that this would be helpful for us here. Here we will need to always force new deployment (with new schema) as a new ENSNode version is available (and actually forced by pushing to a specific branch).
On the other hand about env variables. I believe that we should stick with DEPLOYMENT_ID for now:
- It will be updated during images push and force new deployment
- It should be updated whenever someone changes ENV var (he should also force new redeployments for affected services)

BanaSeba: I assume you are also looking into the future where there will be two services (indexer and API) for ENSNode. Right now I'm not sure about this but I would expect that we will need to set ENV variables in this CI as well

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
+name: Build and deploy ENSNode
```
lightwalker-eth, line 1: As I understand, this workflow is dedicated to Railway.

A few questions / suggestions:
1. Should we rename this workflow file so that "railway" is in the name of the .yml file?
2. Should we refine this `name` field at the top of the file to include "Railway"?
3. When we use "ENSNode" here, can you please follow up with me how this should be interpreted? Is it "ENSNode" the overall solution (both the ENSRainbow service and the ENSNode service) or just the ENSNode service on its own? We're working on giving the "ENSNode service" a distinct name of its own. Until that's finished this can be a source of confusion.

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+
+on:
+  push:
+    branches:
+      - feat/railway-ensnode-deployments
+      - main
```
lightwalker-eth, line 7: Can you help me understand what it means for the `main` branch to be listed here? What is the proposed strategy for how we're actually supposed to build release processes? It would help a lot if you could write up a description of the operational processes you're proposing to enable here. Without that big picture understanding its difficult for me to understand.

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+      docker_image_id: ${{ steps.determine_tag.outputs.docker_image_id }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Extract Branch Name
+        run: echo "BRANCH=${GITHUB_REF#refs/heads/}" >> $GITHUB_ENV
+    
+      # We need to recognize which branch had push incomin
```
lightwalker-eth, line 26:
```suggestion
      # We need to recognize which branch had an incoming push
```

Is that correct? Just trying to fix the typo

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+      # Based on that create Docker image with corresponding label - use TAG variable for this
+      # For alhpa/subgraph TAG we need also to specify corresponding Railway SVC_ID which is used for service redeployment
```
lightwalker-eth, line 28:
```suggestion
      # For alpha/subgraph TAG we need also to specify corresponding Railway SVC_ID which is used for service redeployment
```

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+      - name: Determine tag
+        id: determine_tag
+        run: |
+            case "$BRANCH" in
+              "main")
+                echo "TAG=latest" >> $GITHUB_ENV
```
lightwalker-eth, line 34: Can you share more info for what the goals are here for the "main" branch? Please see related questions above.

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+                ;;
+              "alpha")
+                echo "TAG=alpha" >> $GITHUB_ENV
+                echo "SVC_ID=${{ secrets.ALPHA_SVC_ID }}" >> "$GITHUB_OUTPUT"
+                ;;
+              "subgraph")
+                echo "TAG=subgraph" >> $GITHUB_ENV
+                echo "SVC_ID=${{ secrets.SUBGRAPH_SVC_ID }}" >> "$GITHUB_OUTPUT"
+                ;;
+              "feat/railway-ensnode-deployments")
```
lightwalker-eth, line 44: Can you help me understand the goal and scope for referencing this branch? I understand this is a reference to the branch I'm currently reviewing a PR for. Is this branch just temporary in your mind (such that any reference to it will be removed in a future version of this script) or is it part of a longer-term version?

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+                echo "TAG=subgraph" >> $GITHUB_ENV
+                echo "SVC_ID=${{ secrets.SUBGRAPH_SVC_ID }}" >> "$GITHUB_OUTPUT"
+                ;;
+              *)
+                echo "Branch not recognized, skipping workflow"
+                exit 1
+            esac
+            echo "TAG=$TAG" >> "$GITHUB_OUTPUT"
+
+      - name: Build & Push Docker image for the ENSNode
```
lightwalker-eth, line 54:
```suggestion
      - name: Build & Push Docker image for ENSNode to GitHub
```

Is that correct? That this is pushing the image to GitHub specifically?

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+        RAILWAY_TOKEN: ${{ secrets.RAILWAY_TOKEN }}
+        RAILWAY_PROJECT_ID: ${{ secrets.RAILWAY_PROJECT_ID }}
+        RAILWAY_ENVIRONMENT_ID: ${{ secrets.RAILWAY_ENVIRONMENT_ID}}
+        TAG: ${{ needs.image-build-and-push.outputs.tag }}
+        SVC_ID: ${{ needs.image-build-and-push.outputs.svc_id }}
+        DOCKER_IMAGE_ID: ${{ needs.image-build-and-push.outputs.docker_image_id }}
+    steps:
+      - name: Deploy if necessary
+        run: |
+          apt install jq
```
lightwalker-eth, line 78: Please check the CI on GitHub for this PR. It looks like some kind of error is being reported here: https://github.com/namehash/ensnode/actions/runs/13335081307/job/37248438895?pr=241

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+          API_RESPONSE=`curl --request POST \
```
lightwalker-eth, line 79: Can we please add some comment describing the goal of this query from a bigger picture perspective?

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+          --header 'Project-Access-Token: '${{ env.RAILWAY_TOKEN }} \
+          --header 'Content-Type: application/json' \
+          --data '{
+            "query": "query variables($projectId: String!, $environmentId: String!) { variables(projectId: $projectId, environmentId: $environmentId) }",
+              "variables": {
+                "projectId": "'${{ env.RAILWAY_PROJECT_ID }}'" ,
+                "environmentId": "'${{ env.RAILWAY_ENVIRONMENT_ID }}'"
+              }
+          }'`
+          VARIABLE_NAME=${${{ env.TAG }}^^}_DATABASE_SCHEMA
```
lightwalker-eth, line 91: A few questions.

I'm confused and trying to understand the difference between different goals I understand we need to achieve:

1. Tell Railway to run the new Docker image we published to GitHub
2. When Railway runs this new Docker image, we need to ensure the new `RAILWAY_DEPLOYMENT_ID` environment variable is distinct for this new Docker image.

CleanShot 2025-02-16 at 17 45 05

Can you please help to clarify how we're managing these goals?

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+          CURRENT_SCHEMA=`${API_RESPONSE} | jq '.data.variables.'${VARIABLE_NAME}`
```
lightwalker-eth, line 92: Can you please share more info about what this code is attempting to do? Please appreciate how I don't understand devops but still need to do my best to ensure we're on the right path. It's very difficult if I get a bunch of dense code to review without some additional info to go with it about what we're doing and why.

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+          echo "Calculated schema variable: ${VARIABLE_NAME}"
+          #TODO FOR NOW JUST REPLACE DOCKER IMAGE ID. EXTEND LOGIC IN FUTURE
+          if [[ "$CURRENT_SCHEMA" != "${{ env.DOCKER_IMAGE_ID }}" ]]; then
+          curl --request POST \
```
lightwalker-eth, line 96: What's the big goal of this query?

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+              "variables": {
+                "projectId": "'${{ env.RAILWAY_PROJECT_ID }}'" ,
+                "environmentId": "'${{ env.RAILWAY_ENVIRONMENT_ID }}'"
+              }
+          }'`
+          VARIABLE_NAME=${${{ env.TAG }}^^}_DATABASE_SCHEMA
+          CURRENT_SCHEMA=`${API_RESPONSE} | jq '.data.variables.'${VARIABLE_NAME}`
+          echo "Calculated schema variable: ${VARIABLE_NAME}"
+          #TODO FOR NOW JUST REPLACE DOCKER IMAGE ID. EXTEND LOGIC IN FUTURE
+          if [[ "$CURRENT_SCHEMA" != "${{ env.DOCKER_IMAGE_ID }}" ]]; then
```
lightwalker-eth, line 95: What is this `if` statement doing in the bigger picture? Its not clear from reviewing this code.

.github/workflows/deploy_ensnode_subgraph.yml
```diff
@@ -0,0 +1,107 @@
...
+          curl --request POST \
+              --silent \
+              --url https://backboard.railway.app/graphql/v2 \
+              --header 'Project-Access-Token: '${{ env.RAILWAY_TOKEN }} \
+              --header 'Content-Type: application/json' \
+              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${{ env.RAILWAY_ENVIRONMENT_ID }}'\", name: \"'${VARIABLE_NAME}'\", value: \"${{ env.DOCKER_IMAGE_ID }}\" }) }"}'
+          else
+              echo "Skipping environment variable update"
```
lightwalker-eth, line 103: What environment variable update are we skipping? And why is it right to skip it?

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: 1. Added railway to file name
2. Added railway to workflow name
3. We are deploying ENSNode service only. Renamed both workflow and file to clearly state that

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Added comment about overall flow for each branch

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Adjusted

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: I've added comments at top sections about each branch flow

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: This will be removed in the future. I just need to test this workflow E2E without pushing to any 'real' branches

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: They are calling it 'Github Container registry'. Adjusted

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Yep, still need to adjust it a little bit

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Added more comments about this query

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: 1. We are setting the Railway shared environment variable by using Railway API.
2. I'm looking further here. We can't rely only on DEPLOYMENT_ID. If we want to introduce a new 'API' version of ENSNode (which is only serving requests and not indexing) we need a way to point both the indexer and API to the same schema. Shared environment variables should do the trick here.

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Added comment. We are filtering out all shared Railway variables except one defining schema

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Added comment. We either update the shared env variable and proceed with any necessary deployment steps or skip deployment part (no redeployment is needed, schema stays the same)

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Added comment. We are updating specific Railway shared env variable which is defining schema

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Schema env variable. If the indexer schema should not change (docker sha image is the same etc.)

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,109 @@
...
+        run: |
+          # Sending Railway API request to update specifc shared variable
+          if [[ ${{env.RAILWAY_SCHEMA}} != ${{ github.sha }} ]]; then
+          curl --request POST \
+              --silent \
+              --fail \
+              --url https://backboard.railway.app/graphql/v2 \
+              --header 'Project-Access-Token: '${{ env.RAILWAY_TOKEN }} \
+              --header 'Content-Type: application/json' \
+              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${{ env.RAILWAY_ENVIRONMENT_ID }}'\", name: \"'${VARIABLE_NAME}'\", value: \"${{ github.sha }}\" }) }"}'
```
tk-o, line 99: Perhaps we should define database schema value on the job `env` so it's easier to notice it and also easier to apply any updates to it (i.e. adding env var hash):
```yaml
jobs:
  image-build-and-push:
    runs-on: ubuntu-latest
    env:
      RAILWAY_TOKEN: ${{ secrets.RAILWAY_TOKEN }}
      # Token for redeploy. Seems lika a bug in Railway ..
      RAILWAY_TEAM_TOKEN: ${{ secrets.RAILWAY_TEAM_TOKEN }}
      RAILWAY_PROJECT_ID: ${{ secrets.RAILWAY_PROJECT_ID }}
      RAILWAY_ENVIRONMENT_ID: ${{ secrets.RAILWAY_ENVIRONMENT_ID}}
      DATABASE_SCHEMA: ${{ github.sha }} # this one could also have some env vars hash attached
```

```suggestion
              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${{ env.RAILWAY_ENVIRONMENT_ID }}'\", name: \"'${VARIABLE_NAME}'\", value: \"${{ env.DATABASE_SCHEMA }}\" }) }"}'
```

BanaSeba: Nice suggestion. Applied

.github/workflows/deploy_ensnode_subgraph.yml
BanaSeba: Applied this logic here

tk-o (approved):
LGTM

shrugs (approved):
looks good to me

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+                "environmentId": "'${{ env.RAILWAY_ENVIRONMENT_ID }}'"
+              }
+          }'`
+          # Filter all env variables from response except one that is defining schema for specific deployment.
+          RAILWAY_SCHEMA=`echo ${API_RESPONSE} | jq '.data.variables.'${{ env.VARIABLE_NAME }}`
+          echo "RAILWAY_SCHEMA=$RAILWAY_SCHEMA" >> "$GITHUB_ENV"
+          
+
+      - name: Update Railway schema variable
+        # Check if newly calculated schema matches old one, if not replace env variable and force deployment. Otherwise skip deploymen.
```
lightwalker-eth, line 90:
```suggestion
        # Check if newly calculated schema matches old one, if not replace env variable and force deployment. Otherwise skip deployment.
```

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+        run: |
+          # Sending Railway API request to update specifc shared variable
```
lightwalker-eth, line 92: Can we clarify this statement to explicitly identify the specific shared variable you're trying to update?

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+  image-build-and-push:
+    runs-on: ubuntu-latest
+    env:
+      RAILWAY_TOKEN: ${{ secrets.RAILWAY_TOKEN }}
+      # Token for redeploy. Seems lika a bug in Railway ..
+      RAILWAY_TEAM_TOKEN: ${{ secrets.RAILWAY_TEAM_TOKEN }}
+      RAILWAY_PROJECT_ID: ${{ secrets.RAILWAY_PROJECT_ID }}
+      RAILWAY_ENVIRONMENT_ID: ${{ secrets.RAILWAY_ENVIRONMENT_ID}}
+      # For now just use github.sha. Should be extended in the future
+      DATABASE_SCHEMA: ${{ github.sha }}
```
lightwalker-eth, line 23: Can you please share background info in Slack about the big idea here? I feel like there's some important assumptions / strategies behind the logic here, but it's not currently clear to me what they are.

lightwalker-eth (changes requested):
@BanaSeba Hey this is looking really clever. I see I forgot to submit the PR comments I started earlier. Submitting for you now. Appreciate your advice! Thank you!

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
+name: Build and deploy to Railway ENSNode service
+
+on:
+  push:
+    branches:
+      # Build and push'latest' Docker image. Skip deployments
+      - main
+      # Build and push 'alpha' Docker image. Update 'alpha' Railway shared environment variables.
```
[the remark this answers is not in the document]
BanaSeba: Here's the official doc about them: https://docs.railway.com/guides/variables#shared-variables
In short, they are being shared across environments ( alpha env does have different envs than beta).
Those variables can be shared between every deployment inside the environment.

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+      - alpha
+      # Build and push 'subgraph' Docker image. Update 'subgraph' Railway shared environment variables.
```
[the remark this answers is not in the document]
BanaSeba: Added comment with doc reference

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Extract Branch Name
+        run: echo "BRANCH=${GITHUB_REF#refs/heads/}" >> $GITHUB_ENV
+    
+      # We need to recognize which branch had an incoming push
+      # Based on that create Docker image with corresponding label - use TAG variable for this
+      # For alpha/subgraph TAG we need also to specify corresponding Railway SVC_ID which is used for service redeployment
```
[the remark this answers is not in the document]
BanaSeba: Adjusted

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
+name: Build and deploy to Railway ENSNode service
+
+on:
+  push:
+    branches:
+      # Build and push'latest' Docker image. Skip deployments
```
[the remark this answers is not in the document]
BanaSeba: Adjusted

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${{ env.RAILWAY_ENVIRONMENT_ID }}'\", name: \"'${VARIABLE_NAME}'\", value: \"${{ env.DATABASE_SCHEMA }}\" }) }"}'
+          # Sending Railway API request to redeploy service
+          curl --request POST \
+              --silent \
+              --url https://backboard.railway.app/graphql/v2 \
+              --header 'Authorization: Bearer '${{ env.RAILWAY_TEAM_TOKEN }}\
+              --header 'Content-Type: application/json' \
+              --data '{"query":"mutation serviceInstanceDeploy($serviceId: String!, $environmentId: String!) { serviceInstanceDeploy(serviceId: $serviceId, environmentId: $environmentId) }","variables":{"environmentId":"'${{ env.RAILWAY_ENVIRONMENT_ID }}'","serviceId":"'${{ env.SVC_ID }}'"}}'
+          else
+          echo "Skipping deployment since schema did not changed"
```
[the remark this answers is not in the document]
BanaSeba: Will followup on Slack about this

.github/workflows/deploy_railway_ensnode_service.yml
```diff
@@ -0,0 +1,110 @@
...
+          # Filter all env variables from response except one that is defining schema for specific deployment.
+          RAILWAY_SCHEMA=`echo ${API_RESPONSE} | jq '.data.variables.'${{ env.VARIABLE_NAME }}`
+          echo "RAILWAY_SCHEMA=$RAILWAY_SCHEMA" >> "$GITHUB_ENV"
+          
+
+      - name: Update Railway schema variable
+        # Check if newly calculated schema matches old one, if not replace env variable and force deployment. Otherwise skip deploymen.
+        run: |
+          # Sending Railway API request to update specifc shared variable
+          if [[ ${{env.RAILWAY_SCHEMA}} != ${{ env.DATABASE_SCHEMA }} ]]; then
```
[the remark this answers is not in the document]
BanaSeba: Will followup on Slack about this

.github/workflows/deploy_railway_ensnode_service.yml
BanaSeba: Will followup on Slack about this

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+          # Filter all env variables from response except one that is defining schema for specific deployment.
+          RAILWAY_SCHEMA=`echo ${API_RESPONSE} | jq '.data.variables.'${{ env.VARIABLE_NAME }}`
+          echo "RAILWAY_SCHEMA=$RAILWAY_SCHEMA" >> "$GITHUB_ENV"
+          
+
+      - name: Update Railway schema variable
+        # Check if newly calculated schema matches old one, if not replace env variable and force deployment. Otherwise skip deployment.
+        run: |
+          # Sending Railway API request to update specifc shared variable
+          if [[ ${{env.RAILWAY_SCHEMA}} != ${{ env.DATABASE_SCHEMA }} ]]; then
```
lightwalker-eth, line 94: Hmm. I'm really confused here. This seems to suggest that the `RAILWAY_SCHEMA` variable is for the same idea as the `DATABASE_SCHEMA` environment variable?

Is the `RAILWAY_SCHEMA` supposed to represent the idea of "the new database schema name to be used for this for this new deployment"? If so, suggest renaming `RAILWAY_SCHEMA` to `NEW_DATABASE_SCHEMA`.

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+          }'`
+          # Filter all env variables from response except one that is defining schema for specific deployment.
+          RAILWAY_SCHEMA=`echo ${API_RESPONSE} | jq '.data.variables.'${{ env.VARIABLE_NAME }}`
+          echo "RAILWAY_SCHEMA=$RAILWAY_SCHEMA" >> "$GITHUB_ENV"
+          
+
+      - name: Update Railway schema variable
+        # Check if newly calculated schema matches old one, if not replace env variable and force deployment. Otherwise skip deployment.
+        run: |
+          # Sending Railway API request to update specifc shared variable
```
lightwalker-eth, line 93: "... to update specific shared variable"

Exactly which "specific shared variable"? Please reference it explicitly here. Please don't require the reader to guess or figure it out manually.

Are we updating the `DATABASE_SCHEMA` environment variable for a specific ENSIndexer deployment?

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${{ env.RAILWAY_ENVIRONMENT_ID }}'\", name: \"'${VARIABLE_NAME}'\", value: \"${{ env.DATABASE_SCHEMA }}\" }) }"}'
+          # Sending Railway API request to redeploy service
+          curl --request POST \
+              --silent \
+              --url https://backboard.railway.app/graphql/v2 \
+              --header 'Authorization: Bearer '${{ env.RAILWAY_TEAM_TOKEN }}\
+              --header 'Content-Type: application/json' \
+              --data '{"query":"mutation serviceInstanceDeploy($serviceId: String!, $environmentId: String!) { serviceInstanceDeploy(serviceId: $serviceId, environmentId: $environmentId) }","variables":{"environmentId":"'${{ env.RAILWAY_ENVIRONMENT_ID }}'","serviceId":"'${{ env.SVC_ID }}'"}}'
+          else
+          echo "Skipping deployment since schema did not change"
```
lightwalker-eth, line 110: Since which schema did not change? Appreciate if we can be more clear

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+          RAILWAY_SCHEMA=`echo ${API_RESPONSE} | jq '.data.variables.'${{ env.VARIABLE_NAME }}`
+          echo "RAILWAY_SCHEMA=$RAILWAY_SCHEMA" >> "$GITHUB_ENV"
+          
+
+      - name: Update Railway schema variable
+        # Check if newly calculated schema matches old one, if not replace env variable and force deployment. Otherwise skip deployment.
+        run: |
+          # Sending Railway API request to update specifc shared variable
+          if [[ ${{env.RAILWAY_SCHEMA}} != ${{ env.DATABASE_SCHEMA }} ]]; then
+          curl --request POST \
```
lightwalker-eth, line 95: Can we add a comment above this `curl` command summarizing what its doing? It looks like it's updating some environment variable or something but it would help if this was written in a highly specific and clear way.

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+              --silent \
+              --fail \
+              --url https://backboard.railway.app/graphql/v2 \
+              --header 'Project-Access-Token: '${{ env.RAILWAY_TOKEN }} \
+              --header 'Content-Type: application/json' \
+              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${{ env.RAILWAY_ENVIRONMENT_ID }}'\", name: \"'${VARIABLE_NAME}'\", value: \"${{ env.DATABASE_SCHEMA }}\" }) }"}'
+          # Sending Railway API request to redeploy service
```
lightwalker-eth, line 102:
```suggestion
          # Sending Railway API request to redeploy service (using the newly updated configuration)
```

Is that fair? In other words, the first `curl` seems to change something, but not necessarily to trigger a redeployment? And therefore we need this second `curl` to trigger the redeployment using the new config?

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+  image-build-and-push:
+    runs-on: ubuntu-latest
+    env:
+      RAILWAY_TOKEN: ${{ secrets.RAILWAY_TOKEN }}
+      # Token for redeploy. Redeploy needs a separate team token. 
+      RAILWAY_TEAM_TOKEN: ${{ secrets.RAILWAY_TEAM_TOKEN }}
+      RAILWAY_PROJECT_ID: ${{ secrets.RAILWAY_PROJECT_ID }}
+      RAILWAY_ENVIRONMENT_ID: ${{ secrets.RAILWAY_ENVIRONMENT_ID}}
+      # For now just use github.sha. Should be extended in the future
+      DATABASE_SCHEMA: ${{ github.sha }}
```
lightwalker-eth, line 24: This solution will meet our needs in the short term, thank you :+1:

lightwalker-eth (changes requested):
@BanaSeba Hey really appreciate the updates here. Reviewed and shared feedback :+1:

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+              "subgraph")
+                echo "TAG=subgraph" >> $GITHUB_ENV
+                echo "SVC_ID="${{ secrets.SUBGRAPH_SVC_ID }} >> "$GITHUB_ENV"
+                ;;
+              *)
+                echo "Branch not recognized, skipping workflow"
+                exit 1
+            esac
+
+      - name: Build & Push Docker image for ENSNode to Github Container registry
```
lightwalker-eth, line 54:
```suggestion
      - name: Build & Push Docker image for ENSIndexer to Github Container registry
```

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+        env:
+          IMAGE_NAME: ghcr.io/${{ github.repository }}/ensnode
```
lightwalker-eth, line 56:
```suggestion
          IMAGE_NAME: ghcr.io/${{ github.repository }}/ensindexer
```

Is that right? Please fix as needed.

.github/workflows/ensnode_railway_deploy.yml
```diff
@@ -0,0 +1,111 @@
...
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+          docker build -f apps/ensnode/Dockerfile -t $IMAGE_NAME:$TAG .
```
lightwalker-eth, line 59:
```suggestion
          docker build -f apps/ensindexer/Dockerfile -t $IMAGE_NAME:$TAG .
```

lightwalker-eth (changes requested):
@BanaSeba Hey made a second follow up review now where I've tried to go deeper towards my understanding of our target.

BanaSeba:
Same purpose achieved in this PR: https://github.com/namehash/ensnode/pull/416

PR #247, opened 2025-02-13
WIP: 21M Diff Fixes

### LINKED ISSUE #246: 21M Diff Fixes
- [x] `Domain.name` is null when it should be something like `[encoded].eth`. only happens when label is not healed
  - cause by `undefined` not being no-op in entity update, likely ponder bug
- [x] wrappedOwner id is wrong for some entities
  - was not materializing in `handleTransfer`
- [x] domain `0xde9c560b24fb00edec8f81d50be948f0e29bd6d8346d92c3e28c44703a06bd97` missing its resolver (should be `0x1da022710df5002339274aadee8d58218e9d6ab5`)
  - resolver record exists but graphql relation isn't working
- [x] domains are missing a subdomainCount
  - domain `0x297836a76312224372ac04e26dd23d1294bb8256598ec113ecc52735f826beff` (`freedomain.eth`) is missing one domain
  - domain `0x297fb7d83d3657ab9c208a2bfc2a073cfbcca3a2a5cc11cd4eba416fe679cb79` (`kolodochka.eth`) is missing one domain
  - perhaps also caused by https://github.com/ponder-sh/ponder/issues/1532 given the strict null check...
- [x] some domain's resolver addresses were their previous address when they should have been nulled
  - caused by https://github.com/ponder-sh/ponder/issues/1532
- [x] domain `0xc84dfbec915db66663670edff66a88e2a74a6e4df271a4d81c165a0c9072a13b` has an incorrect `owner`
  - the correct event exists, but the `ownerId` record wasn't updated?
  - solved in 0.9.16
- [x] ENSNode heals `daohq` when subgraph does not (domain `0x0093b7095a35094ecbd246f5d5638cb094c3061a5f29679f5969ad0abcfae27f`). doesn't seem to be healed by ensrainbow, so must be in namewrapper somewhere?
  - the `testing` subdomain is created before the `daohq` name is wrapped, so it _should_ use its parents encoded name value, and ENSNode is getting that part right.
  - it's the `coffee` subdomain that's created after

```gql
query Blah {
	domain(id:"0x0093b7095a35094ecbd246f5d5638cb094c3061a5f29679f5969ad0abcfae27f", block:{number:21000000}) {
    name
    parent {
      name
      parent {
        name
      }
    }
  }
}
```

```jsonc
// subgraph
{
  "data": {
    "domain": {
      "name": "coffee.testing.[f09e939dd7d59be2dee24a0fa03110c96910351c12f63abb35506d3778b1809b].dappcraft.eth",
      "parent": {
        "name": "testing.[f09e939dd7d59be2dee24a0fa03110c96910351c12f63abb35506d3778b1809b].dappcraft.eth",
        "parent": {
          "name": "daohq.dappcraft.eth"
        }
      }
    }
  }
}

// ensnode
{
  "data": {
    "domain": {
      "name": "coffee.testing.daohq.dappcraft.eth",
      "parent": {
        "name": "testing.[f09e939dd7d59be2dee24a0fa03110c96910351c12f63abb35506d3778b1809b].dappcraft.eth",
        "parent": {
          "name": "daohq.dappcraft.eth"
        }
      }
    }
  }
}
```

shrugs:
closes #246

in debugging `Domain.name` being null I madesome minor changes to the comments & logic in those functions but at the end of the day our logic was correct and it was ponder setting `name` to null if it was `undefined` that caused the bug in the diff

### DIFF
--- a/apps/ensnode/src/handlers/NameWrapper.ts
+++ b/apps/ensnode/src/handlers/NameWrapper.ts
@@ -54,12 +54,9 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
     await upsertAccount(context, to);
     const node = tokenIdToNode(tokenId);
 
-    // TODO: remove this if it never fires: subgraph upserts domain but shouldn't be necessary
-    const domain = await context.db.find(schema.domain, { id: node });
-    if (!domain) {
-      console.log("NameWrapper:handleTransfer called before domain existed");
-      console.table({ ...event.args, node });
-    }
+    // NOTE: subgraph technically upserts domain with `createOrLoadDomain()` here, but domain
+    // is guaranteed to exist, so we need not re-implement this check
+    // https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/nameWrapper.ts#L197C18-L197C36
 
     // upsert the WrappedDomain, only changing owner iff exists
     await context.db
@@ -76,6 +73,9 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
       .onConflictDoUpdate({
         ownerId: to,
       });
+
+    // materialize `Domain.wrappedOwner`
+    await context.db.update(schema.domain, { id: node }).set({ wrappedOwnerId: to });
 
     // log DomainEvent
     await context.db
--- a/apps/ensnode/src/handlers/NameWrapper.ts
+++ b/apps/ensnode/src/handlers/NameWrapper.ts
@@ -55,10 +55,15 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
     const node = tokenIdToNode(tokenId);
 
     // NOTE: subgraph technically upserts domain with `createOrLoadDomain()` here, but domain
-    // is guaranteed to exist, so we need not re-implement this check
+    // is guaranteed to exist. we encode this stricter logic here to illustrate that fact.
     // https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/nameWrapper.ts#L197C18-L197C36
+    const domain = await context.db.find(schema.domain, { id: node });
+    if (!domain) {
+      console.table({ ...event.args, node });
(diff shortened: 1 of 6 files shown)


apps/ensnode/src/handlers/NameWrapper.ts
```diff
@@ -54,14 +54,16 @@ export const makeNameWrapperHandlers = (ownedName: OwnedName) => {
     await upsertAccount(context, to);
     const node = tokenIdToNode(tokenId);
 
-    // TODO: remove this if it never fires: subgraph upserts domain but shouldn't be necessary
+    // NOTE: subgraph technically upserts domain with `createOrLoadDomain()` here, but domain
```
lightwalker-eth, line 57: Nice :+1:

lightwalker-eth (approved):
@shrugs Looks good 🚀

apps/ensnode/src/handlers/Registrar.ts
```diff
@@ -92,8 +92,11 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
       await context.db.update(schema.domain, { id: node }).set({
         registrantId: owner,
         expiryDate: expires + GRACE_PERIOD_SECONDS,
-        labelName: validLabel,
-        name,
+        // TODO: pass these values as normal once ponder treats undefined as no-op
+        // labelName: validLabel,
+        // name,
+        ...(validLabel && { labelName: validLabel }),
+        ...(name && { name }),
```
tk-o: Nice one with catching the edge case where undefinds are not relly underfineds 😅

apps/ensnode/src/handlers/Registry.ts
```diff
@@ -143,8 +143,7 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
             name,
             // NOTE: only update Domain.labelName iff label is healed and valid
             // via: https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L113
-            // TODO: pass labelName as normal once ponder treats undefined as no-op
-            ...(validLabel && { labelName: validLabel }),
```
tk-o: Oh, so it `undefind`s do not have to be treated this way after all?

PR #242, opened 2025-02-13
feat(ensrainbow-sdk): use in memory cache

tk-o:
This PR introduces a time-based in-memory cache that can be enabled on the ENSNode while using ENSRainbowClient.

### TODO

- [ ] ensure 400 & 404 responses for the `heal` endpoint get cached, as the endpoint can be considered pure (same params, same answer).

<details><summary>ENSRainbow logs with cache enabled</summary>

```
2025-02-13 18:33:57 Successfully healed labelhash 0x6c9c64b7760e6a7df5e2e8c518e1812366da9db100f55236c3c0390cac6cd36b to label "arkatechbeatz"
2025-02-13 18:33:57 Successfully healed labelhash 0x92e827cd64c396b9818077b174d91d6f01c6d556c71d8dda48d1fd074da6ea0b to label "kingzero"
2025-02-13 18:33:57 Unhealable labelhash request: 0x4bcd09e20edba5b5a8cfa576da2fecc85b17fe49653ae4620b23a5d23aaf1c8e
2025-02-13 18:33:57 Unhealable labelhash request: 0x4a52540196241744d3d006f7a9913ee7828766d6a87adbf7d186c1a6da4b3e06
2025-02-13 18:33:57 Unhealable labelhash request: 0xde61d0c5cdfb649ff67dbdaa9de1256276b36aede9d15798799a813977ea3d05
2025-02-13 18:33:57 Unhealable labelhash request: 0x99dc823606a91969107538b6308a1d2e8c367bebd345a540c8d69d2962cc9be5
2025-02-13 18:33:57 Unhealable labelhash request: 0x66bbb6d162aad92ebb4067887d364cd56ad5eefb42af4d9a1203838753a17fd6
2025-02-13 18:33:57 Unhealable labelhash request: 0x5a1b3933c26246774435d43e8c5167a67b908fa92d29458de908641f6f6caf1b
2025-02-13 18:33:57 Unhealable labelhash request: 0xdbbc6e1997e42e60cbad999af2b803416ff9c77a8c524212bb0340270f25aa0c
2025-02-13 18:33:57 Unhealable labelhash request: 0x96aaf0a948b70ec770e6b371e0461bf0e05fee7e67304768be83c237c52925e9
2025-02-13 18:33:57 Successfully healed labelhash 0x01f7b9b3d004b9b1551d68e986a51441db02d9ebf517dbd0d8caa89302c7dc76 to label "ameensol"
[...]

PR #248, opened 2025-02-13
Refine cache

lightwalker-eth:
Proposal for merging into PR https://github.com/namehash/ensnode/pull/242 (not main)

- Creates a simple LRU caching strategy with O(1) operations
- Changes `EnsRainbowAPIClient` to always have a cache so that it exposes a more simple interface. Caching can still be effectively disabled by setting `cacheCapacity` to 0.
- Prohibits caching of server errors. These should be retried.
- Clarify the difference between the size vs capacity of a cache.
- Refines type definitions for stronger typechecking of ENSRainbow HealResponses.
- Open Question: Is there a way to use `Map` (helpful for the O(1) operations in `LruCache`) without changing the `tsconfig.json`?


packages/ensnode-utils/src/cache.spec.ts
```diff
@@ -1,161 +1,78 @@
...
-    // Both items should be cleaned up
-    expect(cache.get("key1")).toBeUndefined();
-    expect(cache.get("key2")).toBeUndefined();
-  });
-
-  it("should handle multiple sets of the same key", () => {
-    cache.set("key1", "value1");
-    cache.set("key1", "value2");
-    expect(cache.get("key1")).toBe("value2");
+describe("LruCache", () => {
```
tk-o, line 4: Nice to see so much simpler software that the previous iteration 🤝

packages/ensnode-utils/src/cache.ts
```diff
@@ -1,99 +1,101 @@
...
 
   /**
-   * Clean up resources and clear the cache
+   * Clear the cache.
    */
-  dispose(): void;
-}
-
-type Timestamp = number;
+  clear(): void;
```
tk-o, line 24: Nice update, now I see why I didn't like the `dispose` method on my `ICache` interface.

packages/ensrainbow-sdk/src/client.spec.ts
```diff
@@ -15,15 +13,20 @@ describe("EnsRainbowApiClient", () => {
...
     });
   });
 
   it("should apply custom options when provided", () => {
     const customEndpointUrl = new URL("http://lb-api.ensrainbow.com");
-    client = new EnsRainbowApiClient({ endpointUrl: customEndpointUrl });
+    client = new EnsRainbowApiClient({
+      endpointUrl: customEndpointUrl,
+      cacheCapacity: 0,
+    });
```
tk-o, line 25: Simplicity vs configurability

I appreciate abstracting configuration details away for users to just pass options that matter, and nothing more. In this case, users of `EnsRainbowApiClient` always inherit the cache implementation, so caching-wise, they can only provide params to the inherited class.

The alternative option would be to allow composition. For example, we could make the `cache` option available as optional constructor parameter with default value set to an `LruCache` instance. This way we would enable users to ship their own cache implementation (i.e. when using the client class in a browser, where the Web Cache API is available.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -23,6 +34,9 @@ export interface EnsRainbowApiClientOptions {
  */
 export class EnsRainbowApiClient {
   private readonly options: EnsRainbowApiClientOptions;
+  private readonly cache: Cache<Labelhash, CacheableHealResponse>;
+
+  public static readonly DEFAULT_CACHE_CAPACITY = 1000;
```
tk-o, line 39: Nice one

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -85,22 +102,20 @@ export class EnsRainbowApiClient {
...
 
-    if (this.options.cache) {
-      this.options.cache.set(labelhash, responseJson);
+    if (isCacheableHealResponse(healResponse)) {
+      this.cache.set(labelhash, healResponse);
     }
 
-    return responseJson;
+    return healResponse;
   }
```
tk-o, line 119: Not optional caching brings us less-nested code, nice 👍

packages/ensrainbow-sdk/src/types.ts
```diff
@@ -8,52 +8,100 @@ export interface HealthResponse {
...
+ * Server errors at not cachable and should be retried.
+ *
+ * @param response the heal response to check
+ * @returns true if the response is cacheable, false otherwise
+ */
+export const isCacheableHealResponse = (
+  response: HealResponse,
+): response is CacheableHealResponse => {
+  return response.status !== StatusCode.Error || response.errorCode !== ErrorCode.ServerError;
+};
```
tk-o, line 79: Perhaps we could move this one into the `client.ts`? 

`types.ts` could then host just types, and client helpers would be kept private to the client. This would further enable us to [test client caching with network calls](https://humanwhocodes.com/blog/2025/01/introducing-mentoss-fetch-mocker/) and observe if we called the network while case was supposed to be in play.

packages/ensrainbow-sdk/src/types.ts
```diff
@@ -8,52 +8,100 @@ export interface HealthResponse {
...
-export interface CountSuccess extends BaseCountResponse<typeof StatusCode.Success> {
+export interface CountSuccess extends BaseCountResponse<typeof StatusCode.Success, never> {
   status: typeof StatusCode.Success;
   /** The total count of labels that can be healed by the ENSRainbow instance. Always a non-negative integer. */
   count: number;
   timestamp: string;
-  error?: undefined;
-  errorCode?: undefined;
+  error?: never;
+  errorCode?: never;
```
tk-o, line 95: Very nice!

apps/ensnode/src/lib/graphnode-helpers.ts
```diff
@@ -1,12 +1,9 @@
 import { EnsRainbowApiClient } from "@ensnode/ensrainbow-sdk/client";
 import { ErrorCode, StatusCode } from "@ensnode/ensrainbow-sdk/consts";
-import { HealResponse } from "@ensnode/ensrainbow-sdk/types";
-import { MemoryCache } from "@ensnode/utils/cache";
 import type { Labelhash } from "@ensnode/utils/types";
 import { ensRainbowEndpointUrl } from "./ponder-helpers";
 
 const ensRainbowApiClient = new EnsRainbowApiClient({
-  cache: new MemoryCache<Labelhash, HealResponse>(),
```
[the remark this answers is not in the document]
lightwalker-eth: Yeah. Everything has tradeoffs, but put higher value on the benefit of keeping the interface of `EnsRainbowApiClient` nice and simple over the benefit of letting people pass in a custom `Cache` implementation. It's difficult for me to imagine why someone would want to do anything other than use a LRU cache.

packages/ensrainbow-sdk/src/client.spec.ts
lightwalker-eth: That's totally true. I just don't see enough benefit from that added complexity.

I think there's a nice benefit to keeping the interfaces we expose more simple. If someone really wanted to build their own custom cache implementation strong enough they could always achieve that by building a wrapper around `EnsRainbowApiClient`.

In my opinion these are highly unlikely usage scenarios. If we see that people are really wanting this later, we can always add it, but until we validate there's a true demand for it, simple has a lot of advantages.

packages/ensrainbow-sdk/src/types.ts
lightwalker-eth: @tk-o Ah Mentoss is interesting for testing purposes. I haven't seen it before 😄 

I don't feel such a need to test the caching at a HTTP layer though, because it seems we can simply rely on the unit tests for `LruCache`? What do you think?

I agree it's a bit awkward to put ideas such as `isCacheableHealResponse` inside a file called `types.ts`. At the same time, I don't think we should move these ideas into the client, because these ideas are more fundamental than any specific client implementation.

For example, if someone wanted to build a totally different client implementation, they should still make use of `isCacheableHealResponse`. Likewise, if someone wanted to implement some middleware on the server-side, this utility might also be relevant there too.

One suggestion could be to move these utility functions out of `types.ts` and instead into something like `responses.ts`. I don't have a strong opinion about a specific file name, but I do think it belongs outside of the client.

tk-o: Putting these helper functions into `response.ts` file should do the trick:
- SDK consumers can still use this functionality
- no runtime code lives in `types.ts`

tk-o (approved):
Merge when ready

packages/ensnode-utils/src/cache.ts
```diff
@@ -1,99 +1,101 @@
...
 
-      this.expiryQueue.pop();
-      // Only delete if the key in cache still has the same timestamp
-      const entry = this.cache.get(oldest[1]);
-      if (entry && entry.timestamp === oldest[0]) {
-        this.cache.delete(oldest[1]);
-      }
+    if (this._cache.size > this._capacity) {
+      // oldestKey is guaranteed to be defined
+      const oldestKey = this._cache.keys().next().value as string;
```
tk-o, line 75: This line is super awesome :D

PR #251, opened 2025-02-14
Switch logger in ENSRainbow to pino

djstrong:


packages/ensnode-utils/src/logger.ts
```diff
@@ -0,0 +1,31 @@
...
+export type LogLevel = (typeof LOG_LEVELS)[number];
+
+let globalLogger: pino.Logger | undefined;
+
+export function getLogger(options?: { level?: LogLevel }): pino.Logger {
+  if (!globalLogger) {
+    globalLogger = pino({
+      level: process.env.LOG_LEVEL || options?.level || "info",
+      transport: {
+        target: "pino-pretty",
```
lightwalker-eth, line 13: As I understand, the best practice is not to use "pino-pretty" in production.

Perhaps consider something like the following oversimplified example:

```
const isProduction = process.env.NODE_ENV === 'production';

const logger = pino({
  level: process.env.LOG_LEVEL || options?.level || 'info',
  ...(isProduction
    ? {} // In production, do not use pino-pretty
    : {
        transport: {
          target: 'pino-pretty',
          options: {
            colorize: true,
            translateTime: 'HH:MM:ss',
            ignore: 'pid,hostname',
          },
        },
      }),
});

```

Assuming we do this, I think we should also document in our readme.md file how `NODE_ENV` changes things.

packages/ensnode-utils/src/logger.ts
```diff
@@ -0,0 +1,31 @@
+import pino from "pino";
+
+export const LOG_LEVELS = ["fatal", "error", "warn", "info", "debug", "trace"] as const;
+export type LogLevel = (typeof LOG_LEVELS)[number];
```
lightwalker-eth, line 4: Instead of creating our own definition for these log levels, I think we can just import `LevelWithSilent` from pino. Please note how this also includes a "silent" option that Pino supports.

lightwalker-eth (changes requested):
@djstrong Hey this is moving in the right direction. Reviewed and shared some suggestions with feedback :+1:

packages/ensnode-utils/package.json
```diff
@@ -37,5 +38,9 @@
   },
   "engines": {
     "node": ">=18.14"
+  },
+  "dependencies": {
```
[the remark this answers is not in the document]
djstrong: Logger is moved to ensrainbow.

apps/ensrainbow/src/cli.ts
```diff
@@ -104,15 +105,15 @@ yargs(hideBin(process.argv))
         })
         .option("log-level", {
           type: "string",
-          description: "Log level (error, warn, info, debug)",
-          choices: Object.keys(logLevels),
+          description: "Log level (fatal, error, warn, info, debug, trace)",
```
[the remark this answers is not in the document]
djstrong: It is common practice to support both ways. However, supporting only env vars is more clear.

packages/ensnode-utils/src/logger.ts
djstrong: Changed to:
```
export type LogLevel = LevelWithSilent;

// The main reason for keeping our own validation, even though we're using Pino's types, is
// to provide a better user experience with clear error messages when invalid log levels are
// provided through environment variables.
export const VALID_LOG_LEVELS: LogLevel[] = [
  "fatal",
  "error",
  "warn",
  "info",
  "debug",
  "trace",
  "silent",
];
```

apps/ensrainbow/src/utils/logger.ts
```diff
@@ -1,34 +1,69 @@
...
+  if (!envLogLevel) {
+    return DEFAULT_LOG_LEVEL;
+  }
+
+  try {
+    return parseLogLevel(envLogLevel);
+  } catch (error) {
+    // Log error to console since we can't use logger yet
+    console.error(
+      `Invalid LOG_LEVEL environment variable value "${envLogLevel}". Valid levels are: ${VALID_LOG_LEVELS.join(", ")}. Defaulting to "${DEFAULT_LOG_LEVEL}".`,
```
lightwalker-eth, line 39:
```suggestion
      `Invalid LOG_LEVEL environment variable value "${envLogLevel}". Valid levels are: ${VALID_LOG_LEVELS.join(", ")}.`,
```

What do you think? It seems nice to me to fail loud.

apps/ensrainbow/src/utils/logger.ts
```diff
@@ -1,34 +1,69 @@
...
+    );
+    return DEFAULT_LOG_LEVEL;
```
lightwalker-eth, line 41: Please see comment above. If you agree, suggest to take the string you currently build inside `console.error` and convert it into a `const errorMessage = "..."` and then using that message in both `console.error` and inside the value passed to `throw new Error(errorMessage);`.

apps/ensrainbow/src/utils/logger.ts
lightwalker-eth, line 39: A few suggestions:

```suggestion
      `Environment variable error: (LOG_LEVEL): ${error.message}`,
```

This way we can reuse the error message you generate inside `parseLogLevel`.

Also suggest to not return `DEFAULT_LOG_LEVEL` and instead fail loudly.

apps/ensrainbow/src/utils/logger.ts
```diff
@@ -1,34 +1,69 @@
...
+  "trace",
+  "silent",
+];
+
+export function parseLogLevel(level: string): LogLevel {
+  const normalizedLevel = level.toLowerCase();
+  if (VALID_LOG_LEVELS.includes(normalizedLevel as LogLevel)) {
+    return normalizedLevel as LogLevel;
+  }
+  throw new Error(`Invalid log level "${level}". Valid levels are: ${VALID_LOG_LEVELS.join(", ")}`);
```
lightwalker-eth, line 25:
```suggestion
  throw new Error(`Invalid log level "${level}". Valid levels are: ${VALID_LOG_LEVELS.join(", ")}.`);
```

lightwalker-eth (changes requested):
@djstrong Great job on those updates! Just a few more small suggestions :+1:

apps/ensrainbow/src/commands/server-command.ts
```diff
@@ -4,55 +4,53 @@ import { Hono } from "hono";
...
-  const db = await openDatabase(options.dataDir, options.logLevel);
+  logger.info(`ENS Rainbow server starting on port ${options.port}...`);
 
-  // Check for incomplete ingestion
-  await exitIfIncompleteIngestion(db, log);
+  const db = await openDatabase(options.dataDir);
 
-  const app = createServer(db, log, options.logLevel);
+  // Check for incomplete ingestion
+  await exitIfIncompleteIngestion(db);
```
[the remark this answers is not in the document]
djstrong: Yes, it is in the other PR.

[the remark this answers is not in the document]
djstrong: We have `process.exit(1)` in `createDatabase` and `openDatabase`.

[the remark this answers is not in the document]
djstrong: We have `process.exit(1)` in `createDatabase` and `openDatabase`.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -11,18 +11,21 @@ export const LABELHASH_COUNT_KEY = new Uint8Array([0xff, 0xff, 0xff, 0xff]) as B
 export const INGESTION_IN_PROGRESS_KEY = new Uint8Array([0xff, 0xff, 0xff, 0xfe]) as ByteArray;
 
 /**
- * Checks if there's an incomplete ingestion and exits with a helpful error message if one is found
+ * Checks if there's an incomplete ingestion and throws an error if one is found
```
lightwalker-eth, line 14:
```suggestion
 * Ensures db does not have an incomplete ingestion.
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -11,18 +11,21 @@ export const LABELHASH_COUNT_KEY = new Uint8Array([0xff, 0xff, 0xff, 0xff]) as B
...
  * @param db The ENSRainbow database instance
+ * @throws Error if an incomplete ingestion is detected
  */
-export async function exitIfIncompleteIngestion(db: ENSRainbowDB): Promise<void> {
+export async function checkIngestionState(db: ENSRainbowDB): Promise<void> {
```
lightwalker-eth, line 18:
```suggestion
export async function ensureIngestionNotIncomplete(db: ENSRainbowDB): Promise<void> {
```

Suggesting this because the function doesn't just check, it ensures. And it specifically ensures that an ingestion is not incomplete.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -11,18 +11,21 @@ export const LABELHASH_COUNT_KEY = new Uint8Array([0xff, 0xff, 0xff, 0xff]) as B
...
   if (await isIngestionInProgress(db)) {
-    logger.error("Error: Database is in an incomplete state!");
-    logger.error("An ingestion was started but not completed successfully.");
-    logger.error("To fix this:");
-    logger.error("1. Delete the data directory");
-    logger.error("2. Run the ingestion command again: ensrainbow ingest <input-file>");
     await db.close();
```
lightwalker-eth, line 20: Worried about calling db.close() in this function. I assume the responsibility for closing is managed at some other layer?

I assume we might want to do this inside the same layer that calls db.open or whatever you think makes sense.

A strategy that can be helpful is:

```
// db.open call
try {
   // ... do a bunch of stuff with
} finally {
   await db.close();
}
```

Above is pseudocode. I might have got the syntax wrong 😅

apps/ensrainbow/src/commands/server-command.ts
```diff
@@ -47,8 +47,8 @@ export async function serverCommand(options: ServerCommandOptions): Promise<void
 
   const db = await openDatabase(options.dataDir);
 
-  // Check for incomplete ingestion
-  await exitIfIncompleteIngestion(db);
+  // Check if there's an incomplete ingestion
+  await checkIngestionState(db); //TODO: remove
```
lightwalker-eth, line 51: Appreciate your help to remember to action this 🙏 Thanks! 😄

lightwalker-eth (changes requested):
@djstrong Very nice 😄 Just 1 remaining open question about how we manage the responsibility for calling `db.close`.

apps/ensrainbow/src/cli.ts
```diff
@@ -78,19 +75,12 @@ yargs(hideBin(process.argv))
           type: "string",
           description: "Directory to store LevelDB data",
           default: getDataDir(),
-        })
-        .option("log-level", {
-          type: "string",
-          description: "Log level (error, warn, info, debug)",
-          choices: Object.keys(logLevels),
-          default: getDefaultLogLevel(),
```
tk-o, line 86: I like us replacing this arg with an env var 👍

apps/ensrainbow/src/utils/logger.ts
```diff
@@ -1,34 +1,70 @@
-export interface Logger {
-  error: (...args: any[]) => void;
-  warn: (...args: any[]) => void;
-  info: (...args: any[]) => void;
-  debug: (...args: any[]) => void;
+import pino, { LevelWithSilent } from "pino";
```
tk-o, line 1: I've had this idea to make logging configurable from the application layer to allow different applications using the logger they need. It's nothing that has to be implemented any time soon, just a rough thought:

1. Move the logger module to a shared package.
2. Use `logger interface` when logging messages.
3. Let application decide and set the actual `logger implementation` that matches the `logger interface`. One implementation will be `pino logger`, suitable for backend apps, and if anyone needs the `logger implementation` to work in web browser, they can provide their own `browser logger`.

lightwalker-eth (approved):
@djstrong Looks awesome 🚀 Great work 🙌

apps/ensrainbow/src/utils/logger.ts
lightwalker-eth: @tk-o We can explore that if / when it becomes a priority, but until we see a clear need for it prefer to avoid adding extra layers of abstractions.

PR #254, opened 2025-02-14
Improve ENSRainbow Docs

djstrong:


apps/ensrainbow/README.md
```diff
@@ -82,50 +80,85 @@ NameHash Labs operates a freely available instance of ENSRainbow for the ENS com
...
-# Get count of healable labels
-curl https://api.ensrainbow.io/v1/labels/count
+### Heal Label
+
+```bash
+curl https://api.ensrainbow.io/v1/heal/0x[labelhash]
 ```
 
-While we aim for high availability, if you need guaranteed uptime or want to keep your requests private, we recommend running your own instance using the instructions above.
+The `labelhash` parameter must be strictly formatted according to these requirements:
```
lightwalker-eth, line 101:
```suggestion
The `labelhash` parameter must be strictly formatted as a "normalized labelhash" according to these requirements:
```

apps/ensrainbow/README.md
```diff
@@ -82,50 +80,85 @@ NameHash Labs operates a freely available instance of ENSRainbow for the ENS com
...
+- Must start with '0x'
+- Must be exactly 66 characters long (including '0x' prefix)
+- Must be lowercase
+- Must be a valid hex string that converts to exactly 32 bytes
 
-## API Endpoints
+For example, this is a valid labelhash:
```
lightwalker-eth, line 107:
```suggestion
For example, this is a valid "normalized labelhash":
```

apps/ensrainbow/README.md
```diff
@@ -2,24 +2,22 @@
...
 - **Storage**:
   - At least 15 GB of free disk space:
     - 6.37 GB for the compressed rainbow tables download
     - ~7 GB for the LevelDB database after ingestion
     - Additional temporary space during build/ingestion
-- **Memory**: At least 4 GB RAM recommended during data ingestion
+- **Memory**: At least 4 GB RAM recommended
 
-### Runtime Requirements
+### Server Requirements (`serve` command)
```
lightwalker-eth, line 20:
```suggestion
### API Server Requirements (`serve` command)
```

apps/ensrainbow/src/lib/server.ts
```diff
@@ -13,18 +13,41 @@ import {
...
+
+    // Check if ingestion is in progress
+    if (await isIngestionInProgress(db)) {
+      throw new Error("Database is in an invalid state: ingestion in progress flag is set");
+    }
+
+    // Verify we can get the rainbow record count
+    const countResponse = await server.labelCount();
+    if (countResponse.status === StatusCode.Error) {
+      throw new Error(`Failed to get rainbow record count: ${countResponse.error}`);
```
lightwalker-eth, line 45:
```suggestion
      throw new Error(`Database is in an invalid state: failed to get rainbow record count: ${countResponse.error}`);
```

apps/ensrainbow/src/lib/server.ts
```diff
@@ -13,18 +13,41 @@ import {
...
+  /**
+   * Creates a new ENSRainbowServer instance
+   * @param db The ENSRainbow database instance
+   * @param logLevel Optional log level
+   * @throws Error if database validation fails
+   */
+  public static async init(db: ENSRainbowDB, logLevel?: LogLevel): Promise<ENSRainbowServer> {
+    const server = new ENSRainbowServer(db, logLevel);
+
+    // Check if ingestion is in progress
```
lightwalker-eth, line 37:
```suggestion
    // Verify that the attached db fully completed its ingestion (ingestion not interrupted)
```

apps/ensrainbow/src/lib/server.ts
```diff
@@ -13,18 +13,41 @@ import {
...
     this.logger = createLogger(logLevel);
   }
 
+  /**
+   * Creates a new ENSRainbowServer instance
+   * @param db The ENSRainbow database instance
+   * @param logLevel Optional log level
+   * @throws Error if database validation fails
+   */
+  public static async init(db: ENSRainbowDB, logLevel?: LogLevel): Promise<ENSRainbowServer> {
```
lightwalker-eth, line 34: I think it's ok to move this logic directly into the constructor and to keep the constructor public.

The main motivation for creating a public static function for creating an instance like this is if you want to have different strategies for creating an instance of an object. Ex: The difference between open vs create. In this situation I think the desired strategy is always the same.

It's nice to keep the JSDoc comments you created. Suggest adjusting them for the constructor.

I think it's ok to make the constructor async too? If that's an issue then I suppose please ignore everything I wrote above because I guess we need these operations to be async.

apps/ensrainbow/src/lib/server.ts
```diff
@@ -13,18 +13,41 @@ import {
...
 
-  constructor(db: ENSRainbowDB, logLevel?: LogLevel) {
+  private constructor(db: ENSRainbowDB, logLevel?: LogLevel) {
     this.db = db;
     this.logger = createLogger(logLevel);
   }
 
+  /**
+   * Creates a new ENSRainbowServer instance
+   * @param db The ENSRainbow database instance
```
lightwalker-eth, line 30:
```suggestion
   * @param db The ENSRainbowDB instance
```

apps/ensrainbow/src/lib/server.ts
```diff
@@ -13,18 +13,41 @@ import {
...
+   * @param logLevel Optional log level
+   * @throws Error if database validation fails
```
lightwalker-eth, line 32:
```suggestion
   * @throws Error if a "lite" validation of the database fails
```

apps/ensrainbow/src/commands/server-command.test.ts
```diff
@@ -23,18 +24,32 @@ import { createServer } from "./server-command";
 describe("Server Command Tests", () => {
   let db: ENSRainbowDB;
   const port = 3223;
```
lightwalker-eth, line 26: Isn't this the default port? From the comments in this file I understand its supposed to be different than the default?

Also suggest renaming `port` to something like `nonDefaultPort`.

lightwalker-eth (changes requested):
@djstrong Hey nice updates here! Reviewed and shared feedback :+1:

apps/ensrainbow/README.md
```diff
@@ -222,13 +237,17 @@ Expected count as of January 30, 2024: 133,856,894 unique label-labelhash pairs
...
-### Data Ingestion Variables
-- `INPUT_FILE`: Path to the gzipped SQL dump file containing ENS rainbow tables (default: './ens_names.sql.gz'). Only used during data ingestion.
+### Global Variables
+These variables affect all commands:
+- `DATA_DIR`: Default directory for the LevelDB database. If not provided, defaults to `data/`
+- `LOG_LEVEL`: Logging level for all operations, one of: "debug", "info", "warn", "error". If not provided, defaults to `info`
+
+### Server Command Variables
+These variables affect the ENSRainbow server operation:
+- `PORT`: Server port (default: 3223)
```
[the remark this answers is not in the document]
djstrong: I think `PORT` should be available by env var so you can change while running docker image without changing the code.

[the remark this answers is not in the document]
lightwalker-eth: @djstrong Ah I see! I also see the work you did on `getDefaultPort `. Cool! As I understand, if a port is passed on the command line, it is given priority over a port set through the `PORT` env var?

Is that right? Suggest documenting how the port actually used is decided.

Or maybe raise an error if both the env var is set and the commandline arg are passed and the values don't agree? I'm a fan of failing loud when people do bad things. If we enforce that these values can't both be defined with conflicting values then I think no need to add more special documentation about this, since if someone tries to do something silly, they just get an error 😄

apps/ensrainbow/README.md
```diff
@@ -241,13 +241,13 @@ The following environment variables can be used to configure different aspects o
 
 ### Global Variables
 These variables affect all commands:
-- `DATA_DIR`: Default directory for the LevelDB database. If not provided, defaults to `data/`
 - `LOG_LEVEL`: Logging level for all operations, one of: "debug", "info", "warn", "error". If not provided, defaults to `info`
```
lightwalker-eth, line 244: Sorry if I'm going across different PRs. Perhaps we should add the extra log level here for silent? Please feel welcome to ignore this feedback if not a good idea based on separate PRs / merge strategy / etc 😅

apps/ensrainbow/README.md
```diff
@@ -241,13 +241,13 @@ The following environment variables can be used to configure different aspects o
...
+- `NODE_ENV`: Standard environment variable (used for logging format)
```
lightwalker-eth, line 245:
```suggestion
- `NODE_ENV`: Standard Node.js environment variable used to indicate the current environment, such as "development, "test", or "production". If "production" a performance optimized logging format is used.
```

apps/ensrainbow/src/cli.ts
```diff
@@ -13,6 +13,14 @@ function getDefaultLogLevel(): LogLevel {
   return envLogLevel && envLogLevel in logLevels ? envLogLevel : "info";
 }
 
+const DEFAULT_PORT = 3223;
```
lightwalker-eth, line 16: Suggest we move the definition of these "port" related ideas into `apps/ensrainbow/src/commands/server-command.ts`. They can be exported from that file and referenced here.

apps/ensrainbow/src/cli.ts
```diff
@@ -13,6 +13,14 @@ function getDefaultLogLevel(): LogLevel {
...
+
+function getDefaultPort(): number {
+  const envPort = process.env.PORT;
+  const port = envPort ? parseInt(envPort, 10) : DEFAULT_PORT;
```
lightwalker-eth, line 20: Suggest making use of `parseNonNegativeInteger` defined in `apps/ensrainbow/src/utils/number-utils.ts`.

apps/ensrainbow/src/cli.ts
```diff
@@ -13,6 +13,14 @@ function getDefaultLogLevel(): LogLevel {
   return envLogLevel && envLogLevel in logLevels ? envLogLevel : "info";
 }
 
+const DEFAULT_PORT = 3223;
+
+function getDefaultPort(): number {
```
lightwalker-eth, line 18: A few ideas for this function. The big idea is to repeat the overall patterns / strategy that we used in PR 227 for `getEnvLogLevel`.
* use `parseNonNegativeInteger` as the "parse" function.
* rename `getDefaultPort` to `getEnvPort`.

lightwalker-eth (changes requested):
@djstrong Nice updates! I think we're super close. Expect these will be my last suggestions for this PR :+1:

apps/ensrainbow/src/cli.ts
```diff
@@ -3,22 +3,46 @@ import type { ArgumentsCamelCase, Argv } from "yargs";
...
+  try {
+    const port = parseNonNegativeInteger(envPort);
+    if (port === null) {
+      throw new Error(`Invalid port number "${envPort}". Port must be a non-negative integer.`);
+    }
+
+    return port;
+  } catch (error: unknown) {
+    const errorMessage = `Environment variable error: (PORT): ${error instanceof Error ? error.message : String(error)}`;
+    // Log error to console since we can't use logger yet
```
lightwalker-eth, line 32: It's safe to use logger here :+1:

lightwalker-eth (approved):
@djstrong Looks awesome 🚀 1 very tiny suggestion. Please take the lead to merge this PR in whenever you're ready :+1:

apps/ensrainbow/src/cli.ts
djstrong: I think we can't without the second PR - I will address it while merging.

Issue #249, opened 2025-02-14
Add missing operations to ENSRainbow SDK


lightwalker-eth:
We should add operations to the ENSRainbow SDK for:
* label count
* health check

Suggest to not take action on this until the following PRs are resolved:
* https://github.com/namehash/ensnode/pull/242
* https://github.com/namehash/ensnode/pull/248

shrugs:
closable?

PR #266, opened 2025-02-17
feat(ensrainbow-sdk): add missing operations

### LINKED ISSUE #210: ENSRainbow SDK exports layout
@lightwalker-eth suggested
> What do you think about making it so that when we import anything from ensrainbow-sdk that we don't need to import from a specific file such as /types? and instead can just import from ensrainbow-sdk.

tk-o:
Adds support for `count` and `health` ENSRainbow endpoints. Also, updates the exports layout to allow upstream package consumers importing directly from SDK root path.

Resolves #210, #249.


packages/ensrainbow-sdk/tsup.config.ts
notrab, line 1: 👀

lightwalker-eth: @notrab Hey, see your comment here 😄 Is there something you're suggesting for us to change?

lightwalker-eth (approved):
Looks good to me! 🚀

packages/ensrainbow-sdk/tsup.config.ts
notrab: I don't think we need this file in this PR, since `main` doesn't have any `tsup.config.ts` files yet.

tk-o: Sure, can drop this one 👍

Issue #252, opened 2025-02-14
ENSRainbow: Add a flag to the validate command to specify if it should be a "full" validation or a "lite" validation


djstrong:
We might then:
- Perform a "full" validation as part of the CI scripts that build the ensrainbow-data image.
- Perform a "lite" validation as part of the CI scripts that build the ensrainbow image.

PR #278, opened 2025-02-19
Add lite validation mode to ENSRainbow database validation

djstrong:
Close https://github.com/namehash/ensnode/issues/252

### DIFF
--- a/apps/ensrainbow/Dockerfile
+++ b/apps/ensrainbow/Dockerfile
@@ -17,6 +17,7 @@ COPY package.json pnpm-lock.yaml ./
 COPY packages ./packages
 COPY apps ./apps
 RUN pnpm install --frozen-lockfile
+RUN cd apps/ensrainbow && pnpm run validate:lite
 
 ###################
 ## App Server — runs server w/ data dir
--- a/apps/ensrainbow/src/cli.ts
+++ b/apps/ensrainbow/src/cli.ts
@@ -3,6 +3,7 @@ import type { ArgumentsCamelCase, Argv } from "yargs";
 import { hideBin } from "yargs/helpers";
 import yargs from "yargs/yargs";
 import { ingestCommand } from "./commands/ingest-command";
+import { purgeCommand } from "./commands/purge-command";
 import { serverCommand } from "./commands/server-command";
 import { validateCommand } from "./commands/validate-command";
 import { getDefaultDataSubDir, getEnvPort } from "./lib/env";
@@ -30,6 +31,10 @@ interface ServeArgs {
 interface ValidateArgs {
   "data-dir": string;
   lite: boolean;
+}
+
+interface PurgeArgs {
+  "data-dir": string;
 }
 
 export interface CLIOptions {
@@ -109,6 +114,22 @@ export function createCLI(options: CLIOptions = {}) {
         await validateCommand({
           dataDir: argv["data-dir"],
           lite: argv.lite,
+        });
+      },
+    )
+    .command(
+      "purge",
+      "Completely wipe all files from the specified data directory",
+      (yargs: Argv) => {
+        return yargs.option("data-dir", {
+          type: "string",
+          description: "Directory containing LevelDB data",
+          default: getDefaultDataSubDir(),
+        });
+      },
+      async (argv: ArgumentsCamelCase<PurgeArgs>) => {
+        await purgeCommand({
+          dataDir: argv["data-dir"],
         });
       },
     )
--- a/apps/ensrainbow/src/cli.ts
+++ b/apps/ensrainbow/src/cli.ts
@@ -29,6 +29,7 @@ interface ServeArgs {
 
 interface ValidateArgs {
   "data-dir": string;
+  lite: boolean;
 }
 
 export interface CLIOptions {
@@ -92,15 +93,22 @@ export function createCLI(options: CLIOptions = {}) {
       "validate",
       "Validate the integrity of the LevelDB database",
       (yargs: Argv) => {
-        return yargs.option("data-dir", {
-          type: "string",
-          description: "Directory containing LevelDB data",
-          default: getDefaultDataSubDir(),
-        });
+        return yargs
+          .option("data-dir", {
+            type: "string",
+            description: "Directory containing LevelDB data",
+            default: getDefaultDataSubDir(),
+          })
+          .option("lite", {
+            type: "boolean",
+            description: "Perform a faster, less thorough validation by skipping hash verification",
+            default: false,
+          });
       },
       async (argv: ArgumentsCamelCase<ValidateArgs>) => {
         await validateCommand({
           dataDir: argv["data-dir"],
+          lite: argv.lite,
         });
       },
     )
--- a/apps/ensrainbow/src/commands/validate-command.ts
+++ b/apps/ensrainbow/src/commands/validate-command.ts
@@ -2,7 +2,7 @@ import { ENSRainbowDB } from "../lib/database";
 
 export interface ValidateCommandOptions {
   dataDir: string;
-  lite?: boolean;
+  lite?: boolean; // defaults to false
 }
 
 export async function validateCommand(options: ValidateCommandOptions): Promise<void> {
--- a/apps/ensrainbow/src/commands/validate-command.ts
+++ b/apps/ensrainbow/src/commands/validate-command.ts
@@ -2,13 +2,14 @@ import { ENSRainbowDB } from "../lib/database";
 
 export interface ValidateCommandOptions {
   dataDir: string;
+  lite?: boolean;
 }
 
 export async function validateCommand(options: ValidateCommandOptions): Promise<void> {
   const db = await ENSRainbowDB.open(options.dataDir);
 
   try {
-    const isValid = await db.validate();
+    const isValid = await db.validate({ lite: options.lite });
     if (!isValid) {
       // Throw error to ensure process exits with non-zero status code for CI/CD and scripts
       throw new Error("Validation failed");
--- a/apps/ensrainbow/src/lib/database.test.ts
+++ b/apps/ensrainbow/src/lib/database.test.ts
@@ -145,6 +145,48 @@ describe("Database", () => {
         await db.close();
       }
     });
+
+    it("should skip labelhash verification in lite mode", async () => {
+      const db = await ENSRainbowDB.create(tempDir);
+
+      try {
+        // Add record with mismatched labelhash (would fail in full validation)
+        const label = "vitalik";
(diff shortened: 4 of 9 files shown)


apps/ensrainbow/src/lib/server.ts
```diff
@@ -19,6 +19,7 @@ export class ENSRainbowServer {
   public static async init(db: ENSRainbowDB): Promise<ENSRainbowServer> {
     const server = new ENSRainbowServer(db);
 
+    //TODO maybe we should call validate lite here instead?
```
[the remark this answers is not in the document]
djstrong: Inconsistency is quite probable because the build of database image is triggered manually and a change of schema will not be detected by running the server and calling endpoints. It is possible that someone will run the ingest with previous schema and update the code to newer schema without running validation - that's why the lite validation should be in init - it's very fast too.
It is also in Dockerfile so it can be detected sooner.
Yes, it is redundant in our production build and runtime but not in other cases and development.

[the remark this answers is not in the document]
tk-o: Fair enough, thank you for extra explanation 🤝

tk-o, line 22: @djstrong did you mean?
```suggestion
    await db.validate({ lite: true });
```

djstrong: Yes, and remove below redundant check.

apps/ensrainbow/Dockerfile
```diff
@@ -17,6 +17,7 @@ COPY package.json pnpm-lock.yaml ./
 COPY packages ./packages
 COPY apps ./apps
 RUN pnpm install --frozen-lockfile
+RUN cd apps/ensrainbow && pnpm run validate:lite
```
lightwalker-eth, line 20: Suggest adding some comment here adding extra visibility to how we're running a lite db validation before starting the server.

Also, I'm confused by the ordering of logic here. Am I correct to understand this is running validation before the data dir is coied from ensrainbow-data?

Don't we need to copy the data dir first before running the lite validation?

apps/ensrainbow/src/cli.ts
```diff
@@ -92,15 +93,22 @@ export function createCLI(options: CLIOptions = {}) {
...
-        });
+        return yargs
+          .option("data-dir", {
+            type: "string",
+            description: "Directory containing LevelDB data",
+            default: getDefaultDataSubDir(),
+          })
+          .option("lite", {
+            type: "boolean",
+            description: "Perform a faster, less thorough validation by skipping hash verification",
```
lightwalker-eth, line 104: Doesn't this skip more than hash verification? Ex: it also skips verifying the the precalculated rainbow record count too? Suggest to invest getting this language nicely polished.

apps/ensrainbow/src/cli.ts
```diff
@@ -92,15 +93,22 @@ export function createCLI(options: CLIOptions = {}) {
...
-          description: "Directory containing LevelDB data",
-          default: getDefaultDataSubDir(),
-        });
+        return yargs
+          .option("data-dir", {
+            type: "string",
+            description: "Directory containing LevelDB data",
+            default: getDefaultDataSubDir(),
+          })
+          .option("lite", {
```
lightwalker-eth, line 102: Please update the docs about the `validate` command in the readme file.

apps/ensrainbow/src/commands/validate-command.ts
```diff
@@ -2,13 +2,14 @@ import { ENSRainbowDB } from "../lib/database";
 
 export interface ValidateCommandOptions {
   dataDir: string;
+  lite?: boolean;
```
lightwalker-eth, line 5:
```suggestion
  lite?: boolean; // defaults to false
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -207,87 +211,136 @@ export class ENSRainbowDB {
   }
 
   /**
-   * Validates the database contents.
-   * @returns boolean indicating if validation passed
+   * Gets the current database schema version.
+   * @returns The current schema version as a number, or null if not set
```
lightwalker-eth, line 215:
```suggestion
   * @returns The current schema version as a non-negative integer, or null if not set
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -207,87 +211,136 @@ export class ENSRainbowDB {
   }
 
   /**
-   * Validates the database contents.
-   * @returns boolean indicating if validation passed
+   * Gets the current database schema version.
```
lightwalker-eth, line 214:
```suggestion
   * Gets the database schema version.
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -46,7 +48,9 @@ export class ENSRainbowDB {
       });
       logger.info("Opening database...");
       await db.open();
-      return new ENSRainbowDB(db, dataDir);
+      const dbInstance = new ENSRainbowDB(db, dataDir);
+      await dbInstance.setDatabaseSchemaVersion(SCHEMA_VERSION);
```
lightwalker-eth, line 52: This looks good to me! A few suggestions:
1. Please update the comment on the `create` function to make note of how this function initializes this.
2. Please apply a similar idea in the static `open` function. Suggest that the `open` function checks the database schema version inside of it before returning the `ENSRainbowDB` and if it doesn't match the expected value that it throws an error. The comments on the `create` function should also be updated to talk about this.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -207,87 +211,136 @@ export class ENSRainbowDB {
...
-    let invalidHashes = 0;
-    let hashMismatches = 0;
+  public async getDatabaseSchemaVersion(): Promise<number | null> {
+    const versionStr = await this.get(SCHEMA_VERSION_KEY);
+    if (versionStr === null) {
+      return null;
+    }
 
-    logger.info("Starting database validation...");
+    try {
```
lightwalker-eth, line 224: I don't believe this logic should be inside a function called `getDatabaseSchemaVersion`. This function name implies it should always just get the value, not validate it.

Suggest moving the validation that the schema version returned matches the expected value somewhere else.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -207,87 +211,136 @@ export class ENSRainbowDB {
...
+  }
+
+  /**
+   * Validates the database contents.
+   * @param options Optional validation options
+   * @param options.lite If true, performs a faster validation by skipping labelhash verification
+   * @returns boolean indicating if validation passed
+   */
+  public async validate(options: { lite?: boolean } = {}): Promise<boolean> {
+    logger.info(`Starting database validation${options.lite ? " (lite mode)" : ""}...`);
```
lightwalker-eth, line 250: Suggest to begin the implementation of this function by fully materializing `lite` into either an explicit `true` or `false` boolean value. All remaining logic in this function shouldn't have to consider the possibility it is undefined.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -207,87 +211,136 @@ export class ENSRainbowDB {
...
+    //TODO should we validate if the count is TOTAL_EXPECTED_RECORDS?
+
+    let rainbowRecordCount = 0;
+    let validHashes = 0;
+    let invalidHashes = 0;
+    let hashMismatches = 0;
 
-      // Verify key is a valid labelhash by converting it to hex string
-      const keyHex = `0x${Buffer.from(key).toString("hex")}` as `0x${string}`;
+    // In lite mode, just check the count
```
lightwalker-eth, line 272: In lite mode, we should just check that the precalculated count is defined as a string that properly converts to a non-negative integer.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -207,87 +211,136 @@ export class ENSRainbowDB {
...
+    }
 
-      // Skip keys not associated with rainbow records
-      if (
-        byteArraysEqual(key, LABELHASH_COUNT_KEY) ||
-        byteArraysEqual(key, INGESTION_IN_PROGRESS_KEY)
-      ) {
-        continue;
-      }
+    //TODO should we validate if the count is TOTAL_EXPECTED_RECORDS?
```
lightwalker-eth, line 265: If we are NOT in lite mode, then we need to fully validate that the true count of rainbow records equals the precalculated count of rainbow records.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -316,7 +369,8 @@ export class ENSRainbowDB {
       // Skip keys not associated with rainbow records
       if (
```
lightwalker-eth, line 370: There's a few places where you've defined logic like this. Perhaps move it into a helper function such as `isRainbowRecordKey`? This could simply check the length of the byte array.

A separate helper function might be `isSystemKey`. This could check if the key is one of the ones you're checking for here.

I see a use case for both of those helper functions.

apps/ensrainbow/src/lib/server.ts
lightwalker-eth: Agreed with implementing a lite validation here and also removing the check below.

@djstrong Could you please suggest how we keep the nice error message you implemented below for the case that the lite validation fails because an ingestion was unfinished?

apps/ensrainbow/src/lib/database.test.ts
```diff
@@ -195,3 +232,64 @@ describe("parseNonNegativeInteger", () => {
...
+
+describe("schema version", () => {
+  test("new database has correct schema version", async () => {
+    const db = await ENSRainbowDB.create(TEST_DB_PATH);
+    try {
+      const version = await db.getDatabaseSchemaVersion();
+      expect(version).toBe(SCHEMA_VERSION);
+    } finally {
+      await db.close();
+      await rm(TEST_DB_PATH, { recursive: true, force: true });
```
lightwalker-eth, line 246: Perhaps in another PR we should add a static function to `ENSRainbowDB` for `purge` that would take a path and execute this internally?

lightwalker-eth (approved):
@djstrong Hey great to see the progress here :+1: Reviewed and shared feedback. Happy for you to decide how you want to action the feedback. I'm open to merging now and following up on suggestions in a separate future PR. Whatever you suggest :+1:

apps/ensrainbow/src/lib/database.ts
djstrong: So no checking against `const TOTAL_EXPECTED_RECORDS`

PR #283, opened 2025-02-19
deps: update to ponder 0.9.18

### LINKED ISSUE #280: some WrappedDomain entities aren't being deleted in handleNameUnwrapped
ponder index has 483,646 wrapped domains, subgraph has 480,283
difference of 3,363

postgres says 6547 domains have been unwrapped and then not wrapped again
of these, 3186 were correctly deleted (not exist in wrapped_domains)
and 3361 were not deleted (still exist in wrapped_domains)

subgraph has 480,283 wrapped domain entities, this ponder index has 483,646
which is a difference of 3,363 domains that ponder has that it should not

this is pretty close to the 3361 that were not deleted correctly, but is off by 2 and i can't think
of where the 2 might come from. likely an issue with postgres queries (claude generated)

---

likely ponder internal issue, they're actively debugging.

shrugs:
tentatively closes #280 , will re-open if that turns out to not be the case

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -22,8 +22,8 @@ catalogs:
       specifier: ^4.6.14
       version: 4.6.17
     ponder:
-      specifier: ^0.9.17
-      version: 0.9.17
+      specifier: ^0.9.18
+      version: 0.9.18
     typescript:
       specifier: ^5.7.3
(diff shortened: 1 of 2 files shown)


lightwalker-eth (approved):
Let's go 🚀

PR #292, opened 2025-02-20
fix: add indexes for reverse domainId relations

shrugs:
speeds up queries that access things like `domain { registrant { id } }` etc

please read this diff with "hide whitespace" on


tk-o (approved):
Very nice, some low hanging fruits 🍒

Issue #309, opened 2025-02-24
Handle ENSRainbow runner out of disk space issue


BanaSeba:
When building ENSRainbow image Github runners are getting warnings/errors about reaching the disk space quota for Github runner.

![Image](https://github.com/user-attachments/assets/0dfa9be2-17cd-44d1-aa09-b8280ebc352d)

This needs to be addressed as this will block any attempts to increase ENSRainbow application size.

Docs considering Github runners and their quota: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

shrugs:
this will be handled as part of #386

PR #310, opened 2025-02-24
Feat/ensrainbow optimize runner disk usage

BanaSeba:
Related to: https://github.com/namehash/ensnode/issues/309
- Removed Android libs from Github runner
- Removed dotnet libs from Github runner


tk-o (approved):
So clever!

PR #316, opened 2025-02-25
feat(ensadmin): introduce Next.js app

tk-o:
This PR introduces a new `ensadmin-next` app. It's going to replace the current `ensadmin` app once confirmed to work well as a Vercel deployment. 

This PR aims to create foundations for further improvements. While reviewing this PR, please focus more on the foundations, and less about the features of the app. Please keep in mind, the a lot of functionality is going to be updated when #319 is completed.

Related to:
- #318 
- #319 
- https://github.com/ponder-sh/ponder/issues/1559

TODO:
- [x] fix a blocker issue that prevents loading Ponder Schema file on the client side due to some nested `node:fs` import done by Ponder lib

Related issues:
- https://github.com/vercel/next.js/issues/28774

<details><summary>Error log</summary>

```
 GET /ponder-client?ensnode=https%3A%2F%2Fensnode.run.tko.box%2F 500 in 84ms
 ⨯ node:fs
Module build failed: UnhandledSchemeError: Reading from "node:fs" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
Import trace for requested module:
node:fs
../../node_modules/.pnpm/ponder@0.9.22_@opentelemetry+api@1.7.0_@types+node@20.17.14_@types+react@19.0.10_hono@4.6.17__wpcb6c2qcm2t5bdromklk5xn7a/node_modules/ponder/dist/esm/indexing/addStackTrace.js
../../node_modules/.pnpm/ponder@0.9.22_@opentelemetry+api@1.7.0_@types+node@20.17.14_@types+react@19.0.10_hono@4.6.17__wpcb6c2qcm2t5bdromklk5xn7a/node_modules/ponder/dist/esm/indexing/index.js
../../node_modules/.pnpm/ponder@0.9.22_@opentelemetry+api@1.7.0_@types+node@20.17.14_@types+react@19.0.10_hono@4.6.17__wpcb6c2qcm2t5bdromklk5xn7a/node_modules/ponder/dist/esm/indexing-store/cache.js
../../node_modules/.pnpm/ponder@0.9.22_@opentelemetry+api@1.7.0_@types+node@20.17.14_@types+react@19.0.10_hono@4.6.17__wpcb6c2qcm2t5bdromklk5xn7a/node_modules/ponder/dist/esm/graphql/index.js
[...]

PR #322, opened 2025-02-27
fix(ensadmin-next): downgrade react from v19 to v18

tk-o:
This is a follow-up PR to:
- #316 

Trying to escape type errors spiral to avoid wasting time on fighting types while functionality is working as expected.


.gitignore
```diff
@@ -19,4 +19,4 @@ dist
 
 # Ponder
 generated
-ponder
```
tk-o, line 22: This change allows to narrow down the ignore path and allow creating `ponder` directories in the monorepo file tree.

apps/ensadmin-next/src/app/gql/ponder/page.tsx
```diff
@@ -0,0 +1,13 @@
+import { GraphiQLPage } from "@/components/graphiql/graphiql-page";
```
tk-o, line 1: Git stopped ignoring this file after changes in top-level `.gitignore` file.

notrab (approved):

PR #331, opened 2025-02-28
fix(ensrainbow): make `bun` runtime to run `vitest` test suite

tk-o:
Makes most of tests to pass, but some fixes are still required.

image

Related:
- https://github.com/oven-sh/bun/issues/4145

### DIFF
--- a/apps/ensrainbow/src/cli.test.ts
+++ b/apps/ensrainbow/src/cli.test.ts
@@ -143,7 +143,7 @@ describe("CLI", () => {
         expect(response.status).toBe(200);
 
         // Cleanup - send SIGINT to stop server
-        process.emit("SIGINT", "SIGINT");
+        // process.emit("SIGINT", "SIGINT");
         await serverPromise;
       });
 
@@ -199,7 +199,7 @@ describe("CLI", () => {
         expect(invalidHealData.error).toBe("Invalid labelhash length 12 characters (expected 66)");
 
         // Cleanup - send SIGINT to stop server
-        process.emit("SIGINT", "SIGINT");
+        // process.emit("SIGINT", "SIGINT");
         await serverPromise;
       });
 
--- a/apps/ensrainbow/vitest.config.ts
+++ b/apps/ensrainbow/vitest.config.ts
@@ -1,8 +1,41 @@
-import { defineConfig } from "vitest/config";
+import { defineConfig, ViteUserConfig } from "vitest/config";
+
+export type Runtime = "node" | "deno" | "bun";
+
+export function getRuntime(): Runtime {
+  if ("bun" in process.versions) return "bun";
+  if ("deno" in process.versions) return "deno";
+
+  return "node";
+}
+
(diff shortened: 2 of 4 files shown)


apps/ensrainbow/vitest.config.ts
```diff
@@ -1,8 +1,41 @@
...
+export type Runtime = "node" | "deno" | "bun";
+
+export function getRuntime(): Runtime {
+  if ("bun" in process.versions) return "bun";
+  if ("deno" in process.versions) return "deno";
+
+  return "node";
+}
+
+export function getPoolOptions(runtime: Runtime): ViteUserConfig["test"] {
```
lightwalker-eth, line 12: Interesting! Could you add a few little comments explaining the goals of this function?

apps/ensrainbow/src/cli.test.ts
```diff
@@ -143,7 +143,7 @@ describe("CLI", () => {
         expect(response.status).toBe(200);
 
         // Cleanup - send SIGINT to stop server
-        process.emit("SIGINT", "SIGINT");
+        // process.emit("SIGINT", "SIGINT");
```
lightwalker-eth, line 146: Can you confirm we wanted to comment this out? Please ignore this comment if not relevant.

apps/ensrainbow/src/cli.test.ts
```diff
@@ -199,7 +199,7 @@ describe("CLI", () => {
         expect(invalidHealData.error).toBe("Invalid labelhash length 12 characters (expected 66)");
 
         // Cleanup - send SIGINT to stop server
-        process.emit("SIGINT", "SIGINT");
+        // process.emit("SIGINT", "SIGINT");
```
lightwalker-eth, line 202: Can you confirm we wanted to comment this out? Please ignore this comment if not relevant.

lightwalker-eth (commented):
@tk-o Hey thanks for this :+1: Shared a few small comments here and will share a little more in Slack :+1:

tk-o:
This PR was an attempt to support @djstrong in getting tests to run for Bun-runtime app. I think we will need to migrate the whole testing suite to use Bun's testing utilities. We can come back to running tests with `vitest` once https://github.com/oven-sh/bun/issues/4145 is resolved.

apps/ensrainbow/src/cli.test.ts
tk-o: I commented it out, as the Bun test runner would got terminated as well. Perhaps we should shape the test cleanup functionality some other way.

apps/ensrainbow/src/cli.test.ts
tk-o: Responded here:
https://github.com/namehash/ensnode/pull/331#discussion_r1979363441

apps/ensrainbow/vitest.config.ts
tk-o: This function was meant to provide a conditional fallback for all other runtimes than `node` trying to execute vitest test suite (i.e. test coverage was being turned off).

PR #333, opened 2025-03-03
feat(ensnode.io): add sidebar topics to documentation

notrab:
I want to propose an idea that is to use an Astro Starlight plugin for sidebar topics. This means the nav doesn't appear deeply nested like you see here:

CleanShot 2025-03-03 at 09 11 38@2x

With the plugin, it enables us to create a cleaner nav that is separated by the parent topics.

CleanShot 2025-03-03 at 09 06 04@2x
CleanShot 2025-03-03 at 09 06 25@2x

To finish this up, we need to create two root pages for ensnode and ensrainbow, which I can do should we decide to proceed with this PR/idea.


docs/ensnode.io/config/integrations/starlight.ts
```diff
@@ -1,10 +1,91 @@
...
+            },
+            {
+              label: "Contributing",
+              collapsed: true,
+              autogenerate: { directory: "ensrainbow/contributing" },
+            },
+          ],
+        },
+        {
+          label: "ENSAdmin",
```
lightwalker-eth, line 71: The preview environment builds for this PR (PR 333) are failing, so I'm not able to test the changes in this PR.

However, I do want to flag that maybe something is wrong somewhere? Saying this because if we look at the current docs site the section for "ENSAdmin" is not appearing as expected:

CleanShot 2025-03-04 at 18 04 03

lightwalker-eth (approved):
@notrab Hey I like the overall strategy here and so approving the PR.

Shared feedback on some suggestions before actually merging the PR.

Please also investigate and resolve the build error: https://vercel.com/namehash/ensnode.io/6t2LJ1Rpuq77NHFRdMUZDaPCsuzf?filter=errors

Thanks :+1:

docs/ensnode.io/config/integrations/starlight.ts
```diff
@@ -1,10 +1,91 @@
...
+          items: [
+            {
+              label: "Using ENSAdmin",
+              collapsed: false,
+              autogenerate: { directory: "ensadmin/usage" },
+            },
+            {
+              label: "Contributing",
+              collapsed: true,
+              autogenerate: { directory: "ensrainbow/contributing" },
```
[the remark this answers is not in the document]
notrab: Good catch @lightwalker-eth, thx!

docs/ensnode.io/config/integrations/starlight.ts
notrab: Yeah so currently we don't have the ensadmin added to docs, which is why it's not on the current site.

lightwalker-eth (approved):
@notrab Great updates! 🚀 Nice work.

PR #341, opened 2025-03-05
deps(ponder): update to v0.9.26

tk-o:
This PR includes Ponder dependency upgrade which includes:
- https://github.com/ponder-sh/ponder/pull/1580
that IMO fixes:
- https://github.com/ponder-sh/ponder/issues/1577


lightwalker-eth (approved):
@tk-o Looks good, thanks :+1:

Issue #332, opened 2025-02-28
ENSRainbow: change the INGESTION_UNFINISHED_KEY to become INGESTION_STATUS_KEY


djstrong:
I'm thinking we change the INGESTION_UNFINISHED_KEY to become INGESTION_STATUS_KEY, with the following potential values:

- (null): Indicates that ingestion has NEVER started. Being able to detect this state might help detect more error states or improve some error messages? Advice appreciated.
- "done": Indicates that an ingestion has started and finished.
- "unfinished": Indicates that an ingestion was started but hasn't finished.

lightwalker-eth:
@djstrong Thanks for creating this! When we make this change, suggest to also increment the database schema number as this represents a schema change.

PR #343, opened 2025-03-05
ENSRainbow: change the INGESTION_UNFINISHED_KEY to become INGESTION_STATUS_KEY

djstrong:
https://github.com/namehash/ensnode/issues/332

### DIFF
--- a/apps/ensrainbow/src/lib/database.ts
+++ b/apps/ensrainbow/src/lib/database.ts
@@ -11,23 +11,22 @@ export const SYSTEM_KEY_PRECALCULATED_RAINBOW_RECORD_COUNT = new Uint8Array([
 /**
  * Key for storing the ingestion status
  * Possible values:
- * - null: Ingestion has never been started
- * - "unfinished": Ingestion was started but hasn't finished
- * - "finished": Ingestion has finished successfully
+ * - IngestionStatus.Unstarted: Ingestion has never been started
+ * - IngestionStatus.Unfinished: Ingestion was started but hasn't finished
+ * - IngestionStatus.Finished: Ingestion has finished successfully
  */
 export const SYSTEM_KEY_INGESTION_STATUS = new Uint8Array([0xff, 0xff, 0xff, 0xfe]) as ByteArray;
 export const SYSTEM_KEY_SCHEMA_VERSION = new Uint8Array([0xff, 0xff, 0xff, 0xfd]) as ByteArray;
 export const SCHEMA_VERSION = 2;
 
 // Ingestion status values
-/**
- * Indicates that an ingestion was started but hasn't finished
- */
-export const INGESTION_STATUS_UNFINISHED = "unfinished";
-/**
- * Indicates that an ingestion has finished successfully
- */
-export const INGESTION_STATUS_DONE = "done";
+export const IngestionStatus = {
+  Unstarted: "unstarted",
+  Unfinished: "unfinished",
+  Finished: "finished",
+} as const;
+
+export type IngestionStatus = (typeof IngestionStatus)[keyof typeof IngestionStatus];
 
 /**
  * Checks if a key is a system key (one of the special keys used for database metadata).
@@ -195,28 +194,40 @@ export class ENSRainbowDB {
   /**
    * Get the current ingestion status
    * @returns The current ingestion status:
-   * - null: Ingestion has never been started
-   * - "unfinished": Ingestion was started but hasn't finished
-   * - "finished": Ingestion has finished successfully
+   * - IngestionStatus.Unstarted: Ingestion has never been started
+   * - IngestionStatus.Unfinished: Ingestion was started but hasn't finished
+   * - IngestionStatus.Finished: Ingestion has finished successfully
+   * @throws Error if the value in the database is not a recognized enum value
    */
-  public async getIngestionStatus(): Promise<string | null> {
-    return await this.get(SYSTEM_KEY_INGESTION_STATUS);
+  public async getIngestionStatus(): Promise<IngestionStatus> {
+    const status = await this.get(SYSTEM_KEY_INGESTION_STATUS);
+    if (status === null) {
+      return IngestionStatus.Unstarted;
+    }
+
+    // Validate that the status is a recognized enum value
+    const validValues = Object.values(IngestionStatus) as string[];
+    if (!validValues.includes(status)) {
+      throw new Error(`Invalid ingestion status value found in database: ${status}`);
+    }
+
+    return status as IngestionStatus;
   }
 
   /**
    * Mark that an ingestion has started and is unfinished
-   * Sets the ingestion status to "unfinished"
+   * Sets the ingestion status to IngestionStatus.Unfinished
    */
   public async markIngestionUnfinished(): Promise<void> {
-    await this.db.put(SYSTEM_KEY_INGESTION_STATUS, INGESTION_STATUS_UNFINISHED);
+    await this.db.put(SYSTEM_KEY_INGESTION_STATUS, IngestionStatus.Unfinished);
   }
 
   /**
    * Mark that ingestion is finished
-   * Sets the ingestion status to "finished"
+   * Sets the ingestion status to IngestionStatus.Finished
    */
   public async markIngestionFinished(): Promise<void> {
-    await this.db.put(SYSTEM_KEY_INGESTION_STATUS, INGESTION_STATUS_DONE);
+    await this.db.put(SYSTEM_KEY_INGESTION_STATUS, IngestionStatus.Finished);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


apps/ensrainbow/src/lib/database.ts
```diff
@@ -8,11 +8,26 @@ import { logger } from "../utils/logger";
...
+ * - null: Ingestion has never been started
+ * - "unfinished": Ingestion was started but hasn't finished
+ * - "done": Ingestion has completed successfully
+ */
+export const SYSTEM_KEY_INGESTION_STATUS = new Uint8Array([0xff, 0xff, 0xff, 0xfe]) as ByteArray;
 export const SYSTEM_KEY_SCHEMA_VERSION = new Uint8Array([0xff, 0xff, 0xff, 0xfd]) as ByteArray;
-export const SCHEMA_VERSION = 1;
+export const SCHEMA_VERSION = 2;
+
+// Ingestion status values
```
lightwalker-eth, line 22: We should define the set of ingestion status values differently.

For an example of the correct strategy, please see the strategy used to define `StatusCode`.

1. https://github.com/namehash/ensnode/blob/332-ensrainbow-change-the-ingestion_unfinished_key-to-become-ingestion_status_key/packages/ensrainbow-sdk/src/consts.ts#L3-L6
2. https://github.com/namehash/ensnode/blob/332-ensrainbow-change-the-ingestion_unfinished_key-to-become-ingestion_status_key/packages/ensrainbow-sdk/src/client.ts#L19

These are the best practice for defining an enum in Typescript. It's very strange that these are split into two separate files currently, but whatever. Normally it's just one line after another.

This strategy should be used here to define a similar enum strategy for `IngestionStatus`.

The values here should be:

`Unstarted`
`Unfinished`
`Done` (or maybe better to use the value "finished" rather than "done" since that is more appropriately related to our use of the language "unfinished").

NOTE: `getIngestionStatus` should never return `null`. Instead, if it finds a `null` value in the db, it should return `Unstarted`, such that the value returned from `getIngestionStatus` is always one of these enum values.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -8,11 +8,26 @@ import { logger } from "../utils/logger";
...
+/**
+ * Indicates that an ingestion was started but hasn't finished
+ */
+export const INGESTION_STATUS_UNFINISHED = "unfinished";
+/**
+ * Indicates that an ingestion has completed successfully
+ */
+export const INGESTION_STATUS_DONE = "done";
```
lightwalker-eth, line 30:
```suggestion
export const INGESTION_STATUS_FINISHED = "finished";
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -8,11 +8,26 @@ import { logger } from "../utils/logger";
...
-export const SCHEMA_VERSION = 1;
+export const SCHEMA_VERSION = 2;
+
+// Ingestion status values
+/**
+ * Indicates that an ingestion was started but hasn't finished
+ */
+export const INGESTION_STATUS_UNFINISHED = "unfinished";
+/**
+ * Indicates that an ingestion has completed successfully
```
lightwalker-eth, line 28:
```suggestion
 * Indicates that an ingestion has finished successfully
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -8,11 +8,26 @@ import { logger } from "../utils/logger";
...
 ]) as ByteArray;
-export const SYSTEM_KEY_INGESTION_UNFINISHED = new Uint8Array([
-  0xff, 0xff, 0xff, 0xfe,
-]) as ByteArray;
+/**
+ * Key for storing the ingestion status
+ * Possible values:
+ * - null: Ingestion has never been started
+ * - "unfinished": Ingestion was started but hasn't finished
+ * - "done": Ingestion has completed successfully
```
lightwalker-eth, line 16:
```suggestion
 * - "finished": Ingestion has finished successfully
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -178,26 +193,30 @@ export class ENSRainbowDB {
...
+   * @returns The current ingestion status:
+   * - null: Ingestion has never been started
+   * - "unfinished": Ingestion was started but hasn't finished
+   * - "done": Ingestion has completed successfully
    */
-  public async isIngestionUnfinished(): Promise<boolean> {
-    const value = await this.get(SYSTEM_KEY_INGESTION_UNFINISHED);
-    return value !== null;
+  public async getIngestionStatus(): Promise<string | null> {
+    return await this.get(SYSTEM_KEY_INGESTION_STATUS);
```
lightwalker-eth, line 203: This function should validate that the value it reads from the db is either `null` or a recognized enum value. If the value it finds is not a recognized enum value it should throw an error.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -178,26 +193,30 @@ export class ENSRainbowDB {
   }
 
   /**
-   * Check if an ingestion is unfinished
-   * @returns true if an ingestion is unfinished, false otherwise
+   * Get the current ingestion status
+   * @returns The current ingestion status:
+   * - null: Ingestion has never been started
```
lightwalker-eth, line 198: Please see my other comments how this function should never return `null`.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -178,26 +193,30 @@ export class ENSRainbowDB {
...
+   * - "unfinished": Ingestion was started but hasn't finished
+   * - "done": Ingestion has completed successfully
```
lightwalker-eth, line 200:
```suggestion
   * - "finished": Ingestion has finished successfully
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -178,26 +193,30 @@ export class ENSRainbowDB {
...
-    return value !== null;
+  public async getIngestionStatus(): Promise<string | null> {
+    return await this.get(SYSTEM_KEY_INGESTION_STATUS);
   }
 
   /**
    * Mark that an ingestion has started and is unfinished
+   * Sets the ingestion status to "unfinished"
    */
   public async markIngestionStarted(): Promise<void> {
```
lightwalker-eth, line 210:
```suggestion
  public async markIngestionUnfinished(): Promise<void> {
```

Seems good to maximally align our use of language.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -178,26 +193,30 @@ export class ENSRainbowDB {
...
-    await this.db.put(SYSTEM_KEY_INGESTION_UNFINISHED, "true");
+    await this.db.put(SYSTEM_KEY_INGESTION_STATUS, INGESTION_STATUS_UNFINISHED);
   }
 
   /**
    * Mark that ingestion is finished
+   * Sets the ingestion status to "done"
```
lightwalker-eth, line 216:
```suggestion
   * Sets the ingestion status to "finished"
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -359,8 +378,15 @@ export class ENSRainbowDB {
   }
 
   /**
-   * Validates the database contents.
-   * @param options Optional validation options
+   * Validates the database to ensure it's in a consistent state.
+   *
+   * Validation includes:
+   * 1. Checking the ingestion status (must be "done" for a valid database)
```
lightwalker-eth, line 384:
```suggestion
   * 1. Checking the ingestion status (must be "finished" for a valid database)
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -359,8 +378,15 @@ export class ENSRainbowDB {
...
+   * 2. Verifying the schema version matches the expected version
+   * 3. In full validation mode: Verifying all keys are valid labelhashes and match their labels
+   * 4. Verifying the precalculated rainbow record count matches the actual count
```
lightwalker-eth, line 387: Isn't this step only done in full validation mode too?

apps/ensrainbow/src/lib/database.ts
```diff
@@ -359,8 +378,15 @@ export class ENSRainbowDB {
...
+   *
+   * @param options Validation options
    * @param options.lite If true, performs a faster validation by skipping labelhash verification
```
lightwalker-eth, line 390: We should identify how `options.lite` also skips validating the precalculated rainbow record count.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -370,14 +396,32 @@ export class ENSRainbowDB {
 
     logger.info(`Starting database validation${isLiteMode ? " (lite mode)" : ""}...`);
     // Verify that the attached db fully completed its ingestion (ingestion not interrupted)
-    if (await this.isIngestionUnfinished()) {
+    const ingestionStatus = await this.getIngestionStatus();
+
+    if (ingestionStatus === null) {
+      const errorMsg = generatePurgeErrorMessage(
```
lightwalker-eth, line 402: Is it appropriate to have a purge error message in this case? Maybe the user just needs to run an ingest first?

lightwalker-eth (changes requested):
@djstrong Thanks for these updates. Reviewed and shared feedback :+1:

apps/ensrainbow/src/lib/database.ts
```diff
@@ -178,26 +192,42 @@ export class ENSRainbowDB {
...
-  public async isIngestionUnfinished(): Promise<boolean> {
-    const value = await this.get(SYSTEM_KEY_INGESTION_UNFINISHED);
-    return value !== null;
+  public async getIngestionStatus(): Promise<IngestionStatus> {
+    const status = await this.get(SYSTEM_KEY_INGESTION_STATUS);
+    if (status === null) {
+      return IngestionStatus.Unstarted;
+    }
+
+    // Validate that the status is a recognized enum value
```
lightwalker-eth, line 208: Suggest this strategy instead: https://claude.site/artifacts/0c39064a-af2d-4752-9a96-1d05b75786aa

lightwalker-eth (approved):
@djstrong Hey thanks for the updates. Looks good. Shared a few small suggestions. Please merge whenever you're ready :+1:

PR #358, opened 2025-03-09
feat(docs/ensnode.io): initial custom starlight ensnode theme

notrab:
I don't suggest we merge this "as is" until we make a few more refinements to the colours.

It currently looks best in "light mode" but there are still some outstanding UI issues with this I need to fix.

CleanShot 2025-03-09 at 18 02 11@2x


docs/ensnode.io/config/integrations/starlight.ts
```diff
@@ -149,7 +149,9 @@ export function starlight(): AstroIntegration {
       replacesTitle: true,
     },
     social: {
+      twitter: "https://x.com/NamehashLabs",
       github: "https://github.com/namehash/ensnode",
+      telegram: "https://t.me/namehash",
```
tk-o, line 154: Let's use official ENSNode tg channel:
```suggestion
      telegram: "https://t.me/ensnode",
```

packages/starlight-theme-ensnode/overrides/ThemeSelect.astro
```diff
@@ -0,0 +1,2 @@
+---
+---
```
tk-o, line 2: Curious about this empty file role? 🙃

packages/starlight-theme-ensnode/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "./package.json": "./package.json",
+    "./overrides/ThemeSelect.astro": "./overrides/ThemeSelect.astro"
+  },
+  "peerDependencies": {
+    "@astrojs/starlight": ">=0.32.0"
+  },
+  "engines": {
+    "node": ">=18"
+  },
+  "packageManager": "pnpm@9.0.4",
```
tk-o, line 20: Let's upgrade these settings to follow changes from #484:

```suggestion
  "engines": {
    "node": ">=22.14.0"
  },
  "packageManager": "pnpm@10.7.1",
```

packages/starlight-theme-ensnode/starlight.ts
```diff
@@ -0,0 +1,39 @@
...
+  logger: AstroIntegrationLogger
+): StarlightUserConfig["components"] {
+  const components = { ...starlightConfig.components };
+  for (const override of overrides) {
+    const name = typeof override === "string" ? override : override.name;
+
+    if (starlightConfig.components?.[name]) {
+      const fallback = `starlight-theme-ensnode/${
+        typeof override === "string" ? "overrides" : "components"
+      }/${typeof override === "string" ? override : override.fallback}.astro`;
```
tk-o, line 16: How does that alternative form look? Was aiming to make it easier reason about.
```suggestion
      const fallback = typeof override === "string"
        ? `starlight-theme-ensnode/overrides/${override}.astro`
        : `starlight-theme-ensnode/components/${override.fallback}.astro`;
```

packages/starlight-theme-ensnode/overrides/ThemeSelect.astro
notrab: Basically this overrides the ThemeSelect to effectively show nothing where this would be - disabling selecting themes.

PR #362, opened 2025-03-10
feat(ensindexer): introduce label healing from reverse registry

### LINKED ISSUE #204: Automatically Heal subnames of "addr.reverse".
# Goal

1. Provide an optional feature in ENSNode to automatically heal subnames of "addr.reverse".
2. Demonstrate an additional key benefit of ENSNode to the ENS community: Over a million unknown names in the ENS Subgraph are known in ENSNode.
3. Support future special features in v2+ of ENSNode that would benefit from an improved representation of ENS primary name configurations.

# Definitions

* "normalized address digits": Given any valid Ethereum address, convert to lowercase and remove the "0x" prefix.

# Background Info

1. ENS primary names are made possible (in part) through ENS reverse resolution.
2. ENS reverse resolution is made possible through subnames of "addr.reverse".
3. Other than "eth", "reverse" is the only TLD ENS adds to become a superset of DNS.
4. A node for "addr.reverse" lives in the ENS Registry.
5. The "owner" field for "addr.reverse" in the ENS Registry is a "ReverseRegistrar" contract with the responsibility for registering subnames of "addr.reverse" into the ENS Registry.
   1. Note how this "ReverseRegistrar" is an example implementation of the more general idea of a "registrar" / "subregistrar".
   2. The ENS DAO retains the ability to change which contract is the owner of the node in the ENS Registry associated with "addr.reverse". This effectively enables an "upgrade mechanism" for how the "ReverseRegistrar" operates.
6. When address X sets their ENS primary name, they send a request to the "ReverseRegistrar" which performs operations including:
   1. Upsert the name "{normalizedAddressDigits(X)}.addr.reverse" into the ENS Registry. Please remember: The ENS Registry doesn't store label values. Therefore this is actually implemented as upserting a subnode of "addr.reverse" with the labelhash `labelhash({normalizedAddressDigits(X)})`.
   2. Configure the resolver in the ENS Registry for the newly upserted node.
   3. Send a request to the resolver asking it to store a special "name" resolver record for the newly upserted node. The value sent for this special "name" resolver record is the primary name that address X is requesting to be set.
      1. Note how primary name values are actually represented as a special "name" resolver record in ENS!
7. There should therefore be a direct relationship between address X that initiated the transaction to set their ENS primary name, and the subname registered beneath "addr.reverse".
8. @djstrong has found that there are almost 1 million (and growing!) subnames of "addr.reverse". Each of these are currently represented as "unknown" inside ENSNode. It would be very nice if we could automatically heal all of these!
9. Automatically healing subnames of "addr.reverse" breaks backwards compatibility with the ENS Subgraph.

# Subtasks

- [ ] Introduce a new environment variable `HEAL_REVERSE_ADDRESSES` to ENSNode.
   - [x] By default, this should be interpreted as true.
   - [x] Validation logic should be defined (similar to our other environment variables) such that if an invalid value (anything not true or false) is set, a helpful error message is given that directs the operator to the specific environment variable name that is misconfigured.
   - [ ] Update environment variable documentation in https://github.com/namehash/ensnode/blob/main/README.md
   - [ ] Update environment variable documentation in https://github.com/namehash/ensnode/blob/main/apps/ensnode/.env.local.example
   - [ ] Ensure all documentation for this new environment variable highlights how turning this feature on breaks backwards compatibility with the ENS Subgraph.
- [ ] Update indexing logic for `handleNewOwner` inside `apps/ensnode/src/handlers/Registry.ts`.
   - [x] Note: Please suggest if we should make updates anywhere else. Ex: I see that there's multiple places we try to heal labels.
   - [x] Note: the current logic in this function has a variable "label" that is incorrectly named. It is actually "labelhash". In the pseudocode below I will refer to it by the name it should have: "labelhash".
   - [ ] if `HEAL_REVERSE_ADDRESSES` is true and `parent.name` is "addr.reverse":
      - [x] `const assumedLabelhash = labelHash(normalizedAddressDigits(event.transaction.from));`
      - [x] if "labelhash" is `assumedLabelhash`:
         - [x] heal the indexed representation of of the new subname as normalizedAddressDigits(event.transaction.from)
      - [ ] else:
         - [ ] throw an error with a helpful error message for debugging purposes. We should validate the assumptions described in the background info above such that if they are not true, we fail loudly.
   - [x] else:
      - [x] perform our standard attempted label healing logic with ENSRainbow
- [x] Update our deployments in Railway
   - [x] Explicitly turn this feature on for the existing `alpha` deployment
   - [x] Explicitly turn this feature off for the existing `subgraph` deployment
   - [x] Create a completely new deployment of ENSNode in Railway that is exactly the same as the existing `subgraph` deployment, except where this feature is explicitly turned on.
      - [x] Goals for this deployment include:
         - [x] Provide an example ENSNode deployment for the community that's 1-for-1 with the ENS Subgraph, with the sole exception of more healed labels.
         - [x] Help @djstrong build metrics to compare how many more labels we can heal in this deployment vs the regular ENS Subgraph.
      - [x] Configure this deployment in Railway such that it automatically redeploys based on our "subgraph" branch in GitHub.
      - [x] This ENSNode instance should use exactly the same Postgres DB / ENSRainbow instance as the others.
      - [x] Configure our DNS such that this new deployment is reachable at healed-subgraph.ensnode.io. Please ask @lightwalker-eth for assistance as needed.
      - [ ] Write a note in https://github.com/namehash/ensnode/issues/109 about the new deployment of ENSNode that also needs to be documented.

tk-o:
closes #204 

## Suggested review order

1. Review `package/ensnode-utils` updates.
2. Review `apps/ensindexer/src/lib` updates.
3. Review `apps/ensindexer/src/plugins` updates.
4. Review `apps/ensindexer/src/handlers` updates
5. Review `ponder.config.ts` updates.

## Scope

This PR introduces a feature of healing labels based on reverse addresses. This feature can be toggled on and off, by setting to `HEAL_REVERSE_ADDRESSES` to `true` or `false`, respectively. Please note, that only the `eth` plugin is going to support the new feature, while `base` and `linea` plugins will not.

To complete the feature, I had to:
- create a parser for `HEAL_REVERSE_ADDRESSES` env var;
- update the ponder plugin handler args type, to include plugin-defined helper methods `canHealReverseAddresses` and `isReverseRootNode`;
- have all plugins to provide implementations for new helper methods (some of them throw error just to manifest lack of support for the feature in question);
- update generic handlers `handleNameRegistered` & `handleNewOwner ` to attempt reverse address label healing before attempting ENSRainbow label healing;
- make ponder config object value altered based of `HEAL_REVERSE_ADDRESSES` env var value

## Pre-merge checklist
- [x] ensure every ENSIndexer and ENSApi service that runs with only `eth` plugin on has `HEAL_REVERSE_ADDRESSES` set to `false`

## Post-merge checklist
- [ ] create an issue for handling errors thrown by `labelByReverseAddress` call
- [ ] create an issue for healing addr.reverse subnames for non-mainnet chains
- [ ] create an issue for https://github.com/namehash/ensnode/pull/362#discussion_r2030766880
- [ ] create an issue for https://github.com/namehash/ensnode/pull/362#discussion_r2030770824

---

Resolves #204


apps/ensindexer/src/handlers/Registry.ts
```diff
@@ -124,7 +139,19 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
 
           // attempt to heal the label associated with labelhash via ENSRainbow
           // https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L112-L116
-          const healedLabel = await labelByHash(labelhash);
+          let healedLabel = await labelByHash(labelhash);
+
+          // if the label has not healed with ENSRainbow query
+          // and healing label from reverse addresses is enabled, give it a go
+          if (!healedLabel && canHealReverseAddresses()) {
```
djstrong, line 146: maybe check that first and if not try healing with ENSRainbow

tk-o: Ah ok, so first we try healing based of the in memory data (using reverse address method), and only if it didn't return any result we should try healing with ENSRainbow. Did I get that right?

djstrong: Yes, maybe it will be faster?

shrugs (approved):
Since only the eth plugin will perform this behavior, the configuration options need not be plugin-specific and you can use a global constant for `namehash('addr.reverse')` and reference the `healReverseAddresses` config in the shared handlers directly using a conditional like this, with the appropriate comments describing its behavior.

```ts
if (ownedName === '.eth' && healReverseAddresses && node === ADDR_REVERSE_NODE) { ... }
```

using this method the plugin definitions and handler args need not be updated at all.

---

alternatively if you'd prefer to keep the logic plugin-specific i'd suggest a simpler api like:

```ts
canHealFromNode?: (node: Node): boolean
```

that is simply not included for base/linea plugin definitions and the implementation for .eth looks like:

```ts
const canHealFromNode = (node: Node) => healReverseAddresses && node === ADDR_REVERSE_NODE;
```

and the conditional in the helpers looks like

```ts
if (canHealFromNode?.(node)) { ... }
```

---

otherwise LGTM

tk-o:
I like the alternative option suggestion as it shows (IMO) better the intent of which plugins support the new healing strategy and which don't.

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -172,7 +173,24 @@ export interface PonderENSPlugin<PLUGIN_NAME extends PluginName, CONFIG> {
...
 export type PonderENSPluginHandlerArgs<OWNED_NAME extends OwnedName> = {
   ownedName: OwnedName;
+
   namespace: ReturnType<typeof createPluginNamespace<OWNED_NAME>>;
+
+  /**
+   * Determines whether the plugin can heal reverse addresses.
+   * Some plugins might not need it at the moment.
+   *
+   * @returns true if the plugin can heal reverse addresses
```
lightwalker-eth, line 183:
```suggestion
   * @returns true if the plugin can heal addr.reverse subnames
```

apps/ensindexer/.env.local.example
```diff
@@ -57,3 +57,12 @@ ENSRAINBOW_URL=https://api.ensrainbow.io
 # The ENSIndexer public service URL
 # It is used to generate self-referential URLs in the API responses
 ENSNODE_PUBLIC_URL=http://localhost:42069
+
+# A feature flag to enable or disable reverse address healing
```
lightwalker-eth, line 61:
```suggestion
# A feature flag to enable or disable healing of addr.reverse subnames
```

apps/ensindexer/.env.local.example
```diff
@@ -57,3 +57,12 @@ ENSRAINBOW_URL=https://api.ensrainbow.io
...
+# If this is set to true, the indexer will attempt to heal reverse addresses
```
lightwalker-eth, line 62:
```suggestion
# If this is set to true, ENSIndexer will attempt to heal subnames of addr.reverse
```

apps/ensindexer/.env.local.example
```diff
@@ -57,3 +57,12 @@ ENSRAINBOW_URL=https://api.ensrainbow.io
...
+# for the ENS names that are owned by the address that is being healed.
```
lightwalker-eth, line 63:
```suggestion

```

Suggest to totally remove this line. We shouldn't get into the implementation details here.

apps/ensindexer/.env.local.example
```diff
@@ -57,3 +57,12 @@ ENSRAINBOW_URL=https://api.ensrainbow.io
...
+# If this is not set, the default value is set to `DEFAULT_HEAL_REVERSE_ADDRESSES`.
+#
+# WARNING: This is a breaking change for the ENS Subgraph compatibility. To avoid
```
lightwalker-eth, line 66:
```suggestion
# WARNING: Setting this to `true` results in indexed data no longer being backwards compatible with the ENS Subgraph. For full
```

apps/ensindexer/.env.local.example
```diff
@@ -57,3 +57,12 @@ ENSRAINBOW_URL=https://api.ensrainbow.io
...
+# breaking changes, set this to `false`.
```
lightwalker-eth, line 67:
```suggestion
# data-level backwards compatibility with the ENS Subgraph, set this to `false`.
```

apps/ensindexer/ponder.config.ts
```diff
@@ -14,7 +14,19 @@ import * as lineaEthPlugin from "./src/plugins/linea/ponder.plugin";
...
-type AllPluginConfigs = MergedTypes<(typeof ALL_PLUGINS)[number]["config"]>;
+type AllPluginConfigs = MergedTypes<(typeof ALL_PLUGINS)[number]["config"]> & {
+  /**
+   * The environment variables that change the behavior of the indexer.
+   * It's important to include all environment variables that change the behavior
+   * of the indexer to ensure Ponder app build ID is updated when any of them change.
+   **/
+  indexingBehaviorDependencies: {
+    /**
+     * Breaking change for ens-subgraph compatibility.
```
lightwalker-eth, line 25:
```suggestion
     * If set to `false` then indexed data will no longer be backwards compatible with the ENS Subgraph.
```

apps/ensindexer/ponder.config.ts
```diff
@@ -14,7 +14,19 @@ import * as lineaEthPlugin from "./src/plugins/linea/ponder.plugin";
 
 const ALL_PLUGINS = [ethPlugin, baseEthPlugin, lineaEthPlugin] as const;
 
-type AllPluginConfigs = MergedTypes<(typeof ALL_PLUGINS)[number]["config"]>;
+type AllPluginConfigs = MergedTypes<(typeof ALL_PLUGINS)[number]["config"]> & {
+  /**
+   * The environment variables that change the behavior of the indexer.
```
lightwalker-eth, line 19: @tk-o Love this! Nice work!

apps/ensindexer/src/handlers/Registry.ts
```diff
@@ -122,9 +132,28 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
+          // if healing label from reverse addresses is enabled,
+          // and the parent node is the reverse root node, give it a go
+          if (canHealReverseAddresses() && isReverseRootNode(node)) {
+            try {
+              healedLabel = labelByReverseAddress({
+                reverseAddress: owner,
+                labelhash,
+              });
+            } catch {
+              // TODO: store event args for analysis and debugging
```
lightwalker-eth, line 146: Suggest we create an issue on GitHub to remind us to come back to this later when we have more time 👍

Also suggest creating another separate GitHub issue to remind us to come back to healing addr.reverse subnames for non-mainnet chains.

apps/ensindexer/src/handlers/Registrar.ts
```diff
@@ -79,9 +88,27 @@ export const makeRegistrarHandlers = (ownedName: OwnedName) => {
 
       const node = makeSubnodeNamehash(ownedNameNode, labelhash);
 
-      // attempt to heal the label associated with labelhash via ENSRainbow
-      // https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ethRegistrar.ts#L56-L61
-      const healedLabel = await labelByHash(labelhash);
+      let healedLabel = null;
+
+      // if healing label from reverse addresses is enabled,
```
lightwalker-eth, line 93: Hmm... I'm pretty sure that when it comes to `Registrar.ts` it will never be true that `isReverseRootNode(node)`. That's because `Registrar.ts` is only for the registrars for direct subnames of .eth, linea.eth, or base.eth.

I believe this logic should only live inside `Registry.ts` -- the registry is triggered for all subname registrations, including for addr.reverse subname registrations.

apps/ensindexer/src/plugins/eth/ponder.plugin.ts
```diff
@@ -6,16 +7,27 @@ import {
...
 const ownedName = "eth" as const;
 
+// namehash('addr.reverse')
+const reverseRootNode: Node = "0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2";
+
 const { chain, contracts } = DEPLOYMENT_CONFIG[pluginName];
 const namespace = createPluginNamespace(ownedName);
 
+// The `eth` plugin supports node checking against its own reverse root node
+const isReverseRootNode = (node: Node): boolean => node === reverseRootNode;
```
lightwalker-eth, line 25: Hmm.. worried about case sensitivity issues? Maybe we should convert both to lowercase before we compare for equality? Is it guaranteed these hex strings will always be all lowercase?

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
 
 export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([node, label]));
 
+// normalize address to match the reverse address format
+// as per https://docs.ens.domains/resolution/names#reverse-nodes
+const normalizedAddressDigits = (address: Hex): string => address.slice(2).toLowerCase();
+
+export interface LabelByReverseAddressArgs {
+  /** The reverse address */
+  reverseAddress: Hex;
```
lightwalker-eth, line 30:
```suggestion
  maybeReverseAddress: Hex;
```

Suggesting this to highlight how this maybe is the reverse address, but it's not guaranteed.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
 
 export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([node, label]));
 
+// normalize address to match the reverse address format
+// as per https://docs.ens.domains/resolution/names#reverse-nodes
+const normalizedAddressDigits = (address: Hex): string => address.slice(2).toLowerCase();
+
+export interface LabelByReverseAddressArgs {
+  /** The reverse address */
```
lightwalker-eth, line 29:
```suggestion
  /** The address that is possibly associated with the addr.reverse subname */
```

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
...
+  reverseAddress: Hex;
+
+  /** The domain's labelhash */
+  labelhash: Labelhash;
+}
+
+/**
+ * Attempt to heal a labelhash to its original label using reverse address.
```
lightwalker-eth, line 37:
```suggestion
 * Attempt to heal the labelhash of an addr.reverse subname using an address that might be related to the subname.
```

We don't perform this healing based on "reverse address". We do it based on "an address that might be related to the subname"

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
...
+export const labelByReverseAddress = (args: LabelByReverseAddressArgs) => {
+  // check if required arguments are valid
+  validateLabelByReverseAddressArgs(args);
+
+  // derive the assumed label from the normalized address
+  const assumedLabel = normalizedAddressDigits(args.reverseAddress);
+
+  // if labelhash of the assumed label matches the provided labelhash
+  if (labelhash(assumedLabel) === args.labelhash) {
+    // return original label found for the labelhash
```
lightwalker-eth, line 53:
```suggestion
    // the assumedLabel successfully heals the labelhash
```

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
...
+    return assumedLabel;
+  }
+
+  // otherwise, healing did not succeed
+  throw new Error(
```
lightwalker-eth, line 58: Noting the comments at the top of this function:

> @returns the original label if healed, otherwise null

I believe that is correct. This should return `null` if it fails, rather than throwing an error. Logic that calls this function should be appropriately updated.

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
...
+    `Failed to heal label for by '${args.reverseAddress}' reverse address for labelhash: '${args.labelhash}'`,
+  );
+};
+
+/**
+ * Validate the arguments for `labelByReverseAddress` function.
+ *
+ * @throws if reverse node subname is not derived from a valid hex address
+ * @throws if labelhash is not a valid hash
```
lightwalker-eth, line 67:
```suggestion
 * @throws if labelhash is not a valid Labelhash
```

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,65 @@ export const ROOT_NODE = namehash("");
...
+  // otherwise, healing did not succeed
+  throw new Error(
+    `Failed to heal label for by '${args.reverseAddress}' reverse address for labelhash: '${args.labelhash}'`,
+  );
+};
+
+/**
+ * Validate the arguments for `labelByReverseAddress` function.
+ *
+ * @throws if reverse node subname is not derived from a valid hex address
```
lightwalker-eth, line 66:
```suggestion
 * @throws if maybeReverseAddress is not a valid Address
```

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -75,3 +77,55 @@ describe("makeSubnodeNamehash", () => {
     );
   });
 });
+
+describe("labelByReverseAddress", () => {
+  const vitalikEthResolvedAddress = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045";
+  const vitalikEthNormalizedAddress = "d8da6bf26964af9d7eed9e03e53415d37aa96045";
+  // `namehash('addr.reverse')`
```
lightwalker-eth, line 84:
```suggestion

```

Seems we should remove this comment?

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -75,3 +77,55 @@ describe("makeSubnodeNamehash", () => {
...
+
+  const validArgs = {
+    // labelhash for `d8da6bf26964af9d7eed9e03e53415d37aa96045`
+    labelhash: "0x535bdae9bb214b3cc583b53384464999f2f7f48625f160728c63e73e766ff71e",
```
lightwalker-eth, line 88:
```suggestion
    labelhash: labelhash(vitalikEthNormalizedAddress),
```

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -75,3 +77,55 @@ describe("makeSubnodeNamehash", () => {
...
+        labelByReverseAddress({
+          ...validArgs,
+          reverseAddress: "0x123",
+        }),
+      ).toThrowError(
+        "Invalid reverse address: '0x123'. Must a valid EVM address, start with '0x' and be 42 characters long",
+      );
+    });
+
+    it("should throw if labelhash is not a valid hash", () => {
```
lightwalker-eth, line 104:
```suggestion
    it("should throw if labelhash is not a valid Labelhash", () => {
```

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -75,3 +77,55 @@ describe("makeSubnodeNamehash", () => {
...
+
+  describe("label healing", () => {
+    it("should return null if the label cannot be healed", () => {
+      const notMatchingLabelhash = labelhash("test.eth");
+      expect(() =>
+        labelByReverseAddress({
+          ...validArgs,
+          labelhash: notMatchingLabelhash,
+        }),
+      ).toThrowError(
```
lightwalker-eth, line 122: This should return null, not throw an error.

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -75,3 +77,55 @@ describe("makeSubnodeNamehash", () => {
     );
   });
 });
+
+describe("labelByReverseAddress", () => {
+  const vitalikEthResolvedAddress = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045";
```
lightwalker-eth, line 82:
```suggestion
  const address = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045";
```

Suggest to keep it simple and more generic.

packages/ensnode-utils/src/subname-helpers.spec.ts
```diff
@@ -75,3 +77,55 @@ describe("makeSubnodeNamehash", () => {
...
+  const vitalikEthNormalizedAddress = "d8da6bf26964af9d7eed9e03e53415d37aa96045";
```
lightwalker-eth, line 83:
```suggestion
  const reverseAddressSubname = "d8da6bf26964af9d7eed9e03e53415d37aa96045";
```

apps/ensindexer/src/handlers/Registrar.ts
tk-o: Now it makes sense to me. When I was testing database state and pulling all nodes from reverse `domains` that couldn't be healed with our new method, I noticed there were no `name_registered` records for those nodes. All those nodes were found in `new_owners` table, though.

`Registrar.ts` writes to `name_registered` table, while `Registry.ts` writes to `new_owners` table.

apps/ensindexer/src/handlers/Registry.ts
tk-o: Will be done as part of Post-merge checklist 👍

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -172,7 +173,24 @@ export interface PonderENSPlugin<PLUGIN_NAME extends PluginName, CONFIG> {
...
+   */
+  canHealReverseAddresses(): boolean;
+
+  /**
+   * Determines whether the plugin supports checking a node against its reverse root node.
```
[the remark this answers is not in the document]
tk-o: I'm not sure I fully understand, so I'll ask some questions below.

#### Question 1

Should I just update comments added for newly introduced methods? And, for example, say:
```
Determines whether the chain supports checking a node against its reverse root node.
```

#### Question 2

Should I introduce a nested structure to capture newly introduced methods? For example:
```ts
/** Defines chain support for the "healing subnames of addr.reverse" feature
chain: {
  /**
   * Determines whether healing subnames of addr.reverse is supported.
   * Note: Some chains might not support it at the moment.
   *
   * @returns true if feature is supported
   */
  canHealReverseAddresses(): boolean;

  /**
   * Determines whether checking a node against its reverse root node.
   *
   * @param node the node to check
   * @returns true if the node is a reverse root node
   */
  isReverseRootNode(node: Node): boolean;
}
```
#### Question 3

Should we rename the `PonderENSPluginHandlerArgs` type to not indicate to a plugin, but instead to a chain? For example:
```ts
export type PonderENSPluginHandlerArgs<OWNED_NAME extends OwnedName> = {
```
would become:
```ts
export type PonderENSChainHandlerArgs<OWNED_NAME extends OwnedName> = {
```

apps/ensindexer/src/plugins/eth/ponder.plugin.ts
tk-o: This comment makes me thinking about [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) method.

In case of TypeScript, we could start defining primitive types as classes. We'd only create an instance of such class, when parsing was successful and we could be sure in all places we use given type, that it carries exactly the value it should (both format and contents).

For example:
```ts
namespace ENSNodeSdk {
  // defining primitive types
  export class Node {
    private constructor(
      public readonly value: string,
    ) {}

    static parse(maybeNodeValue: string): Node {
      if (maybeNodeValue.startsWith("0x") === false) {
        throw new Error(`Invalid node value: ${maybeNodeValue}`);
      }

      if (maybeNodeValue.length !== 66) {
        throw new Error(`Invalid node value: ${maybeNodeValue}`);
      }

      return new Node(maybeNodeValue.toLowerCase());
    }
  }
}

// using primitive types
const reverseRootNode: ENSNodeSdk.Node = ENSNodeSdk.Node.parse("0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2");

function isReverseRootNode(node: ENSNodeSdk.Node): boolean {
  return node.value === reverseRootNode.value;
}
```

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,63 @@ export const ROOT_NODE = namehash("");
...
+/**
+ * Validate the arguments for `labelByReverseAddress` function.
+ *
+ * @throws if maybeReverseAddress is not a valid Address
+ * @throws if labelhash is not a valid Labelhash
+ */
+const validateLabelByReverseAddressArgs = (args: LabelByReverseAddressArgs) => {
+  if (!isHex(args.maybeReverseAddress) || args.maybeReverseAddress.length !== 42) {
+    throw new Error(
+      `Invalid reverse address: '${args.maybeReverseAddress}'. Must a valid EVM address, start with '0x' and be 42 characters long.`,
```
djstrong, line 70:
```suggestion
      `Invalid reverse address: '${args.maybeReverseAddress}'. Must be a valid EVM address, start with '0x' and be 42 characters long.`,
```

lightwalker-eth (changes requested):
@tk-o Thanks for updates. Reviewed and shared feedback!

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,63 @@ export const ROOT_NODE = namehash("");
...
+ * @throws if labelhash is not a valid Labelhash
+ */
+const validateLabelByReverseAddressArgs = (args: LabelByReverseAddressArgs) => {
+  if (!isHex(args.maybeReverseAddress) || args.maybeReverseAddress.length !== 42) {
+    throw new Error(
+      `Invalid reverse address: '${args.maybeReverseAddress}'. Must be a valid EVM address, start with '0x' and be 42 characters long.`,
+    );
+  }
+
+  if (!isHex(args.labelhash) || args.labelhash.length !== 66) {
```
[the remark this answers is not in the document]
shrugs: unnecessary, using `isHash`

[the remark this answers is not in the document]
shrugs: resolved by `isHash`

apps/ensindexer/src/plugins/eth/ponder.plugin.ts
```diff
@@ -7,16 +8,28 @@ import {
...
+// namehash('addr.reverse')
+const reverseRootNode: Node = "0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2";
+
 const { chain, contracts } = DEPLOYMENT_CONFIG[pluginName];
 const namespace = createPluginNamespace(ownedName);
 
+// `eth` plugin can heal reverse addresses from its reverse root node
+const canHealReverseAddressFromParentNode = (parentNode: Node): boolean => {
+  const isReverseRootNode = (maybeReverseRootNode: Node): boolean =>
+    maybeReverseRootNode.toLowerCase() === reverseRootNode.toLowerCase();
```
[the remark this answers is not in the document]
shrugs: the complexity here was unwarranted, changed to simple equality check, and helper not necessary imo

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,63 @@ export const ROOT_NODE = namehash("");
 
 export const makeSubnodeNamehash = (node: Hex, label: Hex) => keccak256(concat([node, label]));
```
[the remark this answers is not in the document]
shrugs: will be resolved in https://github.com/namehash/ensnode/pull/532

packages/ensnode-utils/src/constants.ts
```diff
@@ -0,0 +1,12 @@
+import { namehash } from "viem";
+import { Node } from "./types";
+
+export const ROOT_NODE: Node = namehash("");
+
+/**
+ * A set of known reverse nodes.
```
lightwalker-eth, line 7:
```suggestion
 * A set of nodes whose children are used for reverse resolution.
```

packages/ensnode-utils/src/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * Useful for filtering new domains by whether they're a reverse node for a specific address.
```
lightwalker-eth, line 9:
```suggestion
 * Useful for identifying if a domain is used for reverse resolution.
```

lightwalker-eth (commented):
Small language refinement suggested

lightwalker-eth (approved):

apps/ensindexer/.env.local.example
```diff
@@ -62,3 +62,11 @@ ENSNODE_PUBLIC_URL=http://localhost:42069
...
 # If this is not set, DEFAULT_ENSADMIN_URL will be used to provide easy access to an ENSAdmin UI.
 ENSADMIN_URL=https://admin.ensnode.io
+
+# A feature flag to enable or disable healing of addr.reverse subnames
+# If this is set to true, ENSIndexer will attempt to heal subnames of addr.reverse
+# If this is not set, the default value is set to `DEFAULT_HEAL_REVERSE_ADDRESSES`.
+#
+# WARNING: Setting this to `true` results in indexed data no longer being backwards compatible with the ENS Subgraph. For full
+# data-level backwards compatibility with the ENS Subgraph, set this to `false`.
+HEAL_REVERSE_ADDRESSES=true
```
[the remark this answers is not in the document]
shrugs: https://github.com/namehash/ensnode/issues/531

packages/ensnode-utils/src/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const ROOT_NODE: Node = namehash("");
+
+/**
+ * A set of nodes whose children are used for reverse resolution.
+ *
+ * Useful for identifying if a domain is used for reverse resolution.
+ * See apps/ensindexer/src/handlers/Registry.ts for context.
+ */
+export const REVERSE_ROOT_NODES: Set<Node> = new Set([namehash("addr.reverse")]);
```
tk-o, line 12: This will enable healing reverse addresses for both, `eth` plugin, and `linea.eth` plugin.

The goal of this PR was to only enable healing process for `eth` plugin.

For context:
- https://github.com/Consensys/linea-ens/issues/287

apps/ensindexer/src/handlers/Registry.ts
```diff
@@ -123,9 +131,24 @@ export const makeRegistryHandlers = (ownedName: OwnedName) => {
...
           const parent = await context.db.find(schema.domain, { id: node });
 
-          // attempt to heal the label associated with labelhash via ENSRainbow
-          // https://github.com/ensdomains/ens-subgraph/blob/c68a889/src/ensRegistry.ts#L112-L116
-          const healedLabel = await labelByHash(labelhash);
+          let healedLabel = null;
+
+          // if healing label from reverse addresses is possible, give it a go
+          if (canHealReverseAddressFromParentNode(node)) {
+            // TODO: if healing failed, log the event args for analysis and debugging
```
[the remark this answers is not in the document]
tk-o: - https://github.com/namehash/ensnode/issues/545

packages/ensnode-utils/src/subname-helpers.ts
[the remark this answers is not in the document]
tk-o: @shrugs is that change going to be covered in your PR?
- https://github.com/namehash/ensnode/pull/532

packages/ensnode-utils/src/subname-helpers.ts
```diff
@@ -17,6 +21,63 @@ export const ROOT_NODE = namehash("");
...
 
+// normalize address to match the format used for addr.reverse subnames
+// as per https://docs.ens.domains/resolution/names#reverse-nodes
+const normalizedAddressDigits = (address: Hex): string => address.slice(2).toLowerCase();
+
+export interface LabelByReverseAddressArgs {
+  /** The address that is possibly associated with the addr.reverse subname */
+  maybeReverseAddress: Hex;
```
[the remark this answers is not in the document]
tk-o: @shrugs is that change going to be covered in your PR?
- https://github.com/namehash/ensnode/pull/532

packages/ensnode-utils/src/constants.ts
tk-o: Logged with:
- https://github.com/namehash/ensnode/issues/616

PR #365, opened 2025-03-12
fix(deps): dedupe dependency overrides

tk-o:
This PR fixes duplicated overrides in NPM file that were causing duplicated entries in PNPM lockfile, resulting in failed `pnpm install --frozen-lockfile` command.


lightwalker-eth (approved):
Looks good! Thanks!

PR #373, opened 2025-03-12
feat: sidebar shadcn layout and breadcrumbs

notrab:
There are a few remaining items that need done in this PR, or a follow-up one:

1. Fix the `Ponder` default breadcrumb
2. Move the GraphQL API endpoint to the breadcrumb section:
CleanShot 2025-03-12 at 22 13 12@2x
3. Add ENSAdmin logo to connection selector


apps/ensadmin-next/src/app/@actions/default.tsx
```diff
@@ -0,0 +1,3 @@
+export default function ActionsDefault() {
```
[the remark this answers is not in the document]
notrab: We need to make suree the parallel route matches to a component that returns `null`. Otherwise we'll get a 404.

apps/ensadmin-next/src/app/gql/ponder/page.tsx
```diff
@@ -1,13 +1,22 @@
...
+export default async function PonderGraphQLPage({ searchParams }: PageProps) {
+  const { ensnode = preferredEnsNodeUrl() } = await searchParams;
+
+  const baseUrl = Array.isArray(ensnode)
+    ? ensnode[0]
+    : typeof ensnode === "string"
+      ? ensnode
+      : preferredEnsNodeUrl();
+
+  const url = new URL(`/ponder`, baseUrl).toString();
```
tk-o, line 19: Same question as here:
https://github.com/namehash/ensnode/pull/373/files#r1995246905

apps/ensadmin-next/src/app/ponder-client-api/page.tsx
```diff
@@ -1,13 +1,49 @@
...
+    ? ensnode[0]
+    : typeof ensnode === "string"
+      ? ensnode
+      : preferredEnsNodeUrl();
+
+  const url = new URL(`/subgraph`, baseUrl).toString();
+
+  return (
+    <Suspense fallback={<Loading />}>
+      <PonderClientProvider url={url}>
```
tk-o, line 26: I love how you moved the `PonderClientProvider` down the components tree. Previously, it'd require almost the whole tree to be made of client components, which was not good for page transitions smoothness.

apps/ensadmin-next/src/app/ponder-client-api/provider.tsx
```diff
@@ -4,16 +4,15 @@ import { Provider as QueryClientProvider } from "@/components/query-client/provi
...
+  url: string;
+  children: React.ReactNode;
+};
 
-export function Provider({ children }: PropsWithChildren) {
-  const searchParams = useSearchParams();
-  const [ponderClient] = useState(() =>
-    createPonderClient(selectedEnsNodeUrl(searchParams), schema),
-  );
+export function Provider({ url, children }: ProviderProps) {
```
tk-o, line 14: Passing URL as a prop, very nice improvement over having it calculated internally with search params.

apps/ensadmin-next/src/components/connections/connection-selector.tsx
```diff
@@ -136,81 +138,87 @@ export function ConnectionSelector() {
...
-                  className="px-2 py-1 text-xs text-muted-foreground hover:text-foreground disabled:opacity-50"
-                >
-                  Cancel
-                </button>
-                <button
-                  onClick={handleAdd}
+            <DropdownMenuSeparator />
+
+            {isAdding ? (
+              <div className="px-3 py-2 space-y-2">
```
tk-o, line 148: I was thinking of changing this one to `form` tag, so we could add `onSubmit` and `onReset` callbacks, and let the "Cancel" button be of `type="reset"` and "Add" button be of `type="submit"`.

With that change, we'd get a native form UX, where hitting `<enter>` while focus within any element of a form, submits the form right away.

notrab: Yeah, I didn't touch the functionality here. I want to move this into a new complete slideover component in a follow up PR.

apps/ensadmin-next/src/app/@actions/default.tsx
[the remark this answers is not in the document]
tk-o: TIL

notrab:
Spoke to @tk-o regarding the comments above and agree that the refactoring can be done separately. I will work on this next.

apps/ensadmin-next/src/app/@actions/gql/subgraph-compat/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export default async function ActionsSubgraphCompatPage({ searchParams }: ActionProps) {
+  const { ensnode = preferredEnsNodeUrl() } = await searchParams;
+
+  const baseUrl = Array.isArray(ensnode)
+    ? ensnode[0]
+    : typeof ensnode === "string"
+      ? ensnode
+      : preferredEnsNodeUrl();
```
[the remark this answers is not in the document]
notrab: @tk-o I think we should make use of the Next.js `searchParams` so we can hook into the loading states given to us out of the box by `await`ing the search params promise. We likely can `useParams` too in other client components so you don't need to manually rewrite what Next.js already has.

Lmkwyt

apps/ensadmin-next/src/components/ui/breadcrumb.tsx
```diff
@@ -0,0 +1,101 @@
+import { Slot } from "@radix-ui/react-slot";
+import { ChevronRight, MoreHorizontal } from "lucide-react";
+import * as React from "react";
+
+import { cn } from "@/lib/utils";
+
+const Breadcrumb = React.forwardRef<
```
lightwalker-eth, line 7: @notrab Sorry I'm sure I'm not leaving this comment on the right place. I'm not sure where to put it.

CleanShot 2025-03-17 at 16 00 06

I'm a little worried about people clicking "APIs" in this breadcrumb which then ultimately results in them being redirected to `/status` from my testing.

Likewise, I see that "APIs" are not directly navigable from the main left nav bar as clicking "APIs" in the main left nav bar is only for toggling the submenu open or collapsed.

This seems to create a UX inconsistency between breadcrumbs / main nav bar.

Is it an easy fix to make it so that "APIs" in the breadcrumb isn't clickable? If there isn't a super quick solution to this then no worries just let me know and we can come back to it another time.

Appreciate your advice. Thanks!

apps/ensadmin-next/src/components/app-sidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+const navItems = [
+  {
+    title: "Status",
+    url: "/status",
+    icon: Activity,
+  },
+  {
+    title: "APIs",
+    url: "#",
+    icon: DatabaseIcon,
```
lightwalker-eth, line 20: Suggest we change this to the `radio-tower` icon:


CleanShot 2025-03-17 at 16 12 01

Or the `radio` icon:

CleanShot 2025-03-17 at 16 12 24


Taking these icons from https://lucide.dev/icons/ where it looks like you're already importing from.

Also open to other suggestions, but main idea is to move away from a `DatabaseIcon` for our APIs.

apps/ensadmin-next/src/components/app-sidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+    isActive: true,
+    items: [
+      {
+        title: "GraphQL",
+        url: "/gql/ponder",
+      },
+      {
+        title: "Subgraph",
```
lightwalker-eth, line 28:
```suggestion
        title: "GraphQL (Subgraph-style)",
```

Trying to find a way to communicate / differentiate the two GraphQL endpoints. Shared my suggestion in 2 comments here.

apps/ensadmin-next/src/components/app-sidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+    icon: Activity,
+  },
+  {
+    title: "APIs",
+    url: "#",
+    icon: DatabaseIcon,
+    isActive: true,
+    items: [
+      {
+        title: "GraphQL",
```
lightwalker-eth, line 24:
```suggestion
        title: "GraphQL (Ponder-style)",
```

apps/ensadmin-next/src/components/app-sidebar.tsx
```diff
@@ -0,0 +1,51 @@
...
+        url: "/gql/ponder",
+      },
+      {
+        title: "Subgraph",
+        url: "/gql/subgraph-compat",
+      },
+      {
+        title: "Ponder",
```
lightwalker-eth, line 32:
```suggestion
        title: "Ponder Client",
```

This isn't a "Ponder" API. Ponder supports any arbitrary APIs and it looks like we'll be making some soon. This is a very specific "Ponder Client" API. It's a very specific API type.

tk-o (approved):
As discussed, all feedback / comments will be applied in a follow up PR 👍

apps/ensadmin-next/src/app/@actions/gql/ponder/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  const baseUrl = Array.isArray(ensnode)
+    ? ensnode[0]
+    : typeof ensnode === "string"
+      ? ensnode
+      : preferredEnsNodeUrl();
+
+  const url = new URL(`/ponder`, baseUrl).toString();
+
+  return (
```
lightwalker-eth, line 23: CleanShot 2025-03-17 at 16 42 52

It would be nice to enhance a few UX details here, including:


1. Showing the full endpoint (not truncated).
2. Prefixing the endpoint with "Endpoint:" or "API endpoint:" so its more clear what the value being displayed here represents.
3. Making the copy to clipboard function operational.

apps/ensadmin-next/src/app/@actions/gql/ponder/page.tsx
```diff
@@ -0,0 +1,31 @@
+import { Button } from "@/components/ui/button";
+import { Input } from "@/components/ui/input";
+import { preferredEnsNodeUrl } from "@/lib/env";
```
lightwalker-eth, line 3: There's a bunch of places in the code / env vars / docs for ENSAdmin where we use the word "preferred". In all of those contexts, shouldn't we use the word "default" instead?

apps/ensadmin-next/src/app/ponder-client-api/page.tsx
```diff
@@ -1,13 +1,49 @@
...
+export default async function PonderClientPage({ searchParams }: PageProps) {
+  const { ensnode = preferredEnsNodeUrl() } = await searchParams;
+
+  const baseUrl = Array.isArray(ensnode)
+    ? ensnode[0]
+    : typeof ensnode === "string"
+      ? ensnode
+      : preferredEnsNodeUrl();
+
+  const url = new URL(`/subgraph`, baseUrl).toString();
```
lightwalker-eth, line 22: I see that this file is for the `ponder-client-api` and therefore find it surprising to see a `/subgraph` route in the code here.

@tk-o I think you're familiar with some of the related code here too? Appreciate your suggestions.

lightwalker-eth (approved):
@notrab Huge upgrades here. Great work ! 🚀 I've shared some feedback, appreciate if we can action that feedback in a separate future PR. Let's merge this!

apps/ensadmin-next/src/components/ui/breadcrumb.tsx
notrab: @lightwalker-eth yeah I had the same thoughts when implementing this, but forgot to make it not a link. I'll do just that.

apps/ensadmin-next/src/app/ponder-client-api/page.tsx
tk-o: `PonderClientProvider` should accept the `ensNodeUrl` as a param. Internally, the provider component sets `/sql` path on the URL object (so `/subgraph` path gets overridden anyway.

apps/ensadmin-next/src/app/ponder-client-api/page.tsx
tk-o, line 26:
```suggestion
      <PonderClientProvider ensNodeUrl={baseUrl}>
```

PR #381, opened 2025-03-13
ENSAdmin: show primary names and avatars of owners

djstrong:


apps/ensadmin-next/src/components/ens-name/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
+import { Skeleton } from "@/components/ui/skeleton";
+import { CHAIN_IDS } from "@/lib/wagmi";
+import { ExternalLink } from "lucide-react";
+import { useEffect, useState } from "react";
+import { useEnsName } from "wagmi";
+
+interface ENSNameProps {
+  address: string;
```
lightwalker-eth, line 11:
```suggestion
  address: `0x${string}`;
```

apps/ensadmin-next/src/components/ens-name/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+  address,
+  showAvatar = false,
+  showExternalLink = true,
+  className = "",
+}: ENSNameProps) {
+  const [mounted, setMounted] = useState(false);
+
+  // Use the ENS name hook from wagmi
+  const { data: ensName, isLoading } = useEnsName({
+    address: address as `0x${string}`,
```
lightwalker-eth, line 31:
```suggestion
    address,
```

apps/ensadmin-next/src/components/ens-name/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    chainId: CHAIN_IDS.MAINNET,
```
lightwalker-eth, line 32: I believe here we need to pass in the ChainId of the ENS deployment that is being indexed by the attached ENSIndexer instance.

ENSAdmin should be able to get this through the `/metadata` endpoint I think?

@tk-o Suggest you take the lead on suggesting how we action this?

Consider the following goals:
* If the attached ENSIndexer is using the mainnet ENS Deployment: ENSAdmin performs this lookup on mainnet.
* If the attached ENSIndexer is using the sepolia ENS Deployment: ENSAdmin performs this lookup on sepolia.
* If the attached ENSIndexer is using the holesky ENS Deployment: ENSAdmin performs this lookup on holesky.

apps/ensadmin-next/README.md
```diff
@@ -18,6 +18,16 @@ cp .env.local.example .env.local
 
 You can update `PREFERRED_ENSNODE_URL` environment variable if you wish ENSAdmin to include a given URL as an initial connection option.
 
+#### RPC URLs for ENS Name Resolution
```
lightwalker-eth, line 21:
```suggestion
#### RPC URLs
```

These will be used for more than just ENS Name Resolution

apps/ensadmin-next/README.md
```diff
@@ -18,6 +18,16 @@ cp .env.local.example .env.local
...
+
+ENSAdmin uses [wagmi](https://wagmi.sh/) to resolve primary ENS names for Ethereum addresses. You need to provide RPC URLs for each chain where ENS is deployed:
```
lightwalker-eth, line 23:
```suggestion
ENSAdmin may use RPC URLs for each chain with a "root" ENS Deployment:
```

These will be used by more than just wagmi or for the resolution of primary ENS names.

apps/ensadmin-next/.env.local.example
```diff
@@ -0,0 +1,13 @@
+# ENSNode URL
```
lightwalker-eth, line 1:
```suggestion
# Preferred ENSNode URL (used by default when ENSAdmin is opened)
```

lightwalker-eth (approved):
@djstrong Awesome! 🚀 Just 1 main issue that I think we need @tk-o to help with. Otherwise I shared a few small suggestions :+1:

apps/ensadmin-next/src/components/ens-name/index.tsx
tk-o: Resolved with #387

PR #394, opened 2025-03-14
fix: update Open Graph and Twitter image URLs

djstrong:


lightwalker-eth (approved):
@djstrong Thanks! Let's try it!

PR #395, opened 2025-03-15
fix: correctly parse args where null bytes values may be encountered

shrugs:
- [x] `NameRegistered` handler wasn't catching null bytes correctly because of ponder stripping them — we fixed in preimage but not at this callsite.
- [x] updates all other callsites with manual arg parsing as well, removes the labelhash comparison in `setNamePreimage`
- [x] need to take a snapshot of mainnet again to make sure these edits keep the same logic, so waiting on that before merging


lightwalker-eth (approved):
@shrugs great work 🙌

PR #399, opened 2025-03-17
fix(ci): include node-gyp deps in Docker image

tk-o:
These are required for `node-gyp` to successfully complete when being triggered via `postinstall` task.


lightwalker-eth (approved):
@tk-o Thanks, let's try it!

PR #409, opened 2025-03-17
feat(recent-registrations): add date handling for latest registrations display

djstrong:


apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -160,6 +172,14 @@ export function RecentRegistrations() {
         .currentIndexingDate
```
lightwalker-eth, line 172: Suggest we base this logic based on a specific block number, rather than on a timestamp.

For this, perhaps @tk-o can help to suggest what block number we should use and also how to access the most recently indexed block in code here.

I imagine we can find the specific block number we want somewhere in the `ens-deployments` package? Bonus points if we import the value we want from that package, rather than redefining it here.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -184,7 +204,16 @@ export function RecentRegistrations() {
         </CardTitle>
       </CardHeader>
       <CardContent>
-        {recentRegistrationsQuery.isLoading ? (
+        {isBeforeRegistrationsStartingDate ? (
+          <div className="py-4 text-center">
+            <p className="text-muted-foreground mb-2">
+              This component displays registrations starting from {formattedStartingDate}.
```
lightwalker-eth, line 210:
```suggestion
              Latest indexed .eth registrations will be displayed here after blocks from {formattedStartingDate} are indexed.
```

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -184,7 +204,16 @@ export function RecentRegistrations() {
...
+            </p>
+            <p className="text-sm text-muted-foreground">
+              The indexed data available is from before this date.
```
lightwalker-eth, line 213:
```suggestion
              While .eth domains domains are indexed before this date, .eth registrations are not.
```

lightwalker-eth (changes requested):
@djstrong Hey nice! Good job here! There's a few special details on this PR that we'll need @tk-o to help with tomorrow. @tk-o Can you please help with the feedback I shared?

apps/ensadmin/src/components/recent-registrations/components.tsx
tk-o: The BaseRegistrar contract was deployed on mainnet on block `9380410`. We have [this block listed in our ENS Deployment configuration](https://github.com/namehash/ensnode/blob/7a3df57/packages/ens-deployments/src/mainnet.ts#L61).

However, the very first registration happened on block `9380428`, which is roughly 4 minutes after the contract was deployed. [Here’s a query](https://api.thegraph.com/subgraphs/name/ensdomains/ens/graphql?query=%7B%0A++%23+take+ID+of+the+very+first+registration%0A++registrations%28orderBy%3A+registrationDate%2C+orderDirection%3A+asc%2C+first%3A+1%29+%7B%0A++++id%0A++++labelName%0A++%7D%0A++%0A++%23+find+the+block+details%0A++nameRegistereds%28where%3A%7B%0A++++registration_%3A+%7B%0A++++++id%3A%220x7926fb9208a20417f0b848065c6fe289da455821d6fc0c117845ad8d5b331d2b%22%0A++++%7D%0A++%7D%29+%7B%0A++++blockNumber%0A++++transactionID%0A++++registration+%7B%0A++++++id%0A++++++labelName%0A++++%7D%0A++%7D%0A%7D) that got me this block number.

shrugs:
update: depends on #436

PR #414, opened 2025-03-18
fix(ensadmin): check registration start block

tk-o:
Get the right block number based on ENS deployment chain config. Also, starts support for listing registrations from Sepolia and Holesky.

Test links:
- [Mainnet](https://adminensnodeio-git-fix-use-registration-start-block-namehash.vercel.app/status?ensnode=https://api.mainnet.blue.ensnode.io/)
- [Sepolia](https://adminensnodeio-git-fix-use-registration-start-block-namehash.vercel.app/status?ensnode=https://api.sepolia.blue.ensnode.io/)
- [Holesky](https://adminensnodeio-git-fix-use-registration-start-block-namehash.vercel.app/status?ensnode=https://api.holesky.blue.ensnode.io/)


apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -65,3 +69,86 @@ export function useRecentRegistrations(searchParams: URLSearchParams) {
     },
   });
 }
+
+export function useRecentRegistrationsViaPonder(skipChildrenOfDomains: Array<string> = []) {
```
tk-o, line 73: Please note this hook didn't work as expected. It was lacking the`domainOwnerId` value in results, even thought the network response included correct values. It might be something with Ponder Client.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
+import { holesky, mainnet, sepolia } from "viem/chains";
+
+/**
+ * Get ENS app URL for wallet address on a supported network
+ * @param chainId
+ * @param nameOrAddress
+ * @returns ENS app URL for supported networks, otherwise undefined
+ */
+export function getEnsAppUrl(
+  chainId: number | undefined,
```
lightwalker-eth, line 10: We should make some changes here:

1. We should remove the option to pass `undefined` here.
2. We should take an ens deployment id as defined in the `ens-deployments` package. Each of the different apps here are not based on chains. They are based more fundamentally on the key enum we defined in `ens-deployments.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Get avatar URL for given ENS name on a supported network
+ * @param chainId
+ * @param ensName
+ * @returns avatar URL for supported networks, otherwise undefined
+ */
+export function getEnsAvatarUrl(
+  chainId: number | undefined,
+  ensName: string | undefined,
```
lightwalker-eth, line 33: We shouldn't write our functions such that they accept undefined params like this unless there's a very special reason to do that.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/**
+ * Get avatar URL for given ENS name on a supported network
+ * @param chainId
+ * @param ensName
+ * @returns avatar URL for supported networks, otherwise undefined
+ */
+export function getEnsAvatarUrl(
+  chainId: number | undefined,
```
lightwalker-eth, line 32: We should make some changes here:

1. We should remove the option to pass undefined here.
2. We should take an ens deployment id as defined in the ens-deployments package. Each of the different ens metadata service API endpoints here are not based on chains. They are based more fundamentally on the key enum we defined in `ens-deployments.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  switch (chainId) {
+    case mainnet.id:
+      return `https://metadata.ens.domains/mainnet/avatar/${ensName}`;
+    case sepolia.id:
+      return `https://metadata.ens.domains/sepolia/avatar/${ensName}`;
+    case holesky.id:
+      // NOTE: there's no metadata api available for holesky network
+      return undefined;
+    default:
```
lightwalker-eth, line 47: We should add a case here for all values defined in the `ens-deployments` package. This will remove any need to define the `default` case.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  switch (chainId) {
+    case mainnet.id:
+      return `https://metadata.ens.domains/mainnet/avatar/${ensName}`;
+    case sepolia.id:
+      return `https://metadata.ens.domains/sepolia/avatar/${ensName}`;
+    case holesky.id:
+      // NOTE: there's no metadata api available for holesky network
+      return undefined;
```
lightwalker-eth, line 46: Suggest we return `null` for cases like this, rather than `undefined`. `null` is better based on how it is more explicitly not there, rather than just being "undefined".

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+): string | undefined {
+  if (!ensName) {
+    return undefined;
+  }
+
+  switch (chainId) {
+    case mainnet.id:
+      return `https://metadata.ens.domains/mainnet/avatar/${ensName}`;
+    case sepolia.id:
+      return `https://metadata.ens.domains/sepolia/avatar/${ensName}`;
```
lightwalker-eth, line 43: Not sure, but I think you may need to URL encode the `ensName` in these routes.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Get ENS app URL for wallet address on a supported network
+ * @param chainId
+ * @param nameOrAddress
+ * @returns ENS app URL for supported networks, otherwise undefined
+ */
+export function getEnsAppUrl(
+  chainId: number | undefined,
+  nameOrAddress: string,
```
lightwalker-eth, line 11: Suggest a few changes here:

1. Always make it an address, never a name.
2. Change the param type to one of the 0x string types.

Prefer how this is more explicit. If this function is exclusively for addresses then we don't need to urlencode the address.

apps/ensadmin/src/components/ens-name/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+): string | undefined {
+  switch (chainId) {
+    case mainnet.id:
+      return `https://app.ens.domains/${nameOrAddress}`;
+    case sepolia.id:
+      return `https://sepolia.app.ens.domains/${nameOrAddress}`;
+    case holesky.id:
+      return `https://holesky.app.ens.domains/${nameOrAddress}`;
+    default:
```
lightwalker-eth, line 20: We should add a case here for all values defined in the ens-deployments package. This will remove any need to define the default case.

lightwalker-eth, line 20: We need to add another case for anvil / the ens-test-env.

For this case we should return `null` rather than `undefined`.

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,16 +2,17 @@
...
 import { ExternalLink } from "lucide-react";
 import { useEffect, useState } from "react";
 import type { Hex } from "viem";
 import { useEnsName } from "wagmi";
+import { getEnsAppUrl, getEnsAvatarUrl } from "./helpers";
 
 interface ENSNameProps {
   address: Hex;
-  chainId: SupportedChainId;
+  chainId: SupportedEnsDeploymentChainId;
```
lightwalker-eth, line 15: This shouldn't be a chain id. It should be one of the enum values we defined in `ens-deployments`.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+      ),
+    );
+  }, [blockNumber, publicClient]);
+
+  return blockInfo;
+}
+
+interface UseIndexedNetworkBlockProps {
+  blockName: keyof NetworkIndexingStatus;
+  chainId: number | undefined;
```
lightwalker-eth, line 168: Why do we allow this to be `undefined`? Suggest making that impossible.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+  ensNodeMetadata: EnsNode.Metadata | undefined;
```
lightwalker-eth, line 169: Why do we allow this to be `undefined`? Suggest making that impossible.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+          }) satisfies BlockInfo,
+      ),
+    );
+  }, [blockNumber, publicClient]);
+
+  return blockInfo;
+}
+
+interface UseIndexedNetworkBlockProps {
+  blockName: keyof NetworkIndexingStatus;
```
lightwalker-eth, line 167: This is the "name" of a block? Is that the optimal name for this field? I'm not sure what a "block name" is.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+
+  const ensDeploymentChain = ensNodeMetadata.env.ENS_DEPLOYMENT_CHAIN;
+
+  return ensDeploymentChain === "mainnet"
+    ? DeploymentConfigs[ensDeploymentChain][subregistryName]
+    : DeploymentConfigs[ensDeploymentChain]["eth"];
+}
+
+interface UseBlockInfoProps {
+  chainId: number | undefined;
```
lightwalker-eth, line 130: Suggest not allowing this to be undefined.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+  blockNumber: number | undefined;
```
lightwalker-eth, line 131: Suggest not allowing this to be undefined.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+}
+
+/**
+ * Hook to fetch the block info from network.
+ *
+ * @returns {BlockInfo|undefined} block info or undefined if ENSMetadata is not available yet
```
lightwalker-eth, line 137: This comment makes a reference to "ENSMetadata". What's this?

And when I review the code below I don't see any reference to an "ENSMetadata"?

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+
+/**
+ * Hook to get ENS Subregistry configuration extracted from ENSNode metadata by its name.
+ *
+ * @param ensNodeMetadata
+ * @param subregistryName
+ * @returns
+ */
+export function useEnsSubregistryConfig(
+  ensNodeMetadata: EnsNode.Metadata | undefined,
```
lightwalker-eth, line 115: Suggest not allowing this to be undefined.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
+  subregistryName: SubregistryName = "eth",
+): SubregistryDeploymentConfig | undefined {
+  if (!ensNodeMetadata) {
+    return undefined;
+  }
+
+  const ensDeploymentChain = ensNodeMetadata.env.ENS_DEPLOYMENT_CHAIN;
+
+  return ensDeploymentChain === "mainnet"
```
lightwalker-eth, line 124: The logic here seems really important but I can't confidently understand what it's doing just through code review.

Appreciate if you could please write some comments for this. Thanks.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
   if (!indexingStatus) {
     return undefined;
   }
 
-  return parseSupportedChainIdByName(indexingStatus.env.ENS_DEPLOYMENT_CHAIN);
+  return parseEnsDeploymentChain(indexingStatus.env.ENS_DEPLOYMENT_CHAIN);
+}
+
+/**
+ * Hook to get ENS Subregistry configuration extracted from ENSNode metadata by its name.
```
lightwalker-eth, line 108: What is an "ENS Subregistry configuration"?

For people in the ENS community they won't know what that means. Appreciate if you could please clarify. Thanks.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -81,19 +88,111 @@ function validateResponse(response: EnsNode.Metadata) {
...
  * @param indexingStatus The ENSNode indexing status.
  * @returns The indexed chain ID or null if the status is not available.
- * @throws Error if the ENS Deployment Chain is not a supported chain.
  */
-export function useIndexedChainId(
+export function useEnsDeploymentChain(
   indexingStatus: UseIndexingStatusQueryResult["data"],
-): SupportedChainId | undefined {
+): SupportedEnsDeploymentChainId | undefined {
   // If the status is not available, return undefined
```
lightwalker-eth, line 99: What could cause the indexingStatus to not be defined? It seems we should remove any possibility for this case?

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -151,69 +140,132 @@ const NAME_WRAPPER_ADDRESS = "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401";
...
+  const ensDeploymentChainId = useEnsDeploymentChain(indexingStatus.data);
+
+  if (indexingStatus.isLoading) {
+    return <p>Loading recent registrations.</p>;
+  }
+
+  if (!ensDeploymentChainId) {
+    return null;
+  }
+
```
lightwalker-eth, line 171: Could you please ensure we gracefully handle the case that the connected ens deployment is anvil / the `ens-test-env`? Thanks!

lightwalker-eth (approved):
@tk-o Thanks for advancing some of the tricky logic here! Reviewed and shared feedback. I'm going to go ahead and approve / merge this PR now. Appreciate if you could follow up on the feedback I've shared in some other future PR :+1: Thanks!

PR #417, opened 2025-03-18
feat(ensadmin): add copy to clipboard actions for api url

notrab:
This PR adds the copy to clipboard functionality with a simple toast on success.

The toast (by sonner) can also be used across other various parts of the app now too.


lightwalker-eth (approved):
@notrab Looks super !

PR #418, opened 2025-03-18
ENSAdmin name indexing coverage

djstrong:


lightwalker-eth:
Closing PR since:

1. In the time since this PR was originally opened, we used other strategies to communicate why ENSNode is superior to the ENS Subgraph. Therefore the goals of this PR are no longer a priority.
2. This PR makes use of Ponder Client, but it's better to move away from Ponder Client API use. See https://github.com/namehash/ensnode/issues/836

PR #419, opened 2025-03-19
feat(ensadmin): new connection dialog

notrab:
I've improved the UI of adding a new connection to be inside of a Dialog component.

I also changed the `Input` to be `text` so we have more control over the validation.

A further improvement when we have React 19 would be to use some of the newer React APIs for managing form status with a server action potentially, and we can revalidate some other pages should we need to.


lightwalker-eth (approved):
@notrab Looks fantastic! Nice work 🙌

PR #429, opened 2025-03-22
fix(docker): include `ghcr.io` in image names

tk-o:
This PR:
1. Allows selectively building local docker images from source code, ie. by running any of the following commands
  - `pnpm docker:build:ensadmin` 
  - `pnpm docker:build:ensindexer` 
  - `pnpm docker:build:ensrainbow`
  - `pnpm docker:build:ensnode`
  
2. Enables auto-pulling docker images from ghcr.io registry that were not found locally. This way, a user can run all services just with the following command:
  - `docker compose up -d`

3. Aligns docker compose example file (`examples/example-docker-compose/docker-compose.yml`) with the root docker compose file (`docker-compose.yml`). Uses only the `latest` tag for service images.

### DIFF
--- a/examples/example-docker-compose/docker-compose.yml
+++ b/examples/example-docker-compose/docker-compose.yml
@@ -12,9 +12,9 @@ services:
       ENSRAINBOW_URL: http://ensrainbow:3223
       # Override ENSADMIN_URL to point to docker compose ensadmin
       ENSADMIN_URL: http://localhost:4321
-      # Override ENSNODE_PUBLIC_URL to point to docker compose ensadmin
+      # Override ENSNODE_PUBLIC_URL to point to docker compose ensindexer
+      # Note: it must be URL accessible from web browser (i.e. it cannot be a hostname in docker network)
       ENSNODE_PUBLIC_URL: http://localhost:42069
-      # NOTE: must define indexing behavior in .env.local (see .env.local.example)
     env_file:
       - ./.env.local
     depends_on:
@@ -23,7 +23,7 @@ services:
 
   ensrainbow:
     container_name: ensrainbow
-    image: ghcr.io/namehash/ensnode/ensrainbow:latest
+    image: ghcr.io/namehash/ensnode/ensrainbow:stable
     pull_policy: always
     ports:
       - "3223:3223"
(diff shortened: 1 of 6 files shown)


examples/example-docker-compose/docker-compose.yml
```diff
@@ -12,9 +12,9 @@ services:
       ENSRAINBOW_URL: http://ensrainbow:3223
       # Override ENSADMIN_URL to point to docker compose ensadmin
       ENSADMIN_URL: http://localhost:4321
-      # Override ENSNODE_PUBLIC_URL to point to docker compose ensadmin
+      # Override ENSNODE_PUBLIC_URL to point to docker compose ensindexer
+      # Note: it must be URL accessible from web browser (i.e. it cannot be a hostname in docker network)
       ENSNODE_PUBLIC_URL: http://localhost:42069
-      # NOTE: must define indexing behavior in .env.local (see .env.local.example)
```
lightwalker-eth, line 17: Can you help me understand the goal for removing this comment?

tk-o: I've moved this comment a bit down the file now, into the `env_file` section 👍

examples/example-docker-compose/docker-compose.yml
```diff
@@ -45,7 +46,9 @@ services:
       # Note: it must be URL accessible from web browser (i.e. it cannot be a hostname in docker network)
       NEXT_PUBLIC_PREFERRED_ENSNODE_URL: http://localhost:42069
     env_file:
-      - ./.env.local
+      # NOTE: can define apps/ensadmin/.env.local (see apps/ensadmin/.env.local.example)
+      - path: ./apps/ensadmin/.env.local
+        required: false
```
lightwalker-eth, line 51: Maybe best to identify this as required? I don’t think ENSAdmin can work properly otherwise. Right? If we change this I suppose we should also update the other file too?

examples/example-docker-compose/docker-compose.yml
```diff
@@ -16,25 +15,27 @@ services:
...
-      - ./.env.local
+      # NOTE: can define apps/ensrainbow/.env.local (see apps/ensrainbow/.env.local.example)
+      - path: ./apps/ensrainbow/.env.local
+        required: false
     restart: unless-stopped
 
   ensadmin:
     container_name: ensadmin
-    image: ghcr.io/namehash/ensnode/ensadmin:stable
-    pull_policy: always
```
lightwalker-eth, line 37: Just double checking we wanted to remove this?

tk-o: Yes, I deleted this as the default pull policy also downloads the image from registry:
https://github.com/compose-spec/compose-spec/blob/main/spec.md#pull_policy

examples/example-docker-compose/docker-compose.yml
tk-o: ENSAdmin can work without an env file — the only required env var is `ENSADMIN_PUBLIC_URL` and it's defined in this compose file. All other env vars have default values assigned.

Of course we can make the env file required here.

`ensrainbow` service does not required any env file at all. It has all env vars covered by default value.

lightwalker-eth (approved):
@tk-o Thank you 🙌

PR #432, opened 2025-03-22
ENSAdmin support ens-test-env

djstrong:


apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -101,7 +101,11 @@ export function useEnsDeploymentChain(
     return undefined;
   }
 
-  return parseEnsDeploymentChain(indexingStatus.env.ENS_DEPLOYMENT_CHAIN);
+  const chainId = parseEnsDeploymentChain(indexingStatus.env.ENS_DEPLOYMENT_CHAIN);
+
+  // Only return chainId if it's one of the supported deployment chain IDs
+  // This handles the case where chainId might be ensTestEnv.id which is not part of SupportedEnsDeploymentChainId
+  return chainId as SupportedEnsDeploymentChainId | undefined;
```
lightwalker-eth, line 108: `as` only influences type system checks. It doesn't influence any values. What I assume your goal is here is not what the code is actually doing.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -101,7 +101,11 @@ export function useEnsDeploymentChain(
     return undefined;
   }
 
-  return parseEnsDeploymentChain(indexingStatus.env.ENS_DEPLOYMENT_CHAIN);
+  const chainId = parseEnsDeploymentChain(indexingStatus.env.ENS_DEPLOYMENT_CHAIN);
+
+  // Only return chainId if it's one of the supported deployment chain IDs
+  // This handles the case where chainId might be ensTestEnv.id which is not part of SupportedEnsDeploymentChainId
```
lightwalker-eth, line 107: Hmm.. This doesn't seem right to me. the chain id of the ens-test-env should be included in the list of supported ENS deployments.

apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -34,7 +35,9 @@ export const config = createConfig({
 /**
  * Supported chain ID for requested ENS Deployment Chain.
  */
-export type SupportedEnsDeploymentChainId = (typeof config.chains)[number]["id"];
+export type SupportedEnsDeploymentChainId =
```
lightwalker-eth, line 38: Hmm. Can you help me understand why we need any special logic here for the `ens-test-env`? Shouldn't it just automatically work?

apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -14,7 +15,7 @@ const getRpcUrl = (chainId: number): string => {
     if (chainId === mainnet.id) return "https://eth.drpc.org";
     if (chainId === sepolia.id) return "https://sepolia.drpc.org";
     if (chainId === holesky.id) return "https://holesky.drpc.org";
-
+    // if (chainId === ensTestEnv.id) return "";
```
lightwalker-eth, line 18:
```suggestion

```

Suggest totally removing that.

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -3,13 +3,26 @@
...
+import { config } from "@/lib/wagmi";
 import { cx } from "class-variance-authority";
 import { ExternalLink } from "lucide-react";
 import { useEffect, useState } from "react";
 import type { Hex } from "viem";
 import { useEnsName } from "wagmi";
 import { getEnsAppUrl, getEnsAvatarUrl } from "./helpers";
 
+// Utility function to check if a chain is configured with an RPC endpoint in wagmi config
+export function isConfiguredChain(chainId: SupportedEnsDeploymentChainId): boolean {
```
lightwalker-eth, line 15: Hmm. I appreciate the idea here, but I don't think it's the right direction.

As I understand, the direction proposed here is to make it so that we say that the `ens-test-env` isn't a "configured chain"? and therefore we disable any wagmi lookups if the ens deployment we're connected to is the `ens-test-env`?

If so, that doesn't seem right.

The `ens-test-env` should include a full deployment of ENS and all lookups in theory should still properly work on it.

What about the following strategy instead?
1. For the ens-test-env, be sure to configure the RPC URL for chain id 1337 in ENSAdmin to be the RPC url of the Anvil node you're running.
2. Couldn't we then just rely on the existing `getRpcUrl` function defined in ENSAdmin? If ENSAdmin is connected to an ENSNode that's connected to an ens-test-env, then either:
   1. Option 1: an environment variable is configured in ENSAdmin for the RPC URL of chain id 1337. This RPC URL is returned by `getRpcUrl` and everything just works. This includes things `useEnsName`, etc.. As I understand they should all still work great.
   2. Option 2: No environment variable is configured in ENSAdmin for the RPC URL of chain id 1337, and therefore ENSAdmin displays an error.

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -30,12 +43,15 @@ export function ENSName({
...
 }: ENSNameProps) {
   const [mounted, setMounted] = useState(false);
+  const isChainConfigured = isConfiguredChain(chainId);
 
-  // Use the ENS name hook from wagmi
-  const { data: ensName, isLoading } = useEnsName({
-    address,
-    chainId,
-  });
+  // Use the ENS name hook from wagmi only if the chain is configured with an RPC endpoint
```
lightwalker-eth, line 48: Please see my other comments. I don't feel good about having states like this spread all across ENSAdmin logic.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -147,7 +148,7 @@ function getTrueOwner(owner: { id: Hex }, wrappedOwner?: { id: Hex }) {
   return getAddress(checksumAddress(owner.id));
 }
 
-const supportedChainIds = [mainnet.id, sepolia.id, holesky.id] as const;
+const supportedChainIds = [mainnet.id, sepolia.id, holesky.id, ensTestEnv.id] as const;
```
lightwalker-eth, line 151: I believe this should be using definitions from our ens-deployments package. I don't believe we should be redefining things here.

lightwalker-eth (changes requested):
@djstrong Thanks for the updates here. The changes proposed here do fix one set of issues, but suggest we take a different direction for fixing instead.

apps/ensadmin/src/lib/chains.ts
```diff
@@ -1,11 +1,17 @@
-import { base, holesky, linea, mainnet, sepolia } from "viem/chains";
+import { anvil, base, holesky, linea, mainnet, sepolia } from "viem/chains";
+
+export const ensTestEnv = {
+  ...anvil,
+  id: 1337,
+};
 
 const chains = {
```
shrugs, line 8: this snippet could be useful 
```ts
const chains = uniqBy(Object.values(DeploymentConfigs).flatMap((config) =>
  Object.values(config).map((c) => c.chain),
), (chain) => chain.id);
```

apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -23,11 +25,12 @@ const getRpcUrl = (chainId: number): string => {
 
 // Create wagmi config with supported chains
 export const config = createConfig({
-  chains: [mainnet, sepolia, holesky],
+  chains: [mainnet, sepolia, holesky, ensTestEnv],
   transports: {
```
shrugs, line 29: could be 

```ts
transports: chains.reduce((memo, chain) => ({ ...memo, [chain.id]: http(getRpcUrl(chain.id)) }
```

tk-o:
@djstrong feel free to close this one, as we also covered this change with:
- #476

lightwalker-eth:
Confirmed that this PR is now safe to close.

PR #398, opened 2025-03-17
chore: add ENSAdmin OG-images

Y3drk:
This PR also includes additions of smaller metadata content for ENSAdmin page

PR #433, opened 2025-03-22
fix(ensadmin): use dynamic metadata

tk-o:
This is a follow up PR to:
- https://github.com/namehash/ensnode/pull/398

This PR support reading env vars during runtime, and avoid doing so during build time. It also allows leveraging Vercel defaults for setting `metadataBase` URL to a correct deployment URL. For other platforms, setting `ENSADMIN_PUBLIC_URL` should be done instead.

Also, resolves:
- https://github.com/namehash/ensnode/security/dependabot/13
- https://github.com/namehash/ensnode/security/dependabot/14

### DIFF
--- a/apps/ensadmin/.env.local.example
+++ b/apps/ensadmin/.env.local.example
@@ -1,7 +1,3 @@
-# The ENSAdmin public service URL
-# The ENSAdmin web application needs this value to reference static assets.
-ENSADMIN_PUBLIC_URL=http://localhost:4173
-
 # Preferred ENSNode URL (used by default when ENSAdmin is opened)
 # If not set, the fallback URL from `PREFERRED_ENSNODE_URL` const is used.
 NEXT_PUBLIC_PREFERRED_ENSNODE_URL=https://alpha.ensnode.io
--- a/apps/ensadmin/.env.local.example
+++ b/apps/ensadmin/.env.local.example
@@ -1,3 +1,8 @@
+# The ENSAdmin public service URL
+# The ENSAdmin web application needs this value to reference static assets.
+# Note: make sure to define this value for all deployments outside Vercel platform.
+ENSADMIN_PUBLIC_URL=http://localhost:4173
+
 # Preferred ENSNode URL (used by default when ENSAdmin is opened)
 # If not set, the fallback URL from `PREFERRED_ENSNODE_URL` const is used.
 NEXT_PUBLIC_PREFERRED_ENSNODE_URL=https://alpha.ensnode.io
--- a/apps/ensadmin/.env.local.example
+++ b/apps/ensadmin/.env.local.example
@@ -1,6 +1,9 @@
 # The ENSAdmin public service URL
 # The ENSAdmin web application needs this value to reference static assets.
-# Note: make sure to define this value for all deployments outside Vercel platform.
+# Note: it's recommended to explicitly set it. If not explicitly set,
+# the application will try guess this value. For example, if hosting in Vercel,
+# the Vercel env and Vercel URLs will be used. If not provided,
+# the default public URL will be based on localhost and PORT.
 ENSADMIN_PUBLIC_URL=http://localhost:4173
 
 # Preferred ENSNode URL (used by default when ENSAdmin is opened)
--- a/apps/ensadmin/src/app/layout.tsx
+++ b/apps/ensadmin/src/app/layout.tsx
@@ -22,17 +22,10 @@ const title = "ENSAdmin";
 const description = "Explore the ENS Protocol like never before";
 
 export async function generateMetadata(): Promise<Metadata> {
-  const metadataBaseUrl = ensAdminPublicUrl();
-
   return {
     title: title,
     description: description,
-    /**
-     * Note: using Vercel platform for ENSAdmin deployments works best when
-     * this function returns undefined and lets default values to be applied.
-     * Read more: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default-value
-     */
-    metadataBase: metadataBaseUrl ? new URL(metadataBaseUrl) : undefined,
+    metadataBase: ensAdminPublicUrl(),
     openGraph: {
       title: {
         template: `${siteName} - %s`,
--- a/apps/ensadmin/src/app/layout.tsx
+++ b/apps/ensadmin/src/app/layout.tsx
@@ -21,32 +21,34 @@ const siteName = "ENSAdmin";
 const title = "ENSAdmin";
 const description = "Explore the ENS Protocol like never before";
 
-export const metadata: Metadata = {
-  title: title,
-  description: description,
-  metadataBase: new URL(ensAdminPublicUrl()),
-  openGraph: {
-    title: {
-      template: `${siteName} - %s`,
-      default: title,
-    },
+export async function generateMetadata(): Promise<Metadata> {
+  return {
+    title: title,
     description: description,
-    url: "/",
-    type: "website",
-    siteName: siteName,
-    images: ["/opengraph-image.png"],
-  },
-  twitter: {
-    title: {
-      template: `${siteName} - %s`,
-      default: title,
+    metadataBase: new URL(ensAdminPublicUrl()),
+    openGraph: {
+      title: {
+        template: `${siteName} - %s`,
+        default: title,
+      },
+      description: description,
+      url: "/",
+      type: "website",
+      siteName: siteName,
+      images: ["/opengraph-image.png"],
     },
-    card: "summary_large_image",
-    site: "@NamehashLabs",
-    creator: "@NamehashLabs",
-    images: ["/twitter-image.png"],
-  },
-};
+    twitter: {
+      title: {
+        template: `${siteName} - %s`,
+        default: title,
+      },
+      card: "summary_large_image",
+      site: "@NamehashLabs",
+      creator: "@NamehashLabs",
+      images: ["/twitter-image.png"],
+    },
+  };
+}
 
 export default function Layout({
   children,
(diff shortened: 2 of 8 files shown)


apps/ensadmin/src/app/layout.tsx
```diff
@@ -21,32 +21,34 @@ const siteName = "ENSAdmin";
...
-    url: "/",
-    type: "website",
-    siteName: siteName,
-    images: ["/opengraph-image.png"],
-  },
-  twitter: {
-    title: {
-      template: `${siteName} - %s`,
-      default: title,
+    metadataBase: new URL(ensAdminPublicUrl()),
```
lightwalker-eth, line 28: It looks like this isn't working

CleanShot 2025-03-23 at 15 30 20

Please verify using the Vercel toolbar in the preview environment. Open the Vercel toolbar and then navigate to "Open Graph".

lightwalker-eth (changes requested):
@tk-o Thanks for updates. Appreciate your help fixing the issue.

apps/ensadmin/src/app/layout.tsx
tk-o: This is fixed now. For example, both URLs for preview deployment include working OG images:
- https://adminensnodeio-git-fix-ensadmin-dynamic-metadata-namehash.vercel.app
  - image
- https://adminensnode-morpel8r4-namehash.vercel.app/
  - image

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,12 +1,16 @@
...
  */
-export function ensAdminPublicUrl() {
+export function ensAdminPublicUrl(): string | undefined {
   const envVarName = "ENSADMIN_PUBLIC_URL";
-  const envVarValue = process.env[envVarName];
+  const envVarValue = process.env.ENSADMIN_PUBLIC_URL;
 
   if (!envVarValue) {
-    throw new Error(`Required "${envVarName}" value was not set`);
+    return undefined;
```
tk-o, line 13: Making it an optional value to allow Vercel applying its env vars automatically:
https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default-value

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,12 +1,16 @@
...
+ *
+ * Note: using Vercel platform for ENSAdmin deployments works best when
+ * this function returns undefined and lets default values to be applied.
+ * Read more: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default-value
  */
-export function ensAdminPublicUrl() {
+export function ensAdminPublicUrl(): string | undefined {
   const envVarName = "ENSADMIN_PUBLIC_URL";
-  const envVarValue = process.env[envVarName];
+  const envVarValue = process.env.ENSADMIN_PUBLIC_URL;
```
lightwalker-eth, line 10:
```suggestion
  const envVarValue = process.env[envVarName];
```

I assume we should keep the old logic here?

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,12 +1,16 @@
 /**
  * Get ENSAdmin service public URL.
+ *
+ * Note: using Vercel platform for ENSAdmin deployments works best when
+ * this function returns undefined and lets default values to be applied.
+ * Read more: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default-value
  */
-export function ensAdminPublicUrl() {
+export function ensAdminPublicUrl(): string | undefined {
```
lightwalker-eth, line 8: A few suggestions:

1. This should never return `undefined`.
2. This should return a `URL` and not a `string`.

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,12 +1,16 @@
...
   const envVarName = "ENSADMIN_PUBLIC_URL";
-  const envVarValue = process.env[envVarName];
+  const envVarValue = process.env.ENSADMIN_PUBLIC_URL;
 
   if (!envVarValue) {
```
lightwalker-eth, line 12: As I understand, we need to work to optimize the logic in this function for purposes beyond just OG images. For example, as I understand this function is also important for when ENSAdmin is trying connecting to an ENSNode instance. Is that fair?

All of the following suggests how we should handle the case that no explicit value is set for the env var.

1. First, we should check ENV vars to determine if we are in a Vercel production environment. If we are, then we should return `VERCEL_PROJECT_PRODUCTION_URL`.
2. Next, we should check ENV vars to determine if we are in a Vercel preview environment. If we are, then we should first check `VERCEL_BRANCH_URL`. If that's defined we should use that. If its not defined we should use `VERCEL_URL`.
3. Finally, if all of the above fails, we should use `http://localhost:${process.env.PORT || 3000}` (please suggest what default port we should use)

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,12 +1,16 @@
 /**
  * Get ENSAdmin service public URL.
+ *
+ * Note: using Vercel platform for ENSAdmin deployments works best when
```
lightwalker-eth, line 4: Please see my comments below. Whatever path you suggest we take, please ensure the final comments here are aligned.

apps/ensadmin/src/app/layout.tsx
```diff
@@ -21,32 +21,41 @@ const siteName = "ENSAdmin";
...
-    url: "/",
-    type: "website",
-    siteName: siteName,
-    images: ["/opengraph-image.png"],
-  },
-  twitter: {
-    title: {
-      template: `${siteName} - %s`,
-      default: title,
+    /**
```
lightwalker-eth, line 30: Whatever path you decide to take, please ensure the final comment here is aligned.

apps/ensadmin/.env.local.example
```diff
@@ -1,5 +1,6 @@
 # The ENSAdmin public service URL
 # The ENSAdmin web application needs this value to reference static assets.
+# Note: make sure to define this value for all deployments outside Vercel platform.
```
lightwalker-eth, line 3: Please see my other comments.

Overall my suggestion is to describe this env var as:
1. We recommend to explicitly set it
2. If not explicitly set, we try to guess if hosting in Vercel, and if that fails, we default to localhost and PORT.

lightwalker-eth (changes requested):
@tk-o Thanks for updates. Reviewed and shared feedback.

apps/ensadmin/src/lib/env.ts
tk-o: Right now, the only use case is OG images.

apps/ensadmin/src/app/layout.tsx
tk-o: I wonder if we need any comment here at all. `ensAdminPublicUrl()` will always return a URL:
```tsx
 metadataBase: ensAdminPublicUrl(),
```

Based on that, all docs could be provided on `ensAdminPublicUrl` function. What do you think, @lightwalker-eth?

apps/ensadmin/src/lib/env.ts
tk-o: Correct. The direct read is needed only when bundling env vars into the client app:
https://nextjs.org/docs/app/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,23 +1,102 @@
...
+
+/**
+ * Tells if application runs on Vercel platform.
+ */
+function isAppOnVercelPlatform(): boolean {
+  return process.env.VERCEL === "1";
+}
+
+/**
+ * Builds a public URL of an application running on Vercel platform.
```
lightwalker-eth, line 77:
```suggestion
 * Builds a public URL of the app assuming it runs on Vercel.
```

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,23 +1,102 @@
...
+
+  if (parsedValue <= 0) {
+    throw new Error(`'${rawValue}' is not a natural number`);
+  }
+
+  return parsedValue;
+}
+
+/**
+ * Tells if application runs on Vercel platform.
```
lightwalker-eth, line 70:
```suggestion
 * Tells if the application runs on Vercel.
```

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,23 +1,102 @@
...
-    return parseUrl(envVarValue).toString();
+    return parseUrl(envVarValue);
   } catch (error) {
     console.error(error);
 
     throw new Error(`Invalid ${envVarName} value "${envVarValue}". It should be a valid URL.`);
   }
 }
 
+/** Build a default URL for ENSAdmin application */
```
lightwalker-eth, line 34:
```suggestion
/** Build a default public URL for ENSAdmin */
```

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,23 +1,102 @@
...
+function defaultEnsAdminPublicUrl(): URL {
+  let applicationPort: number;
+
+  try {
+    applicationPort = parseApplicationPort(process.env.PORT);
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : "Unknown error";
+
+    console.error(`Could not build default URL for ENSAdmin application due to: ${errorMessage}`);
```
lightwalker-eth, line 43:
```suggestion
    console.error(`Error building default public URL for ENSAdmin: ${errorMessage}`);
```

apps/ensadmin/src/lib/env.ts
```diff
@@ -1,23 +1,102 @@
...
+ * @throws when Vercel platform was detected but could not determine application hostname
  */
-export function ensAdminPublicUrl() {
+export function ensAdminPublicUrl(): URL {
   const envVarName = "ENSADMIN_PUBLIC_URL";
   const envVarValue = process.env[envVarName];
 
   if (!envVarValue) {
-    throw new Error(`Required "${envVarName}" value was not set`);
+    // if public URL was provided and the application runs on Vercel platform
```
lightwalker-eth, line 15:
```suggestion

```

Suggest to completely remove that comment

lightwalker-eth (approved):
@tk-o Great to see this! Could you please action the small content refinements I suggested and then merge this PR? Thank you!

PR #436, opened 2025-03-23
refactor(ensadmin): wrap failable code within useQuery

tk-o:
This is a follow up PR to:
- https://github.com/namehash/ensnode/pull/414

This PR addresses issues of lots of undefined state spreading across ENSAdmin application.

The main technique used to limit unknown state was to apply `useQuery` hook to wrap failable logic (where errors can be thrown). With this, we've go a nice abstraction layer to control errors.

Also, some of the queries are now dependent on others, which allows composing sequences of correct state transitions. For example, the `useRecentRegistrations` query will not be executed until `lastIndexedBlockInfo` and `registrationsStartBlockInfo` query results are known and can be compared one to another. This allows saving some extra code branches in the presentation layer, as fetching data is state-driven.


apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/**
+ * Get avatar URL for given ENS name on a supported network
+ * @param ensDeploymentChain
+ * @param ensName
+ * @returns avatar URL for supported networks
+ * @throws error when ensDeploymentChain cannot be mapped into an ENS App URL
+ */
+export function getEnsAvatarUrl(ensDeploymentChain: ENSDeploymentChain, ensName: string): URL {
```
lightwalker-eth, line 35:
```suggestion
export function getEnsAvatarUrl(ensDeploymentChain: ENSDeploymentChain, name: string): URL {
```

apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+}
+
+// Create error object for ENS App URL exception
+function ensAppUrlUnsupportedDeploymentChainError(ensDeploymentChain: ENSDeploymentChain): Error {
+  return new Error(`ENS app is not supported on "${ensDeploymentChain}" ENS Deployment Chain`);
+}
+
+/**
+ * Get avatar URL for given ENS name on a supported network
```
lightwalker-eth, line 29:
```suggestion
 * Gets the avatar URL for given name on the provided ensDeploymentChain
```

apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param ensDeploymentChain
+ * @param ensName
+ * @returns avatar URL for supported networks
```
lightwalker-eth, line 32:
```suggestion
 * @returns avatar URL for the given name on the provided ensDeploymentChain
```

apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @throws error when ensDeploymentChain cannot be mapped into an ENS App URL
```
lightwalker-eth, line 33:
```suggestion
 * @throws error when the Avatar URL is not known for the provided ensDeploymentChain
```

apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+      throw ensAvatarUrlUnsupportedDeploymentChainError("ens-test-env");
+  }
+}
+
+// Create error object for Avatar URL exception
+function ensAvatarUrlUnsupportedDeploymentChainError(
+  ensDeploymentChain: ENSDeploymentChain,
+): Error {
+  return new Error(
+    `ENS Avatar URL is not supported on "${ensDeploymentChain}" ENS Deployment Chain`,
```
lightwalker-eth, line 56:
```suggestion
    `No Avatar URL is known on "${ensDeploymentChain}"`,
```

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,103 +2,164 @@
...
-    <div className={cx("flex items-center gap-2", className)}>
+    <>
       {showAvatar && (
-        <Avatar className="h-6 w-6">
-          {avatarUrl && ensName && <AvatarImage src={avatarUrl} alt={ensName} />}
-          <AvatarFallback className="text-xs">
-            {displayName.slice(0, 2).toUpperCase()}
-          </AvatarFallback>
-        </Avatar>
+        <ENSNameAvatar ensAvatarUrlQuery={ensAvatarUrlQuery} ensDisplayName={ensDisplayName} />
```
lightwalker-eth, line 60: Is there a special reason why we pass `ensAvatarUrlQuery` into this component rather than having the internal logic inside this component build `ensAvatarUrlQuery`?

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,103 +2,164 @@
...
-          {showExternalLink && <ExternalLink size={14} className="inline-block" />}
-        </a>
-      ) : (
-        <span className={ensName ? "font-medium" : "font-mono text-xs"}>
-          {isLoading ? <Skeleton className="h-4 w-24" /> : displayName}
-        </span>
-      )}
-    </div>
+      <ENSDisplayName
+        ensAppPath={address}
```
lightwalker-eth, line 64: This is really confusing to me. Why `ensAppPath={address}`?

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,103 +2,164 @@
...
+        ensAppUrlQuery={ensAppUrlQuery}
```
lightwalker-eth, line 65: Shouldn't all these hooks be triggered inside of `ENSDisplayName` rather than passing them in like this? I'm not a frontend expert.

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,103 +2,164 @@
...
+    <Avatar className="h-6 w-6">
+      <AvatarImage src={ensAvatarUrlQuery.data.toString()} alt={ensDisplayName} />
+      <AvatarFallback className="text-xs">{ensDisplayName}</AvatarFallback>
+    </Avatar>
+  );
+}
+
+interface ENSDisplayNameProps {
+  /** either wallet address or domain name  */
+  ensAppPath: string;
```
lightwalker-eth, line 122: Why is this called `ensAppPath`? That's very confusing. I don't understand this param.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -55,24 +58,51 @@ async function fetchRecentRegistrations(baseUrl: string): Promise<RecentRegistra
   return data.data;
 }
 
+interface UseRecentRegistrationsProps {
+  searchParams: URLSearchParams;
```
lightwalker-eth, line 62: Are these `searchParams` related to the connected ENSNode instance? If so, it really feels wrong to me that our logic is setup this way.

I left a related comment with an idea for a `useEnsNode` hook. Please check it, thanks.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -55,24 +58,51 @@ async function fetchRecentRegistrations(baseUrl: string): Promise<RecentRegistra
...
+  lastIndexedBlockInfo: UseQueryResult<BlockInfo, Error>;
```
lightwalker-eth, line 63: As I understand it, all of these fields here (and more) should be returned as part of a `useEnsNode` hook.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -55,24 +58,51 @@ async function fetchRecentRegistrations(baseUrl: string): Promise<RecentRegistra
...
+> {
   const ensNodeUrl = selectedEnsNodeUrl(searchParams);
 
   return useQuery({
-    queryKey: ["recent-registrations", ensNodeUrl],
-    queryFn: () => fetchRecentRegistrations(ensNodeUrl),
-    throwOnError(error) {
-      throw new Error(`ENSNode request error at '${ensNodeUrl}'. Cause: ${error.message}`);
+    enabled() {
+      // this query will only fetch registrations if last indexed block timestamp on or after the registrations start block timestamp
```
lightwalker-eth, line 84: Should we move this idea into a helper function?

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -55,24 +58,51 @@ async function fetchRecentRegistrations(baseUrl: string): Promise<RecentRegistra
...
+      // otherwise, there's no point to fetch registrations
+      return false;
     },
+    queryKey: ["ensnode", ensNodeUrl, "recent-registrations"],
+    queryFn: () => fetchRecentRegistrations(ensNodeUrl),
   });
 }
 
-export function useRecentRegistrationsViaPonder(skipChildrenOfDomains: Array<string> = []) {
+export function useRecentRegistrationsViaPonder(
```
lightwalker-eth, line 103: I'm thinking we should totally avoid any use of Ponder Client in ENSAdmin. I don't feel comfortable with its architecture. The only exception can be inside the API helper page in ENSAdmin giving some small demo of Ponder Client.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -18,17 +19,318 @@ import {
...
 import { Provider as PonderClientProvider } from "@/components/providers/ponder-client-provider";
+import { parseEnsDeploymentChainIntoChainId } from "@/lib/chains";
+import DeploymentConfigs, {
+  ENSDeploymentChain,
+} from "@ensnode/ens-deployments";
+import { BlockInfo } from "@ensnode/ponder-metadata";
+import { UseQueryResult } from "@tanstack/react-query";
+
+export function RecentRegistrations() {
+  const searchParams = useSearchParams();
```
lightwalker-eth, line 45: Please see my other comments about a `useEnsNode` hook.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -18,17 +19,318 @@ import {
...
+      </Card>
+    );
+  }
+
+  if (indexingStatus.isError) {
+    return <ErrorMessage error={indexingStatus.error} />;
+  }
+
+  return (
+    <PonderClientProvider url={selectedEnsNodeUrl(searchParams)}>
```
lightwalker-eth, line 66: Oh wow, this is a surprise for me. I didn't know we were implementing recent registrations through Ponder client? I thought we were implementing it through GraphQL? I preferred GraphQL as I don't feel comfortable with the architecture of Ponder client.

lightwalker-eth (changes requested):
@tk-o Thanks for your efforts on this PR. Reviewed and shared feedback.

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,103 +2,164 @@
...
 import type { Hex } from "viem";
 import { useEnsName } from "wagmi";
-import { getEnsAppUrl, getEnsAvatarUrl } from "./helpers";
+import { useEnsAppUrlQuery, useEnsAvatarUrlQuery } from "./hooks";
+import { formatEnsAccountName, nullToUndefined } from "./utils";
 
 interface ENSNameProps {
   address: Hex;
-  chainId: SupportedEnsDeploymentChainId;
+  ensDeploymentChain: ENSDeploymentChain;
```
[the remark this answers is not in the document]
tk-o: I think of React hooks as service layer, where one service can call other services in a composable fashion. Should I change my my perspective here to some particular one?

I understood your comment as:
1. Define React Context (i.e. `ENSNodeContext`) that would include ENSNode Metadata. I this context all relevant validations after connection change would be performed.
2. Have all components that rely on ENSNode Metadata to read form `ENSNodeContext` so we can avoid passing too many props around.

apps/ensadmin/src/components/ens-name/components.tsx
```diff
@@ -2,103 +2,164 @@
...
  * Falls back to a truncated address if no ENS name is found.
  */
-export function ENSName({
-  address,
-  chainId,
-  showAvatar = false,
-  showExternalLink = true,
-  className = "",
-}: ENSNameProps) {
+export function ENSName({ address, ensDeploymentChain, showAvatar = false }: ENSNameProps) {
```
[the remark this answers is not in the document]
tk-o: We can avoid passing extra `ensDeploymentChain` arg by switching to using React Context. More on that [here](https://github.com/namehash/ensnode/pull/436#discussion_r2009499197).

apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -1,21 +1,29 @@
...
   if (!url) {
     console.warn(`No RPC URL found for chain ID ${chainId} (${envVar}). Using fallback.`);
 
     // Fallbacks for development - should be replaced with proper RPC URLs in production
-    if (chainId === mainnet.id) return "https://eth.drpc.org";
-    if (chainId === sepolia.id) return "https://sepolia.drpc.org";
-    if (chainId === holesky.id) return "https://holesky.drpc.org";
-
-    throw new Error(`No fallback RPC URL available for chain ID ${chainId}`);
+    switch (ensDeploymentChain) {
```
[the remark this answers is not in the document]
tk-o: I don't understand why ENSAdmin would require any RPC URL from its operator. ENSAdmin is served as a browser web app, so each user of this app could use default RPC URLs from `wagmi` and it would allow using ENSAdmin just fine.

Is there any particular reason we can't use `wagmi`'s default behaviour?

apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -25,32 +33,9 @@ const getRpcUrl = (chainId: number): string => {
 export const config = createConfig({
   chains: [mainnet, sepolia, holesky],
```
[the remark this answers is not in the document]
tk-o: Good catch, anvil should be included.

apps/ensadmin/src/lib/chains.ts
```diff
@@ -17,3 +18,19 @@ export function getBlockExplorerUrl(chainId: number, blockNumber: number): strin
   if (!chain?.blockExplorers?.default?.url) return null;
   return `${chain.blockExplorers.default.url}/block/${blockNumber}`;
 }
+
+/**
+ * Parse ENS Deployment Chain into Chain ID.
+ */
+export function parseEnsDeploymentChainIntoChainId(ensDeploymentChain: ENSDeploymentChain): number {
+  switch (ensDeploymentChain) {
```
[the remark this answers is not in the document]
tk-o: Correct, we could read directly from `ENSDeploymentConfig` 👍

apps/ensadmin/src/components/ens-name/components.tsx
tk-o: Yes, the reason here is to call a query once at the container component (which I think `ENSName` is with all hooks being called here) and only pass the query object into the presentational components (i.e. `ENSNameAvatar`, `ENSDisplayName`). This way, presentational components don't have to know all dependencies to required to construct query objects.

Does that make sense?

apps/ensadmin/src/components/ens-name/components.tsx
tk-o: It's because https://app.ens.domains allows either `address` or `name` to be searched for:
- https://app.ens.domains/vitalik.eth
- https://app.ens.domains/0x220866b1a2219f40e72f5c628b65d54268ca3a9d

Open to suggestions on how define this case better in code.

apps/ensadmin/src/components/ens-name/components.tsx
tk-o: We can do one way (as is) or another (call hooks in leaf components).

What I did is to apply inversion of control and have all relevant query object constructed at one level and have the query objects provided to presentational components, so these could handle loading/error/success states as they needed.

If we decide to call queries on the leaf components, we'd need to provide all configuration to each query constructor anyway. I think it's more straight-forward to pass the query object as a prop and let the presentational component to decide how to present given UI slice based on the query state (pending/loading/error/success).

Thoughts on that?

apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+}
+
+export function formatEnsAccountName(address: Hex, ensName: string | undefined) {
```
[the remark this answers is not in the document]
tk-o: I can use `Address` param if you like. From what I understand, it's and alias to `0x{string}`.

apps/ensadmin/src/components/ens-name/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (!ensName) {
+    // fallback to truncated address
+    return `${address.slice(0, 6)}...${address.slice(-4)}`;
+  }
+
+  // ENS Name
+  return ensName;
+}
+
+// narrow down the input value type
```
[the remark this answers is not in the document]
tk-o: Sure, will expand that comment.

apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -1,21 +1,29 @@
+import { ENSDeploymentChain } from "@ensnode/ens-deployments";
 import { http } from "viem";
-import { holesky, mainnet, sepolia } from "viem/chains";
+import { anvil, holesky, mainnet, sepolia } from "viem/chains";
 import { createConfig } from "wagmi";
+import { parseEnsDeploymentChainIntoChainId } from "./chains";
 
 // Get RPC URLs from environment variables
-const getRpcUrl = (chainId: number): string => {
+const getRpcUrl = (ensDeploymentChain: ENSDeploymentChain): string => {
```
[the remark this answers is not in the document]
tk-o: I don't think we need this function in the first place. Please see [my comment here](https://github.com/namehash/ensnode/pull/436#discussion_r2009534619).

apps/ensadmin/src/components/ens-name/components.tsx
tk-o: It's because https://app.ens.domains allows either `address` or `name` to be searched for:
- https://app.ens.domains/vitalik.eth
- https://app.ens.domains/0x220866b1a2219f40e72f5c628b65d54268ca3a9d

Open to suggestions on how define this case better in code.

apps/ensadmin/src/components/recent-registrations/hooks.ts
tk-o: Definitely we can improve here by tracking ENSNodeContext in React components tree, so that all nested components have well defined access to, for example, selected ENSNode URL on the connections list.

apps/ensadmin/src/components/recent-registrations/hooks.ts
tk-o: Yes, we can include these in a single place, for example, `ENSNodeContext`, and have all nested components to read the result using `useENSNodeContext()` hook at their respective level.

apps/ensadmin/src/components/recent-registrations/hooks.ts
tk-o: Sure, will do that and include a relevant code comment.

apps/ensadmin/src/components/recent-registrations/hooks.ts
tk-o: I'll drop this hook if we don't plan using Ponder Client.

apps/ensadmin/src/components/recent-registrations/components.tsx
tk-o: We don't really need this Provider component here as we don't call any `usePonderQuery` hook down the React component tree. Will drop it.

apps/ensadmin/src/components/recent-registrations/components.tsx
tk-o: Yes, we'll address that as [described here](https://github.com/namehash/ensnode/pull/436#discussion_r2009531202).

apps/ensadmin/src/components/ens-name/components.tsx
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Here's an example analogy. Imagine we defined a hook for the currently signed in user. Such a hook should sit nearish to the root of the DOM. All child components in the DOM should be able to ask for state about the currently logged in user without needing to explicitly pass some user object all over the code.

Additionally, a bunch of UI components are going to be dependent on the currently signed in user being defined. If you try to create an instance of component X where the hook for the currently signed in user does not return the necessary state, component X might just throw an exception.

I think we have a similar situation here, where we should be defining both a provider / hook / etc.. for the state of the connected ENSNode instance. Huge areas of the ENSAdmin UI should be totally dependent on that hook returning the expected state, such that if the returned state is null or undefined, they should just throw an exception because we messed something up.

apps/ensadmin/src/lib/wagmi.ts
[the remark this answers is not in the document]
lightwalker-eth: Sure:
1. Wagmi's default RPC URLs have rate limiting.
2. RPC URLs for ens-test-env (Anvil) need a strategy for being defined, so seems we should extend that pattern to all chains that need RPCs.
3. Soon we'll want to test with Namechain testnets and those likely won't be in wagmi yet.

What do you think?

I'm open to coming back to this later if its a problem now, but on principle I think it should be done.

apps/ensadmin/src/lib/wagmi.ts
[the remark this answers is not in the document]
lightwalker-eth: I don't feel comfortable relying on the limitations of the default RPCs provided by viem / wagmi.

apps/ensadmin/src/components/ens-name/utils.ts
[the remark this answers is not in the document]
lightwalker-eth: Yes. An `Address` symbolically represents a more specialized `Hex` value.

apps/ensadmin/src/components/ens-name/components.tsx
lightwalker-eth: I think I understand the goal.

But why are we passing the entire query object into presentation components?

We use a query to get a response. Shouldn't we just pass in the more lightweight response data structure?

apps/ensadmin/src/components/ens-name/components.tsx
lightwalker-eth: Sure. Rename `ensAppPath` to `nameOrAddress`

apps/ensadmin/src/components/ens-name/components.tsx
lightwalker-eth: I think I understand the goal.

But why are we passing the entire query object into presentation components?

We use a query to get a response. Shouldn't we just pass in the more lightweight response data structure?

I'm not a frontend expert.

apps/ensadmin/src/components/recent-registrations/hooks.ts
djstrong: This was made to enable filtering unknown names and other plugin names. GraphQL does not allow for such queries. Also Ponder client will be needed for getting counts of names.

tk-o:
This PR got replaced by:
- #475 
- #476

PR #437, opened 2025-03-24
fix: ensnode.io Twitter image

Y3drk:
as title suggests


lightwalker-eth (approved):
@Y3drk Let's try it, thanks!

PR #438, opened 2025-03-24
Update switch wokflow to include ENSRainbow

BanaSeba:
During "switch environment" workflow run also update routing for ENSRainbow


lightwalker-eth (approved):
@BanaSeba Looks good!

PR #443, opened 2025-03-24
feat(api): add version endpoint to return current version and schema …

djstrong:
…version


lightwalker-eth (approved):
@djstrong Looks good! 😄

Issue #445, opened 2025-03-25
Add versioning for ENSNode


BanaSeba:
ENSNode needs versioning of packages/apps so that specific versions of components can be picked for Github/NPM/Docker

PR #447, opened 2025-03-25
Feat/445 add versioning for ensnode

BanaSeba:
Related to: https://github.com/namehash/ensnode/issues/445

Merged release workflows considering apps/packages into one general.
Added Github action for publishing a Docker image ( used in workflow )
Added "publish-meta" logic. It will gather info about changeset version bumps and tell which Docker images with specified versions should be created.
For now build docker image supports only ENSIndexer, can be easily extended for the next packages in the future.


.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release:
+    name: Release
+    if: github.repository == 'namehash/ensnode'
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+
+    steps:
+      - uses: actions/checkout@v3
```
shrugs, line 18: necessary if using checkoutv4 below?

.github/actions/build_docker_image/action.yml
```diff
@@ -0,0 +1,51 @@
...
+  
+  registry_token:
+    description: 'Registry token for Docker registry authentication'
+    required: true
+    type: string
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout repository
```
shrugs, line 27: i'm not familiar, but does the action need to include actions/checkout if the workflow already does?

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
shrugs, line 21: could use fetch-depth: 1 here, right?

shrugs (approved):

.github/workflows/release.yml
BanaSeba: Nope, some duplication came here. Removed

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+              esac
+            done
+          
+      - name: Build ENSIndexer Docker image
+        if: steps.publish-meta.outputs.ensindexer_version != ''
+        uses: ./.github/actions/build_docker_image
+        env:
+          ENSINDEXER_IMAGE_NAME: ghcr.io/${{ github.repository }}/ensindexer:${{ steps.publish-meta.outputs.ensindexer_version }}
+        with:
+          image_name: ${{ env.ENSINDEXER_IMAGE_NAME }}
```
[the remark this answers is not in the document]
BanaSeba: Skipped env assignment

.github/actions/build_docker_image/action.yml
BanaSeba: No, it's not necessary if the workflow already does that. However, I would want to leave checkout here as well. In the future, we may want to reuse all image-related workflows to use this action. We can then skip this step inside of each one of them

.github/workflows/release.yml
BanaSeba: Yep, added that

PR #448, opened 2025-03-26
feat: move to dynamic import for the react-flow client

notrab:
I've improved `Math.sqrt()` that may produce slightly different results between environments due to floating-point precision.

Also I moved React Flow to use `next/dynamic`.


Y3drk (approved):
Looks good to me :)

* Glad that you improved the math, since it was really sloppy on my side
* Thanks for investigating the issue with "hydration error", I think this should solve the problem. I'll monitor that in the future.

PR #475, opened 2025-04-02
refactor(ensadmin): simplified ensnode metadata fetching

tk-o:
This PR changes the way data is fetched for the Status page. I suggest checking the _Hide whitespaces_ option while reviewing:
> image
 

Before, the data was fetched on the client side only which required us to show a loading screen first. Now, after applying React Query hydration provider, the Status page gets rendered on the server side and is later refetched on the client side when needed.

This PR introduces schema validation for response data using [Valibot](https://valibot.dev/).

In summary, this PR meets the following goals outlined by @lightwalker-eth:
> ENSAdmin strongly validates invariants for any ENSNode it connects to. If any invariant is violated it should refuse to connect.
  - We use schema-based parsing of response data that comes from ENSNode `/metadata` query
  - We check if the ENSNode Version is in a version range supported by ENSAdmin
  - We validate `ENS_DEPLOYMENT_CHAIN` value
  - We validate information about block
    - `firstBlockToIndex` & `latestSafeBlock` are required
    - `lastIndexedBlock` & `lastSyncedBlock` are optional (but if present, still need to maintain correct form)

> Ideally, the responsibility for validating state invariants for a connected ENSNode would have that responsibility concentrated into a single layer and not spread all over.

- We parse response data in right after it was fetched, and right before it is passed into the UI layer
- All UI components access this data via a single hook
- The data is fetched and parsed on the server-side when page is loaded, and then client-side on each data refresh.

### A demo of what ENSNode version mismatch error would look like

image


shrugs (approved):
cool!

packages/ponder-metadata-api/src/types/api.ts
```diff
@@ -27,7 +27,7 @@ export interface PonderMetadataMiddlewareOptions<AppInfo, EnvVars extends Ponder
   };
 
   /** Public clients for each blockchain network fetching data */
-  publicClients: Record<number, PublicClient>;
+  publicClients: Record<string, PublicClient>;
```
tk-o, line 30: Changing record key type to string to follow idiomatic JS coding style.

Valibot expects the keys of a record to be of string type. Object.entires and Object.keys would also include keys as string types. Serialized object has its numeric keys formatted as strings.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -8,30 +8,9 @@ import { BlockInfo, NetworkIndexingStatus } from "@ensnode/ponder-metadata";
...
-    },
-  });
-
-  if (!response.ok) {
-    console.error("Failed to fetch ENSNode status", response);
-    throw new Error("Failed to fetch ENSNode status");
-  }
-
-  return response.json();
-}
```
tk-o, line 33: Moved into `metadata-query.ts` file.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -44,47 +23,7 @@ export function useIndexingStatusQuery(
...
-      return data;
-    },
-    throwOnError(error) {
-      throw new Error(`ENSNode request error at '${ensNodeUrl}'. Cause: ${error.message}`);
-    },
-  });
-}
-
-/**
- * Checks if the response has the expected structure.
```
tk-o, line 62: Moved into `metadata-query.ts` file and replaced with schema-based parsing.

apps/ensadmin/src/components/providers/react-query-client.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Create a query client to be used on the server and browser.
+ *
+ * @returns A query client.
+ */
+function makeQueryClient() {
+  return new QueryClient({
+    defaultOptions: {
+      queries: {
+        staleTime: 10 * 1000,
```
tk-o, line 12: Not sure what is the right value here. Assuming this one will prevent queries from being refetched client-side for 10 seconds.

PR #476, opened 2025-04-02
refactor(ensadmin): simplify recent registrations fetching

tk-o:
This PR addresses recommendations from @lightwalker-eth:

> ENSAdmin must support the `ens-test-env`, which doesn't have a default RPC endpoint defined in viem. It must also be setup for the future to connect to Namechain testnets that also won't be defined in viem until later. The operator of ENSAdmin must always maintain control over the RPC endpoints used by ENSAdmin for all chains.

- We now require ENSAdmin operator to specifically define RPC URLs via environment variables
- We now support `ens-test-env` as by default it's hosted with Anvil node on a local network

> ENSAdmin logic should be maintainable and readable code. Massive volumes of `undefined` values and conditional logic to validate invariants shouldn't be repeated in 10,000 places. Likewise, state that is required across a significant number of UI components ideally wouldn't need to be explicitly passed all over -- but flexible on this point.

- We now have simplified the recent registrations component to rely less on hooks and more on regular selector function calls. This allows using conditionals before calling selector functions, which was not possible with hooks, that must not occur after any conditional statement.
- All UI elements that allow variable input params also handle when some input params are not defined (which is expected to be handled at the UI component level).

Related to:
- #504 
- #380


apps/ensadmin/src/lib/wagmi.ts
```diff
@@ -1,56 +1,60 @@
...
-  chains: [mainnet, sepolia, holesky],
+  chains: [mainnet, sepolia, holesky, anvil],
   transports: {
-    [mainnet.id]: http(getRpcUrl(mainnet.id)),
-    [sepolia.id]: http(getRpcUrl(sepolia.id)),
-    [holesky.id]: http(getRpcUrl(holesky.id)),
+    [mainnet.id]: http(getEnsDeploymentRpcUrl("mainnet").toString()),
+    [sepolia.id]: http(getEnsDeploymentRpcUrl("sepolia").toString()),
+    [holesky.id]: http(getEnsDeploymentRpcUrl("holesky").toString()),
+    [anvil.id]: http(getEnsDeploymentRpcUrl("ens-test-env").toString()),
```
tk-o, line 58: Try applying ens-deployments package config:
```ts
Object.values(DeploymentConfigs).flatMap(config => Object.values(config).map(c => c.chain))
```

lightwalker-eth:
Closing as ideas can be found in the following other PRs:

- [PR 409](https://github.com/namehash/ensnode/pull/409)
- [PR 812](https://github.com/namehash/ensnode/pull/812)

Issue #281, opened 2025-02-19
remove unnecessary .onConflictDoNothing() calls


shrugs:
we included `.onConflictDoNothing(); // upsert for successful recovery when restarting indexing` for every 'event' entity creation due to some re-org related bug in early ponder 0.9.x versions

now that we're on `>0.9.17`, this bug is likely fixed and these calls shouldn't be necessary.

PR #478, opened 2025-04-02
chore: remove unnecessary on-conflict-do-nothing for events

shrugs:
closes https://github.com/namehash/ensnode/issues/281

the reason we added these was apparently related to resuming indexing after a crash, though based on ponder semantics they shouldn't be necessary at all (each event gets its own UUID, so there's no case in which that event with that id already exists) and we're now past the release that apparently required this hotfix

to verify that this only updates the correct `.onConflictDoNothing()` calls, check that each update is also marked with `// upsert for successful recovery when restarting indexing`


tk-o (approved):
LGTM

Issue #174, opened 2025-02-05
remove the manual snake_case for our column names?


shrugs:
- seems like the default should do it for us?

PR #479, opened 2025-04-02
chore: remove manual snake_case of column names, unnecessary

shrugs:
closes https://github.com/namehash/ensnode/issues/174

ponder always `snake_case`s column names, so specifying them is unnecessary. i confirmed schema definition equivalence with `pg_dump "postgresql://..." -n public --schema-only > local.sql` between main and this PR and they're identical, so the generate table names are the same


tk-o (approved):
LGTM

lightwalker-eth (approved):
@shrugs Looks good to me too :+1:

Issue #92, opened 2025-01-28
add tsconfig basepath to make imports look nicer


shrugs:
- and update imports
- can be used to import from workspace root?

PR #480, opened 2025-04-03
WIP: tsconfigs, path aliases

shrugs:
closes https://github.com/namehash/ensnode/issues/92

- adds tsconfig paths config for apps
  - can't use paths configs in packages because we want typescript to import them from source and it gets confused if two things try to import `@/lib/types` for example
- fixes some minor package.json spec things
- adds vitest configs so tests can resolve `@/` as well


tk-o (approved):
Awesome updates 🚀 LGTM

lightwalker-eth (approved):
@shrugs Hey looks great! Thank you! I see the CI is throwing an error for some dependency. Could you please investigate and fix that before we merge this in? Thank you 🙌

PR #487, opened 2025-04-04
feat: ens based theme for docs

notrab:


docs/ensnode.io/config/integrations/starlight.ts
```diff
@@ -149,7 +150,10 @@ export function starlight(): AstroIntegration {
       replacesTitle: true,
     },
     social: {
+      twitter: "https://x.com/NamehashLabs",
       github: "https://github.com/namehash/ensnode",
+      telegram: "https://t.me/namehash",
+      email: "mailto:hello@namehashlabs.org",
```
lightwalker-eth, line 156: Let's remove this email.

docs/ensnode.io/config/integrations/starlight.ts
```diff
@@ -149,7 +150,10 @@ export function starlight(): AstroIntegration {
       replacesTitle: true,
     },
     social: {
+      twitter: "https://x.com/NamehashLabs",
       github: "https://github.com/namehash/ensnode",
+      telegram: "https://t.me/namehash",
```
lightwalker-eth, line 155:
```suggestion
      telegram: "https://t.me/ensnode",
```

We've created a telegram group specifically for ENSNode support.

docs/ensnode.io/src/components/overrides/ThemeSelect.astro
```diff
@@ -0,0 +1,2 @@
+---
```
lightwalker-eth, line 1: My understanding is that creating this file is what turns off the light / dark mode picker? Is that right?

Could we comment the purpose of this file inside of this file? Appreciate your advice.

Also, what happens if a user's browser is configured by default to prefer dark mode? Will dark mode still be activated by default, even though we hide the theme picker at a UI level?

docs/ensnode.io/config/integrations/starlight.ts
lightwalker-eth, line 155: This telegram icon is a bit funny. It's also an inconsistent style as the other icons up here. The other icons use a "fill" style but this one is an "outline" style. Can we change this Telegram icon to look like the one we use in the footer at https://namehashlabs.org/ ?

lightwalker-eth (approved):
@notrab Thanks for these updates :+1: Shared a few suggestions with feedback. Please merge this when ready ✅

docs/ensnode.io/src/components/overrides/ThemeSelect.astro
notrab: Good point. I've added the `ThemeProvider` override so it's light mode only.

The component overrides "override" the default behaviour yeah. Empty `ThemeSelect` means there is no theme select now.

PR #488, opened 2025-04-04
chore: update GitHub workflows to trigger on specific paths for ENSRainbow and ENSIndexer images

djstrong:
1. It builds images only when appropriate files are changed. The drawback is that there is a lot of dependencies (paths) - maybe we can exclude some of them.
2. It also enables building test data ENSRainbow image (it is fast).
3. Scheduled testing ENSRainbow images on Mac and arm once per week.

### DIFF
--- a/.github/workflows/build_ensadmin_image.yml
+++ b/.github/workflows/build_ensadmin_image.yml
@@ -4,6 +4,19 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'apps/ensadmin/**'
+      - 'packages/ponder-metadata/**'
+      - 'packages/ponder-schema/**'
+      - 'packages/ens-deployments/**'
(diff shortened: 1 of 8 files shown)


shrugs:
are there any issues with running the builds all the time? are there resources that we're paying for and stuff like that? i feel like keeping this list of dependencies up to date will be difficult. unfortunately i can't find a simple solution to using pnpm to determine if a package has changed between git commits, the closest thing i can find is https://github.com/mkoreo/pnpm-filter/blob/v2/action.yml but it doesn't seem _simple_ and the action itself looks out of date.

how do you feel about the dependencies specified here, and how important is it for us to avoid building images on every pr?

tk-o (commented):
I wonder if we need to track specific files being changes within apps, or we could track some kind of build ID and build a docker image for the app only if the build ID has changed?

PR #494, opened 2025-04-04
feat: allow runtime config of start/end blocks

### LINKED ISSUE #489: END_BLOCK config + waiting in snapshot tools
- add END_BLOCK configuration to ensindexer
- add ready waiting to snapshot tool

shrugs:
closes #489 

- allows for usage like `END_BLOCK=... ponder start` to ensure ponder stops indexing at specific number for snapshot purposes


apps/ensindexer/src/lib/ponder-helpers.ts
```diff
@@ -34,6 +35,35 @@ export const constrainBlockrange = (
   endBlock: end,
 });
 
+/**
+ * Constrain indexing between the following start/end blocks, configured by START_BLOCK and END_BLOCK
+ * https://ponder.sh/docs/contracts-and-networks#block-range
+ *
+ * NOTE: it only really makes sense to use start/end blocks when running a single plugin
```
lightwalker-eth, line 42: @shrugs Could you please suggest some stronger "rails" for how these are configured? For example:

1. As I understand, the issue here isn't about running a single plugin. It's about indexing only a single chain. We shouldn't assume that there's a 1-to-1 relationship between chains and plugins.
2. Should we either:
   1. Enforce that setting a custom `START_BLOCK` or `END_BLOCK` while indexing multiple chains results in a helpful error message as soon as ENSIndexer starts up? OR;
   2. Modify the way these environment variables work such a chain id might be passed into `getBlockRange` which then would work to read from chain id suffixed environment variable names, similar as how we do things with RPC endpoints?
3. Appreciate your help to ensure we always keep our docs up to date with changes in our code. As I understand, we should ensure all possible environment variables we support are documented in places that include:
   1. https://ensnode.io/ensindexer/usage/configuration/
   2. https://github.com/namehash/ensnode/blob/main/apps/ensindexer/.env.local.example
   3. https://ensnode.io/ensnode/reference/subgraph-compatibility-tooling/
   4. https://github.com/namehash/ens-subgraph-transition-tools/blob/main/README.md

apps/ensindexer/src/lib/ponder-helpers.ts
```diff
@@ -34,6 +35,35 @@ export const constrainBlockrange = (
...
+export const getBlockRange = () => ({
+  startBlock: parseBlockheightEnvVar("START_BLOCK"),
+  endBlock: parseBlockheightEnvVar("END_BLOCK"),
+});
+
+/**
+ * Parses an env var into a blockheight for ponder.
+ *
+ * @param envVarName Name of the environment variable to parse
+ * @returns The parsed block number if valid, undefined otherwise
```
lightwalker-eth, line 56: @shrugs Suggest we document in more detail how `undefined` will ultimately be interpreted.

apps/ensindexer/src/lib/ponder-helpers.ts
```diff
@@ -34,6 +35,35 @@ export const constrainBlockrange = (
...
+/**
+ * Constrain indexing between the following start/end blocks, configured by START_BLOCK and END_BLOCK
+ * https://ponder.sh/docs/contracts-and-networks#block-range
+ *
+ * NOTE: it only really makes sense to use start/end blocks when running a single plugin
+ * (namely the eth plugin, in order to take snapshots). setting start/end blocks
+ * while running multiple plugins (which results in ponder indexing multiple chains) should be
+ * considered undefined behavior.
+ */
+export const getBlockRange = () => ({
```
lightwalker-eth, line 47: @shrugs If an ENSNode instance is operating under a constrained block range it should be visible in ENSAdmin.

We'll continue evolving the "Status" dashboard in ENSAdmin. Suggest for now we add a UI for visualizing these fields in the following panel:

CleanShot 2025-04-05 at 16 41 23

Advice appreciated on the possible permutations of start / end block range values that might be returned per-chain that an ENSIndexer instance is configured with. Ex: both undefined, either undefined, neither undefined, etc..

Also appreciate advice on how those state permutations should translate into values displayed in the UI.

Please feel welcome to transfer these ideas into a new issue for us to follow up on separately.

lightwalker-eth, line 47: @shrugs A few thoughts about this `getBlockRange` function:

1. Seems we should be validating certain invariants here about the relationship between start block and end block values. Suggest ensuring any error thrown would include a helpful error message.
2. Suggest adding some unit tests that validate our implementation of the invariant checking.
3. Suggest in general making the return types of our functions explicit. Appreciate your advice.

lightwalker-eth (changes requested):
@shrugs Thanks for updates here :+1: Reviewed and shared feedback 🙌

apps/ensindexer/src/lib/ponder-helpers.ts
shrugs: i've updated the docs for subgraph transition tools — personally i don't think that the ensnode docs should include this, as it's intended solely for snapshot purposes and developers using the snapshot tool will know about it from its documentation — wdyt?

apps/ensindexer/src/lib/ponder-helpers.ts
shrugs: tracking here for ENSAdmin updates https://github.com/namehash/ensnode/issues/531

lightwalker-eth (approved):

apps/ensindexer/ponder.config.ts
```diff
@@ -35,6 +40,30 @@ const activePluginsMergedConfig = activePlugins
...
+ACTIVE_PLUGINS=${requestedPluginNames().join(",")}
+START_BLOCK=${globalBlockrange.startBlock || "n/a"}
+END_BLOCK=${globalBlockrange.endBlock || "n/a"}
+
+The usage you're most likely interested in is:
+  ENS_DEPLOYMENT_CHAIN=(mainnet|sepolia|holesky) ACTIVE_PLUGINS=eth END_BLOCK=x pnpm run start
+which runs just the eth plugin with a specific end block, suitable for snapshotting ENSNode and comparing to Subgraph snapshots.
+
+In the future, indexing multiple networks with network-specific blockrange constraints may be possible.
+`,
```
tk-o, line 62: Great context 👍

PR #495, opened 2025-04-04
fix: snake_case db column names for intersection tables

shrugs:
- removing the manual snake_case accidentally broke an internal assumption in the subgraph compatible api


djstrong (approved):

lightwalker-eth (approved):
@shrugs Looks good :+1:

PR #497, opened 2025-04-04
fix: await handler activation

shrugs:
there was no bug caused by this behavior but it bothered me that we weren't actually awaiting the registration of the handlers before passing control back to ponder — so this makes sure that every ponder.on handler is registered before returning the config to ponder


lightwalker-eth (approved):
@shrugs Looks good :+1:

PR #498, opened 2025-04-04
chore: update readability of workflow names

### LINKED ISSUE #490: tidy workflows in actions tab

shrugs:
closes #490 

mostly so they'll sort (alphabetically) and be easier to read in the github actions page.

removed `type: string` from actions input definition because it's not a valid param https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#inputs


.github/workflows/release.yml
```diff
@@ -1,4 +1,4 @@
-name: Release
+name: Release Packages
```
lightwalker-eth, line 1: @shrugs @BanaSeba I'm not 100% clear right now on if this action is just related to releasing packages, or if its also related to releasing apps?

lightwalker-eth (approved):
@shrugs Thanks for updates :+1:

PR #500, opened 2025-04-04
Use CUSTOM_GITHUB_TOKEN for release

BanaSeba:
Mitigating problem described here: https://github.com/changesets/action/issues/187#issuecomment-1220791777


.github/workflows/release.yml
```diff
@@ -44,5 +39,5 @@ jobs:
           title: "chore(release): version apps"
           publish: pnpm release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
```
lightwalker-eth, line 42: @BanaSeba Could you help with a few follow ups here:

1. Could you add a description to PRs to help communicate what the goal of the PR is. I'm currently not so clear what the goal of PR 500 is.
2. Suggest we pick a more specific and self documenting name for `CUSTOM_GITHUB_TOKEN`. How is it custom exactly? I assume it's a special because it's used for publishing (write operations, not just read)? If so, maybe call it something like `GITHUB_PUBLISHING_TOKEN` or something else that's more self documenting.

lightwalker-eth (changes requested):
@BanaSeba Thanks for your update. Reviewed and shared feedback :+1:

.github/workflows/release.yml
BanaSeba: 1. Added problem description in PR summary
2. Renamed variable name

shrugs (approved):

lightwalker-eth (approved):

shrugs:
i'm going to assume this secret exists (can't see it myself) and merge this now

Issue #511, opened 2025-04-05
Allow switching connection after ENSNode invalid data error


tk-o:
### Steps to reproduce

1. Connect to a URL that's not going to reply with the correct data.
2. Try to switch to a different ENSNode connection.

### Actual result

User is no longer able to switch to a different ENSNode connection.

### Expected result

Each time a user picks a connection, ENSAdmin should try connecting to the selected ENSNode.

Issue #317, opened 2025-02-25
Create ENSNode SDK package


tk-o:
It includes:
- [all Ponder reserved routes](https://ponder.sh/docs/query/api-functions#reserved-routes):
  - `GET /health` returns 200 immediately when ponder starts running
  - `GET /ready` returns 200 when ponder completed historical indexing
  - `GET /status` returns status object (that we later replay with /GET metadata)
  - `GET /metics` returns prometheus metrics (we use some with GET /metadata)
- all ENSNode API routes
  - `GET /metadata` returns ENSNode metadata object ([dev-preview](https://ensnode.run.tko.box/metadata))
  - `POST /subgraph` handles Subgraph-compatible GQL queries
  - `POST /ponder` handles Ponder GQL queries

tk-o:
Ensure ENSNode SDK supports literal types and not use primitives if possible. For example:

> The type of ENS_DEPLOYMENT_CHAIN should change. It is not just any arbitrary string. It should be importing a a typed enum definition from the ens-deployments package.
>
> We should not be erasing important type information.

tk-o:
Please address this refactoring when introducing ENSNode SDK into `ponder-metadata-api` package:
- https://github.com/namehash/ensnode/pull/342/files#r1989651595

lightwalker-eth:
Suggestion from @shrugs https://github.com/namehash/ensnode/pull/524/files#r2033812435

tk-o:
Make sure `@ensnode/ens-deployments` package is validating chain values used for datasources:
- it has to ensure a chain object defines a block explorer URL
  - related comment: https://github.com/namehash/ensnode/pull/740/files#r2109381475

PR #524, opened 2025-04-07
fix(ensadmin): allow switching connections at all times

tk-o:
This PR introduces error messages management following these points:
1. A top-level `apps/ensadmin/src/app/error.tsx` file that catches all errors from nested app routes.
2. A single component handles error messages formatting for ENSNode APIs (see `apps/ensadmin/src/components/ensnode/components.tsx`)
3. A single component defines how error messages look like (see `apps/ensadmin/src/components/ui/error-message.tsx`).
4. ENSNode errors are defined with granular types (see `apps/ensadmin/src/components/ensnode/errors.ts`)

Also, to simplify search params management, the custom format of searchParams defined by Next.js is mapped into a native URLSearchParams object.

A note on passing URL-type values from server-components to client-components. Next.js does not support that, so as a workaround, whenever URL-type values are passed from server-components to client-components, we call `.toString()` method on the URL value so we could pass the value as a string primitive type.

### Preview

image

Resolves:
- #511


apps/ensadmin/src/app/gql/ponder/page.tsx
```diff
@@ -8,15 +8,8 @@ type PageProps = {
...
-  const baseUrl = Array.isArray(ensnode)
-    ? ensnode[0]
-    : typeof ensnode === "string"
-      ? ensnode
-      : defaultEnsNodeUrl();
-
-  const url = new URL(`/ponder`, baseUrl).toString();
-
-  return <GraphiQLEditor url={url} />;
+  return <GraphiQLEditor graphQlApiUrl={ensNodePonderApiUrl.toString()} />;
```
notrab, line 14: If I am to be pedantic, I'm not sure why we need to prefix this url with `graphQlApi` since graphiQL is a GraphQL UI, the context is already that the URL is for GraphQL API URLs only.

tk-o: Right, will take care of that 👍

apps/ensadmin/src/lib/env.ts
```diff
@@ -168,3 +168,20 @@ function parseUrl(maybeUrl: string): URL {
 export async function ensAdminVersion(): Promise<string> {
   return import("../../package.json").then(({ version }) => version);
 }
+
+/**
+ * Parses a custom Next.js search params object into a URLSearchParams object.
+ * @param nextJsPageSearchParams
+ * @returns {URLSearchParams}
+ */
+export function parseNextJsPageSearchParams(
```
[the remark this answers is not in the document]
notrab: I need to look at the code that invokes this function to understand the requirement better but Next.js provides `useSearchParams` which can be used inside of client components, and it provides the `searchParams` as part of page props, so it's uncertain why this specific function is needed.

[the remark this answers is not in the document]
tk-o: This function was needed to map `searchParam` object from Next.js page component props and pass it into a helper function that requires `URLSearchParams` type input.

apps/ensadmin/src/app/ponder-client-api/page.tsx
```diff
@@ -11,19 +11,12 @@ type PageProps = {
...
-
-  const baseUrl = Array.isArray(ensnode)
-    ? ensnode[0]
-    : typeof ensnode === "string"
-      ? ensnode
-      : defaultEnsNodeUrl();
-
-  const url = new URL(`/subgraph`, baseUrl).toString();
+  const ensNodeUrl = selectedEnsNodeUrl(parseNextJsPageSearchParams(await searchParams));
+  const ensNodePonderSqlApiUrl = new URL(`/sql`, ensNodeUrl);
```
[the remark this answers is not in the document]
tk-o: Keen to have this simplification achieved with:
- https://github.com/namehash/ensnode/issues/317

apps/ensadmin/src/lib/env.ts
```diff
@@ -168,3 +168,24 @@ function parseUrl(maybeUrl: string): URL {
 export async function ensAdminVersion(): Promise<string> {
   return import("../../package.json").then(({ version }) => version);
 }
+
+type NextJsPageSearchParams = Record<string, string | string[] | undefined>;
+
+/**
+ * Parses a custom Next.js search params object into a URLSearchParams object.
```
lightwalker-eth, line 175: Ok. But why do we want to do this? What's special about a URLSearchParams object?

Where does this type definition come from?

If our goal is to translate search params into a data model that is meaningful for our app, then why don't we define an interface for the data model that's meaningful for our app and have this function return that instead of whatever this URLSearchParams is supposed to be?

apps/ensadmin/src/components/graphiql-editor.tsx
```diff
@@ -3,13 +3,17 @@
...
-export function GraphiQLEditor({ url }: { url?: string }) {
-  if (!url || typeof window === "undefined") {
-    return null;
-  }
+const GraphiQL = dynamic(() => import("graphiql").then((m) => m.GraphiQL), {
+  ssr: false,
+});
 
+interface GraphiQLEditorProps {
+  url: string;
```
lightwalker-eth, line 13: What url is this? The URL to an ENSNode instance, or the URL to a specific GraphQL API endpoint served up by an ENSNode instance?

We should name this field more clearly.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -56,23 +56,25 @@ export function useIndexingStatusQuery(
  * @param response
  * @throws Error if the response is invalid
  */
-function validateResponse(response: EnsNode.Metadata) {
+function validateResponse(response: EnsNode.Metadata): EnsNode.Metadata {
```
lightwalker-eth, line 59: I thought we were using a library like zod to more deeply validate the response?

apps/ensadmin/src/components/ensnode/errors.ts
```diff
@@ -0,0 +1,26 @@
+export class ENSNodeConnectionError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "ENSNodeConnectionError";
+  }
+}
+
+export class ENSNodeResponseValidationError extends Error {
```
lightwalker-eth, line 8: Should this be changed to something like `UnsupportedENSNodeError`?

apps/ensadmin/src/components/ensnode/errors.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(message: string) {
+    super(message);
+    this.name = "ENSNodeResponseValidationError";
+  }
+}
+
+export const ENSNodeError = {
+  ensNodeConnectionError(message?: string) {
```
lightwalker-eth, line 16:
```suggestion
  connectionError(message?: string) {
```

apps/ensadmin/src/components/ensnode/errors.ts
```diff
@@ -0,0 +1,26 @@
...
+    return new ENSNodeConnectionError(
+      `Failed to connect to ENSNode${message ? `: ${message}` : ""}`,
+    );
+  },
+  ensNodeResponseValidationError(message?: string) {
```
lightwalker-eth, line 21:
```suggestion
  unsupportedENSNodeError(message?: string) {
```

apps/ensadmin/src/app/ponder-client-api/provider.tsx
```diff
@@ -3,20 +3,16 @@
 import * as ponderSchema from "@ensnode/ponder-schema";
 import { createClient } from "@ponder/client";
 import { PonderProvider } from "@ponder/react";
-import { useState } from "react";
+import { PropsWithChildren, useState } from "react";
 
-type ProviderProps = {
+interface ProviderProps extends PropsWithChildren {
   url: string;
```
lightwalker-eth, line 9: A url to what exactly? There's many ways to interpret this. Let's please be more specific in field naming and/or documentation.

apps/ensadmin/src/app/ponder-client-api/page.tsx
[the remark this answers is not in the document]
lightwalker-eth: Logged: https://github.com/namehash/ensnode/issues/317#issuecomment-2795182344

apps/ensadmin/src/app/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }, [error]);
+
+  if (ENSNodeErrorMessage.isHandledError(error)) {
+    return <ENSNodeErrorMessage error={error} reset={reset} />;
+  }
+
+  return (
+    <ErrorMessage>
+      <ErrorMessage.Header>Unknown error</ErrorMessage.Header>
+      <ErrorMessage.Message>An unknown error occurred.</ErrorMessage.Message>
```
lightwalker-eth, line 31: Should we first check if the `error` has a message defined, and if so, display that message?

apps/ensadmin/src/app/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { ErrorMessage } from "@/components/ui/error-message";
+import { useEffect } from "react";
+
+interface ErrorProps {
+  error: Error;
+  reset: () => void;
+}
+
+/**
+ * Renders an error message for unknown errors at the root of the app.
```
lightwalker-eth, line 13:
```suggestion
 * Renders an error message for any errors at the root of the app.
```

Is that right? And is it correct to really say we render it at the literal "root" of the app? Is this 100% technically correct that it's really at the root of the whole app?

apps/ensadmin/src/app/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+ * Note: It catches all unhandled errors from nested app routes.
+ *
+ * @param error The error to render.
+ * @returns The error message.
```
lightwalker-eth, line 17:
```suggestion
 * @returns a rendering of information about the error.
```

apps/ensadmin/src/app/error.tsx
```diff
@@ -0,0 +1,34 @@
...
+ */
+export default function Error({ error, reset }: ErrorProps) {
```
lightwalker-eth, line 19: Hmm.. seems we should rename this? This isn't an Error itself, right?

Why not call it something like `ErrorRenderer`? and rename this file accordingly?

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { useEffect } from "react";
+import { ENSNodeConnectionError, ENSNodeResponseValidationError } from "./errors";
+
+const handledErrors = [ENSNodeConnectionError, ENSNodeResponseValidationError] as const;
+
+type HandledError = (typeof handledErrors)[number];
+
+interface ENSNodeErrorMessageProps {
+  error: HandledError;
+  reset: () => void;
```
lightwalker-eth, line 11: Can we please document what this function is supposed to do?

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
+import { ErrorMessage } from "@/components/ui/error-message";
+import { useEffect } from "react";
+import { ENSNodeConnectionError, ENSNodeResponseValidationError } from "./errors";
+
```
lightwalker-eth, line 4: Hmm. should this file be named something more specific than "components.tsx"?

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+type HandledError = (typeof handledErrors)[number];
+
+interface ENSNodeErrorMessageProps {
+  error: HandledError;
+  reset: () => void;
+}
+
+/**
+ * Renders an error message for ENSNode errors.
```
lightwalker-eth, line 15:
```suggestion
 * Renders an error message for HandeledError errors.
```

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * @param error The error to render.
```
lightwalker-eth, line 16:
```suggestion
 * @param error The HandledError to render.
```

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * @returns The error message.
+ */
+export function ENSNodeErrorMessage({ error, reset }: ENSNodeErrorMessageProps) {
```
lightwalker-eth, line 19: I don't think it's right to say these are all going to be "ENSNode errors".

Shouldn't we rename "ENSNode..." to "ENSAdmin..."? This logic lives in ENSAdmin, not ENSNode.

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+      window.removeEventListener("ensnode/connection/set", reset);
+    };
+  }, [error, reset]);
+
+  if (error instanceof ENSNodeConnectionError) {
+    return (
+      <ErrorMessage>
+        <ErrorMessage.Header>ENSNode connection error</ErrorMessage.Header>
+        <ErrorMessage.Message>
+          ESNAdmin could not switch to selected ENSNode connection due to the following error:{" "}
```
lightwalker-eth, line 36:
```suggestion
          ENSAdmin could not connect to the selected ENSNode due to the following error:{" "}
```

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <em>{error.message}</em>.
+        </ErrorMessage.Message>
+        <ErrorMessage.Message>
+          Please try to switch to a different ENSNode connection.
```
lightwalker-eth, line 40:
```suggestion
          Please connect to a different ENSNode server.
```

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </ErrorMessage.Message>
+      </ErrorMessage>
+    );
+  }
+
+  if (error instanceof ENSNodeResponseValidationError) {
+    return (
+      <ErrorMessage>
+        <ErrorMessage.Header>ENSNode response validation error</ErrorMessage.Header>
```
lightwalker-eth, line 49: Please see my related comment suggesting this isn't just some generic response validation error, but rather that it's the case of an UnsupportedENSNode?

Do I understand that right or maybe I missed something.

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+  reset: () => void;
+}
+
+/**
+ * Renders an error message for ENSNode errors.
+ * @param error The error to render.
+ * @returns The error message.
+ */
+export function ENSNodeErrorMessage({ error, reset }: ENSNodeErrorMessageProps) {
+  // Reset the error when the ENSNode connection is changed
```
lightwalker-eth, line 20: What happens if there was an error that was not a HandledError? Would it still be correctly reset?

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
...
+  if (error instanceof ENSNodeResponseValidationError) {
+    return (
+      <ErrorMessage>
+        <ErrorMessage.Header>ENSNode response validation error</ErrorMessage.Header>
+        <ErrorMessage.Message>{error.message}</ErrorMessage.Message>
+      </ErrorMessage>
+    );
+  }
+
+  throw new Error(`Cannot render error message for unknown error: ${error.name}`);
```
lightwalker-eth, line 55:
```suggestion
  throw new Error(`Cannot render error that is not a HandledError: ${error.name}`);
```

apps/ensadmin/src/components/ensnode/components.tsx
```diff
@@ -0,0 +1,62 @@
+import { ErrorMessage } from "@/components/ui/error-message";
+import { useEffect } from "react";
+import { ENSNodeConnectionError, ENSNodeResponseValidationError } from "./errors";
+
+const handledErrors = [ENSNodeConnectionError, ENSNodeResponseValidationError] as const;
+
+type HandledError = (typeof handledErrors)[number];
```
lightwalker-eth, line 7: Suggest we rename this idea. We should handle all errors, not just `HandledError`.

What makes our special error types important isn't that they are handled vs not handled. It is that they are `ENSAdminError`. In other words, suggesting to rename `HandledError` to `ENSAdminError` because those are errors specifically known to ENSAdmin.

lightwalker-eth (changes requested):
@tk-o Thanks for updates. Suggest we put more effort into refining our ideas and language.

apps/ensadmin/src/lib/env.ts
```diff
@@ -168,3 +168,24 @@ function parseUrl(maybeUrl: string): URL {
 export async function ensAdminVersion(): Promise<string> {
   return import("../../package.json").then(({ version }) => version);
 }
+
+type NextJsPageSearchParams = Record<string, string | string[] | undefined>;
```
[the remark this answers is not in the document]
tk-o: This is a type that is defined by Next.js Page API:
https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional

I'll add that as a comment over the type definition 👍

apps/ensadmin/src/lib/env.ts
```diff
@@ -168,3 +168,24 @@ function parseUrl(maybeUrl: string): URL {
...
+ * @returns {URLSearchParams}
+ */
+export function parseSearchParams(searchParams: NextJsPageSearchParams): URLSearchParams {
+  const searchParamsNative = new URLSearchParams();
+
+  Object.entries(searchParams).forEach(([key, value]) => {
+    if (Array.isArray(value)) {
+      value.forEach((v) => searchParamsNative.append(key, v));
+    } else {
+      searchParamsNative.append(key, `${value}`);
```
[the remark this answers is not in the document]
tk-o: Of course, I'll add a comment that this function parses [the Next.js Page API property of `searchParams`](https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional) into a native `URLSearchParams` object.

apps/ensadmin/src/lib/env.ts
tk-o: Yeah, I think we should just rely on the value the URLSearchParams will need and not about the whole structure that holds that value within itself 👍

apps/ensadmin/src/components/graphiql-editor.tsx
tk-o: We had it the other way, and [other PR feedback](https://github.com/namehash/ensnode/pull/524#discussion_r2033801000) requested a change to the generic `url` prop name. I guess I'll come back to my initial version?

apps/ensadmin/src/components/ensnode/hooks.ts
tk-o: Not yet, we introduced schema-based parsing in a draft PR that is still waiting for me to come back to it:
- #476

tk-o: FWIW I've logged an issue to address the parsing problem:
- https://github.com/namehash/ensnode/issues/571

apps/ensadmin/src/components/ensnode/errors.ts
tk-o: Yes, it'd more descriptive this way 👍

apps/ensadmin/src/app/ponder-client-api/provider.tsx
tk-o: I've got other feedback on this PR that requested [simplifying naming](https://github.com/namehash/ensnode/pull/524/commits/16e0d4c70c9e9655b55983122a6ee1fd4b874ca2) of URL related props.

I guess I'll come back to my initial proposition of `ponderSqlApiUrl`

apps/ensadmin/src/app/error.tsx
tk-o: I believe it is correct, it's the `app/error.tsx`. If `app/` is the application root, then the `app/error.tsx` file is in the root of the whole app.

Next.js uses `app/` directory an an entrypoint to all app routes.

apps/ensadmin/src/app/error.tsx
tk-o: Well, any React component is a renderer, and when we have, for example, a `Button` component, it's technically a button renderer but yet is named after what it renders.

Perhaps we could call this component `ErrorSummary` or something similar?

apps/ensadmin/src/app/error.tsx
tk-o: Yes, we can do that. Assuming:
```ts
const errorMessage = error instanceof Error && error.message
  ? error.message
  : "Unknown error occurred.";
```

```suggestion
      <ErrorMessage.Message>{errorMessage}</ErrorMessage.Message>
```

apps/ensadmin/src/components/ensnode/components.tsx
tk-o: Ok, we can alter this name to be `ENSAdminError`, but I'd also not that this is not any ENSAdmin error, but a specific group of errors ENSAdmin could observe when interacting with ENSNode.

apps/ensadmin/src/components/ensnode/components.tsx
tk-o: In fact, these are the errors ENSAdmin can observe when interacting with ENSNode. It's hard for me to see `ENSAdminENSNodeError` being a good name. What other options do we have?

I believe that is ENSNodeError lives inside ENSAdmin app, the relation between these is that ENSAdmin app has its own understanding of ENSNode errors suitable for the ENSAdmin's domain.

apps/ensadmin/src/components/ensnode/components.tsx
tk-o: Connecting to another ENSNode requires ENSAdmin to reset presented errors, and let the ENSNode response to be parsed again. If parsing output is error, then we will see ENSAdmin to present error message again.

apps/ensadmin/src/components/ensnode/components.tsx
tk-o: I agree, saying explicitly the selected ENSNode connection was not supported is the correct way to go 👍

apps/ensadmin/src/app/ponder-client-api/page.tsx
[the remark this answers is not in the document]
tk-o: Thank you 🤝

PR #535, opened 2025-04-08
fix: move publish key around

shrugs:
- checks out repo with the PAT

### DIFF
--- a/.github/workflows/deploy_publish_changesets.yml
+++ b/.github/workflows/deploy_publish_changesets.yml
@@ -10,14 +10,14 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   release:
-    name: Release
-    if: github.repository == 'namehash/ensnode'
+    name: Publish Changesets
     runs-on: blacksmith-4vcpu-ubuntu-2204
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
+          token: ${{ secrets.PUBLISHING_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy_publish_changesets.yml
```diff
@@ -10,14 +10,14 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
   release:
-    name: Release
-    if: github.repository == 'namehash/ensnode'
+    name: Publish Changesets
```
lightwalker-eth, line 13: Does this job make a release to NPM? If so I see that as being different than "Publish Changesets". Appreciate your advice on the most precise language we could use to describe what this job does.

.github/workflows/deploy_publish_changesets.yml
```diff
@@ -31,13 +31,13 @@ jobs:
...
+      - name: Create Release PR and Publish to NPM
         id: changesets
         uses: changesets/action@v1
         with:
           commit: "chore(release): version apps"
           title: "chore(release): version apps"
           publish: pnpm release
         env:
-          GITHUB_TOKEN: ${{ secrets.PUBLISHING_GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
lightwalker-eth, line 42: Maybe better to switch this to be `PUBLISHING_GITHUB_TOKEN` and on the "Checkout repository" step use the regular `GITHUB_TOKEN`?

PR #540, opened 2025-04-08
fix: changeset linting

shrugs:
- disable biome for package.json
- format all package.jsons as changesets would (i.e. vscode default


lightwalker-eth (approved):
Nice fix!

PR #549, opened 2025-04-09
Tidy up Changesets Integration

### LINKED ISSUE #546: tidy up changesets
each build workflow triggers on main push, release create and release publish, and all of those happen at once when merging a changeset pr, which creates 9 build workflows, and only 3 are really necessary

shrugs:
closes #546

### DIFF
--- a/.github/actions/build_docker_image/action.yml
+++ b/.github/actions/build_docker_image/action.yml
@@ -1,9 +1,9 @@
 name: Build Docker image
-description: Builds docker image
+description: builds multi-arch docker image
 
 inputs:
   image_name:
-    description: 'Target Docker image name with tag'
+    description: 'Target Docker image name'
     required: true
 
   dockerfile_path:
@@ -26,11 +26,11 @@ runs:
       with:
         fetch-depth: 1
 
-    - name: Set up QEMU (for cross-architecture builds)
-      uses: docker/setup-qemu-action@v3
-
-    - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v3
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        images: ${{ inputs.image_name }}
 
     - name: Log in to GitHub Container Registry
       uses: docker/login-action@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/build_ensnode_images.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  workflow_dispatch:
+  release:
+    types: [published]
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  ensnode-build-and-push:
+    runs-on: blacksmith-4vcpu-ubuntu-2204
```
tk-o, line 11: I wonder if this CI Worker will have enough disk space to deal with the images to be built for ENSRainbow? 

I guess one way of figuring it out is to go live with this PR and see.

package.json
```diff
@@ -9,7 +9,7 @@
     "test": "pnpm -r --parallel test run",
     "typecheck": "pnpm -r --parallel typecheck",
     "changeset": "changeset",
-    "release": "changeset publish --filter='!@ensnode/shared-configs'",
+    "publish": "changeset publish",
```
tk-o, line 12: `changeset publish` creates one release per package:

https://github.com/changesets/changesets/blob/26c8ba9f58d3b123a19106771b4bdcb2800235cc/packages/assemble-release-plan/src/index.ts#L172-L174

tk-o (commented):
If we need a single github release including information about multiple packages release notes, we cannot use Changesets to handle that for us. Changesets create a release per tag, and since each released package gets its own tag, we end up having multiple releases created by Changesets at the same time (effectively, one per package).

To end up with a single Github Release including all packages that were updated, we'll need to opt-out from the flow where Changesets manage Github Releases. We will need a replacement automation that will track if Changesets'-created Pull Request got merged, and if so, read the pull request contents and create a single Github Release.

For example, [this Changesets-generated PR](https://github.com/namehash/ensnode/pull/537) include information about three newly created package tags. We can simply copy this PR contents below _Releases_ heading and put in to a new Github Release. All app packages are versioned with the same version number, so we could extract that number and included in to the Github Release title.

Thoughts on that approach?

.github/actions/build_docker_image/action.yml
```diff
@@ -18,19 +18,18 @@ inputs:
...
+    required: false
+
 runs:
   using: composite
   steps:
-    - name: Checkout repository
-      uses: actions/checkout@v4
+    - name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
```
BanaSeba, line 30: Very cool

BanaSeba (approved):
Let's give it a try

.github/workflows/build_ensnode_images.yml
```diff
@@ -0,0 +1,45 @@
+name: "Build: ENSNode Images"
+on:
+  workflow_dispatch:
+  release:
```
BanaSeba, line 4: I'm worried that we will be swarmed with both packages and app releases here. All of them will trigger this, and we will be building [sub_repos_count * 3] images at once.

.github/workflows/build_ensnode_images.yml
```diff
@@ -0,0 +1,45 @@
...
+    types: [published]
```
BanaSeba, line 5: Also not sure if metadata step is working well with release trigger

Issue #521, opened 2025-04-05
Refactor `ponder-metadata` package


lightwalker-eth:
Background info:

Approximately a week ago we were rushing to ship enhancements related to this panel:

![Image](https://github.com/user-attachments/assets/09dfd70b-4acf-4371-b9b1-ad147959803c)

As I remember, the way we implemented those enhancements made it so that the `ponder-metadata` package would no longer be "generic".

@tk-o Could you please review the `ponder-metadata` package and let us know if you suggest we make any changes to it? Please ensure there continues to be some mechanism available for ENSIndexer to share state about its connected ENSRainbow instance back with ENSAdmin.

Thanks

tk-o:
Related to #513

tk-o:
### Proposed next steps

- keep `ponder-metadata` package so the `ENSIndexerConfig` object can be built upon ENSIndexer start. It'll need access to Ponder metrics parser (`PrometheusMetrics.parse`), Ponder Status (`queryPonderStatus`), and Ponder Metadata (`queryPonderMeta`).


### Context

There's a new idea for `ponder-metadata` package role in the future, based on [this feedback](https://github.com/namehash/ensnode/pull/551/files#r2038314541). The package should focus on building Chain Indexing Status object for each configured chain when requested.

Also, there's a related issue:
- https://github.com/namehash/ensnode/issues/523

which aims at building `ENSIndexerConfig` object upon application start. This config object would need to include some values (i.e. ponder metrics, or ponder codebase build ID) that are currently produced by `ponder-metadata` package.

Finally, this issue:
- https://github.com/namehash/ensnode/issues/525

requires splitting up ENSNode endpoints so that we could served `ENSIndexerConfig` (fixed for ENSIndexer instance), and `ENSIndexerChainStatus`

lightwalker-eth:
@tk-o Thanks for update.

There's misunderstandings on important details here.

1. Why `chain-status` and not `indexing-status`? As I understand we want `indexing-status`. This status is about more than chains. It's more generally about the indexing status of those chains. We need to get the details of our language right.
[...]

PR #551, opened 2025-04-10
feat(ponder-metadata): make package generic again

tk-o:
This PR abstracts fetching ENSRainbow version away from the `ponder-metadata` package and moves it into the ENSIndexer application layer.

### Preview

| ENSRainbow info included in ENSNode response | ENSRainbow info not included in ENSNode response |
| - | - |
| image | image |

Resolves:
- #521

### DIFF
--- a/.changeset/eleven-parents-sing.md
+++ b/.changeset/eleven-parents-sing.md
@@ -1,0 +1,8 @@
+---
+"@ensnode/ponder-metadata": minor
+"ensindexer": minor
+"ensadmin": minor
+---
+
+Generalize `ponder-metadata` package and let ENSAdmin to ensure ENSRainbow version info is available.
+  
--- a/apps/ensadmin/src/components/ensnode/hooks.ts
+++ b/apps/ensadmin/src/components/ensnode/hooks.ts
@@ -90,6 +90,10 @@ function validateResponse(response: EnsNode.Metadata) {
         .join(", ")}`,
     );
   }
+
+  if (typeof response.deps.ensRainbow === "undefined") {
+    throw new Error(`ENSRainbow version not found in the response.`);
+  }
 }
 
 /**
--- a/apps/ensadmin/src/components/ensnode/types.ts
+++ b/apps/ensadmin/src/components/ensnode/types.ts
@@ -1,4 +1,5 @@
 import type { ENSDeploymentChain, ENSDeploymentConfig } from "@ensnode/ens-deployments";
+import type { EnsRainbow } from "@ensnode/ensrainbow-sdk";
 import type * as PonderMetadata from "@ensnode/ponder-metadata";
 
 /**
@@ -12,7 +13,12 @@ export namespace EnsNode {
   /**
    * The status of the ENS node.
    */
-  export interface Metadata extends Omit<PonderMetadata.MetadataMiddlewareResponse, "env"> {
+  export interface Metadata extends PonderMetadata.MetadataMiddlewareResponse {
+    /** Application dependencies info */
+    deps: PonderMetadata.MetadataMiddlewareResponse["deps"] & {
+      ensRainbow: EnsRainbow.VersionInfo;
+    };
+
     // override the `env` field to include the fields required by the ENSAdmin client
     env: {
       ACTIVE_PLUGINS: Array<keyof ENSDeploymentConfig>;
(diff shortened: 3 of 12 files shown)


apps/ensadmin/src/components/ensnode/types.ts
```diff
@@ -12,7 +13,12 @@ export namespace EnsNode {
   /**
    * The status of the ENS node.
    */
-  export interface Metadata extends Omit<PonderMetadata.MetadataMiddlewareResponse, "env"> {
+  export interface Metadata extends PonderMetadata.MetadataMiddlewareResponse {
+    /** Application dependencies info */
+    deps: PonderMetadata.MetadataMiddlewareResponse["deps"] & {
+      ensRainbow: EnsRainbow.VersionInfo;
+    };
```
tk-o, line 20: We declare application-level type to specify a generic type coming from the `ponder-metadata` package.

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -90,6 +90,10 @@ function validateResponse(response: EnsNode.Metadata) {
         .join(", ")}`,
     );
   }
+
+  if (typeof response.deps.ensRainbow === "undefined") {
+    throw new Error(`ENSRainbow version not found in the response.`);
+  }
```
tk-o, line 96: Including response validation to detect if ENSRainbow version info is available. If it's not, ENSAdmin would refuse to connect.

.changeset/eleven-parents-sing.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ensnode/ponder-metadata": minor
+"ensindexer": minor
+"ensadmin": minor
+---
+
+Generalize `ponder-metadata` package and let ENSAdmin to ensure ENSRainbow version info is available.
```
lightwalker-eth, line 7:
```suggestion
Generalize `ponder-metadata` package. Add rules to ENSAdmin to ensure ENSRainbow version info is available.
```

apps/ensadmin/src/components/ensnode/hooks.ts
lightwalker-eth: @tk-o This isn't a full validation of the data model. I expected we would completely verify each field and value.

apps/ensindexer/src/lib/ponder-helpers.ts
```diff
@@ -194,9 +194,10 @@ export const createEnsRainbowVersionFetcher = () => {
     endpointUrl: new URL(ensRainbowEndpointUrl()),
   });
 
-  return async () => {
+  return async function ensRainbowVersionFetcher(): Promise<EnsRainbow.VersionInfo> {
     try {
       const versionResponse = await client.version();
+
       return {
```
lightwalker-eth, line 201: Flagging how several lines below this we can potentially return junk values if an error is thrown. This doesn't seem right. If we can't query the ENSRainbow version for some reason then it seems we should handle that in a smarter way that helps to build up our state invariants.

State invariants are CRITICAL for us. We need to focus on them much more.

packages/ponder-metadata/src/middleware.ts
```diff
@@ -23,6 +22,9 @@ interface PonderMetadataModule {
     version: string;
   };
 
+  /** Application dependencies info */
+  DepsInfo: Record<string, unknown>;
```
lightwalker-eth, line 26: Really don't like type definitions like this that use `unknown`.

I assume a type definition like this is very temporary until we split our `/metadata` endpoint into separate APIs for `/indexing-status` and `/config`?

`/config` should not be generalized for any Ponder app. It should be 100% specific to ENSIndexer.

Likewise, the `ponder-metadata` package maybe should be renamed such that it is only for supporting the `/indexing-status` API.

lightwalker-eth (changes requested):
@tk-o Thanks for updates. Reviewed and shared feedback :+1:

apps/ensindexer/src/lib/ponder-helpers.ts
tk-o: I've started working on list of invariants. More details here:
- https://github.com/namehash/ensnode/issues/571#issuecomment-279998625

packages/ponder-metadata/src/middleware.ts
tk-o: If we assume `ponder-metadata` only deals with indexing status, then we can delegate the config type definition to ENSIndexer app, so `ponder-metadata` knows nothing about it and only passes-through the data (& its application-defined type).

tk-o (commented):
Thanks for feedback, @lightwalker-eth 🤝 I'll park this PR until #571 is done.

PR #605, opened 2025-04-14
fix: add _meta to subgraph compat endpoint

### LINKED ISSUE #178: bring the _meta field back
- need access to metadata store again

shrugs:
closes #178 
- uses data provider pattern via tko

### DIFF
--- a/apps/ensindexer/src/lib/ponder-metadata-provider.ts
+++ b/apps/ensindexer/src/lib/ponder-metadata-provider.ts
@@ -1,0 +1,95 @@
+import packageJson from "@/../package.json";
+
+import { db, publicClients } from "ponder:api";
+import type { PublicClient } from "viem";
+
+import {
+  createEnsRainbowVersionFetcher,
+  createFirstBlockToIndexByChainIdFetcher,
+  createPrometheusMetricsFetcher,
+  getEnsDeploymentChainId,
+  ponderDatabaseSchema,
+  ponderPort,
+} from "@/lib/ponder-helpers";
+import { PrometheusMetrics, queryPonderMeta, queryPonderStatus } from "@ensnode/ponder-metadata";
+import type { PonderMetadataProvider } from "@ensnode/ponder-subgraph";
+
+// setup block indexing status fetching
+export const fetchFirstBlockToIndexByChainId = createFirstBlockToIndexByChainIdFetcher(
+  import("@/../ponder.config").then((m) => m.default),
+);
+
+// setup ENSRainbow version fetching
+export const fetchEnsRainbowVersion = createEnsRainbowVersionFetcher();
+
+// setup prometheus metrics fetching
+export const fetchPrometheusMetrics = createPrometheusMetricsFetcher(ponderPort());
+
+export const makePonderMetdataProvider = (): PonderMetadataProvider => {
+  // get the chain ID for the ENS deployment
+  const ensDeploymentChainId = getEnsDeploymentChainId();
+
+  /**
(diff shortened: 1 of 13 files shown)


shrugs:
@tk-o does ponder meta/status export any value i could use for `hasIndexingErrors`? don't see anything obvious

packages/ponder-subgraph-api/src/graphql.ts
```diff
@@ -584,6 +600,48 @@ export function buildGraphQLSchema(
...
+            }),
+          ),
+        },
+        deployment: { type: new GraphQLNonNull(GraphQLString) },
+        hasIndexingErrors: { type: new GraphQLNonNull(GraphQLBoolean) },
+      },
+    }),
+    resolve: async (_source, _args, context) => {
+      const status = await queryPonderStatus(metaConfig.schema, context.drizzle);
+      const mainnetStatus = status.find((status) => status.network_name === "1");
```
tk-o, line 627: Does that mean we only need to support metadata for the mainnet ENS Deployment, and we don't need support for Sepolia or Holesky deployments?

tk-o (approved):
Looks awesome to me 🚀

apps/ensindexer/src/lib/ponder-metadata-provider.ts
```diff
@@ -0,0 +1,95 @@
...
+    return meta.build_id;
+  };
+
+  /**
+   * Check if there are any indexing errors logged in the prometheus metrics
+   * @returns true if there are no indexing errors, false otherwise
+   */
+  const hasIndexingErrors = async () => {
+    const metrics = PrometheusMetrics.parse(await fetchPrometheusMetrics());
+    return metrics.getValue("ponder_indexing_has_error") === 1;
```
lightwalker-eth, line 86: Very interesting!

Suggest we create an issue to remind us to follow up on this with the Ponder team to verify our understanding of what this really signifies is correct.

My understanding is that the way this field works in GraphNode / the ENS Subgraph is it is used to identify that the indexing has stopped and is broken. When Ponder encounters an indexing error, as I understand it just terminates the `ponder start` process. GraphNode / the ENS Subgraph works differently. If there's an indexing error, the GraphNode process doesn't terminate itself. It just stops any effort to index the subgraph that has an indexing error and reflects that state in this field.

apps/ensindexer/src/lib/ponder-metadata-provider.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    return publicClient;
+  }
+
+  /**
+   * Get the last block indexed by Ponder.
+   *
+   * @returns the block info fetched from the public client
+   */
+  const getLastIndexedBlock = async () => {
```
lightwalker-eth, line 53: Suggest we strengthen the docs / function naming here to make it more clear how this function as currently written gets the last indexed block just for the root ENS deployment chain.

In other words, I understand we're working on goals right now that are for subgraph compat which is single chain only. In our context with Ponder suggest we take special care to describe how we're mapping from multichain contexts to a single chain context for subgraph compat.

lightwalker-eth (approved):
@shrugs Cool! Nice work 🚀 Shared a few small suggestions 🙌

apps/ensindexer/src/lib/ponder-metadata-provider.ts
shrugs: this might be different in the case of the api service for `ponder serve` — not sure how this behaves #613

PR #608, opened 2025-04-15
fix(ensnode.io): browser svg warning

notrab:
This PR fixes a console warning for an invalid React SVG, and removes the pagination shadow.


Y3drk (approved):
Looks legit to me. 

Checked the browser preview also, and all is well

PR #609, opened 2025-04-15
feat(subgraph): _meta resolver data provider

tk-o:
This PR introduces a data provider interface for `ponder-subgraph-api` package. With this change, it's the application's (i.e. ENSIndexer's) responsibility to provide async methods to fetch current data for Subgraph's `_meta` field. This data includes :
- the last indexed block info (deriver from Ponder status DB table and RPC request),
- Ponder build ID (derived from Ponder metadata DB table),
- if there were any indexing errors (derived from Ponder metrics).

With these changes, the `ponder-subgraph-api` package does not a direct dependency on the `ponder-metadata` packages. They are integrated on the ENSIndexer application level.

Also, `_meta` field can now be built for all supported ENS Deployments: `mainnet`, `sepolia`, `holesky`, and `ens-test-env`.

### DIFF
--- a/apps/ensindexer/src/api/index.ts
+++ b/apps/ensindexer/src/api/index.ts
@@ -2,39 +2,33 @@ import packageJson from "@/../package.json";
 
 import { db, publicClients } from "ponder:api";
 import schema from "ponder:schema";
-import {
-  createEnsRainbowVersionFetcher,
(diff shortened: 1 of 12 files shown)


shrugs (approved):
thanks!

PR #611, opened 2025-04-15
deps: ponder and friends to latest

### LINKED ISSUE #396: ENSNode: Remove Manual Arg Parsing for Null Bytes Name Argument
ponder intentionally removes null bytes to spare users the footgun of inserting null bytes into postgres. We don't like this behavior, though, because it's important that 'vitalik\x00'.eth and vitalik.eth are differentiable.

https://github.com/ponder-sh/ponder/issues/1456

we mark these locations with 

```
// NOTE(name-null-bytes): manually decode args that may contain null bytes
```

and if/when ponder fixes this issue we should remove the manual decoding at those sites

### LINKED ISSUE #474: update ponder to 0.1+
- update to ponder 0.1+
- confirm via snapshots
- also #396

shrugs:
closes #474 
closes #396 

- [x] confirm via snapshot


tk-o (approved):
LGTM

pnpm-workspace.yaml
```diff
@@ -5,17 +5,17 @@ packages:
...
   "@vitest/coverage-v8": ^3.1.1
-  drizzle-orm: ^0.39.3
-  hono: ^4.6.14
+  drizzle-orm: =0.41.0
+  hono: ^4.7.6
   typescript: ^5.7.3
   viem: ^2.22.13
   vitest: ^3.1.1
-  ponder: ^0.9.27
+  ponder: 0.10.14
```
tk-o, line 18: Let's update `ponder` lib and its releated dependencies to `0.10.18` — this release include index creation timeout bugfix.

lightwalker-eth (approved):
@shrugs Nice :+1:

Issue #386, opened 2025-03-13
ENSRainbow: Rearchitect w/ Data Dir Downloads


shrugs:
data docker image too large

instead of publishing a data docker image we publish various data dirs (`lite`, `subgraph`, `v2`, ...etc?) on R2 via `git annex`, make app image download/cache the appropriate data dir to a well-known directory, allow users to mount a persistent volume at that location. git workflow that downloads/ingests/uploads each data dir to R2. update documentation to suggest persistent volume in prod. local dev docker compose should mount local directory. example docker compose for external devs should use persistent temp dir ala postgres.

Issue #162, opened 2025-02-03
ENSRainbow: Add versioning of set of healable labels


djstrong:
ENSRainbow should support multiple versions, where each version defines a specific set of healable labels:

- V1: Strict compatibility with ENS rainbow tables used by the ENS Subgraph
- Future versions will expand the healable label set to include:
  - Current DNS names
  - All known ENS labels
  - Dynamic label generators (e.g., combinations of words and emojis)

This versioning approach allows for controlled expansion of healable labels while maintaining compatibility with existing ENS Subgraph functionality in V1.

lightwalker-eth:
@djstrong Thanks for recording this opportunity :+1:

Our key goals on this issue include:
1. Ensure that an ENSNode deployment always receives a consistent representation of healable labels from ENSRainbow.
   1. This includes enabling ENSNode to access the representation of healable labels from ENSRainbow that give full backwards compatibility with the ENS Subgraph.
2. Enable the set of healable labels within ENSRainbow to grow across time.

These goals are not contradictory if we design this right 😄 

Current general idea is (writing this quick and dirty for now):
1. Define standardized file format for an "ENSRainbow Label Set".
2. Build script that takes the ENS Subgraph .sql file and converts it to our "ENSRainbow Label Set" format.
3. Modify ingest logic inside ENSRainbow so that it stops reading .sql files from the ENS Subgraph and instead exclusively reads files with the "ENSRainbow Label Set" format.
4. Apply simplifying assumption: NameHash Labs will assume responsibility for being the exclusive "official" publisher of "ENSRainbow Label Set" files. All files we publish are uploaded to our R2 bucket.
5. Each "ENSRainbow Label Set" includes an "id" field that (uniquely) identifies the label set. This might be a simple incrementing integer, for example.
6. Apply simplifying assumption: If ENSRainbow has ingested "ENSRainbow Label Set" N, it is guaranteed that it has also ingested all "ENSRainbow Label Set" from 1...(N-1).
7. The first time ENSRainbow ingests a label, it stores info about the id of the "ENS Rainbow Label Set" it was first published in.
8. Add configuration to ENSNode for the "ENSRainbow Label Set" id it wishes to use. Ex: id 1 might represent the "ENSRainbow Label Set" with backwards compatibility to the ENSSubgraph.
9. Add a param to the ENSRainbow heal API to identify what "ENSRainbow Label Set" should be used to generate the response:
[...]

PR #612, opened 2025-04-15
ENSRainbow v2 data format

djstrong:
Main problems: 
* ENSRainbow docker images are too large, building and downloading is too slow.
* Need to support the ability to grow the set of healable labels across time while also maintaining the ability for ENSIndexer to achieve deterministic indexing results across time.
* Need to support any number of namespaces of healable label sets to be defined, allowing for different categorizations and sources of healable labels.

Solution: 
* Build only one ENSRainbow image without data and download data on first startup
* Introduce new data format supporting label set evolution and multi-namespace support

Implementation:
* new data format (`.ensrainbow`) has been designed instead of using SQL dumps - it allows storing all labels (also including null bytes); for now the only way to create the `.ensrainbow` file is to convert from SQL dump
* each `.ensrainbow` file has assigned `namespace` name and `label_set` number - this allows publishing incremental .ensrainbow files, e.g. every month with new healed names
* `.ensrainbow` ingestion process needs correct namespace and label_set
* `.ensrainbow` files and databases (with ingested names) are published on R2 - docker image downloads ready to use database so long ingestion process is not needed
* database stores information about label_set for each label so we can heal labels with limited the highest label set
* if you want to run new container with different namespace or labelset but the same persistent directory then you need to clean the directory

* .ensrainbow files are stored on R2 under path `labelsets/${NAMESPACE}_${LABEL_SET}.ensrainbow`, the file format stores metadata: format version, namespace and label_set - if we change the file format, e.g. with field indicating if a label is normalized (https://github.com/namehash/ensnode/issues/163) then we need to change namespace name (e.g. by adding `v2` suffix)
* databases are stored on R2 under path `databases/${SCHEMA_VERSION}/${NAMESPACE}_${LABEL_SET}.tgz` - the archive contains directory called 
`data-${NAMESPACE}_${LABEL_SET}` and database files in it
* `entrypoint.sh`, `download-ensrainbow-files.sh`, `download-database-artifact.sh` are not unit tested
* railway configuration needs to be adjusted
* extended label set is not published yet

Check documentation in `docs/ensnode.io/src/content/docs/ensrainbow`

Issue: https://github.com/namehash/ensnode/issues/162
Discussion on Slack: https://namehash.slack.com/archives/C086Z6FNBHN/p1744020659189489

Related issue: https://github.com/namehash/ensnode/issues/386

### Follow up tasks:
- [ ] create the searchlight label set
- [ ] use labelSet info in ENSAdmin


docker-compose.yml
```diff
@@ -94,3 +96,5 @@ services:
 volumes:
   postgres_data:
     driver: local
+  ensrainbow_data:
```
BanaSeba, line 99: We should probably add this directory to git ignore

.github/workflows/test_ensrainbow_image_on_mac.yml
```diff
@@ -33,12 +33,21 @@ jobs:
...
+          CONTAINER_ID=$(docker run -d \
+            -v ~/my_ensrainbow_test_data:/app/apps/ensrainbow/data \
+            -e SCHEMA_VERSION="v3" \
+            -e NAMESPACE="testenv" \
+            -e LABEL_SET="0" \
+            -p 3223:3223 \
+            ghcr.io/namehash/ensnode/ensrainbow:latest)
           echo "Container ID: $CONTAINER_ID"
 
           # Wait for container to start up
```
BanaSeba, line 53: Now it will certainly take much longer than 30 seconds to start

apps/ensrainbow/Dockerfile
```diff
@@ -1,48 +1,54 @@
...
+
+# Default port, can be overridden by PORT env var for the entrypoint/serve command
 EXPOSE 3223
 
-# serve
-CMD ["pnpm", "run", "serve"]
+# Set the entrypoint to the new script
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+
+# CMD is effectively handled by the exec call in entrypoint.sh
```
BanaSeba, line 53: I guess we can remove those two commented lines?

.github/workflows/test_ensrainbow_image_on_mac.yml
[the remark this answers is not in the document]
djstrong: With testenv it takes 8 seconds to start on my machine.

djstrong:
BugBot run

apps/ensrainbow/download-database-artifact.sh
```diff
@@ -0,0 +1,153 @@
...
+    usage
+fi
+
+SCHEMA_VERSION="$1"
+NAMESPACE="$2"
+LABEL_SET="$3"
+
+# Configuration
+OUT_DIR="${OUT_DIR:-.}" # Default output directory, can be overridden e.g., OUT_DIR="data" ./script.sh ...
+BASE_URL="https://bucket.ensrainbow.io"
```
lightwalker-eth, line 26: We need to remove this hostname as a point of hardcoded centralization of ENSRainbow.

In other words, we need to:
1. Define terminology for this idea. Ex: An ENSRainbow Labelset Server. Open to other ideas.
2. Make it dynamically configurable across all the relevant code which ENSRainbow Labelset Server is used. For example, this should become an environment variable or otherwise become configurable in the relevant places.
3. Note how anyone should be able to run their own instance of an ENSRainbow Labelset Server. Ideally we would include lots of docs for this.

apps/ensrainbow/download-database-artifact.sh
```diff
@@ -0,0 +1,153 @@
...
+# Derived local target paths
+LOCAL_ARTIFACT_DIR_PATH="$OUT_DIR/databases/${SCHEMA_VERSION}"
+TARGET_DATA_FILE_PATH="${LOCAL_ARTIFACT_DIR_PATH}/${DATA_FILE_BASENAME}"
+TARGET_CHECKSUM_FILE_PATH="${LOCAL_ARTIFACT_DIR_PATH}/${DATA_FILE_BASENAME}.sha256sum"
+TARGET_LICENSE_FILE_PATH="$OUT_DIR/${SERVER_LICENSE_PATH}" # License in base OUT_DIR
+
+# Create data directories if they don't exist
+mkdir -p "$LOCAL_ARTIFACT_DIR_PATH" # This will also create $OUT_DIR and $OUT_DIR/databases if they don't exist
+# If OUT_DIR is just ".", license will be in current dir, which is fine.
+# If OUT_DIR is "data", license will be in "data/THE_GRAPH_LICENSE.txt". mkdir -p on LOCAL_ARTIFACT_DIR_PATH handles "data/"
```
lightwalker-eth, line 52: Please see my related comment about making the license file more general.

apps/ensrainbow/download-database-artifact.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script downloads a specific ENS database artifact (.tgz), its checksum,
```
lightwalker-eth, line 4: Can we please document the goal / motivation for creating this script or having these prebuilt databases?

apps/ensrainbow/download-database-artifact.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
lightwalker-eth, line 1: Should we move all of these scripts into a subdirectory such as `/scripts`? Appreciate your advice.

apps/ensrainbow/download-database-artifact.sh
```diff
@@ -0,0 +1,153 @@
...
+else
+    echo "❌ CRITICAL ERROR: Checksum FAILED after download for $DATA_FILE_BASENAME using $(basename "$TARGET_CHECKSUM_FILE_PATH")."
+    echo "The downloaded data file may be corrupted or incomplete."
+    # Consider cleaning up $TARGET_DATA_FILE_PATH here as well
+    exit 1
+fi
+
+echo ""
+echo "---------------------------------------------------"
+echo "ENS Database Artifact download and verification complete."
```
lightwalker-eth, line 145:
```suggestion
echo "ENS Database Artifact download and checksum verification complete."
```

Suggesting this as we only verified the checksum. In theory it might still not be a valid database for ENSRainbow.

apps/ensrainbow/download-ensrainbow-files.sh
```diff
@@ -0,0 +1,154 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script downloads a specific ENS Rainbow labelset, its checksum, and a license file.
```
lightwalker-eth, line 4:
```suggestion
# This script downloads a specific ENS Rainbow labelset, its checksum, and license file.
```

apps/ensrainbow/download-ensrainbow-files.sh
```diff
@@ -0,0 +1,154 @@
+#!/bin/bash
```
lightwalker-eth, line 1: Please apply all the feedback I shared in `apps/ensrainbow/download-database-artifact.sh` into this file as well. Thanks!

apps/ensrainbow/entrypoint.sh
```diff
@@ -0,0 +1,127 @@
+#!/bin/bash
```
lightwalker-eth, line 1: Should we be referencing this script (or any of the other scripts) inside `apps/ensrainbow/package.json`?

lightwalker-eth (changes requested):
@djstrong So excited by this PR! 🚀 Lots of huge upgrades. Here's some initial feedback 😄

apps/ensrainbow/.env.local.example
```diff
@@ -1,3 +1,9 @@
 # Define log level to manage log output
 # Allowed values: fatal, error, warn, info, debug, trace, silent
 LOG_LEVEL=info
+# The database schema version (e.g., `v3`, `v2`).
+SCHEMA_VERSION=v3
+# The data namespace identifier that categorizes the type of ENS data (e.g. `subgraph`, `searchlight`, `discoveryA`, `testenv`
```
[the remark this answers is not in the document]
djstrong: We agreed to rename `namespace` to `LABEL_SET_ID`.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -224,14 +252,32 @@ export class EnsRainbowApiClient implements EnsRainbow.ApiClient {
...
    */
-  async heal(labelHash: LabelHash): Promise<EnsRainbow.HealResponse> {
+  async heal(
+    labelHash: LabelHash,
+    options?: {
+      label_set_version?: number;
+      label_set_id?: string;
+    },
+  ): Promise<EnsRainbow.HealResponse> {
     const cachedResult = this.cache.get(labelHash);
```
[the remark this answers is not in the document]
djstrong: @lightwalker-eth Should we address this problem?

apps/ensrainbow/src/lib/database.ts
```diff
@@ -590,9 +822,23 @@ export class ENSRainbowDB {
...
 
-  public async addRainbowRecord(label: string) {
+  /**
+   * Adds a rainbow record to the database. Labelhash is computed from the label.
+   *
+   * @param label The label to add (without label set prefix)
+   * @param labelSetVersion The label set version number to associate with this label
+   * @throws Error if labelSetVersion is invalid
+   */
+  public async addRainbowRecord(label: string, labelSetVersion: number): Promise<void> {
```
lightwalker-eth, line 832: Suggest defining an interface for `RainbowRecord` with two fields: `label` and `labelSetVersion`. Then we can document the invariants and data model constraints of those fields on the interface definition.

We can then have a little utility function such as `buildRainbowRecord` that performs the validation logic that's currently in this `addRainbowRecord` function. Unit tests can then be defined on `buildRainbowRecord`.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -29,6 +41,33 @@ export const IngestionStatus = {
...
 
+/**
+ * Splits a label string into its label set version number and actual label components.
+ * Format of input is expected to be "{labelSet}:{actualLabel}"
+ *
+ * @param label The label string to split
+ * @returns An object containing the label set version number and the actual label
+ * @throws Error if the label format is invalid or the label set version is not a valid number
+ */
+export function splitLabelString(label: string): { labelSetVersion: number; label: string } {
```
lightwalker-eth, line 52: Several suggestions here:

1. the value passed into this function shouldn't be called `label`. Suggest to call it something else such as an `EncodedRainbowRecord`.
2. Please see my other comments suggesting the creation of a `RainbowRecord` interface containing the `labelSetVersion` and `label` fields.
3. The `splitLabelString` function could then be renamed to something like `decodeRainbowRecord` and it could return a `RainbowRecord` from an `EncodedRainbowRecord` string.
4. Suggest moving all logic that's specifically about these "RainbowRecord" type and related helper functions to a distinct file separate from database.ts.

docs/ensnode.io/src/content/docs/ensrainbow/deploying/docker.mdx
```diff
@@ -7,38 +7,168 @@ sidebar:
...
+      - ensrainbow_data:/app/apps/ensrainbow/data
+    ports:
+      - "3223:3223"
+    restart: unless-stopped
+
+volumes:
+  ensrainbow_data:
+```
+
+For test environment:
```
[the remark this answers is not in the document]
djstrong: It was meant for ENS Labs team to use with ens-test-env.

apps/ensrainbow/src/lib/database.ts
djstrong: I have created `RainbowRecordValue` or maybe it should be `RainbowRecordLabelWithLabelSetVersion`?
Because we have definied `RainbowRecord` in .ensrainbow files, which consists of only labelhash and label.

apps/ensrainbow/src/utils/protobuf-schema.ts
```diff
@@ -0,0 +1,39 @@
...
+  const root = new protobuf.Root();
+
+  // Define the RainbowRecord message type as a Type
+  const RainbowRecord = new protobuf.Type("RainbowRecord");
+  RainbowRecord.add(new protobuf.Field("labelhash", 1, "bytes"));
+  RainbowRecord.add(new protobuf.Field("label", 2, "string"));
+
+  // Define the RainbowRecordCollection message type as a Type
+  const RainbowRecordCollection = new protobuf.Type("RainbowRecordCollection");
+  RainbowRecordCollection.add(new protobuf.Field("data_format_version", 1, "uint32"));
```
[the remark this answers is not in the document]
djstrong: I have changed name of the field to `ensrainbow_file_format_version`.

docs/ensnode.io/src/content/docs/ensrainbow/deploying/docker.mdx
```diff
@@ -7,38 +7,168 @@ sidebar:
...
 ```
 
 The service will be available at `http://localhost:3223`.
 
 :::note[Image Size]
-The ENSRainbow Docker image embeds a set of rainbow tables for healing unknown labels, bringing the image size to `8.6GB`.
+The ENSRainbow storage needed is from 1 MB to dozens GB depending on the label set ID and label set version.
+## Production Setup
+
+For ENS Subgraph compatibility use the `subgraph` label set ID:
```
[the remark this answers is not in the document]
djstrong: This `searchlight` label set is not prepared/uploaded yet.

docs/ensnode.io/src/content/docs/ensrainbow/deploying/docker.mdx
```diff
@@ -7,38 +7,168 @@ sidebar:
...
-    - "3223:3223"
-  restart: unless-stopped
+version: '3.8'
+
+services:
+  ensrainbow:
+    container_name: ensrainbow_test
+    image: ghcr.io/namehash/ensnode/ensrainbow:latest
+    environment:
+      - DB_SCHEMA_VERSION=v3
```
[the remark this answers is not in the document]
djstrong: Yes, changed.

docs/ensnode.io/src/content/docs/ensrainbow/deploying/docker.mdx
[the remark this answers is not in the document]
lightwalker-eth: @djstrong Ok cool. Could you please update the PR description with a list of known follow-up tasks to be completed (perhaps after this PR is merged), including creating the `searchlight` label set.

apps/ensrainbow/src/utils/protobuf-schema.ts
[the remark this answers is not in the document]
lightwalker-eth: @djstrong Ah I see. It's the version of the .ensrainbow file, not the version of the db schema. Ok your suggestion sounds good :+1:

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @param label The label string.
+ * @param labelSetVersion The label set version number.
+ * @returns The encoded versioned rainbow record string.
+ * @throws Error if labelSetVersion is not a non-negative integer.
+ */
+export function buildEncodedVersionedRainbowRecord(
+  label: string,
+  labelSetVersion: number,
+): EncodedVersionedRainbowRecord {
+  if (!Number.isInteger(labelSetVersion) || labelSetVersion < 0) {
```
lightwalker-eth, line 35: Please see my related comment about a `LabelSetVersion` type alias and a utility function that would validate any `maybeLabelSetVersion` `number` value into a validated `LabelSetVersion`.

If we follow that approach then this validation logic should not happen each time we call `buildEncodedVersionedRainbowRecord`. Instead we should be "pushing validation to the edge" where we first capture a `maybeLabelSetVersion` value from input.

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
...
+export type EncodedVersionedRainbowRecord = string;
+
+/**
+ * Builds an encoded versioned rainbow record string from a label and its set version.
+ * Performs validation on the label set version.
+ *
+ * @param label The label string.
+ * @param labelSetVersion The label set version number.
+ * @returns The encoded versioned rainbow record string.
+ * @throws Error if labelSetVersion is not a non-negative integer.
```
lightwalker-eth, line 29: Please see my other comment on this function. This function should not throw any errors.

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * A type alias for the string-encoded representation of a VersionedRainbowRecord.
+ * Format is "{labelSetVersion}:{label}".
+ */
+export type EncodedVersionedRainbowRecord = string;
+
+/**
+ * Builds an encoded versioned rainbow record string from a label and its set version.
+ * Performs validation on the label set version.
```
lightwalker-eth, line 24: We can remove this sentence. Please see my related comments on this function.

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
+import { parseNonNegativeInteger } from "@/utils/parsing";
+import { Label } from "@ensnode/ensnode-sdk";
+
+/**
+ * Represents a decoded versioned rainbow record.
```
lightwalker-eth, line 5:
```suggestion
 * A versioned rainbow record.
```

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * It contains the original label and the label set version it belongs to.
+ * This is the structured representation of the data stored as a string
```
lightwalker-eth, line 8: We should remove this sentence.

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
+import { parseNonNegativeInteger } from "@/utils/parsing";
+import { Label } from "@ensnode/ensnode-sdk";
+
+/**
+ * Represents a decoded versioned rainbow record.
+ *
+ * It contains the original label and the label set version it belongs to.
```
lightwalker-eth, line 7: We should transfer the ideas in this sentence into comments that are specifically on the `label` and `labelSetVersion` fields in this interface.

apps/ensrainbow/src/lib/rainbow-record.ts
```diff
@@ -0,0 +1,74 @@
...
+    throw new Error(
+      `Invalid encoded versioned rainbow record format (missing label set version prefix): "${encodedVersionedRainbowRecord}"`,
+    );
+  }
+
+  const labelSetVersionStr = encodedVersionedRainbowRecord.substring(0, colonIndex);
+  const label = encodedVersionedRainbowRecord.substring(colonIndex + 1);
+
+  try {
+    const labelSetVersion = parseNonNegativeInteger(labelSetVersionStr);
```
lightwalker-eth, line 65: Please see my related suggestions about a utility function that would parse a `maybeLabelSetVersion` into a validated `LabelSetVersion`. We could use that here.

docs/ensnode.io/src/content/docs/ensrainbow/contributing/index.mdx
```diff
@@ -37,8 +37,8 @@ cd apps/ensrainbow
 # Example: Download 'discoveryA' label set ID, label set version 0 (initially empty for discovered labels)
 ./scripts/download-ensrainbow-files.sh discoveryA 0
 
-# Example: Download 'testenv' label set ID, label set version 0 (for test environment)
-./scripts/download-ensrainbow-files.sh testenv 0
+# Example: Download 'ens-test-env' label set ID, label set version 0 (for the ens-test-env)
```
lightwalker-eth, line 40:
```suggestion
# Example: Download the 'ens-test-env' label set ID, label set version 0 (for the ens-test-env)
```

apps/ensrainbow/src/lib/database.ts
```diff
@@ -328,9 +307,9 @@ export class ENSRainbowDB {
...
   /**
-   * Set the highest label set number directly
-   * @param labelSet The label set number to set
-   * @throws Error if the label set is not a valid non-negative integer
+   * Set the highest label set version directly
+   * @param labelSetVersion The label set version to set
+   * @throws Error if the label set version is not a valid non-negative integer
    */
   public async setHighestLabelSetVersion(labelSetVersion: number): Promise<void> {
     if (!Number.isInteger(labelSetVersion) || labelSetVersion < 0) {
```
lightwalker-eth, line 315: Please see my related comments suggesting moving such validation "to the edge" where we capture input values rather than keeping it "in the core".

apps/ensrainbow/src/lib/database.ts
```diff
@@ -340,15 +319,15 @@ export class ENSRainbowDB {
...
    */
-  public async incrementHighestLabelSet(): Promise<number> {
-    const currentValue = await this.getHighestLabelSetVersion();
-    const newValue = currentValue + 1;
-    await this.db.put(SYSTEM_KEY_HIGHEST_LABEL_SET_VERSION, newValue.toString());
-    return newValue;
+  public async incrementHighestLabelSetVersion(): Promise<number> {
+    const currentVersion = await this.getHighestLabelSetVersion();
+    const newVersion = currentVersion + 1;
+    await this.db.put(SYSTEM_KEY_HIGHEST_LABEL_SET_VERSION, newVersion.toString());
```
lightwalker-eth, line 329: Maybe call `setHighestLabelSetVersion` here?

apps/ensrainbow/src/lib/database.ts
```diff
@@ -825,67 +794,17 @@ export class ENSRainbowDB {
   /**
    * Adds a rainbow record to the database. Labelhash is computed from the label.
    *
-   * @param label The label to add (without label set prefix)
+   * @param label The label to add (without label set version prefix)
    * @param labelSetVersion The label set version number to associate with this label
    * @throws Error if labelSetVersion is invalid
```
lightwalker-eth, line 799: Please see my related comments about validating labelSetVersion "at the edge".

lightwalker-eth (changes requested):
@djstrong Hey great updates! Reviewed and shared feedback :+1:

packages/ensrainbow-sdk/src/utils/labelset.ts
```diff
@@ -0,0 +1,74 @@
...
+  labelSetId: string;
+  highestLabelSetVersion: number;
+}
+
+/**
+ * Represents the label set preferences of an ENSRainbow client.
+ */
+export interface EnsRainbowClientLabelSet {
+  /**
+   * Optional label set ID that the ENSRainbow server is expected to use. If provided, heal operations will validate the ENSRainbow server is using this labelSetId.
```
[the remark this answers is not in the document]
lightwalker-eth: Maybe others on the team know a better way? I'm not sure how to automate.

packages/ensrainbow-sdk/src/client.ts
```diff
@@ -198,18 +194,17 @@ export class EnsRainbowApiClient implements EnsRainbow.ApiClient {
   }
 
   constructor(options: Partial<EnsRainbow.ApiClientOptions> = {}) {
+    const { labelSet, ...rest } = options;
+    const newLabelSet = labelSet
```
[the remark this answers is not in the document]
lightwalker-eth: Ok cool. That's a nice idea 👍

apps/ensrainbow/src/lib/server.ts
```diff
@@ -41,15 +35,22 @@ export class ENSRainbowServer {
       throw new Error("Database is in an invalid state");
     }
```
lightwalker-eth, line 37: At the very beginning of the implementation of this `init` function, suggest to document something like the following:

```
// using the Factory method pattern to workaround the limitation of Javascript not supporting `await` within a constructor. We do all async work in this `init` function and then make the synchronous call to the constructor when ready.
```

Goal: Make it more clear why we do things the way we are doing them here.

apps/ensrainbow/src/lib/server.ts
```diff
@@ -41,15 +35,22 @@ export class ENSRainbowServer {
...
-    const labelSetId = await db.getLabelSetId();
-    const highestLabelSetVersion = await db.getHighestLabelSetVersion();
+    const serverLabelSet = await db.getLabelSet();
 
-    return new ENSRainbowServer(db, { labelSetId, highestLabelSetVersion });
+    return new ENSRainbowServer(db, serverLabelSet);
+  }
+
+  /**
+   * Returns the full server label set object
```
lightwalker-eth, line 44:
```suggestion
   * Returns the server's EnsRainbowServerLabelSet.
```

apps/ensrainbow/src/lib/server.ts
```diff
@@ -64,22 +65,19 @@ export class ENSRainbowServer {
...
-        }
-      }
+      validateSupportedLabelSet(this.serverLabelSet, clientLabelSet);
+    } catch (error) {
+      logger.info(
+        `Label set mismatch: requested=${clientLabelSet.labelSetId}, actual=${this.serverLabelSet.labelSetId}`,
+      );
+      return {
+        status: StatusCode.Error,
+        error: "Label set mismatch",
```
lightwalker-eth, line 75: Please see a suggestion here: https://github.com/namehash/ensnode/pull/612#discussion_r2171288509

apps/ensrainbow/src/lib/server.ts
```diff
@@ -92,13 +90,13 @@ export class ENSRainbowServer {
...
       if (
-        clientLabelSet?.labelSetVersion !== undefined &&
+        clientLabelSet.labelSetVersion !== undefined &&
         labelSetVersionNumber > clientLabelSet.labelSetVersion
       ) {
         logger.info(
-          `Label set version ${labelSetVersionNumber} for ${labelHash} exceeds highestLabelSetVersion ${clientLabelSet.labelSetVersion}`,
+          `Label set version ${labelSetVersionNumber} for ${labelHash} exceeds client's requested label set version ${clientLabelSet.labelSetVersion}`,
         );
         return {
```
lightwalker-eth, line 101: It makes me a bit nervous if we implement this `NotFound` error twice in this function. In the future we might update something like the error message in one place but not the other.

Here's a suggestion:

Several lines above, here's the current code:

```
try {
      const versionedRainbowRecord = await this.db.getVersionedRainbowRecord(labelHashBytes);
      if (versionedRainbowRecord === null) {
        logger.info(`Unhealable labelHash request: ${labelHash}`);
        return {
          status: StatusCode.Error,
          error: "Label not found",
          errorCode: ErrorCode.NotFound,
        } satisfies EnsRainbow.HealError;
      }
```

What if we changed it to something approximating the following?

```
try {
      const versionedRainbowRecord = await this.db.getVersionedRainbowRecord(labelHashBytes);
      if (versionedRainbowRecord === null || needToSimulateAsUnhealable(versionedRainbowRecord, clientLabelSet)) {
        logger.info(`Unhealable labelHash request: ${labelHash}`);
        return {
          status: StatusCode.Error,
          error: "Label not found",
          errorCode: ErrorCode.NotFound,
        } satisfies EnsRainbow.HealError;
      }
```

And then you could transfer some the logic that's currently implemented in this `heal` function into `needToSimulateAsUnhealable`.

Advice appreciated 👍

apps/ensrainbow/src/lib/database.ts
```diff
@@ -293,65 +299,44 @@ export class ENSRainbowDB {
...
-    if (labelSetVersion === null) {
-      throw new Error("Highest label set version not found");
-    }
-    return parseNonNegativeInteger(labelSetVersion);
-  }
-
   /**
    * Set the highest label set version directly
    * @param labelSetVersion The label set version to set
    * @throws Error if the label set version is not a valid non-negative integer
```
lightwalker-eth, line 305: We can now remove this @throws comment.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -293,65 +299,44 @@ export class ENSRainbowDB {
...
+    const labelSetIdStr = await this.get(SYSTEM_KEY_LABEL_SET_ID);
+    if (labelSetIdStr === null) {
       throw new Error("Database label set ID is null");
     }
-    return labelSetId;
+    const labelSetId = buildLabelSetId(labelSetIdStr);
+
+    const labelSetVersionStr = await this.get(SYSTEM_KEY_HIGHEST_LABEL_SET_VERSION);
+    if (labelSetVersionStr === null) {
+      throw new Error("Highest label set version not found");
```
lightwalker-eth, line 325: Suggest we better align how we structure our language in both of the error messages in this function. Both are basically the same error, however each of the error messages are written in a different way.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -455,7 +440,7 @@ export class ENSRainbowDB {
     }
 
     try {
-      const count = parseNonNegativeInteger(countStr);
+      const count = buildLabelSetVersion(countStr);
```
lightwalker-eth, line 443: I believe this should stay as a call to `parseNonNegativeInteger`.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -487,7 +472,7 @@ export class ENSRainbowDB {
     }
 
     try {
-      return parseNonNegativeInteger(versionStr);
+      return buildLabelSetVersion(versionStr);
```
lightwalker-eth, line 475: I believe this should stay as a call to `parseNonNegativeInteger`.

apps/ensrainbow/src/lib/database.ts
```diff
@@ -576,33 +561,17 @@ export class ENSRainbowDB {
       return false;
     }
 
-    // 3. Check Label Set ID Existence
+    // 3. Check Label Set ID and Highest Label Set Version Existence and Validity
+    let labelSetId: LabelSetId;
+    let highestLabelSetVersion: LabelSetVersion;
```
lightwalker-eth, line 566: Seems no need to create `labelSetId` or `highestLabelSetVersion` as distinct variables and instead just make use of `labelSet. labelSetId` and `labelSet. highestLabelSetVersion`?

apps/ensrainbow/src/lib/database.ts
```diff
@@ -576,33 +561,17 @@ export class ENSRainbowDB {
...
-      const errorMsg = generatePurgeErrorMessage(`Error checking label set ID: ${error}`);
-      logger.error(errorMsg);
-      return false;
-    }
-
-    // 4. Check Highest Label Set Version Existence and Validity
-    let highestLabelSetVersion: number;
-    try {
-      highestLabelSetVersion = await this.getHighestLabelSetVersion();
       logger.info(`Highest label set version verified: ${highestLabelSetVersion}`);
```
lightwalker-eth, line 572: Suggest combining both of these logger.info calls into a single call to logger.info that gives visibility into both fields in `labelSet`.

packages/ensrainbow-sdk/src/utils/labelset.ts
```diff
@@ -1,42 +1,109 @@
...
+
 /**
- * Represents the label set information provided by an ENSRainbow server.
+ * The state of label sets managed by an ENSRainbow server.
  */
 export interface EnsRainbowServerLabelSet {
-  labelSetId: string;
-  highestLabelSetVersion: number;
+  labelSetId: LabelSetId;
+  highestLabelSetVersion: LabelSetVersion;
```
lightwalker-eth, line 74: Suggest we provider per-field documentation on `EnsRainbowServerLabelSet`.

For inspiration, please refer to `EnsRainbowClientLabelSet`.

packages/ensrainbow-sdk/src/utils/labelset.ts
```diff
@@ -1,42 +1,109 @@
...
+export type LabelSetId = string;
+
+/**
+ * Builds a valid LabelSetId from a string.
+ * @param maybeLabelSetId - The string to validate and convert to a LabelSetId.
+ * @returns A valid LabelSetId.
+ * @throws If the input string is not a valid LabelSetId.
+ */
+export function buildLabelSetId(maybeLabelSetId: string): LabelSetId {
+  if (typeof maybeLabelSetId !== "string") {
```
lightwalker-eth, line 18: We can remove this check because we're strict with ourselves about not violating the type system. Therefore the param passed in can be safely assumed to be a string.

packages/ensrainbow-sdk/src/utils/labelset.test.ts
```diff
@@ -3,9 +3,85 @@ import {
...
+      "LabelSetVersion must be a non-negative integer.",
+    );
+  });
+
+  it("should return a valid label set version for a string input", () => {
+    expect(buildLabelSetVersion("1")).toBe(1);
+  });
+
+  it("should throw an error for a non-number/non-string input", () => {
+    expect(() => buildLabelSetVersion({} as any)).toThrow(
```
lightwalker-eth, line 79: Please see my related comment about how we shouldn't need this test anymore because we avoid violating the type system in such a way anywhere in our code.

packages/ensrainbow-sdk/src/utils/labelset.ts
```diff
@@ -1,42 +1,109 @@
...
+    try {
+      versionNumber = parseNonNegativeInteger(maybeLabelSetVersion);
+    } catch (error) {
+      throw new Error(
+        `Invalid label set version: ${maybeLabelSetVersion}: ${
+          error instanceof Error ? error.message : error
+        }`,
+      );
+    }
+  } else if (typeof maybeLabelSetVersion === "number") {
```
lightwalker-eth, line 57: I believe it should be ok to just put this in an `} else {` block (no `else if`).

Goal: put our trust in the type system. We say the param must be a number or a string. If it's not a string, therefore it must be a number.

packages/ensrainbow-sdk/src/utils/labelset.ts
```diff
@@ -1,42 +1,109 @@
...
+    if (maybeLabelSetVersion < 0 || !Number.isInteger(maybeLabelSetVersion)) {
+      throw new Error(`LabelSetVersion must be a non-negative integer.`);
+    }
+    versionNumber = maybeLabelSetVersion;
+  } else {
```
lightwalker-eth, line 62: We can remove this else block because we avoid ever violating the type system this much.

packages/ensrainbow-sdk/src/utils/labelset.test.ts
```diff
@@ -3,9 +3,85 @@ import {
...
+  });
+
+  it("should throw an error for a string with symbols", () => {
+    expect(() => buildLabelSetId("subgraph_1")).toThrow(
+      "LabelSetId can only contain lowercase letters (a-z) and hyphens (-).",
+    );
+  });
+
+  it("should throw an error for a non-string input", () => {
+    expect(() => buildLabelSetId(123 as any)).toThrow("LabelSetId must be a string.");
```
lightwalker-eth, line 52: Please see my related comment about how we shouldn't need this test anymore because we avoid violating the type system in such a way anywhere in our code.

packages/ensrainbow-sdk/src/utils/labelset.test.ts
```diff
@@ -3,9 +3,85 @@ import {
...
+  });
+});
+
+describe("buildLabelSetVersion", () => {
```
lightwalker-eth, line 56: Suggest to also add a test for:

1. `NaN`
2. `Infinity`

Both should fail.

apps/ensrainbow/src/lib/api.ts
```diff
@@ -56,9 +59,12 @@ export async function createApi(db: ENSRainbowDB): Promise<Hono> {
...
     }
 
-    let clientLabelSet: EnsRainbowClientLabelSet | undefined = undefined;
+    let clientLabelSet: EnsRainbowClientLabelSet;
     try {
-      clientLabelSet = buildEnsRainbowClientLabelSet(labelSetIdParam, labelSetVersion);
+      const labelSetId: LabelSetId | undefined = labelSetIdParam
+        ? buildLabelSetId(labelSetIdParam)
+        : undefined;
+      clientLabelSet = buildEnsRainbowClientLabelSet(labelSetId, labelSetVersion);
```
lightwalker-eth, line 67: I believe we can move the definition and the initialization of `clientLabelSet` outside of the try / catch here.

lightwalker-eth (changes requested):
@djstrong Great updates 👍 Reviewed and shared feedback 😄

apps/ensrainbow/src/lib/api.ts
djstrong: buildEnsRainbowClientLabelSet can throw an error and the try block is for this.

djstrong: `buildEnsRainbowClientLabelSet` can throw an error and the try block is for this.

apps/ensrainbow/src/lib/server.ts
```diff
@@ -41,13 +43,28 @@ export class ENSRainbowServer {
...
    */
   public getServerLabelSet(): EnsRainbowServerLabelSet {
     return this.serverLabelSet;
   }
 
+  /**
+   * Determines if a versioned rainbow record should be treated as unhealable
+   * based on the client's label set version requirements.
+   */
+  private needToSimulateAsUnhealable(
```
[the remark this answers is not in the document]
djstrong: The function ignores the label set ID, so in the unit tests, the label set IDs of the client and record are always the same.

PR #615, opened 2025-04-16
feat(ci): update github release workflow

tk-o:
This PR makes Changesets to not create a Github Release and instead, has a separate Github Action to read the changesets-made PR and create the new Github Release from its description.

Also, this PR updates `ACTIVE_PLUGINS` values to match recent refactorings.


shrugs (approved):
looks like it should work, nice

PR #627, opened 2025-04-17
docs: fix outdated links

notrab:
Three links weren't updated when the docs folder got renamed. This fixes that.


lightwalker-eth (approved):
@notrab Thanks 😄

PR #657, opened 2025-04-21
3DNS Support (Optimism, Base)

shrugs:
- implements ThreeDNS Base and Optimism Datasources
- implements ThreeDNS Plugin that indexes both base and optimism
  - every (?) name on optimism is correctly healed due to the on-chain metadata usage of the fqdn
  - can't seem to find a way to get the fqdns of mystery.box subnames, so those fall back to ensrainbow as normal
- refactored resolver behavior
  - resolver config is shared between plugins, since the abi is well-defined
  - custom `mergePonderConfigs` deepmerge handles ensuring the indexer uses the earliest possible `startBlock` for the Resolvers per-network
  - custom `mergePonderConfigs` deepmerge handles merging abis (for any multi-network contract definition like `Resolver`) using `viem#mergeAbi` instead of default concatenation
  - resolvers are now idempotent per-chain using ponder's built-in multinetwork behavior, instead of instantiating resolvers per-plugin. this would have caused double-indexing of resolver records if multiple active plugins needed to index Resolvers on the same network
  - this ended up not being strictly necessary, as ThreeDNS uses an immutable resolver for all of their managed names, but it's still a good refactor to include regardless, and it correctly isolates Resolvers as a per-network indexing behavior rather than a per-plugin contract config

---

### Checklist

- [x] test snapshot again mainnet to ensure resolver refactor didn't affect subgraph compatibility
- [ ] before releasing to production, please ensure `RPC_URL_10` and `RPC_REQUEST_RATE_LIMIT_10` env vars are set properly across production services that will have `threedns` plugin enabled


apps/ensindexer/src/handlers/ThreeDNSToken.ts
```diff
@@ -0,0 +1,302 @@
...
+        id: node,
+        ownerId: owner,
+        parentId: parentNode,
+        createdAt: event.block.timestamp,
+        labelhash: labelHash,
+
+        // NOTE: threedns has no concept of registry migration, so domains managed by this plugin
+        // are always considered migrated
+        isMigrated: true,
+      });
```
tk-o, line 74: Noting that we didn't include `subdomainCount` value updates.

apps/ensindexer/src/plugins/threedns/handlers/ThreeDNSOptimism.ts
```diff
@@ -0,0 +1,80 @@
+import { Context, ponder } from "ponder:registry";
```
tk-o, line 1: Perhaps we should apply a factory function to create the default export here as it's almost exactly the same function as in `ThreeDNSBase.ts` file?

shrugs: good call - i've deduplicated this with ponder multi-network behavior, like with Resolver

apps/ensindexer/src/handlers/ThreeDNSToken.ts
shrugs: implemented, thanks for the heads up

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -194,12 +194,42 @@ export function networkConfigForContract<CONTRACT_CONFIG extends ContractConfig>
 ) {
   return {
     [chain.id.toString()]: {
-      ...contractConfig,
-      ...constrainContractBlockrange(contractConfig.startBlock),
+      address: contractConfig.address, // provide per-network address if available
```
shrugs, line 197: note: this explicitly only passes `address`, and the blockrange (`startBlock`, `endBlock`) to the per-network contract config overrides, to avoid bloating the ponder config with unnecessary values (like `abi`)

tk-o (approved):
LGTM

PR #664, opened 2025-04-23
feat(ensnode): custom saved queries graphiql plugin

notrab:
This PR begins some groundwork for "Saved Queries" plugin that we can present to users of Ponder & Subgraph APIs with some demo queries.

The idea of how this would appear is like below:

CleanShot 2025-04-23 at 10 02 58@2x

There are a few bugs though:

* Clicking on a query doesn't actually update the context
* Passing `plugins` as a prop on the `gql/page.tsx` causes an issue because the page is server rendered. So the plugins are hardcoded to be the same now for all GraphiQL instances.

### DIFF
--- a/apps/ensadmin/package.json
+++ b/apps/ensadmin/package.json
@@ -22,8 +22,9 @@
   },
   "dependencies": {
     "@ensnode/ponder-metadata": "workspace:*",
-    "@ensnode/utils": "workspace:*",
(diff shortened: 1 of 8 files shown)


tk-o (approved):
@notrab reviewed my changes and was happy with the updates included in this PR. Will merge this one soon 🚀

PR #666, opened 2025-04-23
fix: release tags and notes

shrugs:
- re-enable git tags so publishing works
- remove shared-configs section from release notes and npm link


tk-o (approved):
LGTM

PR #667, opened 2025-04-24
fix(ensadmin): always show teaser badge on inspector page

notrab:
This always shows the `teaser` badge on the inspector page but still hides it in the breadcrumbs. A separate fix is needed for that so we can better stack those items for mobile.


lightwalker-eth (approved):
@notrab Nice, that fixed it :+1:

PR #668, opened 2025-04-24
feat(ensadmin): ui for ai generated queries

tk-o:
This PR includes a new UI component on the Subgraph API Page that allows generating GQL queries with AI.

Note: we decided not to roll this feature out on the Ponder API Page.

### Preview

image

PR #669, opened 2025-04-24
fix(AdapterAnthropic): use prompt caching, caches 95% of tokens

djstrong:


tk-o:
Moved all changes into #668

PR #676, opened 2025-04-29
Feat/631 add slack notifications upon key ci actions

BanaSeba:
Added Github notifications on switch/deploy/release

### DIFF
--- a/.github/workflows/deploy_ensnode.yml
+++ b/.github/workflows/deploy_ensnode.yml
@@ -174,3 +174,15 @@ jobs:
             redeploy_service  ${RAILWAY_ENVIRONMENT_ID} ${HOLESKY_INDEXER_SVC_ID}
             #ENSRAINBOW
             redeploy_service  ${RAILWAY_ENVIRONMENT_ID} ${ENSRAINBOW_SVC_ID}
+
+
+      - name: Slack Notification on Success
+        if: success()
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
+          SLACK_USERNAME: "ENSNode Bot"
+          SLACK_ICON: "https://ensnode.io/favicon.svg"
+          SLACK_TITLE: "✅ Deploy ENSNode completed"
+          SLACK_MESSAGE: "ENSNode deployed: to ${{ env.get-pr.outputs.version }}"
+          SLACK_COLOR: good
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy_ensnode.yml
```diff
@@ -174,3 +174,15 @@ jobs:
             redeploy_service  ${RAILWAY_ENVIRONMENT_ID} ${HOLESKY_INDEXER_SVC_ID}
             #ENSRAINBOW
             redeploy_service  ${RAILWAY_ENVIRONMENT_ID} ${ENSRAINBOW_SVC_ID}
+
+
+      - name: Slack Notification on Success
+        if: success()
```
shrugs, line 180: i think `if: success()` unnecessary? 

> A default status check of success() is applied unless you include one of these functions.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#status-check-functions

.github/workflows/deploy_ensnode.yml
```diff
@@ -174,3 +174,15 @@ jobs:
...
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK_URL }}
```
shrugs, line 183: would it be beneficial to extract this into a shared action where the default values can be applied? username and icon are static and title/message/color(default good) can be the args. not sure if the secret can be set in the action or whether only the main workflow file has access to `secrets.` but otherwise that can be provided as a default as well to cut down on the duplication in each workflow we want to notify slack in

.github/workflows/deploy_ensnode.yml
```diff
@@ -174,3 +174,15 @@ jobs:
             redeploy_service  ${RAILWAY_ENVIRONMENT_ID} ${HOLESKY_INDEXER_SVC_ID}
             #ENSRAINBOW
             redeploy_service  ${RAILWAY_ENVIRONMENT_ID} ${ENSRAINBOW_SVC_ID}
+
+
+      - name: Slack Notification on Success
```
shrugs, line 179: also happy to remove the `name` key here, as it'll use the action name by default which is a-ok to me and minimizes our configuration

.github/workflows/deploy_ensnode.yml
BanaSeba: Removed

.github/workflows/deploy_ensnode.yml
BanaSeba: Added new action for Slack notification. We can extend it further in the future if needed

.github/workflows/deploy_ensnode.yml
BanaSeba: Removed

shrugs (approved):

PR #679, opened 2025-04-29
fix: allow rpc endpoints to be undefined at runtime and checked at config gen time

shrugs:
panics if there are no custom rpc endpoints defined for the networks that the config ends up using and provides verbose error message


tk-o (approved):
LGTM

Issue #663, opened 2025-04-22
ENSAdmin: error showing status page before all chains have begun indexing


shrugs:
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/fdfd9887-fd1f-4358-8597-f175a265ba7e" />

tk-o:
Resolved with #686

PR #686, opened 2025-04-30
fix(ensadmin): allow displaying networks that have not yet been indexed

### LINKED ISSUE #468: ENSAdmin: Bug When API is down
mainnet blue is currently down and there's a client side parsing error in ensadmin because of it. looks like it's caught by the error boundary. can't get a stacktrace because we minify — maybe worth enabling source maps in production for the near future, since we don't have a bug reporting tooling set up?

https://admin.ensnode.io/status?ensnode=https%3A%2F%2Fapi.mainnet.blue.ensnode.io%2F

Image

tk-o:
This PR allows ENSAdmin to display indexing status for chains that have not yet been indexed.

### Before

image

### After

Resolves #468

### DIFF
--- a/apps/ensadmin/src/components/ensnode/hooks.ts
+++ b/apps/ensadmin/src/components/ensnode/hooks.ts
@@ -78,18 +78,6 @@ function validateResponse(response: EnsNode.Metadata) {
         .join(", ")}`,
     );
   }
-
-  const networksWithoutLastIndexedBlock = Object.entries(networkIndexingStatusByChainId).filter(
-    ([, network]) => network.lastIndexedBlock === null,
-  );
-
-  if (networksWithoutLastIndexedBlock.length > 0) {
-    throw new Error(
-      `Missing last indexed block for some networks with the following chain IDs: ${networksWithoutLastIndexedBlock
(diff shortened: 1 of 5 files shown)


shrugs (approved):

apps/ensadmin/src/components/ensnode/hooks.ts
```diff
@@ -78,18 +78,6 @@ function validateResponse(response: EnsNode.Metadata) {
...
-    ([, network]) => network.lastIndexedBlock === null,
-  );
-
-  if (networksWithoutLastIndexedBlock.length > 0) {
-    throw new Error(
-      `Missing last indexed block for some networks with the following chain IDs: ${networksWithoutLastIndexedBlock
-        .map(([chainId]) => chainId)
-        .join(", ")}`,
-    );
-  }
```
tk-o, line 92: There's a legit case for a chain not to include the last indexed block value. For example, the chain might be waiting for its turn to be indexed. For the time being, I'll make the viewmodel responsible for mapping proper UI output.

tk-o:
Resolves #663

PR #690, opened 2025-05-05
chore: onboarding changes

0xBison:
chore: update .gitignore, add missing val in .env.local.example, force dev command to use public schema and small doc updates

- Added .history to .gitignore for VSCode local history extension.
- Added missing PORT environment variable which is required
- Changed DATABASE_SCHEMA to 'production' to avoid conflicts between dev and start commands and default dev command to use 'public' schema
- Fixed some broken links in docs


apps/ensindexer/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "homepage": "https://github.com/namehash/ensnode/tree/main/apps/ensindexer",
   "scripts": {
-    "dev": "ponder dev",
+    "dev": "ponder dev --schema=public",
```
0xBison, line 15: The reason i suggest this change is that when running `pnpm run dev` followed by `pnpm run start` we will see the following error:

```
2:59:09 PM ERROR process    Caught unhandledRejection event
NonRetryableError: Schema 'public' was previously used by a different Ponder app. Drop the schema first, or use a different schema. Read more: https://ponder.sh/docs/getting-started/database#database-schema
```

By adding this change we isolate the `dev` and `start` commands to use independent schemas always (so long as the environment file doesn't use `public` for the `DATABASE_SCHEMA` environment variable).

This is in line with the default behaviour of ponder which is to use the `public` schema for the dev command - https://ponder.sh/docs/getting-started/database

apps/ensindexer/.env.local.example
```diff
@@ -32,7 +35,11 @@ RPC_REQUEST_RATE_LIMIT_59144=20
 #
 # Read more about database schema rules here:
 # https://ponder.sh/docs/api-reference/database#database-schema-rules
-DATABASE_SCHEMA=public
+#
+# Avoid using the `public` schema as we force that in the `dev` command. Using `public`
+# cause conflicts as you interchange between dev and start commands so use literally
+#anything else.
+DATABASE_SCHEMA=production
```
0xBison, line 42: `public` is used for the dev command (see other comment) so set this to a different value which will be used when we run the `start` command

apps/ensindexer/.env.local.example
```diff
@@ -1,3 +1,6 @@
+# The port ENSIndexer listens on for HTTP requests.
+PORT=42069
```
0xBison, line 2: Without an explicit `PORT`, we get some validation failure:

```
Error: Error parsing environment variable 'PORT': Expected value not set.
```

Maybe we are over-validating and can default to the default ponder port which is `42069`. I know there's a separate story to uplift the config and maybe the scope of that includes making some fields optional but as the current codebase stands without this value i see the error above.

apps/ensindexer/package.json
shrugs: i think this might cause issues, as ENSIndexer reads the DATABASE_SCHEMA variable to power api metadata endpoints (for subgraph api and ensadmin) — so the only way we should configure the schema is with the DATABASE_SCHEMA variable

this should be changed to `DATABASE_SCHEMA=public ponder dev` and it should work

apps/ensindexer/.env.local.example
```diff
@@ -32,7 +35,11 @@ RPC_REQUEST_RATE_LIMIT_59144=20
 #
 # Read more about database schema rules here:
 # https://ponder.sh/docs/api-reference/database#database-schema-rules
-DATABASE_SCHEMA=public
+#
+# Avoid using the `public` schema as we force that in the `dev` command. Using `public`
+# cause conflicts as you interchange between dev and start commands so use literally
+#anything else.
```
shrugs, line 41:
```suggestion
# anything else.
```

apps/ensindexer/.env.local.example
0xBison: this is handled in https://github.com/namehash/ensnode/pull/698 with a default now

shrugs (approved):

Issue #678, opened 2025-04-29
3DNS Text Records Not showing up?


shrugs:
- `TextChanged` event isn't emitted perhaps they emit a different one

https://optimistic.etherscan.io/tx/0xf32db67e7bf2118ea2c3dd8f40fc48d18e83a4a2317fbbddce8f741e30a1e8d7#eventlog

shrugs:
can't find implementation or abi anywhere and signature isn't known by 4byte directory so we have nothing to go off of

PR #697, opened 2025-05-06
ThreeDNS DNSRecordChanged Parsing

shrugs:
closes https://github.com/namehash/ensnode/issues/678

the threedns resolvers don't emit `TextChanged`, they emit `DNSRecordChanged` which we previously didn't parse. this implements RRSet parsing to bring the behavior in-line with the other plugins'


0xBison:
see: https://github.com/0xBison/3dns-record-decoding

apps/ensindexer/src/handlers/Resolver.ts
```diff
@@ -323,10 +325,72 @@ export const makeResolverHandlers = ({ pluginName }: { pluginName: PluginName })
...
+          default: {
+            // no-op unhandled records
+            // NOTE: should never occur due to resource id check above
+            console.warn(
+              `Invariant: received answer ${JSON.stringify(answer)} that is not type === TXT despite resource === 16 check!`,
+            );
+            break;
+          }
+        }
+      }
```
tk-o, line 393: Making exceptions based on the plugin makes me think we could inject the new logic behind an interface from a particular plugin. Meaning, the `subgraph` plugin would implement a no-op for that interface, and both `basenames` and `lineanames` would implement the interface with the exact logic you proposed above.

Thoughts on that approach?

apps/ensindexer/src/plugins/threedns/handlers/ThreeDNSToken.ts
```diff
@@ -70,7 +71,19 @@ export default function ({
   ponder.on(namespace("ThreeDNSResolver:InterfaceChanged"), handleInterfaceChanged);
   ponder.on(namespace("ThreeDNSResolver:AuthorisationChanged"), handleAuthorisationChanged);
   ponder.on(namespace("ThreeDNSResolver:VersionChanged"), handleVersionChanged);
-  ponder.on(namespace("ThreeDNSResolver:DNSRecordChanged"), handleDNSRecordChanged);
+  ponder.on(
```
0xBison, line 74: I assume this is an old event on a proxy that got updated and the below one is the new event?

Do we have any documentation somewhere where we track updates to contracts that we index? if not i would suggest adding a comment here to at least explain why we have an overloaded event and which one is newer, specifically because these contracts are not verified on optimismscan.

apps/ensindexer/src/plugins/threedns/handlers/ThreeDNSToken.ts
```diff
@@ -70,7 +71,19 @@ export default function ({
...
+    namespace(
+      "ThreeDNSResolver:DNSRecordChanged(bytes32 indexed node, bytes name, uint16 resource, bytes record)",
```
0xBison, line 76: I assume this is the old event on a proxy that got updated and the below one is the new event?

Do we have any documentation somewhere where we track updates to contracts that we index? if not i would suggest adding a comment here to at least explain why we have an overloaded event and which one is newer, specifically because these contracts are not verified on optimismscan.

apps/ensindexer/src/handlers/Resolver.ts
```diff
@@ -323,10 +325,72 @@ export const makeResolverHandlers = ({ pluginName }: { pluginName: PluginName })
         node: Node;
         name: Hex;
         resource: number;
+        ttl?: number;
```
0xBison, line 328: We don't use this field anywhere - I would just double check that is intentional and this field has no value. I see the old event didn't have this field but worth checking

apps/ensindexer/src/lib/dns-helpers.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * parses an RRSet encoded as Hex string into a set of Answer records.
+ *
+ * the only relevant node library capable of this seems to be dns-packet, and its un-exported
+ * `answers.decode` function, which we leverage here.
+ *
+ * @param record the hex representation of an RRSet
+ */
+export function parseRRSet(record: Hex) {
```
0xBison, line 12: i think it is worth adding a few tests here. i know the logic is being offloaded to `dns-packet` but would be good for developer readability to see an example case of the input and output in this function. particularly because it takes in `Hex`.

apps/ensindexer/src/handlers/Resolver.ts
```diff
@@ -338,10 +402,47 @@ export const makeResolverHandlers = ({ pluginName }: { pluginName: PluginName })
...
+        .update(schema.resolver, { id })
+        .set({ texts: (resolver.texts ?? []).filter((text) => text !== key) });
+
+      // log ResolverEvent
+      await context.db.insert(schema.textChanged).values({
+        ...sharedEventValues(context.network.chainId, event),
+        resolverId: id,
+        key,
+        value: null,
+      });
```
tk-o, line 445: Making exceptions based on the plugin makes me think we could inject the new logic behind an interface from a particular plugin. Meaning, the `subgraph` plugin would implement a no-op for that interface, and both `basenames` and `lineanames` would implement the interface with the exact logic you proposed above.

Thoughts on that approach?

tk-o (approved):
Awesome update, @shrugs & @0xBison. Merge when ready 🚀

apps/ensindexer/src/handlers/Resolver.ts
shrugs: now that we have the tooling to confirm that subgraph compatibility isn't affected by changes to the shared handlers, i'm less worried about including per-plugin logic here

Issue #696, opened 2025-05-06
Remove ENSRainbow.io default API url in ENSIndexer


shrugs:
the .env.local.example should _not_ provide a default `ENSRAINBOW_API` value and should encourage the user to run ENSRainbow locally or on the same private network for performance reasons.

- update `ensindexer/.env.local.example`
  - can include in the comments a note like "while not recommended, you can temporarily configure ENSIndexer to use the public ENSRainbow API here:" or whatever
- update any documentation references
- ENSIndexer should panic if no ENSRAINBOW_API endpoint configured (or is not valid url)

Issue #523, opened 2025-04-07
Build `ENSIndexerConfig` upon startup


lightwalker-eth:
We should improve the maturity of how we build the config used by ENSIndexer.

Currently we have logic spread in many places across the code that reads some environment variable. This includes places that are inside core indexing logic loops, etc.

A better strategy is:

1. Define a type / interface for an entire `ENSIndexerConfig`.
2. When ENSIndexer starts up, it should parse environment variables as needed to build an `ENSIndexerConfig`.
   1. This should include strong invariant validation logic. Ex: We need to not only validate that each individual field in this config is set in a valid way, but also that no set of values in the combined config are in conflict with each other.
3. Environment variables should only be read once during the ENSIndexer startup and never again while it is actually running.
4. There should be an easy mechanism for anywhere in ENSIndexer to get the active config (after it is created) and build logic based on it.
5. Unit tests that are based on one config or another should use dependency injection, such that we can easily test different potential config cases.

shrugs:
use zod4 and stringbool, for example https://v4.zod.dev/v4

shrugs:
invariant: only activate a plugin if RPCs were defined for all of the plugin's datasource deps

lightwalker-eth:
Completed via https://github.com/namehash/ensnode/pull/698

PR #698, opened 2025-05-07
feat: centralized indexer config with zod validations

0xBison:
Resolves the following issues:

https://github.com/namehash/ensnode/issues/523
https://github.com/namehash/ensnode/issues/696

Synopsis:

Consolidates most of the config related and error related code into a single `config` folder. We now have a singleton that is created and validated at startup. Some initial type validations are performed at startup and there are another set of validations that run after which themself need the config values. We can add validations as we wish to the validations file. On startup failed validations will show like this:

issues

Potential improvements:

- Currently logs aren't very prominent due to them being `console.log` statements. we dont have access to a ponder logger as they dont expose it. i've asked them about this and will follow up. for now it may be ok or we can add some color to validations with the `chalk` package to make it clear.


apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+  ensAdminUrl: customUrlSchema("ENSADMIN_URL").default(DEFAULT_ENSADMIN_URL),
+  ponderDatabaseSchema: z
+    .string({
+      error: (issue) => {
+        if (issue.input === undefined) return "DATABASE_SCHEMA is required.";
+        if (String(issue.input).trim() === "") return "DATABASE_SCHEMA cannot be empty.";
+        return "DATABASE_SCHEMA must be a string.";
+      },
+    })
+    .default(DEFAULT_DATABASE_SCHEMA),
```
shrugs, line 107: i don't think we should default the DATABASE_SCHEMA variable, were we doing so before?

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+          error: "Each plugin name in ACTIVE_PLUGINS must be a string.",
+        }),
+        { error: "ACTIVE_PLUGINS must resolve to a list of plugin names." },
+      )
+      .min(1, {
+        error:
+          "ACTIVE_PLUGINS must be set and contain at least one valid plugin name (e.g. 'subgraph' or 'subgraph,basenames').",
+      }),
+  ),
+  healReverseAddresses: z.preprocess(
```
shrugs, line 133: use https://v4.zod.dev/api#stringbool

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+      if (val === "true") return true;
+      if (val === "false") return false;
+      return val;
+    },
+    z.boolean({
+      error: "HEAL_REVERSE_ADDRESSES must be 'true' or 'false'.",
+    }),
+  ),
+  ponderPort: z.preprocess(
+    (val) => (val === undefined || val === "" ? DEFAULT_PORT : Number(val)),
```
shrugs, line 147: coerce with default instead of custom preprocess

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+      .min(1, {
+        error: "Ponder port (PORT env var) must be a number between 1 and 65535.",
+      }),
+  ),
+  ensRainbowEndpointUrl: customUrlSchema("ENSRAINBOW_URL"),
+  chains: z
+    .record(
+      z
+        .string({
+          error: (issue) => {
```
shrugs, line 163: this validation unnecessary if we avoid including `NaN` as a possible chainId when constructing `chains`

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,76 @@
...
+   * Used to avoid rate limiting by the RPC provider.
+   */
+  rpcMaxRequestsPerSecond: number;
+}
+
+/**
+ * The complete configuration object for the ENSIndexer application.
+ * This object is typically constructed at startup and used throughout the app.
+ */
+export interface ENSIndexerConfig {
```
shrugs, line 26: remove this type entirely in favor of `z.infer<>`

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,76 @@
+import { Blockrange } from "@/lib/types";
+import { ENSDeploymentChain } from "@ensnode/ens-deployments";
+import { Chain } from "viem";
+
+/**
+ * Configuration for a single blockchain network (chain) used by the ENSIndexer.
+ */
+export interface ChainConfig {
```
shrugs, line 8: replace with `z.infer<>`

apps/ensindexer/src/config/validations.ts
```diff
@@ -0,0 +1,64 @@
...
+    
+    The following RPC_URL_* environment variables must be defined for nominal indexing behavior:
+    ${allChainIds
+      .map((chainId) => `RPC_URL_${chainId}: ${rpcEndpointUrl(chainId) || "N/A"}`)
+      .join("\n")}
+    `);
+  }
+};
+
+export function validateConfig(allChainIds: number[], networks: MergedPluginConfig["networks"]) {
```
shrugs, line 50: this function wants to be like `validatePonderConfig(config: Config, ponderConfig: MergedPluginConfig)`

and each of the internal validation functions should be pure functions over the same arguments, no globals

apps/ensindexer/ponder.config.ts
```diff
@@ -68,51 +35,14 @@ const ponderConfig = activePlugins
...
-  throw new Error(`ENSNode has been configured with the following ACTIVE_PLUGINS: ${requestedPluginNames.join(", ")}.
-These plugins, collectively, index events from the following chains: ${allChainIds.join(", ")}.
 
-The following RPC_URL_* environment variables must be defined for nominal indexing behavior:
-${allChainIds.map((chainId) => `RPC_URL_${chainId}: ${rpcEndpointUrl(chainId) || "N/A"}`).join("\n")}
-`);
-}
+// Validate the config before we activate the plugins. These validations go beyond simple type
+// validations and ensure any relationships between environment variables are correct.
+validateConfig(allChainIds, ponderConfig.networks);
```
shrugs, line 45: this would read better as `validatePonderConfig(config, ponderConfig)` where config is the zod-parsed `Config` and `ponderConfig` is the runtime-constructed ponder config

apps/ensindexer/src/handlers/Registry.ts
```diff
@@ -73,13 +77,15 @@ export const makeRegistryHandlers = ({ pluginName }: { pluginName: PluginName })
 
         // if the domain doesn't yet have a name, attempt to construct it here
         if (!domain.name) {
-          const parent = await context.db.find(schema.domain, { id: parentNode });
+          const parent = await context.db.find(schema.domain, {
```
shrugs, line 80: nit: make sure your editor is properly respecting biome's lineWidth setting — install the biome vscode plugin if you don't have it — this diff should disappear

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+let _config: ReturnType<typeof buildENSIndexerConfig> | undefined;
+
+/**
+ * Returns the ENSIndexer configuration object. If it doesnt exist will instantiate it
+ * via the buildENSIndexerConfig function. This function ensures that the configuration
+ * is built only once and cached for reuse.
+ */
+export function getConfig() {
```
shrugs, line 85: does this singleton need to be deferred, or can we do something like

```ts
export default buildENSIndexerConfig()
```

so that consumers can do

```ts
import { healReverseAddreses } from '@/config/app-config'
```

apps/ensindexer/src/lib/globals.ts
```diff
@@ -1,7 +1,10 @@
-import { getEnsDeploymentChain } from "@/lib/ponder-helpers";
+import { getConfig } from "@/config/app-config";
 import { ENSDeployments } from "@ensnode/ens-deployments";
 
-export const SELECTED_ENS_DEPLOYMENT = ENSDeployments[getEnsDeploymentChain()];
+const config = getConfig();
+const ensDeploymentChain = config.ensDeploymentChain;
+
+export const SELECTED_ENS_DEPLOYMENT = ENSDeployments[ensDeploymentChain];
```
shrugs, line 7: nit: inline `export const SELECTED_ENS_DEPLOYMENT = ENSDeployments[config.ensDeploymentChain];`

apps/ensindexer/src/plugins/index.ts
```diff
@@ -0,0 +1,26 @@
+import { MergedTypes } from "@/lib/plugin-helpers";
```
shrugs, line 1: I prefer this in ponder.config.ts — better to centralize this information rather than splitting to another file

shrugs (changes requested):
could we use zod transforms to make the schema a bit tighter? in particular where preprocess is used or where a custom error function that includes validation is used (which seems like an antipattern)? https://v4.zod.dev/api#transforms

this feels heavily llm-generated, which is fine in general but i'm worried i had to spend a lot of time reviewing these changes, in particular zod-related behavior where it's not using zod4 changes/features

lightwalker-eth (commented):
@0xBison Hey excited to see this PR, thank you! I've only partially been able to complete a review of this PR so far, but sharing initial feedback 🫡

apps/ensindexer/src/config/app-config.ts
0xBison: we cant destructure a default export but otherwise followed this approach. much cleaner

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+  rpcEndpointUrl: customUrlSchema("RPC_URL"),
+  rpcMaxRequestsPerSecond: z
+    .number({ error: "RPC max requests per second must be a number." })
+    .int({ error: "RPC max requests per second must be an integer." })
+    .min(1, { error: "RPC max requests per second must be at least 1." })
+    .default(DEFAULT_RPC_RATE_LIMIT),
+});
+
+const ENSDeploymentChainSchema = z
+  .enum(Object.keys(ENSDeployments) as [keyof typeof ENSDeployments], {
```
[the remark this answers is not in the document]
0xBison: This is an object, not an enum which is why I have to do this.

The best I can get is:

const ENSDeploymentChainSchema = z
  .enum(Object.keys(ENSDeployments) as [keyof typeof ENSDeployments])
  .default(DEFAULT_DEPLOYMENT);

Which comes out with the error message:  `Invalid option: expected one of "mainnet”, “sepolia”, “holesky”, “ens-test-env"`

It’s not terrible but it’s also not good. I dont think we need a change here

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+          error: `${envVarKey} must be a valid URL string (e.g., http://localhost:8080 or https://example.com).`,
+        },
+      )
+  );
+};
+
+const ChainConfigSchema = z.object({
+  rpcEndpointUrl: customUrlSchema("RPC_URL"),
+  rpcMaxRequestsPerSecond: z
+    .number({ error: "RPC max requests per second must be a number." })
```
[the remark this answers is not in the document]
0xBison: `Invalid input: expected number, received NaN` in this case.

Most of this overly verbose code is because we need a relationship between the environment variables and the error messages. otherwise we get generic unhelpful messages which is why i've approached it like this.

I really wanted to ensure we are helping users understand what is actually happening with their config instead of trying to make this as “clean” as possible. This was one of the goals of the original issue.

apps/ensindexer/src/config/config.schema.ts
0xBison: nope, removed

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+        val.startBlock === undefined ||
+        val.endBlock === undefined ||
+        val.startBlock <= val.endBlock,
+      { error: "END_BLOCK must be greater than or equal to START_BLOCK." },
+    ),
+  ensNodePublicUrl: customUrlSchema("ENSNODE_PUBLIC_URL"),
+  ensAdminUrl: customUrlSchema("ENSADMIN_URL").default(DEFAULT_ENSADMIN_URL),
+  ponderDatabaseSchema: z
+    .string({
+      error: (issue) => {
```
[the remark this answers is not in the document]
0xBison: Error messages are like:

`Too small: expected string to have >1 characters"`

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,76 @@
...
+ * This object is typically constructed at startup and used throughout the app.
+ */
+export interface ENSIndexerConfig {
+  /**
+   * The ENS deployment chain identifier (e.g. "mainnet", "sepolia").
+   */
+  ensDeploymentChain: ENSDeploymentChain;
+
+  /**
+   * The global block range to index (start and end block numbers).
```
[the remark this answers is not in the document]
0xBison: I believe this is some dev only function, reached out to investigate from others.

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Configuration for a single blockchain network (chain) used by the ENSIndexer.
+ */
+export interface ChainConfig {
+  /**
+   * The RPC endpoint URL for the chain.
+   * Example: "https://eth-mainnet.g.alchemy.com/v2/..."
+   */
+  rpcEndpointUrl: string;
```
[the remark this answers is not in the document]
0xBison: Invariants are done but they may be slightly redundant now as I've removed types in favour of `z.infer` so the comments live alongside the schema and validations.

Regarding using `URL`, i think we should stick to using a string as we only use this once and it's to pass it to `viem` which takes in a `string` and does not accept a `URL`. Using `URL` would just introduce extra casting to and from string

apps/ensindexer/src/config/config.schema.ts
0xBison: this is the case now.

apps/ensindexer/src/api/index.ts
```diff
@@ -18,19 +20,14 @@ import {
   buildGraphQLSchema as buildSubgraphGraphQLSchema,
   graphql as subgraphGraphQL,
 } from "@ensnode/ponder-subgraph";
-import {
-  addDocStringsToIntrospection,
-  extendWithBaseDefinitions,
-  generateTypeDocSet,
-} from "ponder-enrich-gql-docs-middleware";
 
 const {
```
shrugs, line 24: you may be planning on doing this in a future commit, and if so sorry for the eager review note, but it may read better if this destructuring were done during import like

```ts
import { requestedPluginNames, ... } from "@/config/app-config"
```

OR we should always

```ts
import config from "@/config/app-config"
```

and reference variables without destructuring like `config.requestedPluginNames` for clarity.

---

whichever pattern is preferred should be used exclusively, for consistence, in my opinion. personally i like the minimalism of the first option where the config values used in the file are stated and destructured in the import statement

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,118 @@
...
+ * Looks for RPC_URL_{chainId} and optional RPC_REQUEST_RATE_LIMIT_{chainId}.
+ */
+function getChainsFromEnv(): Record<number, ChainConfig> {
+  const chains: Record<number, ChainConfig> = {};
+
+  Object.entries(process.env).forEach(([key, value]) => {
+    const match = key.match(/^RPC_URL_(\d+)$/);
+    if (!match || !value) return;
+
+    const chainId = Number(match[1]);
```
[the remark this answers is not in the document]
shrugs: ah, i forgot about the regex restricting the input to only valid numbers — with that context then we shouldn't need to catch any NaNs and the existing logic makes sense

apps/ensindexer/src/config/app-config.ts
```diff
@@ -96,23 +98,32 @@ export function getConfig() {
  * @returns the rate limit in requests per second (rps)
  */
 export const rpcMaxRequestsPerSecond = (chainId: number): number => {
-  const config = getConfig();
-  if (!config.chains[chainId]?.rpcMaxRequestsPerSecond) {
-    throw new Error(`RPC max requests per second not found for chain ID ${chainId}`);
+  const chainConfig = config.indexedChains[chainId];
```
shrugs, line 101: this can be simplified to `return config.indexedChains[chainId]?.rpcMaxRequestsPerSecond || DEFAULT_RPC_RATE_LIMIT`

apps/ensindexer/src/config/app-config.ts
```diff
@@ -96,23 +98,32 @@ export function getConfig() {
...
+ * will only flag a single missing RPC URL.
+ *
  * @param chainId the chain ID to get the RPC URL for
  * @returns the URL of the RPC endpoint
  */
-export const rpcEndpointUrl = (chainId: number): string => {
-  const config = getConfig();
-  if (!config.chains[chainId]?.rpcEndpointUrl) {
-    throw new Error(`RPC endpoint URL not found for chain ID ${chainId}`);
+export const rpcEndpointUrl = (chainId: number): string | undefined => {
```
shrugs, line 121: this can be simplified to `return config.indexedChains[chainId].rpcEndpointUrl`

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -47,45 +49,76 @@ const customUrlSchema = (envVarKey: string) => {
...
-    },
-  })
-  .default(DEFAULT_DEPLOYMENT);
+// Schema for a variable for a block number
+// Invariant: The value must be a number greater than 0
+const BlockNumberSchema = (envVarKey: string) =>
+  z.coerce
+    .number({ error: `${envVarKey} must be a number.` })
+    .int({ error: `${envVarKey} must be an integer.` })
+    .min(0, { error: `${envVarKey} must be a non-negative number.` })
```
shrugs, line 90: nit: "must be a positive number"

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,178 @@
...
+          return `${envVarKey} has an invalid string value. Please check the format.`;
+        },
+      })
+      // Step 1: Ensure the string, if provided, is not empty or just whitespace.
+      // The .trim() method is applied first, then .min(1) checks the length of the trimmed string.
+      .trim()
+      .min(1, { error: `${envVarKey} is required and cannot be empty.` })
+      // Step 2: If the string is non-empty, then try to parse it as a URL.
+      // This .refine() will only execute if the .min(1) check (on the trimmed string) passes.
+      .refine(
```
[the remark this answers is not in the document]
lightwalker-eth: @0xBison What do you think about creating an issue for this in Zodv4? Ideally they should fix this on their side eventually so that we don't need special logic for this ourselves in the long-term.

apps/ensindexer/src/config/app-config.ts
shrugs: and since it's a rather simple getter now, perhaps the getter is unnecessary and consumers can read `indexedChains` themselves.

in fact the zod schema could/should ensure that if a given chain is defined in the map, `rpcEndpointUrl` is guaranteed to exist and be valid and rpcMaxRequestsPerSecond is guaranteed to exist and be defaulted to DEFAULT_RPC_RATE_LIMIT already.

so callers could do some check to see if the relevant chain was configured `if (config.indexedChains[chainId]) { ... }` and if so have guaranteed access to `config.indexedChains[chainId].rpcEndpointUrl` and `config.indexedChains[chainId].rpcMaxRequestsPerSecond`

---

for example the "does chain exist" behavior is a bit duplicated in validation.ts `doesRpcUrlExistForChain` and the pattern of accessing `config.indexedChains` directly is already used in `validateChainConfigs` so i think it makes sense to go all the way and remove these two getters

apps/ensindexer/src/config/app-config.ts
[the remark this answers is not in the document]
lightwalker-eth: Prefer we apply the following ideas everywhere:

1. All our code is absolutely bulletproof for any possible case. This includes `RPC_URL_SOMESTRING` or other weird cases.
2. We write our code with an optimization for efficiency in reading by other humans over efficiency of writing or the CPU (with special exceptions only when absolutely necessary).
3. If some statements may potentially cause concern or confusion, ex: This call to `Number(...)` I also would have had concern this didn't test for NaN or other use cases, since I didn't look up at the regex until reading your reply in the PR comment. Suggest adding some comment in the code there about the range of possible values of `match[1]` based on the regex.

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,279 @@
...
+  z.coerce
+    .number({ error: `${envVarKey} must be a number.` })
+    .int({ error: `${envVarKey} must be an integer.` })
+    .min(0, { error: `${envVarKey} must be a non-negative number.` })
+    .optional();
+
+/**
+ * The complete runtime configuration for an ENSIndexer application instance.
+ */
+export const ENSIndexerConfigSchema = z.object({
```
lightwalker-eth, line 96: @0xBison Quick feedback: As I understand this is combining the ideas of validation of a particular object with the definition of the object itself?

My first instinct would be to put the definition of the object itself separately. Ex: define an explicit interface for these types separately from the zod logic for validating an instance of that object?

Appreciate your advice.

apps/ensindexer/src/config/types.ts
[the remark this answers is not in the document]
0xBison: Added documentation around this. I've omitted this in the `.env.local.example` file but do you think we should include it there (commented) too?

Just concerned people might accidentally set this and it will appear as everything is going ok but their index could be incorrect.

apps/ensindexer/test/config.test.ts
```diff
@@ -0,0 +1,437 @@
...
+    it("throws if ENS_DEPLOYMENT_CHAIN is an invalid string value", async () => {
+      process.env.ENS_DEPLOYMENT_CHAIN = "not-a-chain";
+      const freshConfigPromise = getFreshConfig();
+      await expect(freshConfigPromise).rejects.toThrow(
+        /Invalid ENS_DEPLOYMENT_CHAIN. Supported chains are: mainnet, sepolia, holesky, ens-test-env/i,
+      );
+    });
+  });
+
+  describe(".requestedPluginNames", () => {
```
lightwalker-eth, line 358: Goals:

1. Improve alignment between environment variable names and config field names. Ex: "active" vs "requested".
   2. For background: previously I suggested adding distinct language for each of these fields because the related code was smashing lots of ideas together in the same place and it was easy to get confused if just "plugin" was used. But now with all the updates we're making I think it's ok to use more simple language.

A few suggestions:

1. Rename the `ACTIVE_PLUGINS` env var to just `PLUGINS`.
   1. NOTE: renaming an ENV var brings a need for a bunch of other updates across our docs and devops scripts / etc. I'm open to moving the idea here into a separate follow up GitHub issue.
2. Rename the `requestedPluginNames` field to just `plugins`.
3. Introduce another validated invariant that that this `plugins` field must never include the same value more than once. In other words, conceptually it should be a set, not a list.

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,303 @@
...
+   */
+  rpcEndpointUrl: url("RPC_URL"),
+
+  /**
+   * The maximum number of RPC requests per second allowed for this chain.
+   * This is used to avoid rate limiting by the RPC provider. This value
+   * is optional and defaults to DEFAULT_RPC_RATE_LIMIT if not specified.
+   *
+   * Invariants:
+   * - The value must be a number greater than 0
```
lightwalker-eth, line 77:
```suggestion
   * - The value must be an integer greater than 0
```

docs/ensnode.io/src/content/docs/ensindexer/usage/configuration.mdx
```diff
@@ -9,6 +9,10 @@ import { LinkCard } from '@astrojs/starlight/components';
 import { Code } from '@astrojs/starlight/components';
 import envExample from '@workspace/apps/ensindexer/.env.local.example?raw';
 
+:::caution[Performance Tip: ENSRainbow Server]
```
lightwalker-eth, line 12: I remember a few days ago we merged a temporary fix for something related to this? Is there something we should do in this PR based on the temporary fix we merged previously?

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -171,8 +175,19 @@ export function networksConfigForChain(chain: Chain) {
...
     [chain.id.toString()]: {
       chainId: chain.id,
-      transport: http(rpcEndpointUrl(chain.id)),
-      maxRequestsPerSecond: rpcMaxRequestsPerSecond(chain.id),
+      // This may return undefined if the RPC URL is not configured for the chain.
+      // This is intentional to allow us to aggregate all the RPC URLs that are missing in a
+      // later validation step allowing us to throw a helpful error message aggregating them
+      // all instead of throwing an error here which will only flag a single missing RPC URL.
+      // The code which does that validation is the `validateChainConfigs` function in
+      // `src/config/validations.ts`.
```
lightwalker-eth, line 183: Really appreciate the comments here!

What do you think about changing the `networksConfigForChain` function such that it always operated off of a fully validated `config`?

It concerns me if ideas related to config validation spread across different parts of our code. Prefer we maximally concentrate that responsibility into a single place, and then help all the other logic across our apps benefit from total confidence that whatever it reads from the config is already maximally validated.

In other words, I understand that Ponder may want some specially formatted config objects of its own. For this, suggest we define functions that take our own validated config object and return back out of them a config object that's needed for a particular Ponder situation.

Appreciate your advice.

apps/ensindexer/test/utils/mockConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+// Default, non-exported mock configuration template
+const _defaultMockConfig: ENSIndexerConfig = {
+  ensDeploymentChain: "mainnet",
+  ensNodePublicUrl: "http://localhost:42069",
+  ensAdminUrl: "http://localhost:3000",
+  ponderDatabaseSchema: "test_schema",
+  requestedPluginNames: [PluginName.Subgraph, PluginName.Basenames, PluginName.Lineanames],
+  ensRainbowEndpointUrl: "https://api.ensrainbow.io",
+  healReverseAddresses: true,
+  port: 42069,
```
lightwalker-eth, line 33: I think we have a named constant for this?

apps/ensindexer/test/utils/mockConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
+      clonedObj[key] = deepClone(obj[key]);
+    }
+  }
+  return clonedObj;
+}
+
+// Default, non-exported mock configuration template
+const _defaultMockConfig: ENSIndexerConfig = {
+  ensDeploymentChain: "mainnet",
```
lightwalker-eth, line 26: Do we have an enum for this?

apps/ensindexer/test/utils/mockConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+ * // Add mainnet configuration
+ * setChainConfig(1, "https://eth-mainnet.g.alchemy.com/v2/1234", 100);
+ *
+ * // Add base chain configuration
+ * setChainConfig(8453, "https://base-mainnet.g.alchemy.com/v2/5678");
+ */
+export function setChainConfig(
+  chainId: number,
+  rpcEndpointUrl: string,
+  rpcMaxRequestsPerSecond: number = 50,
```
lightwalker-eth, line 180: Suggest using `DEFAULT_RPC_RATE_LIMIT`

lightwalker-eth (changes requested):
@0xBison Hey great to see the updates here! Reviewed and shared feedback :+1:

apps/ensindexer/ponder.config.ts
```diff
@@ -1,35 +1,26 @@
...
-import { NetworkConfig } from "ponder";
+import { DatasourceName } from "@ensnode/ens-deployments";
 
 ////////
 // First, generate MergedPluginConfig type representing the merged types of each plugin's `config`,
 // so ponder's typechecking of the indexing handlers and their event arguments is correct.
 ////////
-
-const AVAILABLE_PLUGINS = [
+export const AVAILABLE_PLUGINS = [
```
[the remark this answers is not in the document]
0xBison: Haha I made this change, to me also feels like this should live in a file in the `plugins` dir but @shrugs suggested i move it back... 🤷‍♂️

apps/ensindexer/test/utils/mockConfig.ts
0xBison: I dont believe so. In the `ens-deployments` package we have `export type ENSDeploymentChain = "mainnet" | "sepolia" | "holesky" | "ens-test-env"` but not an enum. This type is also used in ensadmin.

Do you think we should introduce an enum too?

docs/ensnode.io/src/content/docs/ensindexer/usage/configuration.mdx
0xBison: I think the previous change you are speaking of is commenting out the default ENSRAINBOW_URL in the `.env.local.example` file. That hasn't impacted anything but we do need to do some devops checks for the environment variables as there are "breaking changes" in this PR (some fields are now mandated and don't have defaults).

The other change you might be thinking of is a change we introduced to return a default port as port was previously not defaulted. That is also relevant to this PR because `PORT` is no longer required as of this PR as well. The config schema now defaults to the default port so the change already exists in this PR. no action required there.

apps/ensindexer/src/api/index.ts
```diff
@@ -57,8 +52,8 @@ app.onError((error, ctx) => {
 // use root to redirect to the environment's ENSAdmin URL configured to connect back to the environment's ENSNode Public URL
 app.use("/", async (ctx) => {
   try {
-    const ensAdminRedirectUrl = new URL(ensAdminUrl());
-    ensAdminRedirectUrl.searchParams.set("ensnode", ensNodePublicUrl());
+    const ensAdminRedirectUrl = new URL(ensAdminUrl);
```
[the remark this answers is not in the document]
0xBison: I don't agree here personally. Had it like that before and @shrugs suggested i change it and it feels much nicer without the `config.` prefix everywhere. but if you feel super strongly about this i can change it back.

apps/ensindexer/src/config/config.schema.ts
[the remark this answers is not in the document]
tk-o: FYI @0xBison, https://github.com/colinhacks/zod/issues/4103 has been resolved just a while ago 🚀

apps/ensindexer/ponder.config.ts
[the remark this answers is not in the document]
shrugs: i preferred this in the `ponder.config.ts` for the readability of that file but moving it to another file feels tidy and modular as well, go for it

lightwalker-eth:
@tk-o Hey thanks for your feedback [here](https://github.com/namehash/ensnode/pull/698#issuecomment-2884778406).

I spent a few hours earlier today reviewing PR 698 and having a discussion with @0xBison on suggested next steps.

I see some benefits to using a builder pattern as you suggested, but for now suggested to @0xBison not to add that as our priority should be on optimizing our code for reading, rather than writing. I think the builder pattern would make it more difficult to read our code and reason about it. The direction I suggested wot @0xBison should hopefully result in highly readable and simple code.

apps/ensindexer/src/config/config.schema.ts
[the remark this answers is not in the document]
0xBison: we now use the `url` from zod directly.

apps/ensindexer/test/config.test.ts
0xBison: descoped as suggested. issue here: https://github.com/namehash/ensnode/issues/731

apps/ensindexer/src/config/config.schema.ts
0xBison: Now moved back to a dedicated type.

apps/ensindexer/src/config/validations.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Validates the chain configurations in the config and ponder config to ensure that the
+ * indexer has the necessary configuration to index the chains it is configured to index.
+ *
+ * For each plugin, the indexer will attempt to source events from the chains it is configured to index.
+ * This function ensures that the RPC_URL_* environment variables are defined for each chain the indexer
+ * is configured to index.
+ */
+export const validateChainConfigs = (
+  config: ENSIndexerConfig,
+  ponderConfig: MergedPluginConfig,
```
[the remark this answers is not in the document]
0xBison: Each plugin returns an independent ponder config which we merge, so type has been renamed.

apps/ensindexer/ponder.config.ts
```diff
@@ -45,15 +36,13 @@ type MergedPluginConfig = MergedTypes<(typeof AVAILABLE_PLUGINS)[number]["config
 // user-specified plugin is unsupported by the Datasources available in SELECTED_ENS_DEPLOYMENT.
 ////////
 
-const requestedPluginNames = getRequestedPluginNames();
-
 // the available Datasources are those that the selected ENSDeployment defines
 const availableDatasourceNames = Object.keys(SELECTED_ENS_DEPLOYMENT) as DatasourceName[];
```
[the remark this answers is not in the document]
0xBison: `selectedEnsDeployment` is now part of the config and datasources can be assessed in the same way.

apps/ensindexer/ponder.config.ts
```diff
@@ -45,15 +36,13 @@ type MergedPluginConfig = MergedTypes<(typeof AVAILABLE_PLUGINS)[number]["config
...
 
 // filter the set of available plugins by those that are 'active'
 const activePlugins = getActivePlugins(
```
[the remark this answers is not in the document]
0xBison: Yes, plugins should be part of config but the way plugins are currently designed, they depend on config.

so we have `config -> plugins -> config` which doesnt work in any capacity without a refactor. I have covered this in more detail in this separate issue and suggest this is a separate optimization after this PR is merged: https://github.com/namehash/ensnode/issues/735

apps/ensindexer/ponder.config.ts
```diff
@@ -45,15 +36,13 @@ type MergedPluginConfig = MergedTypes<(typeof AVAILABLE_PLUGINS)[number]["config
 // user-specified plugin is unsupported by the Datasources available in SELECTED_ENS_DEPLOYMENT.
```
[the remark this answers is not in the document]
0xBison: this comment is still relevant as of this current pr, suggest it is moved as part of https://github.com/namehash/ensnode/issues/735

apps/ensindexer/ponder.config.ts
```diff
@@ -68,51 +57,12 @@ const ponderConfig = activePlugins
...
-  throw new Error(`ENSNode has been configured with the following ACTIVE_PLUGINS: ${requestedPluginNames.join(", ")}.
-These plugins, collectively, index events from the following chains: ${allChainIds.join(", ")}.
-
-The following RPC_URL_* environment variables must be defined for nominal indexing behavior:
-${allChainIds.map((chainId) => `RPC_URL_${chainId}: ${rpcEndpointUrl(chainId) || "N/A"}`).join("\n")}
-`);
-}
+// Validate the config before we activate the plugins. These validations go beyond simple type
+// validations and ensure any relationships between environment variables are correct.
+validateConfig(config, ponderConfig);
```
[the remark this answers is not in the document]
0xBison: see: https://github.com/namehash/ensnode/issues/735

`ponderConfig` is passed as we require the networks but the plugin refactor will cater for this

apps/ensindexer/src/api/index.ts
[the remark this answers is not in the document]
0xBison: has been reverted.

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -167,14 +156,25 @@ export const activateHandlers =
...
+      transport: http(config.indexedChains[chainId]?.rpcEndpointUrl),
+      // This can only return undefined if there is no RPC url for the chain. This
+      // means the `validateChainConfigs` function in `src/config/validations.ts` will
+      // throw an error so we don't need to handle undefined here with a default.
+      // That is already handled in our schema validation so when the RPC url is added
+      // this will not be undefined.
+      maxRequestsPerSecond: config.indexedChains[chainId]?.rpcMaxRequestsPerSecond,
+      // NOTE: disable cache on local chains (e.g. Anvil, Ganache)
+      ...((chainId === 31337 || chainId === 1337) && { disableCache: true }),
```
0xBison, line 177: @shrugs suggest you check this as i use `chainId` instead of chain name here as it was prior.

anvil defaults to chain `31337` but it our README it says: `Represents a local testing deployment running on an Anvil chain (chain id 1337) with deterministic configurations` so i have included both

apps/ensindexer/ponder.config.ts
```diff
@@ -1,35 +1,26 @@
...
   subgraphPlugin,
   basenamesPlugin,
   lineaNamesPlugin,
   threednsPlugin,
 ] as const;
 
-type MergedPluginConfig = MergedTypes<(typeof AVAILABLE_PLUGINS)[number]["config"]> & {
+export type MergedPluginConfig = MergedTypes<(typeof AVAILABLE_PLUGINS)[number]["config"]> & {
   /**
    * The environment variables that change the behavior of the indexer.
```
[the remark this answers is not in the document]
lightwalker-eth: I suppose my concern here comes from uncertainty over exactly how Ponder checks for code changes. I'm not sure how exhaustive that is. But happy to keep this comment in the code unchanged for now only to reference environment variables.

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,212 @@
...
+    port: parsePort(),
+    ensRainbowEndpointUrl: parseEnsRainbowEndpointUrl(),
+    indexedChains: parseIndexedChains(),
+    databaseUrl: parseDatabaseUrl(),
+    selectedEnsDeployment: parseSelectedEnsDeployment(),
+  })
+  // Add selectedEnsDeployment to the config after as it's a derived value
+  .transform((config) => ({
+    ...config,
+    selectedEnsDeployment: ENSDeployments[config.ensDeploymentChain] as ENSDeploymentGlobalType,
```
tk-o, line 208: I can see this spot as a nice place to include active plugin projection onto `ENSIndexerConfig` object.
In this very spot, we already have access to the parsed `config` value. This enables us to read `config.plugins` and `availableDatasourceNames` (which value is returned from `Object.keys(config.selectedEnsDeployment) as DatasourceName[];` call):
```ts
const activePlugins = getActivePlugins(AVAILABLE_PLUGINS, config.plugins, availableDatasourceNames);
```

This approach would let us to move the declaration of `const AVAILABLE_PLUGINS` from `ponder.config.ts` into `config.schema.ts`.

Alternatively, if using this `transform` callback on the `ENSIndexerConfigSchema` turns out not to be the best place (however, I think it's a place for the purpose), we can move the transformation logic into `buildConfigFromEnvironment` function call of `app-config.ts` file.

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+  /**
+   * Configuration for each indexed chain, keyed by chain ID.
+   *
+   * Invariants:
+   * - Each key (chainId) must be a number
+   * - For any requested plugin, the config will have a key for chainId in this
+   *   indexedChains object which is ensured by the validateChainConfigs
+   *   function in `validations.ts` and not part of the schema validation here.
+   */
+  indexedChains: Record<number, ChainConfig>;
```
shrugs, line 187: nit: `chainConfigs`?

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,97 @@
...
+
+function getValidationErrors(issues: z.core.$ZodIssue[]) {
+  // loop over and concatenate the messages. precede with a line break due
+  // to the way ponder prints errors
+  return `\n` + issues.map((issue) => issue.message).join("\n");
+}
+
+// loads the relevant environment variables in the shape of the zod schema
+function parseEnvironment(): ENSIndexerEnvironment {
+  const config: ENSIndexerEnvironment = {
```
shrugs, line 54: nit: just `return { ... }` should suffice with the return type specified in the function, or perhaps `return { ... } satisfies ENSIndexerEnvironment` to be explicit

shrugs (approved):
this looks GREAT to me, well done. config.schema.ts feels lovely as well and reads so cleanly. excited to have this in the codebase

apps/ensindexer/src/config/types.ts
0xBison: i like this, @lightwalker-eth what are your thoughts, as this was previously `chains` and you suggested we change it to `indexedChains`.

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   *   function in `validations.ts` and not part of the schema validation here.
+   */
+  plugins: Set<PluginName>;
+
+  /**
+   * A feature flag to enable or disable healing of addr.reverse subnames.
+   * If this is set to true, ENSIndexer will attempt to heal subnames of
+   * addr.reverse.
+   *
+   * If this is not set, the default value is set to `DEFAULT_HEAL_REVERSE_ADDRESSES`.
```
lightwalker-eth, line 148: This idea should move to where we translate from environment variables into an ENSIndexerConfig.

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   * compatibility with the ENS Subgraph, this should be set to `false`.
+   *
+   * Invariant: The value must be 'true' or 'false'
+   */
+  healReverseAddresses: boolean;
+
+  /**
+   * The network port ENSIndexer listens for http requests on. ENSIndexer
+   * exposes various APIs, most notably the GRAPHQL API. All APIs are accessible
+   * on the same port. This defaults to DEFAULT_PORT if not specified.
```
lightwalker-eth, line 161: The following idea should move to where we build an ENSIndexerConfig, not on the interface itself.

> This defaults to DEFAULT_PORT if not specified.

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   * Invariant: The URL must be a valid URL. localhost urls are allowed,
+   * and expected.
+   */
+  ensRainbowEndpointUrl: string;
+
+  /**
+   * Configuration for each indexed chain, keyed by chain ID.
+   *
+   * Invariants:
+   * - Each key (chainId) must be a number
```
lightwalker-eth, line 182: We should make this more strict. For example:

```suggestion
   * - Each key (chainId) must be a non-negative integer
```

I'm not sure if it must be positive or non-negative?

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   * - For any requested plugin, the config will have a key for chainId in this
```
lightwalker-eth, line 183: @shrugs Hey could you please help to suggest refinements to how we document the invariant relationships here?

`indexedChains` should be a function of `plugins` and the ENS deployment.

Appreciate if we can describe how data sources fit into this narrative too. Thanks!

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * Represents the raw unvalidated environment variables for a single chain.
+ *
+ * Since an RPC_URL_<chainId> is checked in `getChainsFromEnv`, the value
+ * for rpcEndpointUrl is guaranteed to be a non-empty string.
+ *
+ * rpcMaxRequestsPerSecond is optional and will be undefined if the
+ * RPC_REQUEST_RATE_LIMIT_<chainId> environment variable is not set.
+ */
+export interface RawChainConfig {
```
lightwalker-eth, line 208:
```suggestion
export interface ChainEnvironment {
```

Not sure, just a quick brainstorming idea. I thought it might be nice to align the naming of this interface with the naming of `ENSIndexerEnvironment`.

apps/ensindexer/src/plugins/index.ts
lightwalker-eth, line 1: Should we remove this file? I note that it's empty.

apps/ensindexer/test/utils/mockConfig.ts
```diff
@@ -0,0 +1,215 @@
...
+  ensDeploymentChain: "mainnet",
+  selectedEnsDeployment: ENSDeployments.mainnet,
+  ensNodePublicUrl: "http://localhost:42069",
+  ensAdminUrl: "http://localhost:3000",
+  ponderDatabaseSchema: "test_schema",
+  plugins: new Set([PluginName.Subgraph, PluginName.Basenames, PluginName.Lineanames]),
+  ensRainbowEndpointUrl: "https://api.ensrainbow.io",
+  healReverseAddresses: true,
+  port: DEFAULT_PORT,
+  indexedChains: {},
```
lightwalker-eth, line 38: It shouldn't be possible for this to be empty when we have the 3 plugins on the mainnet ENS deployment.

Please see other related comments I'll share in this PR review :+1:

apps/ensindexer/test/utils/mockConfig.ts
lightwalker-eth: Suggest using a named constant here.

apps/ensindexer/test/utils/mockConfig.ts
```diff
@@ -0,0 +1,215 @@
...
+  const clonedObj = {} as T;
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
+      clonedObj[key] = deepClone(obj[key]);
+    }
+  }
+  return clonedObj;
+}
+
+// Default, non-exported mock configuration template
```
lightwalker-eth, line 26: Suggest building this from the environment interface you created for a config.

In other words, all instances of an `ENSIndexerConfig`, including this mock instance, should be guaranteed to always go through the validation logic we're building.

This will be helpful as over time we'll continue evolving the invariants and schema. It's good that our unit tests fail here if we forgot something.

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -93,7 +74,15 @@ export function getActivePlugins<PLUGIN extends ENSIndexerPlugin>(
 
     if (!hasRequiredDatasources) {
```
lightwalker-eth, line 75: In my understanding, this check should happen as part of building and validating an `ENSIndexerConfig` and not here.

It should be impossible to have an `ENSIndexerConfig` that fails this test anywhere in our code.

Please see several other related comments I've shared on this PR.

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -61,28 +58,12 @@ export function makePluginNamespace<PLUGIN_NAME extends PluginName>(pluginName:
  */
 export function getActivePlugins<PLUGIN extends ENSIndexerPlugin>(
   availablePlugins: readonly PLUGIN[],
-  requestedPluginNames: string[],
+  requestedPluginNames: Set<PluginName>,
   availableDatasourceNames: DatasourceName[],
```
lightwalker-eth, line 62: Should we include these DatasourceName values as a field in the config object too? I feel like the config we build should include a data model that's aligned with the plugin -> data source relationship.

Appreciate if this is too complex for this PR. If it is, but you think this is a nice idea, can we please create a follow up GitHub Issue to remember the idea for the future? Thanks

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -167,14 +156,25 @@ export const activateHandlers =
...
-      ...(chain.name === "Anvil" && { disableCache: true }),
+    [chainId.toString()]: {
+      chainId: chainId,
+      // This may return undefined if the RPC URL is not configured for the chain.
+      // This is intentional to allow us to aggregate all the RPC URLs that are missing in a
+      // later validation step allowing us to throw a helpful error message aggregating them
+      // all instead of throwing an error here which will only flag a single missing RPC URL.
+      // The code which does that validation is the `validateChainConfigs` function in
+      // `src/config/validations.ts`.
+      transport: http(config.indexedChains[chainId]?.rpcEndpointUrl),
```
lightwalker-eth, line 169: This function should first validate that `config.indexedChains[chainId]` is defined. If it isn't, it should throw an exception.

After validating that, there should be no need for the `?` here.

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -167,14 +156,25 @@ export const activateHandlers =
...
+      // This can only return undefined if there is no RPC url for the chain. This
+      // means the `validateChainConfigs` function in `src/config/validations.ts` will
+      // throw an error so we don't need to handle undefined here with a default.
+      // That is already handled in our schema validation so when the RPC url is added
+      // this will not be undefined.
+      maxRequestsPerSecond: config.indexedChains[chainId]?.rpcMaxRequestsPerSecond,
```
lightwalker-eth, line 175: Similar feedback here as above.

lightwalker-eth (changes requested):
@0xBison Really super work here! 🚀 This is a partial review but about to hop on some long calls and didn't want to further delay sharing some feedback 🫡

apps/ensindexer/src/config/types.ts
shrugs: currently this property in the config defines "the set of chain configurations (`rpcEndpoint` & `maxRequests`) provided by the environment" and _isn't_ filtered by those that are explicitly indexed.

to derive "the set of all actively indexed chains" in ensadmin, for example, the `networks` property of the ponder config should be used, as that's the actual set of chains ponder is indexing at runtime. i.e. the plugins determine which networks they require by specifying as such in the `networks` property of their ponder config.

i don't think it's important for this property in the `ENSIndexerConfig` (current: `indexedChains`, proposed: `chainConfigs`) to only represent explicitly indexed chains, and i think the fact that it represents how the user is configuring ENSIndexer is ideal, actually. the decision making of which chains are actively indexed is done by each of the active plugins, not the ENSIndexerConfig, which is just providing the _configuration options_ for each of the possible networks ensindexer can index from.

so i'm still in favor of `chainConfigs` (and possibly even changing the object to `{ endpoint: string, maxRequestsPerSecond: number }` for minimalism) and don't think any additional re-architecture of this parameter is necessary. and the documentation should be updated to read more like `"Configuration for each indexable chain"...`

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   * This is used to avoid rate limiting by the RPC provider. This value
+   * is optional and defaults to DEFAULT_RPC_RATE_LIMIT if not specified.
+   *
+   * Invariants:
+   * - The value must be an integer greater than 0
+   */
+  rpcMaxRequestsPerSecond: number;
+}
+
+export interface GlobalBlockrange {
```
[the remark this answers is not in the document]
shrugs: i believe that's handled on `apps/ensindexer/src/config/config.schema.ts` line 56

```ts
.refine(
      (val) =>
        val.startBlock === undefined ||
        val.endBlock === undefined ||
        val.startBlock <= val.endBlock,
      { error: "END_BLOCK must be greater than or equal to START_BLOCK." },
    );
```

BUT i just tested this with ponder and if start and end blocks are the same, nothing is indexed, even though it feels like just that exact block should be indexed. will log bug with ponder but in the meantime we should switch this comparison to be just `<` and update the error message.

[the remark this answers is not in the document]
shrugs: sounds good to me, actually we should use re-use the `Blockrange` type from `apps/ensindexer/src/lib/types.ts` here

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   * (see `@ensnode/ens-deployments` for available deployments)
+   */
+  ensDeploymentChain: keyof typeof ENSDeployments;
+
+  /**
+   * The selected ENS Deployment for the indexer which could be "mainnet", "sepolia", etc.
+   *
+   * (see `@ensnode/ens-deployments` for available deployments)
+   */
+  selectedEnsDeployment: ENSDeploymentGlobalType;
```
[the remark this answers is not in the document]
shrugs: i see, one is the key and the other is the value.

perhaps just the key can be kept in the config, called `ensDeployment` (and we can remove the `.transform` to include `selectedENSDeployment`) and instead consumers can use a helper like the following to access the selected `ENSDeployment` object, typed correctly

```ts
export const getENSDeployment = (deployment: keyof typeof ENSDeployments) => ENSDeployments[deployment] as ENSDeploymentGlobalType

// then in a plugin...
const { chain, contracts } = getEnsDeployment(config.ensDeployment)[DatasourceName.Basenames];
```

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   * If this is not set, DEFAULT_ENSADMIN_URL will be used to provide easy access to an ENSAdmin UI.
+   *
+   * Invariants:
+   * - The URL must be a valid URL (localhost urls are allowed)
+   */
+  ensAdminUrl: string;
+
+  /*
+   * This is a namespace for the tables that the indexer will create to store indexed data.
+   * It should be a string that is unique to the running indexer instance.
```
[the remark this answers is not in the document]
shrugs: here's my attempt

```ts
  /**
   * A Postgres database schema name. This instance of ENSIndexer will write indexed data to the
   * tables in this schema.
   *
   * The `ponderDatabaseSchema` must be unique per running instance of ENSIndexer (ponder will
   * enforce this with database locks). If multiple instances of ENSIndexer with the same
   * `ponderDatabaseSchema` are running, only the first will successfully acquire the lock and begin
   * indexing: the rest will crash.
   *
   * If an ENSIndexer instance with the same configuration (including `ponderDatabaseSchema`) is
   * started, and it successfully acquires the lock on this schema, it will continue indexing from
   * the current state.
   *
   * Many clients can read from this Postgres schema during or after indexing.
   *
   * Read more about database schema rules here:
   * https://ponder.sh/docs/api-reference/database#database-schema-rules
   *
   * Invariant: Must be a non-empty string that is a valid Postgres database schema identifier.
   */
```

apps/ensindexer/src/config/types.ts
```diff
@@ -0,0 +1,236 @@
...
+   */
+  ponderDatabaseSchema: string;
+
+  /**
+   * Identify which indexer plugins to activate (see `src/plugins` for available plugins)
+   * This is a set of one or more available plugin names.
+   *
+   * Invariants:
+   * - A set of valid plugin names with at least one value
+   * - For any requested plugin, the config will have a key for chainId in the
```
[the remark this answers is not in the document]
shrugs: here's what we should encode into the zod schema:

1) the set of selected plugins must be 'available' within the specified `ENSIndexerConfig.ensDeployment`
2) the set of datasources required by each plugin must be available within the `ENSDeployments[ENSIndexerConfig.ensDeployment]`
3) the set of chains required by each datasource must be specified within the `ENSIndexerConfig.chainConfigs` (or `ENSIndexerConfig.rpcConfigs`)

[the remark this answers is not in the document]
shrugs: so combining some of the active plugins and other validations inline with the schema. let me take a shot at it

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+  return chains;
+}
+
+function getValidationErrors(issues: z.core.$ZodIssue[]) {
+  // loop over and concatenate the messages. precede with a line break due
+  // to the way ponder prints errors
+  return `\n` + issues.map((issue) => issue.message).join("\n");
+}
+
+// loads the relevant environment variables in the shape of the zod schema
```
lightwalker-eth, line 52:
```suggestion
// loads the relevant environment variables in the shape of a ENSIndexerEnvironment
```

Is that fair?

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * This function scans all environment variables for keys matching the pattern
+ * "RPC_URL_{chainId}", where {chainId} must be a string of digits (e.g., "1", "10", "8453").
+ *
+ * It then looks for matching environment variables for the rate limit of requests per second
+ * for the given chain, using the pattern "RPC_REQUEST_RATE_LIMIT_{chainId}".
+ *
+ * This function returns a raw chain config which is not yet validated against the zod schema.
+ */
+export function getChainsFromEnv(): Record<number, RawChainConfig> {
```
lightwalker-eth, line 16: Maybe no need to export?

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+  };
+}
+
+/**
+ * Builds the ENSIndexer configuration object from an ENSIndexerEnvironment object
+ *
+ * This function then validates the config against the zod schema ensuring that the config
+ * meets all type checks and invariants.
+ *
+ * **Note:** This function does **not** perform deep validation of the configuration values
```
lightwalker-eth, line 78: Appreciate your advice for what we would need to do so that this function would perform this deep validation?

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function getValidationErrors(issues: z.core.$ZodIssue[]) {
+  // loop over and concatenate the messages. precede with a line break due
+  // to the way ponder prints errors
+  return `\n` + issues.map((issue) => issue.message).join("\n");
+}
+
+// loads the relevant environment variables in the shape of the zod schema
+function parseEnvironment(): ENSIndexerEnvironment {
```
lightwalker-eth, line 53: I'm open to exporting this.

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * This function then validates the config against the zod schema ensuring that the config
+ * meets all type checks and invariants.
+ *
+ * **Note:** This function does **not** perform deep validation of the configuration values
+ * beyond basic type checks and presence. Comprehensive validation (such as checking for
+ * required environment variables, correct formats, or logical consistency) is handled
+ * separately by dedicated validation utilities elsewhere in the codebase.
+ */
+function buildConfigFromEnvironment(environment: ENSIndexerEnvironment): ENSIndexerConfig {
```
lightwalker-eth, line 83: Suggest we export this.

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (!parsed.success) {
+    throw new Error(
+      "Failed to parse environment configuration: " + getValidationErrors(parsed.error.issues),
+    );
+  }
+
+  return parsed.data;
+}
+
+export default buildConfigFromEnvironment(parseEnvironment());
```
lightwalker-eth, line 95: I note how in some files we have code like the following (case 1):

```
import config from "@/config/app-config";
```

and in other files we have code like the following (case 2):

```
import { default as appConfig } from "@/config/app-config";
```

Is there a special reason case 2 couldn't be made the same as case 1? It would be ideal if all looked like case 1 as best as possible. It's more clean.

apps/ensindexer/src/config/app-config.ts
```diff
@@ -0,0 +1,95 @@
...
+      endBlock: process.env.END_BLOCK,
+    },
+    indexedChains: getChainsFromEnv(),
+  };
+}
+
+/**
+ * Builds the ENSIndexer configuration object from an ENSIndexerEnvironment object
+ *
+ * This function then validates the config against the zod schema ensuring that the config
```
lightwalker-eth, line 75: As I understand zod is an implementation detail of this function and therefore it would be best for us not to mention it in the comments here.

In other words, in theory we should be able to swap out zod for a different parsing / validation framework without it impacting "external" code. As I understand it, zod is an internal implementation detail for how an `ENSIndexerInvironment` is converted into an `ENSIndexerConfig`.

apps/ensindexer/test/config.test.ts
```diff
@@ -0,0 +1,514 @@
+import { ENSIndexerConfig } from "@/config/types";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Helper to set and unset environment variables for each test
```
lightwalker-eth, line 4: The current approach all looks like it's working :+1: Nice work here!

What do you think about specifically testing the `buildConfigFromEnvironment` function in this file (and refining the unit tests so that they modify a "base" `ENSIndexerEnvironment`)?

Or do you think it's better to build all the unit tests around parsing real env vars?

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -167,14 +156,25 @@ export const activateHandlers =
 /**
  * Defines a ponder#NetworksConfig for a single, specific chain.
```
lightwalker-eth, line 157:
```suggestion
 * Builds a ponder#NetworksConfig for a single, specific chain.
```

apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -167,14 +156,25 @@ export const activateHandlers =
...
  */
-export function networksConfigForChain(chain: Chain) {
+export function networksConfigForChain(config: ENSIndexerConfig, chainId: number) {
```
lightwalker-eth, line 159: I note how here we explicitly pass in a `config` here while in general in other places we are reading from the "global" config defined in `app-config.ts`. If there's a special reason for this, cool, just wanted to ask if we should remove the `config` param here?

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,212 @@
...
+    .enum(Object.keys(ENSDeployments) as [keyof typeof ENSDeployments], {
+      error: (issue) => {
+        return `Invalid ENS_DEPLOYMENT_CHAIN. Supported chains are: ${Object.keys(
+          ENSDeployments,
+        ).join(", ")}`;
+      },
+    })
+    .default(DEFAULT_DEPLOYMENT);
+
+const parseGlobalBlockrange = () =>
```
lightwalker-eth, line 50: Maybe just `parseBlockrange`? This relates to another comment I shared about this interface.

lightwalker-eth (changes requested):
@0xBison Once again nice work here :+1: Finished my review with additional feedback. Appreciate your advice! Thanks 🫡

apps/ensindexer/src/config/types.ts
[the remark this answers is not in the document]
lightwalker-eth: Ok cool. Will be good to also refine the docs we put on this type to fully define all the expected invariants.

apps/ensindexer/src/config/types.ts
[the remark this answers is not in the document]
lightwalker-eth: That would be nice!

apps/ensindexer/src/config/types.ts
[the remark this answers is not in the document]
lightwalker-eth: @shrugs Thanks! That helps but I should have written my concern here more clearly. Specifically I'm referring to the difference between `ponder start` vs `ponder serve`. Many ENSIndexer instances can actually share the same `ponderDatabaseSchema` when running `ponder serve` (read only mode) while only a single ENSIndexer instance can use a given `ponderDatabaseSchema` at a time when running `ponder start` (read and write mode).

I was hoping we could explain this idea here without an expectation for the reader to have studied all of Ponder's docs first. Ideally people who run ENSNode only ever need to read our docs and not to also read Ponder's as well.

apps/ensindexer/src/config/app-config.ts
shrugs: sadly it's cause the word `config` is overloaded as a ponder config in the plugin files. thinking about solutions.

apps/ensindexer/test/config.test.ts
shrugs: in this file i think parsing env is helpful test case, but in other test files like `ponder-helpers.test.ts` they can use the `updateMockConfig` or `setGlobalBlockrange` etc helpers to play with the config directly rather than changing the env variables and re-parsing the environment

apps/ensindexer/src/lib/plugin-helpers.ts
0xBison: i was trying to remove the `config` import from the file to fix the circular dependency and instead inject the config via params but we have now split this file to resolve the circular dependency so we no longer need to remove the `config` import

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -0,0 +1,212 @@
...
+import { parse as parseConnectionString } from "pg-connection-string";
+import { z } from "zod/v4";
+
+export const DEFAULT_RPC_RATE_LIMIT = 50;
+export const DEFAULT_ENSADMIN_URL = "https://admin.ensnode.io";
+export const DEFAULT_PORT = 42069;
+export const DEFAULT_HEAL_REVERSE_ADDRESSES = true;
+export const DEFAULT_DEPLOYMENT = "mainnet";
+
+const parseBlockNumber = (envVarKey: string) =>
```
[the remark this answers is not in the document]
0xBison: Are you sure? That will look like:

`const parseBlockNumber = (
  envVarKey: string
): z.ZodOptional<z.ZodType<number>>`

and

`const parseDatabaseUrl = (): z.ZodUnion<[z.ZodString, z.ZodUndefined]>`

and

`const parseEnsAdminUrl = (): z.ZodDefault<z.ZodURL>`

to give some examples

lightwalker-eth (approved):
@0xBison @shrugs Looks great! Appreciate all of the special efforts here! 🫡 🚀 🚀 Let's ship it!

Issue #695, opened 2025-05-06
Ensure any errors for Ponder runtime integrity checks cause CI to fail


lightwalker-eth:
Goals to fix here include:

- [ ] Ensure errors on this "static-analysis" job correctly result in an error in our CI.
- [ ] Resolve this error, so that all jobs in our CI pass successfully.

# Goal 1

https://github.com/namehash/ensnode/actions/runs/14857221835/job/41713316573?pr=694#step:11:35

Here our Ponder runtime integrity checks had an error, however, as you can see, the CI didn't error as expected.

![Image](https://github.com/user-attachments/assets/48c256e9-c9eb-4307-a22b-2aca830a93d7)

Here's related code: https://github.com/namehash/ensnode/blob/main/.github/workflows/test_ci.yml#L56-L82

# Goal 2

The background info here seems to be related to ENSIndexer needing to provide a URL back to itself in some API responses. That URL needs to include the port that ENSIndexer is running on.

https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/ponder-helpers.ts#L346-L375

So it looks like this code is expecting a PORT environment variable to always be explicitly set, however we aren't explicitly setting it here: https://github.com/namehash/ensnode/blob/main/.github/workflows/test_ci.yml#L63-L70

We should always look at every issue from the perspective of all of our users and docs, too. 

For example, if ENSIndexer requires an explicit PORT environment variable to be set, shouldn't this idea be reflected in other places too such as:
1. https://github.com/namehash/ensnode/blob/main/examples/example-docker-compose/docker-compose.yml#L2-L22
2. https://github.com/namehash/ensnode/blob/main/docker-compose.yml#L2-L22
3. https://github.com/namehash/ensnode/blob/main/apps/ensindexer/.env.local.example
4. And maybe also https://github.com/namehash/ensnode/blob/main/apps/ensindexer/Dockerfile

Advice appreciated!

PR #702, opened 2025-05-08
ci: add healthcheck script for checking ensindexer properly starts or…

0xBison:
… terminate with a failure

Resolves: https://github.com/namehash/ensnode/issues/695

The root cause of this is `pnpm dev` which runs the ponder `dev` command doesn’t exit if there’s issues it just hangs and waits for changes to hot reload. So the timeout was hitting and then it was passing. I think what we were trying to check here is the app starts with no issues so I’ve done this in a slightly different way. We now check for the app healthcheck to return a 200 meaning the app is up and running - the ci will then pass. If we don’t hit the healthcheck within the timeout then something is up. The goal here is to check the app boots and nothing more. We are not checking for errors beyond app start. I actually did a lot of tests locally and in ci and sometimes we do get errors because we’re using public rpcs and requests fail due to rate limits.

So you can see in the commits of this PR the 2 cases - pass and fail. First and last commit are the current ci which is failing as no port is configured and the current codebase requires it. The middle commit is when I hardcoded the port in the script, it passes. So the CI failing right now is actually the correct state of things.

### DIFF
--- a/.github/scripts/run_ensindexer_healthcheck.sh
+++ b/.github/scripts/run_ensindexer_healthcheck.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
-# Simple script to verify Ponder starts up correctly
+# Simple script to verify ENSIndexer starts up correctly
 # This script is used in CI and can be run locally
 
 # Set default timeout if not provided by environment
-: "${HEALTH_CHECK_TIMEOUT:=60}"  # Use env var if set, otherwise default to 60
+# Use env var if set, otherwise default to 60 seconds
+: "${HEALTH_CHECK_TIMEOUT:=60}" 
 
 # Detect if running from CI or local
 if [ -n "$GITHUB_WORKSPACE" ]; then
@@ -13,7 +14,8 @@ if [ -n "$GITHUB_WORKSPACE" ]; then
 else
   # Running locally - determine path relative to script location
   SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-  REPO_ROOT="$( cd "$SCRIPT_DIR/../.." && pwd )"  # Adjusted for .github/scripts path
+  # Adjusted for .github/scripts path
+  REPO_ROOT="$( cd "$SCRIPT_DIR/../.." && pwd )"
   ENSINDEXER_DIR="$REPO_ROOT/apps/ensindexer"
 fi
 
@@ -23,17 +25,18 @@ cd "$ENSINDEXER_DIR" || {
   exit 1
 }
 
-echo "Starting Ponder in the background from $(pwd)..."
+echo "Starting ENSIndexer in the background from $(pwd)..."
 
 # Create a temporary log file
 LOG_FILE=$(mktemp)
 echo "Logging output to $LOG_FILE"
 
-# Run the Ponder dev command in background and redirect output to log file
+# Run ENSIndexer using the Ponder dev command in background
+# and redirect output to log file
 pnpm dev --disable-ui -vv > "$LOG_FILE" 2>&1 &
 PID=$!
 
-echo "Ponder started with PID: $PID"
+echo "ENSIndexer started with PID: $PID"
 
 # Wait for health check to pass
 echo "Waiting for health check to pass (up to $HEALTH_CHECK_TIMEOUT seconds)..."
@@ -53,7 +56,7 @@ while true; do
 
   # Check if the process is still running
   if ! ps -p $PID > /dev/null; then
-    echo "Ponder process exited before health check passed"
+    echo "ENSIndexer process exited before health check passed"
     wait $PID
     EXIT_CODE=$?
     echo "Exit code: $EXIT_CODE"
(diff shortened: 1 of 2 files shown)


.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
...
+    # Clean up env file
+    [ -f "$ENV_FILE" ] && rm -f "$ENV_FILE"
+    exit 1
+  fi
+
+  # Check for health ready message
+  if grep -q "Started returning 200 responses from /health endpoint" "$LOG_FILE"; then
+    echo "Health check passed! Ponder is up and running."
+    echo "Test successful - terminating Ponder"
+    # Force kill the Ponder process
```
lightwalker-eth, line 72:
```suggestion
    # Force kill the ENSIndexer process
```

.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
+#!/bin/bash
+
+# Simple script to verify Ponder starts up correctly
+# This script is used in CI and can be run locally
```
lightwalker-eth, line 4: I'm really not an expert here, just sharing some quick brainstorm ideas. Please attack them!

Would any of the following make sense?
1. Creating a new directory in our repo for `/scripts` and moving this file there instead of nested under `/.github/scripts`?
2. Adding something to the root `package.json` file for executing this script via `pnpm`?

Appreciate your advice :+1:

.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# Set default timeout if not provided by environment
+: "${HEALTH_CHECK_TIMEOUT:=60}"  # Use env var if set, otherwise default to 60
```
lightwalker-eth, line 7:
```suggestion
: "${HEALTH_CHECK_TIMEOUT:=60}"  # Use env var if set, otherwise default to 60 seconds
```

.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
...
+  ENSINDEXER_DIR="$REPO_ROOT/apps/ensindexer"
+fi
+
+# Navigate to the ensindexer directory
+cd "$ENSINDEXER_DIR" || {
+  echo "Error: Could not navigate to $ENSINDEXER_DIR"
+  exit 1
+}
+
+echo "Starting Ponder in the background from $(pwd)..."
```
lightwalker-eth, line 26:
```suggestion
echo "Starting ENSIndexer in the background from $(pwd)..."
```

.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# Create a temporary log file
+LOG_FILE=$(mktemp)
+echo "Logging output to $LOG_FILE"
+
+# Run the Ponder dev command in background and redirect output to log file
```
lightwalker-eth, line 32:
```suggestion
# Run ENSIndexer using the Ponder dev command in background and redirect output to log file
```

.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
...
+pnpm dev --disable-ui -vv > "$LOG_FILE" 2>&1 &
+PID=$!
+
+echo "Ponder started with PID: $PID"
```
lightwalker-eth, line 36:
```suggestion
echo "ENSIndexer started with PID: $PID"
```

.github/scripts/run_ensindexer_healthcheck.sh
```diff
@@ -0,0 +1,101 @@
...
+  if [ $((current_time - last_log_check)) -ge 15 ]; then
+    echo "Still waiting for health check at $(date) (elapsed: $((current_time - health_check_start)) seconds)..."
+    echo "Recent log entries:"
+    tail -n 10 "$LOG_FILE"
+    last_log_check=$current_time
+  fi
+
+  # Check if the process is still running
+  if ! ps -p $PID > /dev/null; then
+    echo "Ponder process exited before health check passed"
```
lightwalker-eth, line 56:
```suggestion
    echo "ENSIndexer process exited before health check passed"
```

.github/scripts/run_ensindexer_healthcheck.sh
0xBison: I actually ran this by claude and he suggested the same as you. personally, i prefer how it is now but don't feel that strongly either way. since this script is primarily for ci feels like it should live there but neither solution feels wrong. i'll wait for someone else's thoughts to be the deciding vote.

shrugs: yeah i like that it lives in the github folder since it's for CI — if that ever changes we can hoist it later

shrugs (approved):

PR #703, opened 2025-05-09
feat: set default for port if not set

0xBison:


lightwalker-eth (approved):
@0xBison Appreciate this fix, thank you :+1:

PR #665, opened 2025-04-23
feat(gql): introduce LLM-based generator for queries

tk-o:
In order to test this new feature:
1. Include `ANTHROPIC_API_KEY` in `apps/ensadmin/.env.local` file if you have it (but it's optional)
2. Start ENSAdmin service (i.e. `pnpm -F ensadmin dev`)
3. Go to https://adminensnodeio-git-feat-gql-query-llm-generator-namehash.vercel.app/gql/api?gqlApiUrl=https://api.alpha.ensnode.io/ponder&prompt=latest%20domain%20names%20ending%20with%20.eth
4. Observe response:
```
{
  "generateQueryDto": {
    "prompt": "latest domain names ending with .eth",
    "gqlApiUrl": "https://api.alpha.ensnode.io/ponder"
  },
  "generatedQuery": {
    "query": "query ($limit: Int, $orderBy: String, $orderDirection: String, $where: domainFilter) {\n  domains(\n    limit: $limit\n    orderBy: $orderBy\n    orderDirection: $orderDirection\n    where: $where\n  ) {\n    items {\n      createdAt\n      id\n      labelName\n      name\n      owner {\n        id\n      }\n      registration {\n        expiryDate\n        registrationDate\n      }\n    }\n    pageInfo {\n      endCursor\n      hasNextPage\n    }\n    totalCount\n  }\n}",
    "variables": {
      "limit": 10,
      "orderBy": "createdAt",
      "orderDirection": "desc",
      "where": {
        "name_ends_with": ".eth"
      }
    }
  }
}
```

PR #708, opened 2025-05-11
refactor(gql): reuse new options for LLM-based generator

danstarns:
I noticed that you made your own Anthropic Adapter in the text to query the implementation: 

* https://github.com/namehash/ensnode/pull/665

[GQLPT alpha.32](https://github.com/rocket-connect/gqlpt/releases/tag/0.0.0-alpha.32) takes your version and exports it - this should increase maintainability for you.


djstrong:
Thank you! I think you have dropped `cache: ephemeral` and this is crucial because it saves >95% tokens. With this enhancement we easily get into rate limits so maybe we will change the API supplier. Also this is the first version of ENSNode query generator and we need full control of the prompt to make it more specific.

danstarns:
> Thank you! I think you have dropped `cache: ephemeral` and this is crucial because it saves >95% tokens. With this enhancement we easily get into rate limits so maybe we will change the API supplier. Also this is the first version of ENSNode query generator and we need full control of the prompt to make it more specific.

Good point about the cache, thanks for the extra pair of eyes. 

I added the cache support in [GQLPT alpha.33](https://github.com/rocket-connect/gqlpt/releases/tag/0.0.0-alpha.33). 

In the name of rebase gods, I reopened this PR with less noise https://github.com/namehash/ensnode/pull/709

PR #713, opened 2025-05-12
fix(ensindexer): use tx sender when healing addresses from reverse registry

### LINKED ISSUE #370: Heal remaining addr.reverse names
About 5% (50k) of all adrr.reverse names are not healed currently. Probably the reason is that labelhashes does not match.

```sql
SELECT 
    DATE_TRUNC('month', TO_TIMESTAMP(created_at)) AS month,
    COUNT(*) AS count
FROM 
    ens.domains
WHERE 
    parent_id = '0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2' 
    AND label_name IS NULL
GROUP BY 
    DATE_TRUNC('month', TO_TIMESTAMP(created_at))
ORDER BY 
    month;
```

Image

```sql
SELECT 
    DATE_TRUNC('month', TO_TIMESTAMP(created_at)) AS month,
    COUNT(*) AS total_count,
    COUNT(CASE WHEN label_name IS NULL THEN 1 END) AS nonhealed_count,
    COUNT(CASE WHEN label_name IS NOT NULL THEN 1 END) AS healed_count
FROM 
    ens5addr8.domains
WHERE 
    parent_id = '0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2'
GROUP BY 
    DATE_TRUNC('month', TO_TIMESTAMP(created_at))
ORDER BY 
    month;
```

Image

tk-o:
This PR addresses a bug introduced in some implementations of Reverse Registry contracts used across time by ENS.

For example, the [ENS: Old Reverse Registrar 2](https://etherscan.io/address/0x084b1c3C81545d370f3634392De611CaaBFf8148) contract would pass the contract's own address as the `NewOwner` event's `owner` value:
image

To work around this issue, we will use `event.transaction.from` which is a proxy for `msg.sender` known from the fixed implementation of the Reverse Registry contract:
image

Resolves #370

### DIFF
--- a/.changeset/moody-bats-trade.md
+++ b/.changeset/moody-bats-trade.md
@@ -1,0 +1,5 @@
+---
+"ensindexer": patch
+---
+
+Use transaction sender address for healing label from reverse registry.


shrugs (approved):
LGTM good find

PR #715, opened 2025-05-13
fix Registrar onlyRegister patch comments

shrugs:
- hoists the invariant inline with documentation
- updated comments reflecting better understanding of context


lightwalker-eth (approved):
@shrugs Nice update :+1: Suggest adding a changeset for this PR before merging it.

Issue #718, opened 2025-05-14
handle null byte in text_changed


shrugs:
3:12:58 AM WARN  indexing   Detected and removed null byte characters from the string '' inserted into the text_changed.value column. Postgres "text" columns do not support null byte characters. Please consider handling this case in your indexing logic.

https://etherscan.io/tx/0x2eb93d872a8f3e4295ea50773c3816dcaea2541f202f650948e8d6efdcbf4599#eventlog

PR #719, opened 2025-05-14
fix: strip null bytes from text changed events

shrugs:
closes https://github.com/namehash/ensnode/issues/718


lightwalker-eth (approved):
@shrugs Looks good :+1: Thanks!

PR #721, opened 2025-05-14
fix: move dnsPacketBytes to dns-helpers and document

### LINKED ISSUE #41: Refine decodeDNSPacketBytes
This issue is a reference to `decodeDNSPacketBytes` in `src/lib/subname-helpers.ts`.

Ideally we would refine the return type here so that it is a named type / interface, where each field / array index is documented for what it represents. We should also document what a null value represents in each field.

shrugs:
closes #41 

i don't think these helpers need to be in ensnode/utils, especially because we were having trouble with the Buffer typings, since ensnode utils specifies browser libs and whatnot. so for simplicity and the tightness of the ensnode/utils lib, i've removed the dnspacketbytes lib and just exported all of the subname helpers from the root index — it's much cleaner now and especially on the consumer end where all imports come from `@ensnode/utils` rather than some some `@ensnode/utils/subname-helpers`

no changeset necessary, imo

### DIFF
--- a/apps/ensindexer/src/handlers/NameWrapper.ts
+++ b/apps/ensindexer/src/handlers/NameWrapper.ts
@@ -1,11 +1,12 @@
 import { type Context } from "ponder:registry";
 import schema from "ponder:schema";
 import { checkPccBurned } from "@ensdomains/ensjs/utils";
-import { type Node, PluginName } from "@ensnode/utils";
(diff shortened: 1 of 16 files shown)


lightwalker-eth (approved):
@shrugs Looks nice! :+1: Thanks!

Issue #534, opened 2025-04-08
ENSIndexer: Analyze & Remove `ownedSubnameNode` usage in NameWrapper.ts


shrugs:
in NameWrapper we use `registrarManagedName` (fka `ownedName`) to determine if, when a WappedDomain is Unwrapped, the corresponding domain's expiry should be set to null (it should be nulled if and only if it is a direct child of `.eth`). 

The plugins implement this behavior as well, but it may be unnecessary, and it depends on whether the plugins expect their names' expiries to work identically to the root plugin's treatment of .eth names.

https://github.com/namehash/ensnode/blob/171cae4cd0ba4306923da73459f8b970ca39bac3/apps/ensindexer/src/handlers/NameWrapper.ts#L182

additionally, this logic should be changed to the following, for readability and simplicity

```ts
// if domain's parent is registrar's managed name, keep existing expiry, otherwise null it out
expiryDate: domain.parentId === registrarManagedNameNode ? undefined : null,
```

PR #722, opened 2025-05-14
docs: add comments about NameWrapper's usage of registrar managed name

shrugs:
closes https://github.com/namehash/ensnode/issues/534

adds discussion for why that value is used in NameWrapper

changeset not necessary for internal comments


lightwalker-eth (approved):
@shrugs Thanks for updates :+1:

PR #726, opened 2025-05-15
fix: only register preminted names

shrugs:
these preminted names are only registered in the Registrar (hence need a Registration) but aren't registered in the Registry (hence are not yet a Domain) — this fix removes the domain creation and simply allows the skipping of the domain.update if the domain does not exist in the basenames plugin. it's much cleaner, more specific, and more accurate.

one downside of this is that the subgraph schema specifies a non-null domainId relationship. in this current patch i've kept the domainId as `node` of the preminted names, which seems alright, and results in api results like this, where the preminted names end and the normally registered names begin

image

the downside of this is that ponder doesn't support where filters over relationships so it's not possible to actually do a filter like

```graphql
registrations(limit: 1000, where:{domain: null}) { ... }
```

in order to identify these preminted names easily.

an alternative solution is to use the zerohash for the domainId of the preminted names, and one could do

```graphql
registrations(limit: 1000, where:{domainId: zerohash}) { ... }
```

and easily filter for them. but i'm not convinced it's strictly necessary ¯\_(ツ)_/¯

### DIFF
--- a/apps/ensindexer/src/handlers/Registrar.ts
+++ b/apps/ensindexer/src/handlers/Registrar.ts
@@ -96,30 +96,22 @@ export const makeRegistrarHandlers = ({
 
       // NOTE: because the mainnet ENS contract _always_ emit Registry#NewOwner _before_
       // Registrar#NameRegistered, the subgraph logic expects a domain entity to exist here.
+      //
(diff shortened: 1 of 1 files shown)


tk-o (approved):
This makes a lot of sense, LGTM!

PR #727, opened 2025-05-16
Draft: plugin docs

0xBison:
Some `TODO`s still need to be resolved. Still draft state


docs/ensnode.io/src/content/docs/ensindexer/indexed-data.mdx
```diff
@@ -0,0 +1,48 @@
...
+sidebar:
+  label: Indexed Data
+  order: 2
+---
+
+:::note
+The specific data points and event types indexed by ENSIndexer can evolve. The best way to explore the currently available data and its structure is by inspecting the GraphQL schema via the GraphiQL playground.
+
+You can access the playground at:
+- Public instance: [https://api.mainnet.ensnode.io/graphql](https://api.mainnet.ensnode.io/graphql)
```
0xBison, line 13: this link doesnt actually work but i think it should as my local has this endpoint and not the `/ponder` endpoint.

I think we migrated recently but havent updated prod? I followed this new `/graphql` endpoint instead of `/ponder` as I believe its the latest.

docs/ensnode.io/config/integrations/starlight.ts
```diff
@@ -27,7 +28,7 @@ export function starlight(): AstroIntegration {
       "@fontsource/inter/900.css",
     ],
     plugins: [
-      starlightLlmsTxt(),
+      // starlightLlmsTxt(),
```
0xBison, line 31: This plugin causes a problem with the page with the react component on. It takes in an `exclude` property and i've tried a lot of patterns to exclude the page with issues with no success:

```
16:21:20   └─ /llms-full.txt[NoMatchingRenderer] Unable to render `EnsFlowDiagram`.

No valid renderer was found for the `.tsx` file extension.
  Hint:
    This issue often occurs when your MDX component encounters runtime errors.
  Error reference:
    https://docs.astro.build/en/reference/errors/no-matching-renderer/
  Location:
    undefined:0:0
  Stack trace:
  ```

0xBison:
Bullet point styling is broken as reported here: https://github.com/namehash/ensnode/issues/739

PR #736, opened 2025-05-20
feat: new sidebar content in ensnode.io/docs

Y3drk:
Change of `TelegramInvite` component that:
- [x] redesigns the Telegram group invite to achieve unified styling with the one from LP
- [x] adds new CTA, inviting visitors to give ensnode repo on GitHub a star

Current components match Figma designs and behave the same way as the rest of the sidebar.


notrab (approved):
@Y3drk this looks great!

This is fine to merge now in my opinion, BUT I do see we changed from using CSS variables to the tailwind classes like `text-gray-500` which makes changing the theme of this in the future require work to update this component again to support any other variants - like dark mode.

No big deal for now I guess, since it's only us using and maintaining it.

Y3drk:
> @Y3drk this looks great!
> 
> This is fine to merge now in my opinion, BUT I do see we changed from using CSS variables to the tailwind classes like `text-gray-500` which makes changing the theme of this in the future require work to update this component again to support any other variants - like dark mode.
> 
> No big deal for now I guess, since it's only us using and maintaining it.

@notrab I've decided to go with a middle-of-the-road solution, that is moving the styles regarding colors to the CSS variables and leaving "shape-related" styles as tailwind. In my opinion that way we can keep the code fairly clean and give the components the flexibility for future changes. 

Let me know what you think :)

lightwalker-eth (approved):
@Y3drk Looks great! Super work 🫡

PR #740, opened 2025-05-21
feat: ENSAdmin - indexed chains panel UI improvements

### LINKED ISSUE #506: ENSAdmin "Indexed Chains" panel UI improvements
Please see related Slack discussion here: https://namehash.slack.com/archives/C086Z6FNBHN/p1743767904699929

Image

Several more suggestions for improving the ENSAdmin UI:
1. Change to "Indexed Chains"
2. Suggest we add a .svg icon for each chain defined in the ens-deployments package, such that we can guarantee that any chain shown in this UI will have a .svg icon available for it. Then, next to the name of each of these chains, we should show an icon for the related chain. This will help to better communicate the multichain capabilities of ENSNode at an emotional level.
3. Can we make each of these block numbers a link to view details about the related block on a block explorer for the related chain? These links should open in a new tab. This logic needs to be written in a way that if a chain doesn't have a block explorer (for example, Anvil in the ens-test-env) that we just don't make the block number a hyperlink.
4. It seems it would be better to display these time values as relative timestamps such as "5 years ago" (as an example for the case that someone has recently started to perform a historical backfill indexing of mainnet) or a value such as "7 seconds ago" or "1 minute ago" or possibly even "just now" for cases where the relative timestamps are very near or directly at the present time. Please note the special case where the user's device could have a time configuration set into the past, such that the "Latest safe block" could theoretically appear as something like "in 3 years" which would look really weird. I'm open to suggestions for how we handle this exotic edge case, my first thought is to keep the handling of this edge case extremely simple and say that if the timestamp of a block is >= (not just ==) to the current time we display that as "just now".

Y3drk:
Closes #506 , which includes:
- [x] New panel title
- [x] SVG icons for each chain
- [x] Make block number a link to block explorer for the chains that allow it
- [x] Change time display from date to relative timestamp


Y3drk:
# For whoever may review
I am a little confused about the display of block's timestamp. 

In the `Latest indexed registrations` panel we leverage react hooks for display, but this approach wasn't employed in the `Indexed chains` panel (have a look at `BlockStats` component)

I was unsure what the right approach here was, and continued with the "static" display for the relative timestamp. 
Changing it to leverage hooks should be super easy, I am just not sure we need them.

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -86,43 +94,75 @@ interface NetworkIndexingStatsCardProps {
   network: NetworkStatusViewModel;
 }
 
+type chainName =
+  | "Base"
+  | "OP Mainnet"
+  | "Ethereum"
+  | "Linea Mainnet"
+  | "Sepolia"
+  | "Anvil"
```
lightwalker-eth, line 103: Please check with Tomasz and Matt what name is appropriate to give to a local ethereum. I'm not sure it is fair to always call it "Anvil" and maybe we need a more generic term.

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -86,43 +94,75 @@ interface NetworkIndexingStatsCardProps {
...
+  | "Holesky"
+  | "3DNS";
```
lightwalker-eth, line 105: Please see my other related comment. 3DNS is not a chain.

lightwalker-eth (changes requested):
@Y3drk Hey great job! 🚀 Reviewed and shared some suggestions with feedback :+1:

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -132,17 +172,56 @@ function BlockStats({ label, block }: BlockSatsProps) {
     );
   }
 
+  let calculatedRelativeTime = formatRelativeTime(block.timestamp.toString());
+
+  if (
+    block.timestamp <= Math.floor(Date.now() / 1000) ||
+    calculatedRelativeTime === "less than a minute ago"
```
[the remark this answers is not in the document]
Y3drk: @lightwalker-eth 
I investigated on such Base blocks:
1. 30742968 - index time: May-26-2025 04:01:23 PM +UTC
2. 30742961 - index time: May-26-2025 04:01:09 PM +UTC

The time difference between them is **14s**, so the fact that both of them had `just now` relative timestamp was in line with the code at the time.

Why? Because of how we set up the `formatRelativeTime` function and the `formatDistanceToNow` helper function inside it. In the version I used till now, it would unify all timestamps that mark time under a minute ago to `less than a minute ago`, which would then be caught in the component and simplified to `just now`.

But your conclusion that the timestamps differ and it should be visible in the panel is very much valid, hence I made some changes, and now `formatRelativeTime` will accept an optional argument that will be responsible for injecting an `includeSeconds` option into the helper function, which will in turn return a more fine-grained relative timestamp. 

Summing up:
* The timestamps will now be more fine-grained and will include texts such as `less than X seconds ago` which should make the panel more detailed and less boring (multitude of descriptions rather than `just now` everywhere)
* The `just now` text will now be solely reserved for the *future* edge case you described in the issue.
* The optional argument allows backwards compatibility so there is no need to change any code from `recent-registrations` component that also uses the `formatRelativeTime` function

Below I'll paste how the panel looks with the aforementioned changes:
image

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -132,17 +172,56 @@ function BlockStats({ label, block }: BlockSatsProps) {
...
+If the chain is supported on blockexplorer.com it will display it as an external link to the block's details
+ */
+function BlockNumber({ networkName, block }: BlockNumberProps) {
+  // for now blockexplorer.com only allows inspection of Ethereum blocks, the list may change if we switch platform or their capabilities expand
+  const blockExplorerAvailableChains: chainName[] = ["Ethereum"];
+
+  if (blockExplorerAvailableChains.includes(networkName)) {
+    return (
+      <a
+        href={`https://www.blockexplorer.com/${networkName.toLowerCase().replace(" ", "")}/block/${block.number}/#overview`}
```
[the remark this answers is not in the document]
tk-o: Spoke with @Y3drk about using the [`Datasource.chain`](https://github.com/namehash/ensnode/blob/main/packages/ens-deployments/src/lib/types.ts#L21) getter to access selected chain configuration object defined by [`viem`: `Chain` type](https://github.com/wevm/viem/blob/48cffbc/src/types/chain.ts#L18-L64).

Please note, the `Chain` value is not fully validated, i.e. it's type allows undefined block explorer.

apps/ensadmin/src/components/indexing-status/components.tsx
tk-o: We can leverage `viem`'s native configuration object for various chains. For example:

https://github.com/namehash/ensnode/blob/95119d2/apps/ensadmin/src/components/indexing-status/view-models.ts#L58

Y3drk: @lightwalker-eth 

Since https://github.com/wevm/viem/tree/main/src/chains/definitions contain definition for `Anvil` and lack some obvious generic replacement for it, I stuck with the current naming convention, all while applying new chain-icon approach that you proposed in the other comment.

@tk-o 
If I missed some blatantly obvious replacement let me know :)

apps/ensadmin/src/components/indexing-status/view-models.ts
```diff
@@ -15,16 +15,27 @@ export interface NetworkIndexingPhaseViewModel {
   endDate: Date;
 }
 
+/**
+ * Information about chain's block explorer, matching viem's definition @ https://github.com/wevm/viem/blob/main/src/types/chain.ts#L266
+ */
+export interface ChainBlockExplorer {
+  name: string;
+  url: string;
+  apiUrl?: string | undefined;
```
tk-o, line 24: We might not need the API URL prop. If that's the case, we could drop it:
```suggestion
```

apps/ensadmin/src/components/indexing-status/view-models.ts
```diff
@@ -87,15 +103,19 @@ export function globalIndexingStatusViewModel(
 
 /**
  * View model for the network indexing status.
+ * @param chainId
  * @param chainName
  * @param networkStatus
  * @param firstBlockToIndexGloballyTimestamp
+ * @param chainsBlockExplorer - optional
```
tk-o, line 110: Let's make this one not optional and provide consistent UX for all chains 👍

tk-o, line 110: Let's make this one not optional and provide consistent UX for all chains 👍 
```suggestion
 * @param chainsBlockExplorer
```

apps/ensadmin/src/components/indexing-status/view-models.ts
```diff
@@ -116,12 +136,14 @@ export function networkIndexingStatusViewModel(
...
 
   return {
+    id: chainId,
     name: chainName,
     latestSafeBlock: blockViewModel(latestSafeBlock),
     firstBlockToIndex: blockViewModel(firstBlockToIndex),
     lastIndexedBlock: lastIndexedBlock ? blockViewModel(lastIndexedBlock) : null,
     lastSyncedBlock: lastSyncedBlock ? blockViewModel(lastSyncedBlock) : null,
     phases,
+    ...(chainsBlockExplorer && { blockExplorer: chainsBlockExplorer }),
```
tk-o, line 146: Let's make this one not optional and provide consistent UX for all chains 👍 
```suggestion
    blockExplorer: chainsBlockExplorer,
```

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -94,15 +94,32 @@ interface NetworkIndexingStatsCardProps {
...
-  | "3DNS";
+/**
+ * Mapping of chain's id to its icon.
+ * Based on chain definitions provided by viem @ https://github.com/wevm/viem/blob/main/src/chains/definitions
+ */
+export const chainIcons = new Map<number, React.ReactNode>([
+  [1, <EthereumIcon width={18} height={18} />],
+  [8453, <BaseIcon width={18} height={18} />],
+  [11_155_111, <EthereumTestnetIcon width={18} height={18} />],
+  [10, <OptimismIcon width={16} height={16} />],
```
lightwalker-eth, line 105:
```suggestion
  [10, <OptimismIcon width={16} height={16} />], // Optimism Mainnet
```

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -94,15 +94,32 @@ interface NetworkIndexingStatsCardProps {
...
-  | "Holesky"
-  | "3DNS";
+/**
+ * Mapping of chain's id to its icon.
+ * Based on chain definitions provided by viem @ https://github.com/wevm/viem/blob/main/src/chains/definitions
+ */
+export const chainIcons = new Map<number, React.ReactNode>([
+  [1, <EthereumIcon width={18} height={18} />],
+  [8453, <BaseIcon width={18} height={18} />],
+  [11_155_111, <EthereumTestnetIcon width={18} height={18} />],
```
lightwalker-eth, line 104:
```suggestion
  [11_155_111, <EthereumTestnetIcon width={18} height={18} />], // Sepolia
```

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -94,15 +94,32 @@ interface NetworkIndexingStatsCardProps {
...
+  [10, <OptimismIcon width={16} height={16} />],
+  [59_144, <LineaIcon width={18} height={18} />],
+  [17000, <EthereumTestnetIcon width={18} height={18} />],
+  [31_337, <EthereumLocalIcon width={18} height={18} />],
```
lightwalker-eth, line 108: Suggest we not add the `_` into these chain id numbers. That complicates cases where we want to search our code for any reference to a particular chainid.

apps/ensadmin/src/components/indexing-status/components.tsx
```diff
@@ -193,22 +206,19 @@ function BlockStats({ networkName, label, block }: BlockSatsProps) {
...
-If the chain is supported on blockexplorer.com it will display it as an external link to the block's details
+If the chain has a designated block explorer it will display it as an external link to the block's details
  */
-function BlockNumber({ networkName, block }: BlockNumberProps) {
-  // for now blockexplorer.com only allows inspection of Ethereum blocks, the list may change if we switch platform or their capabilities expand
-  const blockExplorerAvailableChains: chainName[] = ["Ethereum"];
-
-  if (blockExplorerAvailableChains.includes(networkName)) {
+function BlockNumber({ blockExplorer, block }: BlockNumberProps) {
+  if (blockExplorer) {
```
lightwalker-eth, line 218: Some chains don't have a blockExplorer, such as Anvil. So we need to prepare for the case it won't be defined.

However, even if the blockExplorer is defined, it seems we maybe still need to support the case that `block.number` might be undefined? I'm not familiar with the related invariants but I'm saying this based on the line of code several lines beneath this.

apps/ensadmin/src/components/indexing-status/view-models.ts
```diff
@@ -55,15 +66,20 @@ export function globalIndexingStatusViewModel(
...
   );
   const firstBlockToIndexGloballyTimestamp = Math.min(...indexingStartDatesAcrossNetworks);
-  const getChainName = (chainId: number) => getChainById(ensDeploymentChain, chainId).name;
 
   const networkStatusesViewModel = Object.entries(networkIndexingStatus).map(
-    ([chainId, networkIndexingStatus]) =>
-      networkIndexingStatusViewModel(
-        getChainName(parseInt(chainId, 10)),
+    ([chainId, networkIndexingStatus]) => {
+      const chainIdAsNumber = parseInt(chainId, 10);
```
lightwalker-eth, line 72: There's some cancer here. We need to identify where this potentially invalid chainId value is coming from that isn't already guaranteed to be a valid chainId and fix it at the source. I'm completely against spreading validation logic like this everywhere across our code. This is a form of cancer. We must enforce strong validation "at the edge" and then keep everything in our core guaranteed perfect invariants.

apps/ensadmin/src/components/indexing-status/view-models.ts
```diff
@@ -55,15 +66,20 @@ export function globalIndexingStatusViewModel(
...
+      const chain = getChainById(ensDeploymentChain, chainIdAsNumber);
```
lightwalker-eth, line 73: Why do we need to pass an `ensDeploymentsChain` into this function? How could that potentially influence what chain is returned? Why isn't this a pure function of `chainId`?

lightwalker-eth (changes requested):
@Y3drk Thanks for updates :+1: Reviewed and shared feedback.

apps/ensadmin/src/components/indexing-status/view-models.ts
```diff
@@ -55,15 +66,20 @@ export function globalIndexingStatusViewModel(
...
+
+      return networkIndexingStatusViewModel(
```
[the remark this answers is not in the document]
tk-o: As just discussed, block explorer link should be optional. For example, `anvil` chain configuration does not include such link and we should expect that on the UI side.

Y3drk:
Added a new `<ChainName />` component that is an initial implementation of `getChainName` function proposed in [the related Slack convo](https://namehash.slack.com/archives/C086Z6FNBHN/p1749645007239009?thread_ts=1749571541.829009&cid=C086Z6FNBHN).

The component is located at `ensadmin/src/components/ui` and so far it's only used in `Indexed Chains` panel (at `ensadmin/src/components/indexing-status/components.tsxt`).

If we preferred to just get a string instead of returning a whole React Node we could just make this a function, it all depends on the conception I guess.

tk-o:
This PR will be unblocked by #828

apps/ensadmin/src/components/ui/ChainName.tsx
```diff
@@ -0,0 +1,44 @@
...
+ */
+const chainNames = new Map<number, string>([
+  //TODO: these prettified versions of names will probably change to better accommodate our needs
+  [mainnet.id, "Ethereum"],
+  [base.id, "Base"],
+  [sepolia.id, "Ethereum Sepolia"],
+  [optimism.id, "Optimism"],
+  [linea.id, "Linea"],
+  [holesky.id, "Ethereum Holesky"],
+  [anvil.id, "Ethereum Local"],
```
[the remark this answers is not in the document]
shrugs: also correct me if i'm wrong but the chain object should have a `.name` as well maybe we can just skip this step

lightwalker-eth (changes requested):
@Y3drk Thanks for the updates 👍 Reviewed and shared suggestions

apps/ensadmin/src/components/ui/ChainName.tsx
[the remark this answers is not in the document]
Y3drk: @shrugs

> for ensadmin we should define a lib/chains.ts or whatever and ens-test-env chain can be defined like const ensTestEnv = { ...anvil, id: 1337 } and used as a chain object anywhere in the codebase

Seems like a good idea, I'll try to implement this and see where the changes are necessary.

> also correct me if i'm wrong but the chain object should have a .name as well maybe we can just skip this step

It was previously decided (I think there is a related discussion somewhere in this PR) that we don't want to rely on names in these mappings, but the `id`s should be the way to go

PR #742, opened 2025-05-21
Feat/validations pr updates

shrugs:
- moves datasource spec to plugins/index to decouple circular dependency between plugins and config
- moves all config validation to the zod schema
  - the individual functions could be extracted further
- updates ENSIndexerConfig and comments based on pr notes
- uses `vi.stubEnv` instead of custom env stubbing in `app-config.test.ts`


apps/ensindexer/src/lib/plugin-helpers.ts
```diff
@@ -154,33 +105,25 @@ export const activateHandlers =
   };
 
 /**
- * Defines a ponder#NetworksConfig for a single, specific chain.
+ * Builds a ponder#NetworksConfig for a single, specific chain.
  */
-export function networksConfigForChain(config: ENSIndexerConfig, chainId: number) {
+export function networksConfigForChain(chainId: number) {
+  // NOTE: config.rpcConfigs[chainId] is guaranteed to exist, see app-config.schema.ts
```
lightwalker-eth, line 111: Any chainId value could be passed here, it's not guaranteed to be constrained only to chainIds with rpcConfigs in the config.

Therefore we should check if `config.rpcConfigs[chainId]` is undefined and if so, throw an error.

lightwalker-eth (approved):
@shrugs Super to see these updates! Nice work 🙌

0xBison (approved):

PR #743, opened 2025-05-22
deps(ponder): update to v0.10.28

tk-o:
Includes cached `debug_traceCall` RPC call method released with:
- https://github.com/ponder-sh/ponder/pull/1767

Only `ponder` dependency got a version bump, as the rest of `@ponder/*` packages with version `0.10.26` do not include any code updates (it's just version bumps after looking at changelogs of those packages) and the version `0.10.28` was not released for these.


shrugs (approved):
LGTM

Issue #723, opened 2025-05-14
Prototype Protocol Accelerated Resolution


shrugs:
- index resolver values but avoid exposing via api
  - could be structured as a separate plugin perhaps, modeled after prototype v2 schema resolver entities
- implement api server, ala prototype v2 branch
- implement forward resolution
  - implement CCIP-Read without any acceleration
  - then implement acceleration heuristics
- implement reverse resolution
- implement primary name resolution & verification
- note support for cross-chain "L2 Primary Names" (ENSIP-19)
- hide the prefixed entity names in the subgraph api, expose via ponder?

PR #746, opened 2025-05-23
Index Resolver Values

shrugs:
beginning of https://github.com/namehash/ensnode/issues/723

- adds a set of `ext_` prefixed entities to track resolver records
  - this is safe for subgraph compatibility at the api layer because ENSIndexer will now serve a superset of the subgraph entities.
  - i don't think it's currently trivial to conditionally include/exclude the `ext_` entities from the schema (and the autogenerated api) based on whether this runtime flag is enabled, so if `INDEX_RESOLVER_RECORDS` is false, these records will just not exist and any api calls will return empty array or null
  - currently just represents address records by coinType, text records by key, and `name` under `records` — can add the others if necessary but they seem not useful at the current moment
  - access via `domain(id: "") { resolver { addressRecords textRecords } }` (can't use `texts` because conflicts with existing `texts` coinType array from subgraph)
- adds a derived `isSubgraphCompatible` flag to the `ENSIndexerConfig` which is derived from the other inputs
  - it's more or less just an internal representation of whether or not the shared behaviors should operate with subgraph compatibility in mind or not
  - i don't _love_ that it's in ENSIndexerConfig but i think it's relatiely aligned, as ENSIndexerConfig is, specifically, an internal configuration object. it's the `ENSIndexerEnvironment` that's user-facing and intended to be used. lmk if you have any other ideas. i didn't like the idea of just adding a global based on `config`, nor did i love a `ENSIndexerInternalConfig` that's built from `ENSIndexerConfig`
- refactors all of the id generation to use the new `config.isSubgraphCompatible` flag
  - this resolves the resolver reference issue i noticed [in slack here](https://namehash.slack.com/archives/C086Z6FNBHN/p1748018943438899)
  - the best way to review this change is via [this commit](https://github.com/namehash/ensnode/commit/e25650717bace6b352e35639336fe04939346bfc) in isolation with hide whitespace turned on
- added tests for the new `isSubgraphCompatible` flag and refactored the `ids.test.ts` tests for the new pattern
- adds pretty logging of the config at startup for easier debugging 
- refactors the config.schema to be more zod-like [in this commit](https://github.com/namehash/ensnode/pull/746/commits/90f8fa95b122b2ad29a7d548efdf351e0bc63655)
---

if you review this pr all at once don't forget to hide whitespace diffs

---

Screenshot 2025-05-23 at 13 56 28

image


apps/ensindexer/src/handlers/NameWrapper.ts
```diff
@@ -3,9 +3,9 @@ import schema from "ponder:schema";
 import { checkPccBurned } from "@ensdomains/ensjs/utils";
 import { type Address, type Hex, hexToBytes, namehash } from "viem";
 
-import { type Node, PluginName, uint256ToHex32 } from "@ensnode/utils";
+import { type Node, uint256ToHex32 } from "@ensnode/utils";
 
-import { makeSharedEventValues, upsertAccount } from "@/lib/db-helpers";
+import { sharedEventValues, upsertAccount } from "@/lib/db-helpers";
```
tk-o, line 8: `sharedEventValues` is not used in this file. Do we need to use that import? Otherwise, we should drop it.

apps/ensindexer/src/handlers/Registrar.ts
```diff
@@ -10,7 +10,7 @@ import {
   makeSubdomainNode,
 } from "@ensnode/utils";
 
-import { makeSharedEventValues, upsertAccount, upsertRegistration } from "@/lib/db-helpers";
+import { sharedEventValues, upsertAccount, upsertRegistration } from "@/lib/db-helpers";
```
tk-o, line 13: I can't see `sharedEventValues` function being called anywhere in this file. Should we drop the import, or use the function somehow?

apps/ensindexer/src/handlers/Registrar.ts
```diff
@@ -59,7 +58,7 @@ export const makeRegistrarHandlers = ({
     // materialize the registration's labelName as well
     await context.db
       .update(schema.registration, {
-        id: makeRegistrationId(pluginName, labelHash, node),
+        id: makeRegistrationId(labelHash, node),
```
tk-o, line 61: Nice one on using the config context to drop explicit `pluginName` dependency 👍

apps/ensindexer/ponder.config.ts
```diff
@@ -31,7 +32,10 @@ export type MergedPonderConfig = MergedTypes<(typeof ALL_PLUGINS)[number]["confi
    *
    * https://ponder.sh/docs/api-reference/ponder/database#build-id-and-crash-recovery
    **/
-  indexingBehaviorDependencies: Pick<ENSIndexerConfig, "healReverseAddresses">;
+  indexingBehaviorDependencies: Pick<
+    ENSIndexerConfig,
+    "healReverseAddresses" | "indexResolverRecords"
```
tk-o, line 37: 👍

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -12,177 +12,173 @@ import {
...
+// parses an env string bool with strict requirement of 'true' or 'false'
+const makeEnvStringBoolSchema = (envVarKey: string) =>
+  z
+    .string()
+    .pipe(
+      z.enum(["true", "false"], {
+        error: `${envVarKey} must be 'true' or 'false'.`,
+      }),
+    )
+    .transform((val) => val === "true");
```
tk-o, line 32: Nice helper schema 👍

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -12,177 +12,173 @@ import {
...
+    plugins: PluginsSchema,
+    healReverseAddresses: HealReverseAddressesSchema,
+    indexResolverRecords: IndexResolverRecordsSchema,
+    port: PortSchema,
+    ensRainbowEndpointUrl: EnsRainbowEndpointUrlSchema,
+    rpcConfigs: RpcConfigsSchema,
+    databaseUrl: DatabaseUrlSchema,
   })
+  // inject ENSIndexerConfig.subgraphCompatibility
+  .transform(derive_subgraphCompatibility)
```
tk-o, line 180: Cool pattern! It makes me keen to learn Zod v4 API 🤓

apps/ensindexer/src/handlers/NameWrapper.ts
shrugs: we still use it, it's just hidden in the diff because those lines didn't change — previously we were constructing this function using the `make*` factory but now it's just the function `sharedEventValues` so all of those events that were logging are still using the same function. applies to similar comments below

lightwalker-eth (commented):
@shrugs Hey super to see this PR! Great work and excited for the future possibilities here 🫡 Reviewed and shared some questions with feedback.

Also will be great to include updates to the following in this PR:

1. https://github.com/namehash/ensnode/blob/main/apps/ensindexer/.env.local.example
2. Suggest we expand the docs for the "Purpose" of the ENSNode 'Alpha' deployment here: https://ensnode.io/docs/usage/hosted-ensnode-instances/#available-ensnode-instances to identify how we are also beginning other experimental work for ENSv2 indexing such as ENS Protocol Acceleration.
3. Could you please review our docs and identify any areas that might be helpful to update or expand upon?
4. How about the GraphQL schema docs?

apps/ensindexer/src/plugins/shared/Resolver.ts
```diff
@@ -11,58 +10,41 @@ import { ENSIndexerPluginHandlerArgs } from "@/lib/plugin-helpers";
...
   // https://github.com/ensdomains/ens-contracts/blob/b6bc1eac9d241b7334ce78a51bacdf272f37fdf5/contracts/resolvers/profiles/IDNSRecordResolver.sol#L6
   ponder.on(
     "Resolver:DNSRecordChanged(bytes32 indexed node, bytes name, uint16 resource, bytes record)",
-    handleDNSRecordChanged,
+    handlers.handleDNSRecordChanged,
   );
   // NOTE: this DNSRecordChanged ABI spec with the included `ttl` parameter is specific to
   // 3DNS' Resolver implementation, but we include it here for theoretical completeness, were a
-  // Resolver indexed by these shared handlers to emit this event.
+  // Resolver indexed by these shared rh.handlers to emit this event.
```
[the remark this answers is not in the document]
shrugs: vestige of previous work in the pr

apps/ensindexer/test/config.test.ts
```diff
@@ -132,17 +132,17 @@ describe("config", () => {
   describe(".ensNodePublicUrl", () => {
     it("throws an error if ENSNODE_PUBLIC_URL is not a valid URL", async () => {
       vi.stubEnv("ENSNODE_PUBLIC_URL", "invalid url");
-      await expect(getConfig()).rejects.toThrow(/ENSNODE_PUBLIC_URL must be a valid URL string/i);
+      await expect(getConfig()).rejects.toThrow(/must be a valid URL string/i);
```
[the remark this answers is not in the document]
shrugs: sure thing!

shrugs:
this has been updated:
- i've folded the `ResolverRecords` entity into `Resolver` and just added the `name` property to it — it seems simplest.
- added a bunch of documentation related to the Resolver records schema that should help clear up any confusion in that file.

shrugs:
also confirmed subgraph compatible

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -12,177 +12,179 @@ import {
...
+    plugins: PluginsSchema,
+    healReverseAddresses: HealReverseAddressesSchema,
+    indexResolverRecords: IndexResolverRecordsSchema,
+    port: PortSchema,
+    ensRainbowEndpointUrl: EnsRainbowEndpointUrlSchema,
+    rpcConfigs: RpcConfigsSchema,
+    databaseUrl: DatabaseUrlSchema,
   })
+  // inject ENSIndexerConfig.isSubgraphCompatible
+  .transform(derive_isSubgraphCompatible)
```
tk-o, line 186: Thinking about deriving the `ensDeployment` config prop. Right now, whenever we need to access the current ENS Deployment object, we'd import `getENSDeployment` function and pass the `config.ensDeploymentChain` as a param:
image

I propose we derive `ensDeployment` property and include it on `ENSIndexerSchema`

```suggestion
  .transform(derive_isSubgraphCompatible)
  .transform(derive_ensDeployment)
```

Where `derive_ensDeployment` is defined as the following:
```ts
const derive_ensDeployment = (
  config: Omit<ENSIndexerConfig, "ensDeployment">,
): ENSIndexerConfig {
  return {
    ...config,
    ensDeployment: getENSDeployment(config.ensDeploymentChain),
  };

Comment (tk-o) [reply apps/ensindexer/src/config/config.schema.ts:186]: If we feel the idea is OK, but it needs a follow up PR, I'm happy to create a GH issue ;) 

Also, thinking of deriving active plugins list in a similar fashion, and later using this list in `ponder.config.ts` to actually activate relevant plugins.

File: apps/ensindexer/src/lib/lib-config.ts [hunk 10]
```diff
@@ -43,3 +43,23 @@ export function getRpcConfigsFromEnv(): Record<number, RpcConfigEnvironment> {
...
+          {
+            ...rpcConfig,
+            url: "*******",
+          },
+        ]),
+      ),
+    } as ENSIndexerConfig,
+    null,
+    2,
+  );
```
tk-o, line 64: Feel like a nice candidate for the `toPublic` method we spoke about some other time: https://github.com/namehash/ensnode/issues/521#issuecomment-2802662311

I'll take a note to come back to this function once ENSNode SDK comes to life 🚀

tk-o (approved):
I left couple of comments, but nothing critical. Merge when ready 🚀

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,24 +1,73 @@
...
+ * - name <- added, implicitly keyed by Node
+ * - abi, contenthash, pubkey, version, ...etc if desired
+ * - has many ResolverAddressRecords by `coinType`
+ *   - coinType
+ *   - address
+ * - has many ResolverTextRecords by `key`
+ *   - key
+ *   - value
+ *
+ * This results in the following query patterns (ponder graphql pseudo code) to look up the records
```
lightwalker-eth, line 47: My attention now is focused on the word "records" in this sentence. I think it will help us to introduce more terminology to help us communicate about these ideas with additional clarity and precision:

* "Lite Indexed Records": The resolver records (keys and selected subset of values) indexed by the ENS Subgraph.
* "Extended Indexed Records": The resolver records (keys and selected subset of values) indexed by ENSNode when we have this optional feature flag activated.
* "Effective Protocol Records": The resolver records that would actually be returned for a node through dynamic lookups using the ENS Protocol.

This terminology can help us in a number of ways. For example:
1. Identify how great caution needs to be taken when reading either "Lite Indexed Records" or "Extended Indexed Records" as these don't recognize ENSIP-10 or CCIP-Read and therefore may differ from "Effective Protocol Records".
2. Explicitly give name to the distinction between "Lite Indexed Records" and "Extended Indexed Records"
3. Explicitly give name to the distinction between "Extended Indexed Records" and "Effective Protocol Records"

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,24 +1,73 @@
...
+ * NOTE: generate the (subgraph entity) Resolver ID as:
+ * - (if running in subgraph-compat mode): `${resolverContractAddress.toLowerCase()}-${node}`
+ * - (if not running in subgraph-compat mode): `${chainId}-${resolverContractAddress.toLowerCase()}-${node}`
+ *
+ * @example
+ * ```gql
+ * query RecordsForNode($id: ID!) {
+ *   resolver(id: $id) {
+ *     name
+ *     addressRecords { totalCount items { coinType address } }
```
lightwalker-eth, line 59: Suggest we document some special details about these addressRecords:

1. Different chains (different cointypes) encode addresses in different ways. For example for Bitcoin or Solana or etc.. the addresses will be encoded differently than Ethereum addresses.
2. I'm not sure how the above translates into our Postgres schema or GraphQL API. This might need some consideration. Not sure.
3. Additional details here: https://docs.ens.domains/ensip/9
4. https://github.com/ensdomains/ensjs/blob/main/docs/public/function.getAddressRecord.md
5. https://github.com/ensdomains/ensjs/blob/main/packages/ensjs/src/functions/public/_getAddr.ts

apps/ensindexer/src/lib/ids.ts
```diff
@@ -90,11 +90,12 @@ export const makeRegistrationId = (labelHash: LabelHash, node: Node) => {
 };
 
 /**
- * Makes a unique ID for any resolver record entity.
+ * Makes a unique ID for any resolver record entity that is keyed beyond Node
```
lightwalker-eth, line 93: Suggest to also direct the reader to learn more about the distinction between resolvers in the protocol and resolver entities as indexed by the subgraph in `packages/ensnode-schema/src/subgraph.schema.ts`. Otherwise it's really easy to get confused by this stuff.

apps/ensindexer/src/lib/ids.ts
```diff
@@ -90,11 +90,12 @@ export const makeRegistrationId = (labelHash: LabelHash, node: Node) => {
...
+ * (i.e. text, address records).
  *
  * @param resolverId the id of the resolver
- * @param key any string value
+ * @param key any string value (ex: coinType.toString(), key)
```
lightwalker-eth, line 97: Suggest we refine this. For example "key any string value ..." can cause confusion because both "key" and "value" are words in this sentence.

Would it work to say something like:
`@param key the unique id of the resolver record within resolverId`

Also I don't feel confident that I understand what the following represents:

`(ex: coinType.toString(), key)`

Is this saying the key will be a tuple?

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -12,177 +12,179 @@ import {
...
 } from "@/lib/lib-config";
 import { uniq } from "@/lib/lib-helpers";
-import { PLUGIN_REQUIRED_DATASOURCES } from "@/plugins";
-import { DatasourceName, ENSDeployments, getENSDeployment } from "@ensnode/ens-deployments";
+import { ENSDeployments } from "@ensnode/ens-deployments";
 import { PluginName } from "@ensnode/utils";
-import { Address, isAddress } from "viem";
 
-const parseBlockNumber = (envVarKey: string) =>
+const chainIdSchema = z.number().int().min(1);
```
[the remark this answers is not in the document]
shrugs: i don't think so as it's intentional that rpcConfig allows any chain to be configured, not just the ones that ensnode supports and not just the ones that are explicitly indexed — it only requires that the indexed chains have configuration at the end of the day

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,24 +1,73 @@
...
  *
- * A more accurate datamodel might be to model Resolver as a contract and a ResolverRecords entity
- * to represent the pairwise relationship between a Resolver and a Node for which it holds records.
- * This should be considered in a novel schema for ENSv2.
+ * Some background context: In the ENS protocol, there exist many Resolver contracts, on many chains,
+ * each of which can track many records for any number of Nodes. Only one Resolver contract for a
+ * given Name is 'active' at any one time, and it is _that_ Resolver contract that one must query to
+ * resolve records for the Node in question.
+ *
+ * In the Subgraph schema, however a Resolver _entity_ does NOT represent a Resolver _contract_:
```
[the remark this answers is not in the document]
shrugs: preferring the original structure of this sentence, but with an added comma after however to be grammatically correct

packages/ensnode-schema/src/resolver-records.schema.ts
shrugs: i'm less enthused about this terminology approach and would instead prefer we effectively say "resolving records for a node via database lookups is not ENSIP-10 nor CCIP-Read compliant" or something — basically using the GraphQL API to do any sort of record resolution is a no-no. I see why the subgraph only tracks values because it's hard to police an uneducated consumer from doing this

packages/ensnode-schema/src/resolver-records.schema.ts
shrugs: I think that information can be deferred to ENSIP-9, i'll add another callout there

apps/ensindexer/src/config/config.schema.ts
shrugs: actually, splitting this into a separate issue

shrugs:
updates:
- chose `indexAdditionalResolverRecords` and `INDEX_ADDITIONAL_RESOLVER_RECORDS` as the config names
- because `indexAdditionalResolverRecords` results in superset of subgraph behavior, data and api, i've removed it from the `isSubgraphCompatible` derivation
- updated comments incorporating most feedback, please give them another look over

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
+ * Resolver (the Resolver-Node relationship)
+ *   - has one `addr` record for a Node
+ *   - has one `contenthash` record for a Node
+ *   - has one set of `coinTypes` storing the keys of all ENSIP-9 address records emitted for a Node
+ *   - has one set of `texts` storing the keys of all ENSIP-5 text records emitted for a Node
  *
- * NOTE: generate the (subgraph entity) Resolver ID as:
- * - (if running in subgraph-compat mode): `${resolverContractAddress.toLowerCase()}-${node}`
- * - (if not running in subgraph-compat mode): `${chainId}-${resolverContractAddress.toLowerCase()}-${node}`
+ * In this file, we extend the subgraph's Resolver _entity_ with the following:
```
lightwalker-eth, line 41:
```suggestion
 * In this file, we extend the subgraph's Resolver _entity_ with the following "additional resolver records":
```

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
+ * Resolver (the Resolver-Node relationship)
+ *   + has one `name` record for a Node (see ENSIP-3)
+ *   + has many `addressRecords` for a Node (by coinType) (see ENSIP-9)
+ *     - ResolverAddressRecord indexes `coinType` and `address`
```
lightwalker-eth, line 45: Sorry could you clarify how this should be interpreted? Not confident I follow. Suggest to write this idea out in more words and detail.

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
+ *   + has many `textRecords` for a Node (by key) (see ENSIP-5)
+ *     - ResolverTextRecord indexes `key` and `value`
```
lightwalker-eth, line 47: Sorry could you clarify how this should be interpreted? Not confident I follow. Suggest to write this idea out in more words and detail.

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
  *
+ * This results in the following query patterns for the autogenerated GraphQL APIs:
  * @example
  * ```gql
- * query RecordsForNode($id: ID!) {
+ * # NOTE: generate the Resolver entity ID as:
+ * # - (if running in subgraph-compat mode): `${resolverContractAddress.toLowerCase()}-${node}`
+ * # - (if not running in subgraph-compat mode): `${chainId}-${resolverContractAddress.toLowerCase()}-${node}`
+ * query RecordsForNodeOnResolver($id: ID!) {
```
lightwalker-eth, line 55: How should this example query be interpreted?

1. Is it a query for all subgraph-compat resolver records?
2. Is it a query for all "additional resolver records"?
3. Is it a query for all of both (1) and (2) listed above?
4. Is it a query for (3) listed above less fields that are redundant between (1) and (2)?

Suggest to refine how we describe and name this query to help make it explicit how it should be interpreted. Is this query case 4?

I would be interested to see an example query for all of these cases if possible 🙏 Or perhaps case 1, 2, and 4 assuming 3 provides no benefit over 4.

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
+ *       items {
+ *         key          # see ENSIP-5, ENSIP-12, ENSIP-18
+ *         value        # see ENSIP-5, ENSIP-12, ENSIP-18
+ *       }
+ *     }
  *   }
  * }
  *```
+ *
+ * NOTE: The addition of these record keys and values does NOT allow the caller to confidently
```
lightwalker-eth, line 77:
```suggestion
 * NOTE: These "additional resolver records" do NOT allow the caller to confidently
```

Suggesting this to help maximize explicit relationships between terminology we introduce.

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
+ * resolve records for names: a direct lookup of a Name's Resolver and retrieval of its values
```
lightwalker-eth, line 78:
```suggestion
 * resolve records for names: a direct query for the indexed values of a name's Resolver records
```

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -1,66 +1,82 @@
...
- *   - key
- *   - value
+ * Naturally, this discrepancy may be confusing, but throughout this codebase a Resolver _entity_
+ * refers to that _relationship_ between a Resolver and Node, _not_ the unique Resolver contract itself.
  *
- * This results in the following query patterns (ponder graphql pseudo code) to look up the records
- * for a Node within a Resolver contract:
+ * The Subgraph describes the following datamodel instead:
+ * Resolver (the Resolver-Node relationship)
+ *   - has one `addr` record for a Node
```
lightwalker-eth, line 36: Is there any guaranteed relationship in the subgraph between the `addr` field being defined and the presence of a particular `coinTypes` key? No matter if that's a yes or no, this would be a nice invariant / expectation to document.

packages/ensnode-schema/src/resolver-records.schema.ts
lightwalker-eth, line 45: Is there any guaranteed relationship between the `addr` field being defined and the presence of a particular `addressRecords` key? No matter if that's a yes or no, this would be a nice invariant / expectation to document.

Additionally, I imagine there is a guaranteed relationship between the subgraph-compat `coinTypes` keys and the `coinType` keys defined here? Will also be nice to document this.

apps/ensindexer/.env.local.example
```diff
@@ -87,11 +87,18 @@ ENSADMIN_URL=https://admin.ensnode.io
...
+# It is unsafe to naively assume that indexed resolver record values are equivalent to the resolver
+# record values that would be returned through dynamic lookups via the ENS protocol. For example, if
+# a resolver implements CCIP-Read, the resolver records may not be discoverable through onchain indexing.
+# This feature is under R&D. At this time we do not recommend anyone directly use indexed resolver
+# record values in their applications. Features are planned in the ENSNode roadmap that will provide
+# safe use of indexed resolver record values (in appropriate contexts).
+#
+# NOTE: Setting this to `true` results in the resulting indexed data being a strict _superset_ of
+# subgraph data. For exact subgraph data compatibility, set this to `false`.
 INDEX_RESOLVER_VALUES=true
```
lightwalker-eth, line 104: I believe we need to rename this. Could you please search all files for other places where this new env var should be renamed or introduced? For example, what about our docker configs, etc? Appreciate your advice.

Could you please also share advice for how we should track the need to update environment variable configurations as part of the deployment of this new release? For example, we should explicitly add configurations to this environment variable across all of our ENSNode instances. Perhaps we should add a GitHub issue for this and assign it to Sebastian and/or Tko? I'm worried we will forget about this important follow up action item when we deploy the next release.

apps/ensindexer/.env.local.example
```diff
@@ -87,11 +87,18 @@ ENSADMIN_URL=https://admin.ensnode.io
...
-# the ENS Subgraph. For full data-level backwards compatibility with the ENS Subgraph, set this to `false`.
+# WARNING: Special care must be taken when interacting with indexed resolver record values.
+# It is unsafe to naively assume that indexed resolver record values are equivalent to the resolver
+# record values that would be returned through dynamic lookups via the ENS protocol. For example, if
+# a resolver implements CCIP-Read, the resolver records may not be discoverable through onchain indexing.
+# This feature is under R&D. At this time we do not recommend anyone directly use indexed resolver
+# record values in their applications. Features are planned in the ENSNode roadmap that will provide
+# safe use of indexed resolver record values (in appropriate contexts).
+#
+# NOTE: Setting this to `true` results in the resulting indexed data being a strict _superset_ of
```
lightwalker-eth, line 102:
```suggestion
# NOTE: Setting this to `true` results in indexed data being a strict _superset_ of
```

lightwalker-eth (changes requested):
@shrugs Great updates 😄 Reviewed and shared a few suggestions :+1:

packages/ensnode-schema/src/resolver-records.schema.ts
shrugs: I'm just going to remove this example, since we don't want people querying it anyway

terraform/modules/ensindexer/main.tf
```diff
@@ -99,6 +99,10 @@ resource "railway_variable_collection" "ensindexer" {
       name  = "HEAL_REVERSE_ADDRESSES"
       value = var.heal_reverse_addresses
     },
+    {
+      name  = "INDEX_ADDITION_RESOLVER_RECORDS"
```
lightwalker-eth, line 103:
```suggestion
      name  = "INDEX_ADDITIONAL_RESOLVER_RECORDS"
```

terraform/modules/ensindexer/main.tf
```diff
@@ -180,6 +184,10 @@ resource "railway_variable_collection" "ensindexer_api" {
...
```
lightwalker-eth, line 188:
```suggestion
      name  = "INDEX_ADDITIONAL_RESOLVER_RECORDS"
```

apps/ensindexer/.env.local.example
```diff
@@ -87,18 +87,16 @@ ENSADMIN_URL=https://admin.ensnode.io
...
 #
-# WARNING: Special care must be taken when interacting with indexed resolver record values.
-# It is unsafe to naively assume that indexed resolver record values are equivalent to the resolver
-# record values that would be returned through dynamic lookups via the ENS protocol. For example, if
-# a resolver implements CCIP-Read, the resolver records may not be discoverable through onchain indexing.
-# This feature is under R&D. At this time we do not recommend anyone directly use indexed resolver
-# record values in their applications. Features are planned in the ENSNode roadmap that will provide
-# safe use of indexed resolver record values (in appropriate contexts).
+# WARNING: It is unsafe to naively assume that indexed resolver record values are equivalent to the
+# result of performing dynamic forward resolution via the ENS protocol, as retrieving records values
```
lightwalker-eth, line 95:
```suggestion
# result of performing dynamic forward resolution via the ENS protocol, as naively retrieving resolver records
```

apps/ensindexer/.env.local.example
```diff
@@ -87,18 +87,16 @@ ENSADMIN_URL=https://admin.ensnode.io
...
+# from an index is not ENSIP-10 nor CCIP-Read compliant. At this time we do _not_ recommend anyone
```
lightwalker-eth, line 96:
```suggestion
# from indexed may not be ENSIP-10 nor CCIP-Read compliant. At this time we do _not_ recommend anyone
```

packages/ensnode-schema/src/resolver-records.schema.ts
```diff
@@ -36,47 +36,23 @@
...
  *
- * In this file, we extend the subgraph's Resolver _entity_ with the following:
+ * In this file, we extend the subgraph's Resolver _entity_ with the following _additional_ records:
  * Resolver (the Resolver-Node relationship)
  *   + has one `name` record for a Node (see ENSIP-3)
+ *     - see 'NOTE(resolver-records)' in subgraph.schema.ts
  *   + has many `addressRecords` for a Node (by coinType) (see ENSIP-9)
- *     - ResolverAddressRecord indexes `coinType` and `address`
+ *     - see ResolverAddressRecord below
+ *     - NOTE: _does_ represent the implcit coinType of 60 for the `addr` record, if set
```
lightwalker-eth, line 48:
```suggestion
 *     - NOTE: _does_ represent the implicit coinType of 60 for the `addr` record, if set
```

lightwalker-eth (approved):
@shrugs Great work on all of this! 🚀 Major new capability 💪

packages/ensnode-schema/src/resolver-records.schema.ts
lightwalker-eth: Is there a simple way for us to separate the idea of indexing this data from it being exposed in our GraphQL APIs? If so that would be great. I like the idea of not making this data available for direct query through our GraphQL APIs at all if possible. No worries if this is complex -- we can come back to it later too.

lightwalker-eth (approved):
Sharing a few more suggestions :+1:

shrugs:
i was able to trivially filter out the ext_ prefixed tables/relations from the schema before passing to the generated apis so that seems to be the simplest approach here for sure

Issue #548, opened 2025-04-09
Refine `Domain` type used for "Recent Registrations" component


lightwalker-eth:
![Image](https://github.com/user-attachments/assets/16d5a144-d71c-43d5-bef4-fecc47a04cf6)

Above is a screenshot taken while reviewing PR 532.

It looks like the above definition of `Domain` is related to the schema returned by GraphQL API queries. The scope of this issue is not to change the GraphQL API. So this type definition will still be needed, but suggest renaming this interface from `Domain` to something like `DomainResult` (based on it being associated with a result returned from the GraphQL API for the latest registrations.

Further, this query is not about domains specifically, it's about registrations. Therefore suggest to further rename it from `DomainResult` to `LatestRegistrationResult`.

Then, suggest the following additional refinements:
1. Define a `LatestRegistration` interface that is "pure" based on a data model that actually makes sense to build logic on top of and is detached from silly details of the GraphQL API response data model from `LatestRegistrationResult`.
   1. There shouldn't be nested objects for owner / wrappedOwner addresses. The data model should be flattened. There should be a field for `owner` of type `Address` and likewise an optional field for `wrappedOwner` of type `Address`.
   2. createdAt and expiryDate: Do we really want these to be strings? If so, we should document exactly what date format those strings are. Or do we want to change them to number and document how they are actually unix timestamps or whatever? Advice appreciated for whatever will be simple and make sense.
3. Define a function that transforms a `LatestRegistrationResult` to a `LatestRegistration`.
4. The logic that queries latest registrations should then return only `LatestRegistration` and never `LatestRegistrationResult`.
5. All the UI component logic or other "business" logic should be refactored to operate on `LatestRegistration` and not `LatestRegistrationResult`.
   1. This is expected to help make related logic more clean and nice as a result.

PR #748, opened 2025-05-26
chore: ENSAdmin - fix data model for `Latest Indexed Registrations` panel

Y3drk:
Solves Issue #548 .


Y3drk:
Overall, I have a feeling that this refinement is quite redundant, and could be replaced with just some additional documentation. 

Maybe I didn't fully lean into the requirements described in the Issue, but I don't know what else to add there.

Awaiting feedback :)

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,10 +15,35 @@ export interface Domain {
...
+/**
+  A "pure" version of LatestRegistrationResult that streamlines data usage in RecentRegistrations component.
+ @params:
+ * name - string - recently registered name
+ * createdAt - number - a UNIX timestamp in seconds of when the name was originally created
+ * expiryDate - number - a UNIX timestamp in seconds of when will the current owner's tenure end
+ * owner - Address -  owner's address
+ * wrappedOwner - Address - optional - wrapped owner's address
+ **/
+export interface LatestRegistration {
```
lightwalker-eth, line 27:
```suggestion
export interface Registration {
```

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,10 +15,35 @@ export interface Domain {
   };
 }
 
+/**
+  A "pure" version of LatestRegistrationResult that streamlines data usage in RecentRegistrations component.
+ @params:
```
lightwalker-eth, line 20: We shouldn't document params here like this.

Instead, the best practice is to document each field directly where each field is defined.

```
/**
 * The registered ENS name
 */
name: string;
```

Make sure you always document these with /** style comments and not // style comments. Otherwise IDEs don't read it properly.

(edit: GitHub is messing up the alignment of th

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -46,6 +51,17 @@ async function fetchRecentRegistrations(baseUrl: URL): Promise<RecentRegistratio
...
 
   const data = await response.json();
+  data.data.registrations = data.data.registrations.map(
+    (latestRegistration: {
+      registrationDate: string;
+      expiryDate: string;
+      domain: LatestRegistrationResult;
+    }) => ({
+      ...latestRegistration,
+      domain: toLatestRegistration(latestRegistration.domain),
```
lightwalker-eth, line 61: A domain is not a registration. We're getting the ideas all mixed up here.

lightwalker-eth (changes requested):
@Y3drk Thanks for updates. Reviewed and shared feedback.

apps/ensadmin/src/components/recent-registrations/types.ts
Y3drk: @lightwalker-eth 

This change conflicts with another type used in that panel, that's named exactly `Registration` in **\recent\registrations\types.ts**...Maybe we could unify them?

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+   */
+  name: string;
+
+  /**
+   * a UNIX timestamp in seconds of when the name was originally created
+   */
+  createdAt: number;
+
+  /**
+   * a UNIX timestamp in seconds of when will the current owner's tenure end
```
lightwalker-eth, line 33:
```suggestion
   * a UNIX timestamp in seconds when the registration is scheduled to expire
```

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+   */
+  expiryDate: number;
+
+  /**
+   * owner's address
+   */
+  owner: Address;
```
lightwalker-eth, line 40: Suggest to rename this to `ownerInRegistry`.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+
+  /**
+   * wrapped owner's address
```
lightwalker-eth, line 43:
```suggestion
   * The owner according to the ENS NameWrapper. If undefined, the domain associated with the registration is unwrapped (not in the ENS NameWrapper).
```

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+   */
+  createdAt: number;
+
+  /**
+   * a UNIX timestamp in seconds of when will the current owner's tenure end
+   */
+  expiryDate: number;
+
+  /**
+   * owner's address
```
lightwalker-eth, line 38:
```suggestion
   * The "true" owner of the domain in the ENS Registry.
```

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
   };
 }
 
+/**
+ Data associated with a Registration event.
+ **/
+export interface LatestRegistration {
+  /**
+   * recently registered ENS name
```
lightwalker-eth, line 23:
```suggestion
   * The registered ENS name
```

This data model should not be specific to a "recent registration". It should be general to any "registration"

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -1,9 +1,9 @@
 import type { Address } from "viem";
 
 /**
- * Types for the recent registrations component
+ * The data model returned by a GraphQL query for the latest registrations.
  */
-export interface Domain {
+export interface LatestRegistrationResult {
```
lightwalker-eth, line 6: This definition should not be here. It should be in `apps/ensadmin/src/components/recent-registrations/hooks.ts` and it should NOT be exported from that file. It should be an internal implementation detail of the hook and not leaked outside of that.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+  /**
+   * a UNIX timestamp in seconds of when will the current owner's tenure end
+   */
   expiryDate: string;
-  domain: Domain;
+
+  /**
+   * Data associated with a Registration event.
+   */
+  registration: LatestRegistration;
```
lightwalker-eth, line 65: This field should not exist. There shouldn't be both a `LatestRegistration` and a `Registration` type. These ideas should be merged together into a single type: `Registration`.

apps/ensadmin/src/components/recent-registrations/types.ts
lightwalker-eth: These should be unified into a single type.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+
+  /**
+   * a UNIX timestamp in seconds of when the name was originally created
+   */
+  createdAt: number;
+
+  /**
+   * a UNIX timestamp in seconds of when will the current owner's tenure end
+   */
+  expiryDate: number;
```
lightwalker-eth, line 35:
```suggestion
  expiresAt: number;
```

It's confusing to call this a "date" as that could be interpreted as just a date ex: "Jan 1, 2025" rather than a full timestamp.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+}
+
+/**
+ * Extended data associated with Registration event
+ */
 export interface Registration {
+  /**
+   * a UNIX timestamp in seconds of when the name was registered by its current owner
+   */
   registrationDate: string;
```
lightwalker-eth, line 55: We shouldn't be mixing types used for timestamps. We should use the same type everywhere for timestamps. Please suggest which we should use.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+
+  /**
+   * a UNIX timestamp in seconds of when will the current owner's tenure end
+   */
   expiryDate: string;
```
lightwalker-eth, line 60: We shouldn't be mixing types used for timestamps. We should use the same type everywhere for timestamps. Please suggest which we should use.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
-  domain: Domain;
+
+  /**
+   * Data associated with a Registration event.
+   */
+  registration: LatestRegistration;
 }
 
+/**
+ * Data about 5 most recent registrations
```
lightwalker-eth, line 69: A few issues here:

1. This shouldn't make any reference to the number 5. The number of records contained should be any number.
2. This type should not live in this file. It should live inside the hooks.ts file and it should not be exported out of that file. It is an internal implementation detail.
3. The `registrations` field here should NOT be a `Registration`. It should be a `RegistrationResponse`. You're mixing up ideas. Please 100% separate the internal implementation details of GraphQL response schemas (which are ugly and shit due to a number of factors) from the 100% clean and perfect data model we want to use across ENSAdmin for Registrations. The ugly implementation details of GraphQL response schemas must be TRAPPED inside a single layer and not allowed to leak out anywhere else across our code. Only the 100% perfect data model of a `Registration` should spread across our code.

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -15,12 +15,59 @@ export interface Domain {
...
+
+  /**
+   * owner's address
+   */
+  owner: Address;
+
+  /**
+   * wrapped owner's address
+   */
+  wrappedOwner?: Address;
```
lightwalker-eth, line 45: Suggest to rename this to `ownerInNameWrapper`.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -65,3 +81,20 @@ export function useRecentRegistrations(searchParams: URLSearchParams) {
     },
   });
 }
+
+/**
+ * Function that transforms a LatestRegistrationResult to a LatestRegistration
+ * @param graphQLResponseData - LatestRegistrationResult - GraphQL compatible representation of recently registered domain, dubed 'domain' in the query
+ * @returns fetched data in new, streamlined LatestRegistration format
+ */
+export function toLatestRegistration(
```
lightwalker-eth, line 90: This function should not be exported. It should be an internal implementation detail.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,6 +1,11 @@
...
-import { RecentRegistrationsResponse } from "./types";
+import {
+  LatestRegistration,
+  LatestRegistrationResult,
+  RecentRegistrationsResponse,
+  Registration,
+} from "./types";
 
 /**
  * Fetches info about the 5 most recently registered .eth domains that have been indexed.
```
lightwalker-eth, line 11: This function should not return a `RecentRegistrationsResponse` this is leaking ugly GraphQL shit all over our code. It needs to return `Promise<Registration[]>`

lightwalker-eth, line 11: It's not a good design to make specific reference to a number such a "5" here. The number of registrations to query should be a param passed into the function.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -46,6 +51,17 @@ async function fetchRecentRegistrations(baseUrl: URL): Promise<RecentRegistratio
   }
 
   const data = await response.json();
+  data.data.registrations = data.data.registrations.map(
```
lightwalker-eth, line 54: We should be completely transforming a `RegistrationResult` to a `Registration`.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -135,18 +136,19 @@ const NAME_WRAPPER_ADDRESS = "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401";
  * Determines the true owner of a domain.
  * If the owner is the NameWrapper contract, returns the wrapped owner instead.
  *
- * @param owner The owner address
- * @param wrappedOwner The wrapped owner address (optional)
- * @returns The true owner address
+ * @param registration - Extended data associated with Registration event
  */
-function getTrueOwner(owner: { id: Address }, wrappedOwner?: { id: Address }) {
+function getTrueOwner(registration: Registration) {
```
lightwalker-eth, line 141: @Y3drk This function shouldn't live in this file.

1. The logic here should move into the hook.
2. This function should not be exported. It is an internal implementation detail.
3. The `Registration` type should have another field added: `owner` (please see my other comments suggesting renaming of the existing `owner` field to `ownerInRegistry`. This NEW `owner` field serves the purpose of identifying the effective owner in a simple way.
4. Whenever a `Registration` is built by the DAL, the DAL should have the responsibility of properly setting this NEW `owner` field by calling this internal implementation detail helper function.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -235,10 +237,7 @@ export function RecentRegistrations() {
                     <TableCell>
                       {indexedChainId ? (
                         <ENSName
-                          address={getTrueOwner(
-                            registration.domain.owner,
-                            registration.domain.wrappedOwner,
-                          )}
+                          address={getTrueOwner(registration)}
```
lightwalker-eth, line 240:
```suggestion
                          address={registration.owner}
```

Please see my related comment above for more details. When we get all this code right, this suggestion shows how the final result should look here.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -135,18 +136,19 @@ const NAME_WRAPPER_ADDRESS = "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401";
...
+ * @param registration - Extended data associated with Registration event
  */
-function getTrueOwner(owner: { id: Address }, wrappedOwner?: { id: Address }) {
+function getTrueOwner(registration: Registration) {
   // Only use wrapped owner if the owner is the NameWrapper contract
-  if (wrappedOwner?.id && isAddressEqual(owner.id, NAME_WRAPPER_ADDRESS)) {
-    return getAddress(wrappedOwner.id);
+  if (
+    registration.registration.wrappedOwner &&
+    isAddressEqual(registration.registration.owner, NAME_WRAPPER_ADDRESS)
```
lightwalker-eth, line 145: FYI: Here what we're really doing is we're checking if the `ownerInRegistry` is equal to the `NAME_WRAPPER_ADDRESS`. If so, then `ownerInNameWrapper` MUST be defined (otherwise we should throw an exception as our expectations are violated) and the `ownerInNameWrapper` should be considered the effective owner.

Y3drk:
# Please read before review

I've implemented the vast majority of @lightwalker-eth 's suggestions, but some things weren't 100% clear to me, so I made some assumptions, which I will describe in detail later. Also, I have several questions regarding those points of uncertainty that I will also write below.

## Assumptions
1. Latest feedback included a suggestion to unify the type that we associate with various `timestamps` in the `Registration` type. I've chosen to go with `string` type, since this is the form these fields have when returned by a GraphQL query, and all of the helper functions that were previously written expect the input timestamp to be in string format.
2. One of the suggestions was to unify `LatestRegistration` and `Registration` types and to do that I made several assumptions:
    a. I injected the `LatestRegistration` type into `Registration` and **flattened** it, assuming that we would want to avoid nested objects.
    b. The aforementioned unification resulted in two separate fields having the same name: `expiresAt`. After some investigation I ruled out reduction to a single field since the values were not identical and decided to rename one of the fields. To my knowledge, when returned as a GraphQL query result, the `expiryDate` field that is returned inside the "domain" object is the value of the "outer" `expiryDate` object plus the [**grace period**](https://docs.ens.domains/terminology#grace-period) set by ENS to 90 days, hence I renamed this field to the `expiresAtWithGracePeriod`.
    c. To avoid confusion about the meaning of `createdAt` field, I renamed it to `domainCreatedAt` so that it's clear to what item it refers to
    d. As suggested I renamed previous owner fields and added a new one: `owner`, but I wasn't certain about its description.
    
Here is how the unified `Registration` type looks:
```
export interface Registration {

  registeredAt: string;

  expiresAt: string;

  name: string;

  domainCreatedAt: string;

  expiresAtWithGracePeriod: string;

  ownerInRegistry: Address;

  ownerInNameWrapper?: Address;

  owner: Address;
}
```
Fields' descriptions were excluded and can be seen in the PR's files.

3. I didn't implement the suggestion about making a hardcoded limit of recent registrations (currently set to 5) a function input, but I plan on doing it.

## Questions
1. Is the proposed type fine with you?
2. Questions about new `Registration` type:
    a. Is the proposed, unified type acceptable, should we go back to its nested version? 
    > Personally I prefer it the way it is now
    
    b. The response from GQL has one more field `labelName` that we don't use. Should we add it to the `Registration` type?
    c. Do we want to remove any of the current fields from this type?
    > As of yet we never use `domainCreatedAt` and `expiresAtWithGracePeriod` fields, but I suppose that, they might come in handy in the future
    

3. Do we want the users to be able to set the limit of registrations themselves or we just want it as a function parameter for the sake of clean-code?

apps/ensadmin/src/components/recent-registrations/types.ts
```diff
@@ -1,26 +1,48 @@
...
 
 /**
- * Types for the recent registrations component
+ * Data associated with Registration event
  */
-export interface Domain {
+// TODO: fix the order of the fields for improved clarity
+export interface Registration {
+  /**
+   * a UNIX timestamp in seconds of when the domain was registered by its current owner
```
lightwalker-eth, line 9: When we write comments about a field, we should read it back to ourselves as if we were a CPU.

This is not actually a UNIX timestamp, it is a string formatted as an integer that represents a UNIX timestamp. Please take special note of how we should document our ideas for accuracy.

Suggestion is for this to just be a `Date` however, so the feedback above is just general advice for the future.

lightwalker-eth (changes requested):
@Y3drk Thanks for updates. Reviewed and shared feedback :+1:

Y3drk:
# Before review (06/17/25)

## New changes
* I've applied all @lightwalker-eth's suggestions from all previous PRs, including:
    * Refactoring and moving helper functions to `recent-recgistrations/utils.tsx` file --> these functions could be moved to `/components/ui` and further refactored if deemed necessary. Need to decide on their final location
    * Creating a new parsing function `transformTimestamp` in `recent-recgistrations/hooks.ts` that allowed to move the validation of the data completely into the `hooks.ts` file, thus making the validation of the GraphQL response an internal app process
    * Investigation about the inclusion of a grace period in one of the `expiresAt` fields (see comment in the review above)
    * Changing & unifying all timestamp-related fields in the `Registration` type to `Date` objects.
    * Renaming some internal types and improvements of docstrings in general
* Fixed edge case for very short durations of registrations, where the component previously displayed `0 months`. Now for these cases, the returned value is `less than a month`. If preferred, we can make it more detailed.

## Note about uncertainties & questions
In cases when I wasn't sure what to do, I left a comment right under the review suggestion. Would appreciate having a look at them first if the changes aren't satisfactory. 

For now, I don't think I have any additional questions, besides those mentioned above.

Awaiting feedback and thanks in advance :)

apps/ensadmin/src/components/identity/index.tsx
```diff
@@ -52,7 +52,7 @@ export function ENSName({
```
lightwalker-eth, line 52: GitHub won't let me put this comment on the correct line. This comment is actually for several lines above:

```
const ensAppUrl = `https://app.ens.domains/${address}`;
```

Please see my other comment about how this logic is not correct as it needs to consider the ENS deployment chain of the connected ENSNode.

apps/ensadmin/src/components/identity/index.tsx
```diff
@@ -52,7 +52,7 @@ export function ENSName({
 
   // If not mounted yet (server-side), show a skeleton
   if (!mounted) {
-    return <EnsNamePlaceholder showAvatar={showAvatar} className={className} />;
+    return <IdentityPlaceholder showAvatar={showAvatar} className={className} />;
   }
 
   return (
```
lightwalker-eth, line 58: GitHub won't let me put this comment on the correct line. This comment is actually for several lines below:

```
            <AvatarImage
              src={`https://metadata.ens.domains/mainnet/avatar/${ensName}`}
              alt={ensName}
            />
```

This logic is not correct:
1. It should not always use `mainnet` in the image path. Depending on the ENS deployment chain it should use holesky or sepolia.
2. The logic needs to consider the case that the ENS deployment chain is the `ens-test-env`. In this case there's no ENS Metadata Service we can talk to and therefore we should always just display the AvatarFallback.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,95 @@
...
+ */
+const NAME_WRAPPER_ADDRESS = "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401";
+
+/**
+ * Determines the effective owner of a domain.
+ * If the owner is the NameWrapper contract, returns the wrapped owner instead.
+ *
+ * @param registrationResult - Data associated with Registration event
+ */
+function getTrueOwner(registrationResult: RegistrationResult) {
```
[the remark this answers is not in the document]
Y3drk: Applied, but left the comment open for future reference

apps/ensadmin/src/components/recent-registrations/utils.tsx
```diff
@@ -0,0 +1,75 @@
...
+    setRelativeTime(formatDistanceToNow(date, { addSuffix: true }));
+  }, [date]);
+
+  return <>{relativeTime}</>;
+}
+
+/**
+ * Helper function to calculate duration length in years or months
+ */
+const formatDuration = (beginsAt: Date, endsAt: Date) => {
```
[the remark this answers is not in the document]
Y3drk: @lightwalker-eth 

Noted. From now on, the `formatDuration` function will be fully replaced with the `formatDistanceStrict` from `date-fns`. 

About leveraging functions options &rarr; As far as I am aware, the default version of this function is perfect for us since it doesn't add any suffixes or force any specific unit, hence it should stick to displaying the duration in the largest unit possible (for our use-case usually years, sometimes months) just like we did until now.

apps/ensadmin/src/components/recent-registrations/utils.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const years = differenceInYears(endsAt, beginsAt);
+
+  // If less than a year, show months instead
+  if (years === 0) {
```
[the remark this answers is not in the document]
Y3drk: I'll keep that in mind, but this particular case was solved by replacing our custom function with the `formatDistanceStrict` from `date-fns` library

apps/ensadmin/src/components/recent-registrations/utils.tsx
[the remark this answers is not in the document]
Y3drk: This suggestion is resolved but kept open for future reference

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+      id: Address;
+    };
+    wrappedOwner?: {
+      id: Address;
+    };
+  };
+}
+
+/**
+ * Numeric string representing a Unix timestamp in seconds.
```
lightwalker-eth, line 27:
```suggestion
 * An integer value (representing a Unix timestamp in seconds) formatted as a string.
```

This states the idea more clearly. It's not so clear what a "Numeric string" is.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+    return getAddress(registrationResult.domain.wrappedOwner.id);
+  }
+
+  // Otherwise, use the regular owner
+  return getAddress(registrationResult.domain.owner.id);
+}
+
+function transformTimestamp(timestamp: UnixTimestampInSeconds): Date {
+  try {
+    return new Date(parseInt(timestamp) * millisecondsInSecond);
```
lightwalker-eth, line 57: Please write unit tests for `transformTimestamp`. I believe you'll see that the logic on this line won't throw an exception on invalid `UnixTimestampInSeconds` values as you expect.

Please see how we've defined other unit tests for examples.

As I understand, this is the correct pattern in Javascript:

```
const d = new Date("foo"); // NOTE: this won't throw an exception!
if (Number.isNaN(d.getTime())) {
  // invalid date -- here's where you actually throw an exception.
}
```

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+      );
+    }
+    return getAddress(registrationResult.domain.wrappedOwner.id);
+  }
+
+  // Otherwise, use the regular owner
+  return getAddress(registrationResult.domain.owner.id);
+}
+
+function transformTimestamp(timestamp: UnixTimestampInSeconds): Date {
```
lightwalker-eth, line 55:
```suggestion
function unixTimestampToDate(timestamp: UnixTimestampInSeconds): Date {
```

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+ * Transforms a RegistrationResult into a Registration
  */
-async function fetchRecentRegistrations(baseUrl: URL): Promise<RecentRegistrationsResponse> {
+function toRegistration(registrationResult: RegistrationResult): Registration {
+  return {
+    registeredAt: transformTimestamp(registrationResult.registrationDate),
+    expiresAt: transformTimestamp(registrationResult.expiryDate),
+    name: registrationResult.domain.name,
+    ownerInRegistry: registrationResult.domain.owner.id,
+    owner: getEffectiveOwner(registrationResult),
```
lightwalker-eth, line 72: When building an instance of object X, it's nice to list the fields in the same order they are defined for object X.

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+    ownerInNameWrapper: registrationResult.domain.wrappedOwner?.id,
+  };
+}
+
+/**
+ * Fetches info about most recent registrations that have been indexed.
```
lightwalker-eth, line 78:
```suggestion
 * Fetches info about the most recent registrations that have been indexed.
```

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -46,22 +115,24 @@ async function fetchRecentRegistrations(baseUrl: URL): Promise<RecentRegistratio
...
-
+export function useRecentRegistrations(ensNodeURL: URL, maxResults: number) {
   return useQuery({
-    queryKey: ["recent-registrations", ensNodeUrl],
-    queryFn: () => fetchRecentRegistrations(ensNodeUrl),
+    queryKey: ["recent-registrations", ensNodeURL],
+    queryFn: () => fetchRecentRegistrations(ensNodeURL, maxResults),
     throwOnError(error) {
-      throw new Error(`Could not fetch ENSNode data from '${ensNodeUrl}'. Cause: ${error.message}`);
+      throw new Error(`Could not fetch ENSNode data from '${ensNodeURL}'. Cause: ${error.message}`);
```
lightwalker-eth, line 135:
```suggestion
      throw new Error(`Could not fetch recent registrations from '${ensNodeURL}'. Cause: ${error.message}`);
```

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -46,22 +115,24 @@ async function fetchRecentRegistrations(baseUrl: URL): Promise<RecentRegistratio
...
 }
 
 /**
- * Hook to fetch info about the 5 most recently registered .eth domains that have been indexed.
- * @param searchParams The URL search params including the selected ENS node URL.
- * @returns React Query hook result.
+ * Hook to fetch info about most recently registered domains that have been indexed.
+ *
+ * @param ensNodeURL The URL of the selected ENS node instance.
+ * @param maxResults number of the maximal number of latest registrations to be retrieved by the query
```
lightwalker-eth, line 128:
```suggestion
 * @param maxResults the max number of recent registrations to retrieve
```

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+ */
+const NAME_WRAPPER_ADDRESS = "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401";
+
+/**
+ * Determines the effective owner of a domain.
+ * If the owner is the NameWrapper contract, returns the wrapped owner instead.
+ */
+function getEffectiveOwner(registrationResult: RegistrationResult): Address {
+  // Only use wrapped owner if the owner is the NameWrapper contract
+  if (isAddressEqual(registrationResult.domain.owner.id, NAME_WRAPPER_ADDRESS)) {
```
lightwalker-eth, line 42: There's unnecessary nesting of `if` statements here. Please see this article that talks about the best practice: https://www.linkedin.com/pulse/avoiding-nested-statements-guard-clauses-danny-olsen-a1e8f/

apps/ensadmin/src/components/recent-registrations/hooks.ts
```diff
@@ -1,17 +1,86 @@
...
+/**
+ * Transforms a RegistrationResult into a Registration
  */
-async function fetchRecentRegistrations(baseUrl: URL): Promise<RecentRegistrationsResponse> {
+function toRegistration(registrationResult: RegistrationResult): Registration {
+  return {
+    registeredAt: transformTimestamp(registrationResult.registrationDate),
+    expiresAt: transformTimestamp(registrationResult.expiryDate),
+    name: registrationResult.domain.name,
+    ownerInRegistry: registrationResult.domain.owner.id,
```
lightwalker-eth, line 71: Please note: You haven't validated that the following fields are valid address values using the `getAddress` function:
- `registrationResult.domain.owner.id`.
- `registrationResult.domain.wrappedOwner?.id`. -- take care to ensure getAddress is only called here if this address value is defined.

lightwalker-eth (changes requested):
@Y3drk Thanks for updates :+1: Identified some other issues for us to fix here. Thanks

apps/ensadmin/src/components/recent-registrations/hooks.ts
Y3drk: # Update (from the sync call) 06/23/25

The function will be moved to the `utils.tsx` file and exported. The related type alias will probably have to be exported as well

Then its test will simply be moved to a separate `utils.test.ts` file as we would usually do with vitest tests

apps/ensadmin/src/components/recent-registrations/utils.tsx
```diff
@@ -51,3 +52,21 @@ export function Duration({
...
   return <>{duration}</>;
 }
+
+/**
+ * An integer value (representing a Unix timestamp in seconds) formatted as a string.
+ */
+export type UnixTimestampInSeconds = string;
+
+/**
+ * Transforms a timestamp formatted as a string to a Date object.
```
lightwalker-eth, line 62:
```suggestion
 * Transforms a UnixTimestampInSeconds to a Date.
```

Best to build upon definitions you've already invested in.

lightwalker-eth (changes requested):
@Y3drk Overall code review looks good now, just 1 very trivial feedback, but we must have missed something. Please check the Vercel preview environment:

CleanShot 2025-06-23 at 18 49 15

PR #749, opened 2025-05-26
fix(ensindexer): use lazy-loaded config

tk-o:
This change prevents premature execution of plugin-specific config factories.

Before this change, running ENSIndexer service with `ENS_DEPLOYMENT_CHAIN` env var set to other valid value than `mainnet` would cause application runtime error. The problem was caused due to all plugin config files being unconditionally executed from `ponder.config.ts` file. For example, the `basenames` plugin would not find a required datasource loaded when `ENS_DEPLOYMENT_CHAIN` was set to either `sepolia`, `holesky`, or `ens-test-env`.

```
ensindexer  | 9:08:53 AM ERROR build      Error while executing 'ponder.config.ts':
ensindexer  | TypeError: Cannot destructure property 'chain' of 'deployment[__vite_ssr_import_3__.DatasourceName.Basenames]' as it is undefined.
ensindexer  |     at /app/apps/ensindexer/src/plugins/basenames/basenames.plugin.ts:24:9
ensindexer  |   22 | // extract the chain and contract configs for Basenames Datasource in order to build ponder config
ensindexer  |   23 | const deployment = getENSDeployment(appConfig.ensDeploymentChain);
ensindexer  | > 24 | const { chain, contracts } = deployment[DatasourceName.Basenames];
ensindexer  |      |         ^
ensindexer  |   25 |
ensindexer  |   26 | export const config = createConfig({
ensindexer  |   27 |   networks: networksConfigForChain(chain.id),
ensindexer  | 9:08:53 AM WARN  process    Failed intial build, starting shutdown sequence
Cleaning up...
 ELIFECYCLE  Command failed with exit code 1.
```

The fix was to make the plugin config object construction logic lazy-evaluated and only called during plugin activation phase (by when all required datasources must be available).

### DIFF
--- a/apps/ensindexer/ponder.config.ts
+++ b/apps/ensindexer/ponder.config.ts
@@ -21,7 +21,7 @@ export const ALL_PLUGINS = [
   threednsPlugin,
 ] as const;
 
-export type MergedPonderConfig = MergedTypes<(typeof ALL_PLUGINS)[number]["config"]> & {
+export type MergedPonderConfig = MergedTypes<(typeof ALL_PLUGINS)[number]["config"]["loader"]> & {
   /**
    * NOTE: we inject additional values (ones that change the behavior of the indexing logic) into the
    * Ponder config in order to alter the ponder-generated build id when these additional options change.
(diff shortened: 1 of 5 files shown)


shrugs:
perhaps we can restructure to something more like this:

```ts
export default {
  get config () {
    // ...
    return createConfig()
  }
  activate: activateHandlers({ ... });
}
```

so we're exporting a single plugin-shaped object from each plugin file

---

the solution i had in my l2 primary name branch was similar to the previous behavior where we merged the configs

```ts
export const getENSDeployment = (ensDeploymentChain: keyof typeof ENSDeployments) =>
  ({
    ...ENSDeployments.mainnet,
    ...ENSDeployments[ensDeploymentChain],
  }) as ENSDeploymentGlobalType;
```

but i think i prefer this deferred execution approach rather than the hacky merged deployment thing

PR #753, opened 2025-05-28
docs(ensindexer): add extra context around reverse address healing strategies

tk-o:
This PR includes additional context behind decisions we made to come up for three distinct label healing strategies for reverse addresses.

Also, this PR replaces logging a console warning with trowing a runtime error to make it very clear that our 100% goal of reverse addresses being healed successfully was not reached and requires attention.

Related to #750

### DIFF
--- a/apps/ensindexer/src/handlers/Registry.ts
+++ b/apps/ensindexer/src/handlers/Registry.ts
@@ -117,7 +117,7 @@ export const makeRegistryHandlers = ({
             //    Contract call:
             //    https://etherscan.io/address/0x084b1c3c81545d370f3634392de611caabff8148#code#L106
             //
-            //    For these transaction, the transaction sender address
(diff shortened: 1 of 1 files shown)


apps/ensindexer/src/handlers/Registry.ts
```diff
@@ -93,40 +93,89 @@ export const makeRegistryHandlers = ({
...
               healedLabel = maybeHealLabelByReverseAddress({
                 maybeReverseAddress: event.args.owner,
                 labelHash,
               });
+
+              if (healedLabel) {
+                console.log(
+                  `Found a healed label: ${healedLabel} with event owner: ${event.args.owner} of tx ${event.transaction.hash}`,
+                );
+              }
```
tk-o, line 152:
```suggestion
```

PR #773, opened 2025-06-06
R&D: ENSAdmin - initial work on Example Queries panel

Y3drk:
Very much a WIP. Just trying things out.

---

## State for **06/06**:
* The UI styling is copied from the `Ponder Client` panel, but, of course, it doesn't have to be that way 
* The documentation links are inserted above the examples (not below as shown in the example in the relevant Slack thread) due to the lack of space. I felt that, if I put them anywhere else, the panel would feel super cramped and messy. I'm eager to discuss that and the whole design of this subpage in general, as said above.
* The main question is how to translate from browsing examples to executing them in the GraphiQL Editor, so far my ideas are:
    * Open a new editor either as a modal or in a subpage (**current version!!!**)
    * Navigate to `GraphQL (Subgraph-style)` panel
    
In both cases proposed in the 3rd point, the real challenge is how to inject the chosen query (and variables if needed) into the Editor. So far, they are passed as query params of the URL. I remember that this was the initial idea for the AI query generator, but it was later implemented differently. That makes me unsure about this approach, but I haven't come up with any alternative yet.
    
I am very open to all suggestions and will work on implementing them (and/or polishing my own idea) on Monday


notrab:
@Y3drk and I discussed moving towards an Examples Directory. My proposal is that this lives on the ensnode.io website outside of ENSAdmin, then links can be "Open in ENSAdmin" instead of "Open in GraphiQL".

The queries are bespoke to ENSAdmin, they are ENSNode, which means they can also be used with the API directly from consumer apps outside ENSAdmin.

CleanShot 2025-06-16 at 11 01 53@2x

PR #777, opened 2025-06-10
chore: ENSNode/docs - enable anchor links for specific hosted ensnode instances

Y3drk:
As the title suggests, a simple change that transferred names of ENSNode's hosted instances from table headers to section headers, thus enabling separate anchor links for them.

## How it was: 
image
---
## How it is:
image
---

## Possible adjustments
We could make the headers smaller if preferred. Below I present a comparison:
ensnode_docs_hosted_instances_headers


notrab (approved):
Approving as it can be merged now as is, but there could be some room for improvement.

Having the tables inside the markdown files look really messy and hard to parse what is what. It might make sense to create a new Astro component for these that can inline, then that file is properly formatted.

PR #780, opened 2025-06-11
feat(ensadmin): add new saved GraphQL queries

djstrong:
- Introduced multiple new GraphQL queries for domain and resolver functionalities, including fetching domains by name, label, and namehash.
- Added a category field to saved queries for better organization.
- Enhanced existing queries to support additional data retrieval options, improving overall functionality.

### DIFF
--- a/.changeset/shy-glasses-sip.md
+++ b/.changeset/shy-glasses-sip.md
@@ -1,0 +1,5 @@
+---
+"ensadmin": minor
+---
+
+add new saved GraphQL queries in ENSAdmin


djstrong:
All review comments are addressed and a few new queries are added.

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
```diff
@@ -27,6 +30,1234 @@ const savedQueries = [
...
+      2,
+    ),
+  },
+  {
+    operationName: "GetDomainByName",
+    id: "2a",
+    name: "Get Domain by Name",
+    category: "Domain",
+    description:
+      "Looks up a domain by its human-readable name (e.g., 'ens.eth'). This is more user-friendly than using namehash but may be slightly less efficient for programmatic access.",
```
lightwalker-eth, line 152: The key issue here is not that it's less efficient. The key issue is that it's not always reliable. See https://ensnode.io/docs/usage/querying-best-practices/#stable-name-identification

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
```diff
@@ -27,6 +30,1234 @@ const savedQueries = [
...
+    variables: JSON.stringify(
+      {
+        name: "ens.eth",
+      },
+      null,
+      2,
+    ),
+  },
+  {
+    operationName: "GetDomainsByLabel",
```
lightwalker-eth, line 173:
```suggestion
    operationName: "GetDomainsByChildmostLabel",
```

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
```diff
@@ -27,6 +30,1234 @@ const savedQueries = [
...
+    ),
+  },
+  {
+    operationName: "GetDomainEvents",
+    id: "5",
+    name: "Get Domain Events Only",
+    category: "Domain",
+    description:
+      "Retrieves only the domain-level events (transfers, ownership changes, wrapping events) without registration or resolver events. Use this when you're specifically interested in domain ownership and management events.",
+    //events(orderBy: blockNumber, orderDirection: desc) is not working
```
lightwalker-eth, line 343: Interesting. Why isn't it working? Is this something we should be concerned about?

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
```diff
@@ -27,6 +30,1234 @@ const savedQueries = [
...
+        labelhash: "0x5cee339e13375638553bdf5a6e36ba80fb9f6a4f0783680884d92b558aa471da", // labelhash("ens")
+      },
+      null,
+      2,
+    ),
+  },
+  {
+    operationName: "GetNameHistory",
+    id: "4",
+    name: "Get Complete Name History",
```
lightwalker-eth, line 222: This is the history of a domain, not a name

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
```diff
@@ -27,6 +30,1234 @@ const savedQueries = [
...
+        tokenId: "0x5cee339e13375638553bdf5a6e36ba80fb9f6a4f0783680884d92b558aa471da", // labelhash("ens")
+      },
+      null,
+      2,
+    ),
+  },
+  {
+    operationName: "getRegistrationByTokenId", //similar to getSubgraphRegistrant - remove?
+    id: "28",
+    name: "Get Registration by Token ID",
```
lightwalker-eth, line 1208: Hmm I don't think it's right to call this a TokenId in all cases

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
djstrong: Corrected.

apps/ensadmin/src/app/gql/subgraph-compat/page.tsx
djstrong: This is right only for:
"For an unwrapped .eth 2LD (second-level domain), the token ID for the ERC-721 NFT is the same as the labelhash for the second-level part of the domain."

PR #785, opened 2025-06-14
fix(docs/ensode): properly split up tailwind styles

### LINKED ISSUE #739: docs: bullet points don't work
I've tried both markdown syntax of using `-` and `*` neither create bullet points, maybe this has something to do with the astro upgrade? Also, new lines don't show up so i had to use `<br />` instead which is a bit annoying.

notrab:
This PR fixes #739 by splitting up styles as they should be.

CleanShot 2025-06-14 at 07 51 56@2x

* `tailwind.css` includes the tailwind import and namekit stylesheet
* `starlight.css` includes the starlight specific import and starlight overwrites
* `globals.css` includes no tailwind imports but some general variables as they're shared in both starlight and the astro site

This means @Y3drk we should not use `@namekit-react` inside the starlight pages. Tailwind only.


docs/ensnode.io/src/styles/globals.css
```diff
@@ -108,91 +70,32 @@ div.header {
...
-    padding-inline-start: initial;
-    flex: 1;
+/* Search styles */
+@media screen and (min-width: 800px) {
+  button[aria-label="Search"] {
+    width: 30vw;
   }
 }
 
-site-search dialog {
```
Y3drk, line 139: I think this needs to stay in `globals.css` because we are using Astro's `<Search />` component on the landing page.

The searchbar itself looks as expected (cause it uses a custom overlay and its styles are left in globals as well) but when you open the `Search dialog` in the LP it looks like this: 
image

So I suspect that these styles not being available in the LP is the reason, especially since when this modal is opened in the `/docs` it looks fine (cause the `starlight.css` is imported there and it contains those styles now)

Y3drk (changes requested):
# In general
The idea and the changes in the code look legit to me, but there is some stuff that I think we should fix before merging. 

Have a look at the additional comment I left and a section below.

---

## Minor UI stuff
Two quite minor issues: 
1. 
image

After the changes, when you hover these buttons (`/molecules/TelegramInvite.astro`) the cursor, doesn't change to `pointer`, but just adding `cursor-pointer` there should be enough  

2. This one might be a little bit more difficult to fix. When in the `/docs` the `info icon` is a little bit misplaced. 
image

In the LP the placement is fine, 
image

thus I am unsure what is the cause of this. Both `DocsSearch` and `SearchKeyboardShortcut` components that are responsible for this do not use `namekit-react`. It would be nice if you could investigate it.

---

## Summary
IMO, after these suggestions are addressed we can go ahead and merge this PR

docs/ensnode.io/src/styles/globals.css
notrab: Fixed.

Y3drk: One more thing, cause I'm actually an idiot. 

The variables that we use here are defined in `starlight.css`, not here, hence this doesn't quite work yet.
So I propose replacing the variables with what we have defined in the `globals.css`, as follows:

```
site-search dialog {
  background-color: var(--color-background);
  border: 1px solid var(--color-border-light);
  box-shadow: none;
}
``` 

When I tried it locally, it worked just fine. If you have any other ideas for the replacements, go ahead

I'll give you an approve in advance, but please include this hotfix here.

PR #788, opened 2025-06-16
improved bullet styles for docs

notrab:
The previous PR made way to actually improve the bullet styles the way we want them. The styles now look much better across the docs:

CleanShot 2025-06-16 at 12 53 50@2x
CleanShot 2025-06-16 at 13 01 09@2x
CleanShot 2025-06-16 at 13 01 18@2x
CleanShot 2025-06-16 at 13 01 56@2x


Y3drk (approved):
Both code & the improved styling look legit to me

PR #800, opened 2025-06-23
fix: test with new commit mode no changesets

shrugs:
- pin to previous version without offending change
- ensure release.yml can execute


shrugs:
merging as hotfix

PR #803, opened 2025-06-23
fix: rename ENSNamespace to ENSNamespaceId, and DatasourceMap to ENSNamespace

shrugs:
- don't think it needs a changelog, is bundled with previous rename, also is internal
- keeping the public-facing configuration value as `namespace` for clarity/readability
- internally using `namespaceId` where appropriate
- updated all docs i could find

### DIFF
--- a/packages/datasources/src/lib/types.ts
+++ b/packages/datasources/src/lib/types.ts
@@ -1,7 +1,7 @@
 import type { Abi, Address, Chain } from "viem";
 
 /**
- * ENSNamespaces encodes the set of well-known ENS namespaces.
+ * ENSNamespaceIds encodes the set of identifiers for well-known ENS namespaces.
  *
  * Each ENS namespace is a single, unified set of ENS names with a distinct onchain root
  * Registry (the ensroot Datasource) and the capability of spanning from that root Registry across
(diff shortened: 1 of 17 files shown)


packages/datasources/src/lib/types.ts
```diff
@@ -95,9 +95,10 @@ export type ContractConfig =
     };
 
 /**
- * DatasourceMap encodes the set of known Datasources within an ENS namespace, keyed by DatasourceName.
+ * ENSNamespace encodes a set of known Datasources, together constituting an ENS namespace.
```
lightwalker-eth, line 98:
```suggestion
 * ENSNamespace encodes a set of known Datasources associated with the same ENS namespace.
```

This is a key distinction.

packages/datasources/src/lib/types.ts
```diff
@@ -95,9 +95,10 @@ export type ContractConfig =
...
+ *
  * The ENSRoot Datasource is required, and all others are optional.
```
lightwalker-eth, line 100:
```suggestion
 * Each ENSNamespace requires an ENSRoot Datasource (this is what forms the namespace, technically). All other Datasources within the ENSNamespace are optional to define.
```

lightwalker-eth (approved):
@shrugs Looks nice! Thank you 🙌 2 small suggestions. Please merge when ready 👍

PR #818, opened 2025-06-30
deps(ponder): update to 0.11

tk-o:
This PR addresses breaking changes introduced in Ponder v0.11:
- [x] https://github.com/ponder-sh/ponder/pull/1791
- [x] https://github.com/ponder-sh/ponder/pull/1704
- [x] explicitly set ordering strategy to `omnichain`
- [x] migrate `context.network` to `context.chain`
- [x] migrate Ponder config props from `networks` to `chains`, and from `network` to `chain`
- [x] update logic reading from `publicClients` of `ponder:api` to make sure the keys are read as chain names (replacing chain IDs)

Blocked by
- https://github.com/ponder-sh/ponder/pull/1863
- https://github.com/ponder-sh/ponder/pull/1866

Closes #741

PR #828, opened 2025-07-02
refactor: update terminology

tk-o:
Replace `network` with `chain`, and `networks` with `chains`. This change aligns well with updates introduced in Ponder 0.11 version.

Related to #818


packages/ponder-metadata/src/types/common.ts
```diff
@@ -21,9 +21,12 @@ export interface BlockInfo {
 }
 
 /**
- * Network indexing status for a chain.
+ * Chain indexing status for a chain.
```
shrugs, line 24: let's make this just 'indexing status for a chain' for readability

shrugs (approved):
LGTM

PR #855, opened 2025-07-07
feat(ensindexer): include `GET /amirealtime` endpoint

### LINKED ISSUE #834: Create HTTP endpoint to monitor realtime indexing status
# ENSNode HTTP Endpoint for Monitoring Realtime Indexing Status

We need ENSNode to support a new HTTP endpoint to enable realtime indexing status monitoring.

## Endpoint Specification

Create a `GET /amirealtime` endpoint that will:

- Return a `503` HTTP response if the lowest timestamp across the last indexed block data for each chain is more than 10 minutes away from the current timestamp
- Otherwise, return a `200` HTTP response

In both cases, the endpoint should return the time difference between these timestamps (current timestamp minus the lowest indexed block timestamp).

tk-o:
This PR:
- introduces a new ENSNode HTTP API endpoint
  - uses a new `realitime-indexing-status` module which includes:
    - `dto.ts` file with functions for parsing request data and composing response data
    - `service.ts` file containing business logic
    - `utils.ts` file for other functions, such as defining how to get ponder status.
- defines a set of new types in the ENSNode SDK
- updates ENSNode documentation to describe the new endpoint

### About the endpoint

The `GET /amirealtime` endpoint allows the monitoring tools to know the current progress of indexing for the ENSNode instance.

The `GET /amirealtime` request accepts an optional `maxAllowedIndexingLag ` query param. It allows the requesting client to pass its own accepted `maxAllowedIndexingLag ` value (in seconds). If `maxAllowedIndexingLag ` query param is not provided, the default value will be applied.

### Suggested review order

1. `packages/ensnode-sdk/src/realtime-indexing-status-monitoring/types.ts`
2. `apps/ensindexer/src/lib/ponder-helpers.ts`
3. `apps/ensindexer/src/api/lib/realtime-indexing-status/*`
4. `apps/ensindexer/src/api/handlers/ensnode-api.ts`
5. `apps/ensindexer/test/realtime-indexing-status.test.ts`
6. `docs/ensnode.io/src/content/docs/docs/usage/api.mdx`

Resolves #834


apps/ensindexer/src/api/index.ts
```diff
@@ -85,6 +86,18 @@ app.get(
   }),
 );
 
+// HTTP endpoint for uptime monitoring toolkit
+app.get(
+  "/amirealtime",
+  uptimeMonitoring({
```
shrugs, line 92: `realtimeMonitoring`, not `uptime`, right?

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -153,6 +155,10 @@ const DatabaseUrlSchema = z.union(
   },
 );
 
+const RealtimeIndexingGapThresholdSchema = positiveIntegerSchema(
```
shrugs, line 158: honestly i don't think this should be configurable in the app, this should just be a constant, and consumers can provide their own gap via the api if necessary. @lightwalker-eth thoughts?

apps/ensindexer/src/lib/uptime-monitoring-middleware.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  /**
+   * Default values to be applied if optional values
+   * were not provided with the {@link RawRequest} object.
+   */
+  export interface RequestDefaults {
+    /** Describes the acceptable range in seconds between
+     * the last known block on a chain and the last indexed block on that chain.
+     */
+    gapThreshold: number;
```
shrugs, line 40: imo just `gap` or `delay`, or `lag` or `maximum` or `maximumLag`? i like the conciseness of the smaller words but we should probably use the more specific `maximum*` naming

apps/ensindexer/src/config/config.schema.ts
tk-o: @BanaSeba the `/ready` endpoint is reserved by Ponder framework. We cannot override its logic.

apps/ensindexer/src/lib/uptime-monitoring-middleware.ts
tk-o: `maximumLag` feels right to me 👍

apps/ensindexer/src/lib/uptime-monitoring-middleware.ts
```diff
@@ -0,0 +1,181 @@
...
+    const currentDate = new Date();
+    const publicClientChainIds = getPublicClientChainIds(options.publicClients);
+    const ponderStatus = await options.query.ponderStatus();
+
+    const lowestLastIndexedBlockTimestamp = getLowestLastIndexedBlockTimestamp(
+      publicClientChainIds,
+      ponderStatus,
+    );
+
+    const lowestLastIndexedBlockDate = fromUnixTime(lowestLastIndexedBlockTimestamp);
```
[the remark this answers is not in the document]
tk-o: Working on the unix timestamp values is just as valid as working on Date values. I'll refactor the code to work directly on the unix timestamp values and skip the conversions part.

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Creates a Hono App object to be used for providing
+ * the realtime indexing status monitoring tools with
+ * information if the requested realtime indexing gap
+ * has been achieved or not.
+ */
+export function realtimeIndexingStatusMonitoringApp(
+  options: RealtimeIndexingStatusMonitoringApp.Options,
+): Hono {
+  const app = new Hono();
```
lightwalker-eth, line 145: I don't understand why we're creating a Hono app just for 1 API endpoint?

I can understand needing to create a Hono app for N endpoints, and perhaps we're starting with just 1. But why is this Hono app so tightly bound with this one API endpoint? Something seems really wrong here.

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Default values to be applied if optional values
+   * were not provided with the {@link RealtimeIndexingStatusMonitoringApp.RawRequest} object.
+   */
+  export interface RequestDefaults {
+    /** Describes the acceptable range in seconds between
+     * the last known block on a chain and the last indexed block on that chain.
+     */
+    maxAllowedIndexingLag: RealtimeIndexingStatusMonitoring.TimeSpanInSeconds;
```
lightwalker-eth, line 28: 1. This value here isn't a time span. It's a duration. A time span is a span in time from timestamp A to timestamp B. This is just a non-negative integer value holding a duration. So better to just call that type something clean like `Duration`.
2. Why is an idea as general as `Duration` only being defined narrowly in the `RealtimeIndexingStatusMonitoring`. This doesn't seem like a good organization of code.

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+import { HTTPException } from "hono/http-exception";
+import { prettifyError, z } from "zod/v4";
+
+export namespace RealtimeIndexingStatusMonitoringApp {
+  /**
+   * Option values to be used as input for {@link realtimeIndexingStatusMonitoringApp} function.
+   */
+  export interface Options {
+    /** Query methods */
+    query: {
```
lightwalker-eth, line 14: I don't understand why this is being passed in like this.

I understand that in the processing of a request to this new endpoint, the new endpoint needs to query the ponderStatus, but why do we have to pass in a function pointer or whatever to this new endpoint? It seems unnecessarily complex.

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+    /** Describes the acceptable range in seconds between
+     * the last known block on a chain and the last indexed block on that chain.
+     */
+    maxAllowedIndexingLag: RealtimeIndexingStatusMonitoring.TimeSpanInSeconds;
+  }
+
+  /**
+   * Request object to be parsed.
+   */
+  export type RawRequest = RealtimeIndexingStatusMonitoring.Request;
```
lightwalker-eth, line 34: Why do we need a raw request type at all?

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * Describes the acceptable lag between
+     * the date of the last known block for a chain and
+     * the date of the last indexed block on that chain.
+     */
+    maxAllowedIndexingLag: RealtimeIndexingStatusMonitoring.TimeSpanInSeconds;
+  }
+}
+
+const NonNegativeIntegerSchema = z.coerce
```
lightwalker-eth, line 49: This is a generic idea. Why are we defining it only in this file? We should better organize our code for reusability. This includes helping other devs on our team more easily discover that some logic they might need has already been defined somewhere else.

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+  });
+
+/**
+ * Get the unix timestamp of the oldest last indexed block across all chains from Ponder Status.
+ *
+ * @param ponderStatus
+ * @returns unix timestamp of the oldest last indexed block
+ * @throws if the oldest indexer block timestamp was not a non-negative integer
+ */
+export function getOldestLastIndexedBlockTimestamp(
```
lightwalker-eth, line 70: Why are we putting this idea here? It seems more broadly useful than to put it only within this API handler.

Why not organize this idea closer to the code that works with `PonderStatus`?

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+  const oldestLastIndexedBlockTimestamp = Math.min(...lastIndexedBlockTimestamps);
+
+  // invariant: block timestamp is a non negative integer
+  if (NonNegativeIntegerSchema.safeParse(oldestLastIndexedBlockTimestamp).error) {
+    throw new Error(
+      `The oldestLastIndexedBlockTimestamp was set to "${oldestLastIndexedBlockTimestamp}". It must be a non-negative integer.`,
+    );
+  }
+
+  return oldestLastIndexedBlockTimestamp as RealtimeIndexingStatusMonitoring.UnixTimestamp;
```
lightwalker-eth, line 85: Why are we defining types like `UnixTimestamp`? Why not just use `Date`?

Please take note to separate our business logic layers from serialization / deserialization layers with the IO associated with handling API requests and responding to them.

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+  return parsed.data;
+}
+
+/**
+ * Get the current realtime indexing status based on the provided Ponder Status.
+ */
+function getRealtimeIndexingStatus(
+  ponderStatus: PonderStatus,
+): RealtimeIndexingStatusMonitoring.RealtimeIndexingStatus {
+  const currentTimestamp = getUnixTime(
```
lightwalker-eth, line 119: This should set off alarms that our code is too complicated.

If this is more complicated then just `const now = new Date();` then it suggests something has gone wrong somewhere.

docs/ensnode.io/src/content/docs/docs/usage/api.mdx
```diff
@@ -7,6 +7,31 @@ sidebar:
...
+
+- **Status Code 503**: The requested realtime indexing gap has not been achieved
+- **Status Code 200**: The indexing lag is within acceptable limits
+
+The response data is returned regardless of the status code. Here's an example response:
+
+```json
+{
+  "currentRealtimeIndexingLag": 42608279,
+  "maxAllowedIndexingLag": 600,
```
lightwalker-eth, line 28: What is the goal for including this field in the response?

docs/ensnode.io/src/content/docs/docs/usage/api.mdx
```diff
@@ -7,6 +7,31 @@ sidebar:
 
 import { LinkCard } from "@astrojs/starlight/components";
 
+## Realtime Indexing Status Monitoring
+
+You can track ENSNode's realtime indexing status using the `GET /amirealtime` endpoint.
+
+HTTP clients can define their acceptable indexing lag using the `maxAllowedIndexingLag` query parameter. For example, if you want to check whether ENSNode is no more than 10 minutes behind the chain's latest block, you would set the maximum allowed lag to 10 minutes.
```
lightwalker-eth, line 14:
```suggestion
Each request to this endpoint can optionally define what it considers to be an acceptable realtime indexing lag in seconds using the `maxAllowedIndexingLag` query parameter. For example, to check if ENSNode's indexing is no more than 1 minute behind the last indexed block of any chain ENSNode is indexing, you would set `maxAllowedIndexingLag=60` (60 seconds = 1 minute).
```

docs/ensnode.io/src/content/docs/docs/usage/api.mdx
```diff
@@ -7,6 +7,31 @@ sidebar:
...
+
+To make this request, send the following HTTP request to ENSNode: `GET /amirealtime?maxAllowedIndexingLag=600`, where `600` represents the time span in seconds (10 minutes = 600 seconds).
```
lightwalker-eth, line 16: We also need to document the default value for `maxAllowedIndexingLag` if it is not defined.

I suggest we reduce it from 10 minutes to just 1 minute. What do you think?

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  app.get("/", async (ctx) => {
+    let maxAllowedIndexingLag: RealtimeIndexingStatusMonitoring.TimeSpanInSeconds;
+
+    try {
+      const rawRequest = {
+        maxAllowedIndexingLag: ctx.req.query("maxAllowedIndexingLag"),
+      } satisfies RealtimeIndexingStatusMonitoringApp.RawRequest;
+
+      const requestDefaults = {
```
lightwalker-eth, line 155: Why define this each time the function is called?

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
```diff
@@ -0,0 +1,192 @@
...
+        maxAllowedIndexingLag: DEFAULT_REALTIME_INDEXING_MAX_LAG,
+      } satisfies RealtimeIndexingStatusMonitoringApp.RequestDefaults;
+
+      const parsedRequest = buildRealtimeIndexingStatusMonitoringRequest(
```
lightwalker-eth, line 159: Is there a special reason why we are passing in `requestDefaults` as a param here? Why not just have this function implement the logic internally for how defaults are applied?

apps/ensindexer/src/lib/realtime-indexing-status-monitoring.ts
tk-o: What would you say about the idea of a private `validation` module inside the ENSNode SDK, that only our `apps/*` could use? It'd not be exported as part of the ENSNode SDK package. It'd be imported into apps as a dev dependency.

Alternatively, we could create a separate private package for shared validation methods.

Thoughts?

docs/ensnode.io/src/content/docs/docs/usage/api.mdx
tk-o: If the client does not provide a custom value in the `maxAllowedIndexingLag` query param, we'd use a default value defined on the route handler level.

Including the value applied as the max allowed indexing lag presents immediately what constraint was applied.

tk-o: As discussed on Slack, we'll keep this field as it might be helpful for requesting clients.

packages/ensnode-sdk/src/realtime-indexing-status-monitoring/types.ts
```diff
@@ -21,34 +9,53 @@ export namespace RealtimeIndexingStatusMonitoring {
...
-    oldestLastIndexedBlockTimestamp: UnixTimestamp;
+    oldestLastIndexedBlockDate: Date;
   }
 
   /**
    * Request object.
    */
   export interface Request {
-    maxAllowedIndexingLag: string | undefined;
+    maxAllowedIndexingLag: Duration;
```
lightwalker-eth, line 29: "lag" is a word that creates negative emotions.

Suggest to rename this idea to something more positive, such as:

`maxRealtimeAllowance` or `maxRealtimeDistance`.

I note how there's other places where this "lag" terminology is being used too. For example:`currentRealtimeIndexingLag`. An alternative (more positive!) framing for this idea could be something like: `distanceFromRealtime`.

apps/ensindexer/src/api/handlers/ensnode-api.ts
```diff
@@ -10,4 +12,60 @@ if (config.experimentalResolution) {
...
+    ponderStatus,
+  });
+
+  // 4. set appropriate HTTP status
+  if (indexingStatus.hasAchievedRequestedRealtimeIndexingGap) {
+    // if the requested realtime indexing gap has been achieved
+    // set HTTP status to `200`
+    c.status(200);
+  } else {
+    // otherwise, set HTTP status to 503
```
lightwalker-eth, line 57:
```suggestion
    // otherwise, set HTTP status to 503 (Service Unavailable)
```

apps/ensindexer/src/api/handlers/ensnode-api.ts
```diff
@@ -10,4 +12,60 @@ if (config.experimentalResolution) {
...
+  const indexingStatus = getRealtimeIndexingStatus({
+    maxAllowedIndexingLag,
+    currentDate,
+    ponderStatus,
+  });
+
+  // 4. set appropriate HTTP status
+  if (indexingStatus.hasAchievedRequestedRealtimeIndexingGap) {
+    // if the requested realtime indexing gap has been achieved
+    // set HTTP status to `200`
```
lightwalker-eth, line 54:
```suggestion
    // set HTTP status to `200` (OK)
```

apps/ensindexer/src/api/handlers/ensnode-api.ts
```diff
@@ -10,4 +12,60 @@ if (config.experimentalResolution) {
...
+    });
+  } catch (error) {
+    const errorMessage = error instanceof Error ? error.message : "Unknown error";
+
+    return c.text(errorMessage, 400);
+  }
+
+  // 2. arrange dependencies
+  const maxAllowedIndexingLag = request.maxAllowedIndexingLag;
+  const currentDate = getCurrentDate();
```
lightwalker-eth, line 41: A few suggestions here:

1. Please see my other comment about where we check the current time.
2. We should get the current time only after we call `getPonderStatus`, right? As I understand, `getPonderStatus` might even take 1 or 2 seconds to do its work as it may involve multiple database lookups, RPC calls, etc..?
   1. Is my memory correct that the RPC calls made here are to get details of the block at the tip of each indexed chain? It would be nice if we could cut down on the need for RPC calls (to potentially several chains) each time we get the indexing status. For example:
      1. If Ponder's indexing status for a given chain is that the latest indexed block timestamp is within the last X seconds (relative to now, where X might be 15 seconds or something) then we might just consider Ponder's latest indexed block to be equal to the latest block on the chain for the purpose of building our indexing status data model.

lightwalker-eth (changes requested):
@tk-o Thanks for your updates. Reviewed and shared suggestions 👍

tk-o:
Will be replaced with PR #896 which is going to allow monitoring realtime indexing status by sending `GET /api/indexing-status?maxRealtimeDistance={nonNegativeIntValue}` request. 

If this param is passed, the response will be either a HTTP error if either the _overallStatus_ is not "following" or the overall realtime distance is greater than the _maxRealtimeDistance_.

tk-o:
Replaced with #896

PR #858, opened 2025-07-07
Stop promoting ponder style graphql

### LINKED ISSUE #857: Stop Promoting Ponder-Style GraphQL API Schemas
# Goals

1. Reduce the surface area of API usage we need to support. It's good to remove these now while people aren't using them yet.
2. Improve our positioning to transition ENSNode users to the new REST-based APIs we will build (because they there's less options that we need to support transitions away from).

# Subtasks

## ENSAdmin

Image

1. Remove.
2. Update.
   1. Change text label from "GraphQL (Subgraph-style)" to "Subgraph (GraphQL)"
   2. Change route from `/gql/subgraph-compat` to `/api/subgraph`
      1. Review and update all relevant links across all of our code.

Image

Update breadcrumb from "APIs > GraphQL > Subgraph-style" to "APIs > Subgraph (GraphQL)". Please note how this changes the breadcrumb from being 3 levels to only 2.

## ENSIndexer

Note: No updates should be made to ENSIndexer for now in relation to this story. We should continue to implement the Ponder-style GraphQL APIs. The goal of this story is just to stop publicly promoting the existence of this API while we work towards phasing it out.

## Docs

Image

Remove the above.

Cleanup / update content on the following pages:

- [x] https://ensnode.io/docs/usage/api/
- [x] https://ensnode.io/docs/concepts/what-is-ensnode

notrab:
Closes #857 

Additional changes:

 - Updates the endpoint `/gql/api` to `/api/ai` so we don't have `/api/api`.
   - Alternative here would be to keep the `/gql/api` folder, but it feels weird having that with `/api/subgraph`
- Redirects
  - `/gql/ponder` to `/`
  - `/gql/subgraph-compat` to `/api/subgraph`


lightwalker-eth (approved):
@notrab Thank you 👍 Please take the lead to merge this after resolving the issue I flagged in Slack.

Issue #821, opened 2025-06-30
New ETHRegistrarController Support


shrugs:
# Key Goals

- Optimize alignment between ENSNode's indexing capabilities and the latest ENS protocol developments.
   - ... demonstrate our responsiveness to supporting new ENS enhancements.
   - ... ensure ENSNode's backwards compatibility with the ENS Subgraph also extends to quickly support any [updates that might be published to the ENS Subgraph](https://github.com/ensdomains/ens-subgraph/pull/90/files).
- Ensure all relevant events emitted by the new ETHRegistrarController are indexed by ENSNode.
   - Index the new referrer data emitted by the new `NameRegistered` and `NameRenewed` events.
      - Enable us (or anyone) to build dashboards on top of ENSNode to visualize referrer data. Ex: View aggregated referrer metrics for a given range of time. The building of these dashboards is outside the scope of this issue. In this issue we are focused on making it possible for such dashboards to be built.

# Key Open Questions

- [x] Review [the proposed Subgraph updates](https://github.com/ensdomains/ens-subgraph/pull/90/files).
   - [x] Any special considerations on updated indexing logic?
   - [x] Any special considerations on updates to the indexed data model?

# Key Subtasks

- [x] Review the new ETHRegistrarController contracts to identity potential indexing impacts and prepare a plan for us to add mature support.
   - [x] [General background info](https://discuss.ens.domains/t/allowing-the-dao-to-manually-issue-eth-2lds-including-1-and-2-character-ones/20851?u=lightwalker.eth)
   - [x] [Technical discussion](https://discuss.ens.domains/t/executable-enable-l2-reverse-registrars-and-new-eth-registrar-controller/20969)
   - [x] [Code for the new contract](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/ethregistrar/ETHRegistrarController.sol) and [its onchain deployment](https://etherscan.io/address/0x59E16fcCd424Cc24e280Be16E11Bcd56fb0CE547).
   - [x] [ENS Subgraph updates](https://github.com/ensdomains/ens-subgraph/pull/90/files)
- [x] Define deployment plan
   - [x] ENSNode rollout
   - [x] Update [our hosted instance of the ENS Subgraph](https://ensnode.io/docs/concepts/what-is-the-ens-subgraph#namehash-labs-hosted-ens-subgraphs) (and related docs)
[...]

PR #867, opened 2025-07-11
UnwrappedEthRegistrarController Support

shrugs:
closes https://github.com/namehash/ensnode/issues/821

parallels the changes to the subgraph [here](https://github.com/ensdomains/ens-subgraph/pull/90/files)

- rename `EthRegistrarControllerOld` and associated things to `LegacyEthRegistrarController`
- rename `EthRegistrarController` to `WrappedEthRegistrarController`
- add new `UnwrappedEthRegistrarController` contract
- only two events, `NameRegistered` and `NameRenewed` which are identical to existing implementation
- slightly refactored the `handleNameRegisteredByController` and `handleNameRenewedByController` arguments to match ENSNode terminology and break with the subgraph's legacy naming pattern (which was inherited by the legacy versions of the registrar controllers that emitted events with incorrectly named arguments)
- added `referrals` plugin to track referral/referrer information
  - in the future, additional referral sources can be integrated
  - consumers can run just the referrals plugin if they only need referral data

the commits are well-structured so you may review commit-by-commit if you'd like

casually confirmed working on sepolia — it picks up the registration details of names registered with the new controller

### Rollout

The changes are forward- and backward-compatible so we can deploy this version of ENSNode ahead of the proposal execution and it should Just Work ™️ 

### TODO
- changes to referrals plugin
- update documentation / example queries showing new referrer information
- snapshot comparison against a sepolia graph instance


apps/ensindexer/src/plugins/lineanames/handlers/Registrar.ts
```diff
@@ -118,6 +124,20 @@ export default function () {
...
+    async ({ context, event }) => {
+      await handleNameRenewedByController({
+        context,
+        event: {
+          ...event,
+          args: {
+            // EthRegistrarController incorrectly names its event arguments, so we re-map them here
+            label: event.args.name,
+            labelHash: event.args.label,
+            // EthRegistrarController#NameRenewed just uses `cost` param
```
lightwalker-eth, line 136: I imagine there's a deeper motivation behind the comment here but not sure I understand what it is?

apps/ensindexer/src/plugins/subgraph/handlers/Registrar.ts
```diff
@@ -78,32 +83,64 @@ export default function () {
...
-      await handleNameRegisteredByController({ context, event });
+      await handleNameRegisteredByController({
+        context,
+        event: {
+          ...event,
+          args: {
+            // LegacyEthRegistrarController incorrectly names its event arguments, so we re-map them here
+            label: event.args.name,
+            labelHash: event.args.label,
+            // LegacyEthRegistrarController just uses `cost` param
```
lightwalker-eth, line 101: Similar to my other comment. Could you expand on the docs here? I imagine there's a deeper idea we're trying to communicate here?

packages/datasources/src/ens-test-env.ts
```diff
@@ -67,16 +68,25 @@ export default {
...
+      LegacyEthRegistrarController: {
+        abi: root_LegacyEthRegistrarController,
         address: deploymentAddresses?.LegacyETHRegistrarController ?? EMPTY_ADDRESS,
         startBlock: 0,
       },
-      EthRegistrarController: {
-        abi: root_EthRegistrarController,
+      WrappedEthRegistrarController: {
+        abi: root_WrappedEthRegistrarController,
+        // TODO: once ens-test-env is updated with the new UnwrappedEthRegistrarController, update
```
lightwalker-eth, line 78: What do you think about updating our code here (and for how we build `deploymentAddresses`) so that all of the following is true:

1. After we merge this PR (PR 867):
   1. Everything in the ens-test-env still works.
   2. The ens-test-env can be independently updated at any time, and once updated, our code here would automatically switch to using the new ens-test-env logic without needing us to update anything.
2. After the ens-test-env is updated, we can come back to clean up the "forward-compatible" logic we temporarily add here.

Goal: Make it easy for the ens-test-env to be independently updated at any time, without us needing to coordinate a "synchronous" update of our datasources definition here.

Happy to discuss more in Slack 👍

packages/datasources/src/ens-test-env.ts
```diff
@@ -67,16 +68,25 @@ export default {
...
+        // this reference here
         address: deploymentAddresses?.ETHRegistrarController ?? EMPTY_ADDRESS,
         startBlock: 0,
       },
+      UnwrappedEthRegistrarController: {
+        abi: root_UnwrappedEthRegistrarController,
+        // TODO: once ens-test-env is updated with the new UnwrappedEthRegistrarController, update
```
lightwalker-eth, line 85: Please see related comment above 👍

shrugs:
the subgraph's `Registration` entity represents the fact that a domain was registered using a registrar (specifically the `ETHRegistrar:NameRegistered`). upon `NameRenewed`, the only field updated in the Registration entity is the expiry. the `Registration` entity tracks the `registrationDate`, `expiryDate`, `cost` (_not_ `baseCost` + `premium`), `registrant` (which is _always_ the original registrant, or the recipient when `Transfer`ed) and it materializes the `labelName` of the domain in question (likely to support searches at the graphql layer). so: `Registration` represents both the initial registration event of the domain but also some on-going data (expiry, namely), and doesn't represent renewals at all.

the unique event of domain registration is the `NameRegistered` entity, and for renewal the `NameRenewed` entity.

in general i think a priority of the referral plugin should be that it can be run independently, without requiring any other plugins, so that referral data is trivially indexable on its own. obviously it can/should be combined with other plugins' data to enrich the datamodel, but it shouldn't be required.

finally, the non-subgraph schema extensions (like these referral-related ones here) are filtered out of the subgraph-compatible graphql api, and aren't accessible. this could be changed (simply remove the `ext_` prefix from the relevant drizzle object names.

---

all that taken into context means i don't know that linking to the `Registration` entity is really worth it, but adding in a reference to it is trivial, so I suppose we can. i don't think that there's anything particularly useful to link to in the context of a renewal. in either case, linking to the event entity is a bit useless (they don't include any information we can't include in the `RegistrationReferral` or `RenewalReferral` event ourselves, and reconstructing those ids is not ergonomic.

summary of my preference would be that we just include all of the information one would like about a domain and its registration within the `RegistrationReferral` (and vise-versa for renewals), and optionally link it to the rest of the datamodel (via a `RegistrationReferral.registrationId` if desired).

---

re: polymorphic registration/renewal events

i originally implemented this with a polymorphic table (i.e. just `Referral` with a `reason` enum for differentiation and encoded the exact `baseCost` + `premium` = `total` logic you suggested. What I ran into is that the datamodel of a registration and a renewal differs enough that the polymorphic table becomes much less ergonomic:
1) a renewal only has `cost`, like you mentioned, but,
2) a renewal can be done by anyone, so the `registrant` field must now be nullable

so i decided to split them back out, and i'm a bit more pleased with the current version, primarily because each event is exactly specified. but if a polymorphic approach is preferred, there's nothing strictly wrong with it

---

re: `Referrer` entity

we need some way to track the aggregate statistics related to referrers' unique id (the `bytes32 referrer` value in the event) — that's what this entity accomplishes. we can remove it if we want to do something else, but as it is it just materializes the total value referred by unique referrer id. this can certainly be dropped

apps/ensindexer/src/plugins/referrals/handlers/UnwrappedEthRegistrarController.ts
```diff
@@ -0,0 +1,55 @@
...
+  ponder.on(
+    namespaceContract(pluginName, "UnwrappedEthRegistrarController:NameRegistered"),
+    async ({ context, event }) => {
+      await handleRegistrationReferral({
+        context,
+        event: {
+          ...event,
+          args: {
+            ...event.args,
+            referee: event.args.owner,
```
lightwalker-eth, line 29: I don't believe it's right to say that the `owner` is the referee? As I understand, you can register a name and set someone else as the owner (not yourself)? Therefore shouldn't the referee actually be the account that executed the transaction?

packages/ensnode-schema/src/referrals.schema.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * A RenewalReferral tracks individual occurences of referrals for ENS name renewals.
+ */
+export const ext_renewalReferral = onchainTable(
+  "ext_renewal_referral",
+  (t) => ({
+    // keyed by any arbitrary unique id, usually `event.id`
+    id: t.text().primaryKey(),
+
+    referrerId: t.hex().notNull(),
```
lightwalker-eth, line 67: Suggest we add a `refereeId` field here for renewal referrals, just as we did for registration referrals.

Please review all the implementation details of the `refereeId` for registration referrals and mirror all those ideas for renewal referrals.

lightwalker-eth (commented):
Sharing remaining open feedback :+1:

PR #871, opened 2025-07-13
ENSIP-19 Updates

shrugs:
- updates forward/reverse resolution logic based on new ENSIP-19 spec
- new `/primary-names/:address` endpoint for batch lookups of well-known chainIds supporting ENSIP-19

## TODO:
- define sepolia datasources for reverse-resolution datasources ala mainnet
- define entity representing `(chainId, registrarAddress, address) -> Name`
- upsert those entities in `apps/ensindexer/src/handlers/multi-chain/StandaloneReverseRegistrar.ts`
- integrate a new acceleration pathway: if the `activeResolver` is one of the well-known `ChainReverseResolver` addresses (get mainnet's from [here](https://discuss.ens.domains/t/executable-enable-l2-reverse-registrars-and-new-eth-registrar-controller/20969), idk where sepolia's are, you can probably just look up the ens names [here](https://docs.ens.domains/ensip/19/#sepolia)) then we know the `chainId` and the `registrarAddress` to use: just go ahead and fetch the `name` record via index via those entities and return it.
- rename the plugin to `reverse-resolution` instead of `reverse-resolvers` (inc all datasource names)


apps/ensindexer/src/api/lib/forward-resolution.ts
```diff
@@ -267,7 +252,7 @@ export async function resolveForward<SELECTION extends ResolverRecordsSelection>
           // create an un-cached publicClient
           const publicClient = createPublicClient({ transport: http(rpcConfig.url) });
 
-          // requireResolver() — validate behavior
+          // 3.1 requireResolver() — verifies that the resolver supports ENSIP-10 if necessary
           await withProtocolStepAsync(
```
lightwalker-eth, line 256: I assume this comment will be better to follow up with in a separate issue / PR?

What if we calculated this (and possibly other metadata) about each resolver as we index them, rather than during resolution time?

Goal: Transfer work from resolution time to indexing time.

apps/ensindexer/src/api/lib/forward-resolution.ts
```diff
@@ -305,8 +290,9 @@ export async function resolveForward<SELECTION extends ResolverRecordsSelection>
 
           span.setAttribute("rawResults", JSON.stringify(replaceBigInts(rawResults, String)));
 
-          // interpret the results beyond simple return values
+          // additional semantic interpretation of the raw resuls from the chain
```
lightwalker-eth, line 293:
```suggestion
          // additional semantic interpretation of the raw results from the chain
```

apps/ensindexer/src/api/lib/find-resolver.ts
```diff
@@ -25,7 +25,11 @@ const NULL_RESULT: FindResolverResult = {
 
 const tracer = trace.getTracer("find-resolver");
 
-export async function findResolver(chainId: number, name: Name) {
+export async function findResolver(
```
lightwalker-eth, line 28: Perhaps I'm missing something here, but was surprised to see `chainId` as a param to `findResolver`.

In my mental model:
1. Each name has 0 or 1 `registryResolver` (the resolver explicitly set for the name as set in the root ENS registry)
2. Each name has 0 or 1 (maybe always 1??) `effectiveResolver` based on the ENSIP-10 algorithm.
3. Each `effectiveResolver` may conditionally delegate to a `delegatedResolver` through CCIP-Read.

It will be very helpful if we can refine our language and docs to separate the different types of resolvers. For example, I'm not sure how to interpret this function. Which of the 3 resolver types listed above is this function returning? Ideally this could be self-documenting from the function name alone.

I also don't see why `chainId` is a param here. Suggest to either remove or document why it is here.

apps/ensindexer/src/api/lib/find-resolver.ts
```diff
@@ -37,16 +41,18 @@ export async function findResolver(chainId: number, name: Name) {
...
+      // the Domain's configured Resolver (see invariant above)
+      return findResolverWithIndex(chainId, name);
+    }
   }
 
-  // otherwise we _must_ have access to the indexed Domain-Resolver relations necessary to look up
-  // the Domain's configured Resolver (see invariant above)
-  return findResolverWithIndex(chainId, name);
+  // query the UniversalResolver on the ENSRoot Chain (via RPC)
+  return findResolverWithUniversalResolver(chainId, name);
```
lightwalker-eth, line 55: Please see my comment above about the name we give to the `findResolver` function.

Please also apply those ideas to the `findResolverWithIndex` and `findResolverWithUniversalResolver` functions.

lightwalker-eth, line 55: Maybe good to rename this from `...WithUniversalResolver` to `...WithRPC`.

I believe the key distinction here is that it uses an RPC call. Which contract the RPC call talks to seems like an internal implementation detail of the function.

apps/ensindexer/src/api/lib/primary-names-resolution.ts
```diff
@@ -0,0 +1,47 @@
...
+import type { Address } from "viem";
+
+import { resolveReverse } from "@/api/lib/reverse-resolution";
+import config from "@/config";
+import { withActiveSpanAsync } from "@/lib/auto-span";
+import { ENSNamespaceIds } from "@ensnode/datasources";
+
+const tracer = trace.getTracer("primary-names-resolution");
+
+const COIN_TYPES_BY_NAMESPACE = {
```
lightwalker-eth, line 12: Perhaps we could build these dynamically by looking at the config and which chainIds are being indexed?

apps/ensindexer/src/api/lib/primary-names-resolution.ts
```diff
@@ -0,0 +1,47 @@
...
+  [ENSNamespaceIds.Mainnet]: [0, 1, 10, 8453, 42161, 59144, 59144],
+
+  // Sepolia via https://docs.ens.domains/ensip/19/#sepolia
+  [ENSNamespaceIds.Sepolia]: [0, 1, 11155420, 59141, 84532, 421614, 534351],
+
+  [ENSNamespaceIds.Holesky]: [0, 1],
+  [ENSNamespaceIds.EnsTestEnv]: [0, 1],
+} as const;
+
+type PrimaryNames = Record<number, Name | null>;
```
lightwalker-eth, line 23: We've defined type aliases for things like `Label` and `Name`.

Seems it would be nice to do the same for `ChainId`.

We should document how a `ChainId` is guaranteed to be a non-negative (or positive??) integer.

Then we could make use of that type alias in places like here where it will be more self-documenting than to use `number`.

apps/ensindexer/src/api/lib/reverse-resolution.ts
```diff
@@ -10,30 +12,28 @@ import { Address, isAddress, isAddressEqual } from "viem";
...
 
 /**
  * Implements ENS Reverse Resolution, including support for ENSIP-19 L2 Primary Names.
  *
- * @see https://docs.ens.domains/ensip/19#primary-name-resolution-process
+ * @see https://docs.ens.domains/ensip/19/#reverse-resolution
  *
  * @param address the adddress to lookup the Primary Name of
- * @param [chainId=1] fetch the Primary Name of the `address` in the context of this `chainId`
+ * @param [chainId=0] the chainId context to fetch the Primary Name in, defaulting to the 'default' Primary Name
```
lightwalker-eth, line 30: As I understand we need to implement the "default" idea differently.

For example:
1. If you ask for the `DEFAULT_EVM_CHAIN_ID` then the actual reverse lookup is:
   1. `DEFAULT_EVM_CHAIN_ID` (only)
2. If you ask for chainId X then the actual reverse lookup is:
   1. chainId X (return whatever this returns if it is successful or if we generated the result through an RPC call (non-accelerated).
   2. `DEFAULT_EVM_CHAIN_ID` (as a fallback if we accelerated the reverse lookup for chainId X and that accelerated lookup didn't find anything).


The logic here isn't part of ENSIP-19 explicitly. Instead it is based on the implementation of the contracts that were deployed as part of ENSIP-19. Internally they switch to the default if they don't have the requested data. If we accelerate those contracts then we need to implement the same default fallback in our accelerated logic.

apps/ensindexer/src/api/lib/reverse-resolution.ts
```diff
@@ -45,69 +45,53 @@ export async function resolveReverse(
...
-            protocolTracingSpan,
-            TraceableENSProtocol.ReverseResolution,
-            ReverseResolutionProtocolStep.DefaultNameRecordExists,
-            !!records.name,
-          );
-          if (!records.name) {
+          if (!name) {
             span.setStatus({
               code: SpanStatusCode.ERROR,
               message: `No Primary Name for coinType "${evmChainIdToCoinType(chainId)}" or default coinType.`,
```
lightwalker-eth, line 72: It would be nice to refine this so that it's directly worded no matter if the chainId is associated to the default coinType or not.

apps/ensindexer/src/api/lib/resolve-calls-and-results.ts
```diff
@@ -24,6 +24,7 @@ const tracer = trace.getTracer("resolve-calls-and-results");
 
 // for all relevant eth_calls here, all Resolver contracts share the same abi, so just grab one from
 // @ensnode/datasources that is guaranted to exist
+// TODO: update this to use datasources' ResolverABI once it's available
```
lightwalker-eth, line 27: Cool I think I saw you added this now during my PR reviews? Or maybe that was a different PR?

lightwalker-eth (changes requested):
@shrugs Great updates! Thank you 😄 Reviewed and shared feedback 👍

packages/datasources/src/mainnet.ts
```diff
@@ -287,6 +306,31 @@ export default {
...
+        address: "0x0000000000D8e504002cC26E3Ec46D81971C1664",
+        startBlock: 31808582,
+      },
+    },
+  },
+
+  /**
+   * The Reverse Resolver on Linea.
+   */
+  [DatasourceNames.ReverseResolverLinea]: {
```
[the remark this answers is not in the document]
shrugs: Datasources describe a collection of contracts within a certain context. In this case, the context is the reverse resolver plugin. the datasources here should be called "ReverseResolver*" because they contains all of the contracts required to index (and accelerate) the _reverse resolution_ on that specific chain. They track a single contract, `L2ReverseRegistrar` which implements the `StandaloneReverseRegistrar` interface/contract.

shrugs:
superseded by https://github.com/namehash/ensnode/pull/902

packages/ensnode-sdk/src/utils/constants.ts
```diff
@@ -18,9 +18,17 @@ export const REVERSE_ROOT_NODES: Set<Node> = new Set([namehash("addr.reverse")])
  */
```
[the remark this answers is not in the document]
shrugs: https://github.com/namehash/ensnode/issues/903

Issue #832, opened 2025-07-03
Seaport Sales Plugin


Zimtente:
Official deliverable description:

> Ship support for name sales history.

Index all seaport sales for the supported contracts

lightwalker-eth:
@Zimtente Thanks for creating this issue 👍 

A few suggestions:

1. For creating a new plugin, suggest taking inspiration from the work that @tk-o is advancing for introducing a new EFP plugin in [PR 771](https://github.com/namehash/ensnode/pull/771).
2. I'm not sure if we should call this plugin "seaport" or "tokenscope". It's a bit difficult to say for sure until we get further into this effort. Happy to proceed with "seaport" for now and we can see how it goes as we build it up 👍 We can always rename later depending on what we learn.
3. As I understand, this plugin will index all seaport sales events across all NFTs and then within our indexing event handler we will filter which NFT contracts we will store sales data for in our database. To determine those contract addresses, it would be nice to reference these addresses indirectly through the definitions inside our [datasources package](https://github.com/namehash/ensnode/tree/main/packages/datasources). Related examples of indirectly getting chainId or address constants from the data models we're defining in this `datasources` package:
   1. https://github.com/namehash/ensnode/blob/main/packages/datasources/src/index.ts#L51-L52
   2. Please see the changes in progress for `packages/datasources/src/index.ts` in [PR 812](https://github.com/namehash/ensnode/pull/812).
   3. Please note the pattern we're using of passing in an `ENSNamespaceId` (such as "mainnet" or "sepolia" or "ens-test'env") to get the specific constants that are relevant for the context.
4. Overall don't suggest getting stuck on any complexity the above suggestions might create. Feel welcome to "just make it work" first and we can always make it more fancy as another iteration 👍

Zimtente:
A few questions and feedback:

1. As Seaport relies on other plugins to index names (ENS, 3DNS etc) what would the best approach be to handle this, as atm it is possible to enable and disable certain plugins? As @lightwalker-eth said we need a way to filter by "Contract" as the Seaport OrderFulfilled Event indexes any NFT sale that happens on Seaport.
2. As for seaport datasource [https://github.com/namehash/ensnode/blob/main/packages/datasources/src/mainnet.ts](url) does it make sense that we put it into the "ENSNamespace" ? 
`  [DatasourceNames.Seaport]: {
    chain: mainnet,
    contracts: {
      Seaport: {
        abi: Seaport,
[...]

PR #771, opened 2025-06-04
feat(ensindexer): introduce the `efp` plugin

tk-o:
This is a plugin demonstrating ENSNode's capability to index the Ethereum Follow Protocol data.

PR #872, opened 2025-07-14
added seaport orderfulfill event

Zimtente:
#832


packages/ensnode-schema/src/seaport.schema.ts
```diff
@@ -0,0 +1,35 @@
...
+    currencyId: t.hex().notNull(),
+    price: t.bigint().notNull(),
+    chainId: t.integer().notNull(),
+    orderHash: t.hex().notNull(),
+  }),
+  (t) => ({
+    idx_from: index().on(t.fromOwnerId),
+    idx_to: index().on(t.newOwnerId),
+    idx_currency: index().on(t.currencyId),
+    idx_compound: index().on(t.fromOwnerId, t.id),
```
lightwalker-eth, line 33: Interesting. What do you see is the goal for this index?

packages/ensnode-schema/src/seaport.schema.ts
```diff
@@ -0,0 +1,35 @@
...
+export const currency = onchainTable("currency", (t) => ({
+  id: t.hex().primaryKey(),
+  name: t.text(),
+  symbol: t.text(),
+  decimals: t.integer().notNull(),
+  contractAddress: t.hex().notNull(),
+  chainId: t.integer().notNull(),
+}));
+
+export const nameSold = onchainTable(
```
lightwalker-eth, line 18: What do you think about adding additional columns to this table for:

1. `domainId` -- this would be the node / namehash of the name that was sold. I believe we should be able to calculate this for all of the token issuing contracts we index? The goal for this is that we want to make it easy to build a database query where given any ENS name we get the list of its sales. It would be good to create an index on this field.
2. Some kind of a compound identifier for chainId + contract address + tokenId. For example, maybe a field that concatenates all 3 of these values together with `:` characters in between? The goal would be also support database queries where we can get the list of all sales of an ENS name specifically through one of its tokenized representations. This might also be nice to create an index on.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,191 @@
...
+  const isValidItemType = item.itemType === ItemType.ERC721 || item.itemType === ItemType.ERC1155;
+  const isSupportedContract = SUPPORTED_CONTRACTS.includes(item.token);
+
+  return isValidItemType && isSupportedContract;
+}
+
+/**
+ * Finds the payment item from offer array
+ */
+function findPaymentInOffer(offer: readonly OfferItem[]): OfferItem | undefined {
```
lightwalker-eth, line 145: Interesting. My feedback here applies to both `findPaymentInOffer` and `findPaymentInConsideration`.

What happens if there's more than 1 `OfferItem` or `ConsiderationItem`?

I'm worried that our indexed data model might not necessarily be correct if there's more than 1?

Should we update `isValidNFTItem` such that it returns false if there's more than 1?

Appreciate your advice!

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,191 @@
...
+  });
+}
+
+/**
+ * Handles NFT listings being fulfilled (someone buying a listed item)
+ * In a listing: NFT owner lists their NFT for sale
+ * - NFT is in offer (what the offerer gives)
+ * - Payment is in consideration (what the offerer wants)
+ */
+async function handleListing(
```
lightwalker-eth, line 96: Should we rename this to something like `handleListingFulfilled`?

Similar feedback for `handleOffer`.

Goal: Make it more clear these functions are associated with the fulfillment of a listing / offer, not the creation of a listing / offer.

Edit: Or maybe better to combine these functions together into one, and instead change the function signature so that the responsibility for unpacking `event.args` and sorting out who is the buyer / seller is handled within `handleOrderFulfilled`.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,191 @@
...
+    offerer: `0x${string}`;
+    zone: `0x${string}`;
+    recipient: `0x${string}`;
+    offer: readonly {
+      itemType: number;
+      token: `0x${string}`;
+      identifier: bigint;
+      amount: bigint;
+    }[];
+    consideration: readonly {
```
lightwalker-eth, line 28: Any special reason not to reuse the `ConsiderationItem` type definition here?

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,191 @@
...
+  "0x0635513f179D50A207757E05759CbD106d7dFcE8",
+];
+
+interface SeaportOrderFulfilledEvent
+  extends EventWithArgs<{
+    orderHash: `0x${string}`;
+    offerer: `0x${string}`;
+    zone: `0x${string}`;
+    recipient: `0x${string}`;
+    offer: readonly {
```
lightwalker-eth, line 22: Any special reason not to reuse the `OfferItem` type definition here?

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,191 @@
...
+// Supported contracts
+const SUPPORTED_CONTRACTS = [
+  "0x57f1887a8bf19b14fc0df6fd9b2acc9af147ea85",
+  "0x0635513f179D50A207757E05759CbD106d7dFcE8",
+];
+
+interface SeaportOrderFulfilledEvent
+  extends EventWithArgs<{
+    orderHash: `0x${string}`;
+    offerer: `0x${string}`;
```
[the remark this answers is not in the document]
Zimtente: yeah I had this pending locally

packages/ensnode-schema/src/seaport.schema.ts
Zimtente: Idk I thought if you want to look up sales for specific accounts this may be interesting. or if you want to filter by currency etc. but was more just examples. I can minimize it and just dont create any if that is cool!

packages/ensnode-schema/src/seaport.schema.ts
```diff
@@ -0,0 +1,35 @@
...
+  "name_sold",
+  (t) => ({
+    ...sharedEventColumns(t),
+    fromOwnerId: t.hex().notNull(),
+    newOwnerId: t.hex().notNull(),
+    currencyId: t.hex().notNull(),
+    price: t.bigint().notNull(),
+    chainId: t.integer().notNull(),
```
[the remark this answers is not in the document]
Zimtente: yes I would do that. I just copied from subgraph

apps/ensindexer/src/handlers/Seaport.ts
Zimtente: event.args is always the same. its basically just about how its interpreted. 
for offers: you are offering payment 
for listings: you are offering a name

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,181 @@
...
+type ConsiderationItem = {
+  itemType: ItemType;
+  token: Address;       // contract address
+  identifier: bigint;   // token id
+  amount: bigint;
+};
+
+interface SeaportOrderFulfilledEvent
+  extends EventWithArgs<{
+    orderHash: Hex;
```
lightwalker-eth, line 32: Interesting. It's nice to optimize our code for reading from other devs who aren't going to be familiar with Seaport already. Does Seaport have some docs on what an `orderHash` is? If so would be cool to link to it here or document the idea of a Seaport orderHash in a summarized way.

lightwalker-eth (commented):
@Zimtente Hey a few other small ideas.

apps/ensindexer/src/handlers/Seaport.ts
Zimtente: I can document all fields if necessary no problem

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,181 @@
...
+    offerer: Address;
+    zone: Address;        // the zone contract this order was created in
```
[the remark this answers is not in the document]
Zimtente: will add more docs

apps/ensindexer/src/handlers/Seaport.ts
Zimtente: imho we should ignore such transactions

packages/ensnode-schema/src/seaport.schema.ts
```diff
@@ -0,0 +1,35 @@
+import { index, onchainTable } from "ponder";
+
+const sharedEventColumns = (t: any) => ({
+  id: t.text().primaryKey(),
```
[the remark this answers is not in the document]
Zimtente: 11155111-5584563-41

apps/ensindexer/src/lib/seaport/seaport-helpers.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export async function lookupDomainId(
+  context: Context,
+  contractAddress: Address,
+  tokenId: string,
+): Promise<Hex | null> {
+  const tokenIdHex = `0x${BigInt(tokenId).toString(16).padStart(64, "0")}` as Hex;
+
+  // OLD ENS Registry: tokenId is labelhash, need to find domain by labelhash
+  if (contractAddress === "0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85") {
```
tk-o, line 19: It'd be great to replace this inline contract address value with a reference to that value from `@ensnode/datasources`.

For example, you can get this address in a following way:
```ts
import { DatasourceNames, getDatasource } from "@ensnode/datasources"
import config from "@/config";

/** ENSRoot Base Registrar Contract Address */
const baseRegistrarContractAddress = getDatasource(config.namespace, DatasourceNames.ENSRoot)
  .contracts["BaseRegistrar"].address;
```

and then use the `baseRegistrarContractAddress` reference 🚀

packages/datasources/src/index.ts
```diff
@@ -217,3 +225,83 @@ export function getChainName(chainId: number): string {
...
+          // NameWrapper Token - Sepolia
+          chainId: sepoliaChain.id,
+          address: "0x0635513f179D50A207757E05759CbD106d7dFcE8",
+        },
+      ];
+    case ENSNamespaceIds.Holesky:
+    case ENSNamespaceIds.EnsTestEnv:
+      return [];
+  }
+};
```
tk-o, line 288: As in the previous comment, I recommend replacing inline contract address values with specific references to contract addresses documented in `@ensnode/datasources` package 👍 

```suggestion
export const getKnownTokenIssuingContracts = (namespaceId: ENSNamespaceId): ChainAddress[] => {
  switch (namespaceId) {
    case ENSNamespaceIds.Mainnet: {
      const rootDatasource = getDatasource(namespaceId, DatasourceNames.ENSRoot);
      const lineanamesDatasource = getDatasource(namespaceId, DatasourceNames.Lineanames);
      const basenamesDatasource = getDatasource(namespaceId, DatasourceNames.Basenames);
      const threeDnsBaseDatasource = getDatasource(namespaceId, DatasourceNames.ThreeDNSBase);
      const threeDnsOptimismDatasource = getDatasource(namespaceId, DatasourceNames.ThreeDNSOptimism);

      return [
        {
          // ENS Token - Mainnet
          chainId: rootDatasource.chain.id,
          address: rootDatasource.contracts['BaseRegistrar'].address,
        },
        // NameWrapper Token - Mainnet
        {
          chainId: rootDatasource.chain.id,
          address: rootDatasource.contracts['NameWrapper'].address,
        },
        // 3DNS Token - Optimism
        {
          chainId: threeDnsOptimismDatasource.chain.id,
          address: threeDnsOptimismDatasource.contracts['ThreeDNSToken'].address,
        },
        // 3DNS Token - Base
        {
          chainId: threeDnsBaseDatasource.chain.id,
          address: threeDnsBaseDatasource.contracts['ThreeDNSToken'].address,
        },
        // Linear Names Token - Base
        {
          chainId: lineanamesDatasource.chain.id,
          address: lineanamesDatasource.contracts['BaseRegistrar'].address,
        },
        // Base Names Token - Base
        {
          chainId: basenamesDatasource.chain.id,
          address: basenamesDatasource.contracts['BaseRegistrar'].address,
        },
      ];
    }

    case ENSNamespaceIds.Sepolia: {
      const rootDatasource = getDatasource(namespaceId, DatasourceNames.ENSRoot);

      return [
        {
          // ENS Token - Sepolia
          chainId: rootDatasource.chain.id,
          address: rootDatasource.contracts['BaseRegistrar'].address,
        },
        {
          // NameWrapper Token - Sepolia
          chainId: rootDatasource.chain.id,
          address: rootDatasource.contracts['NameWrapper'].address,
        },
      ];
    }
    case ENSNamespaceIds.Holesky:
    case ENSNamespaceIds.EnsTestEnv:
      return [];
  }
};
```

tk-o (commented):
@Zimtente very nice work 🚀 I shared some feedback, feel free to follow up.

lightwalker-eth (changes requested):
@Zimtente Reviewed and shared feedback.

packages/ensnode-schema/src/seaport.schema.ts
```diff
@@ -0,0 +1,32 @@
...
+  chainId: t.integer().notNull(),
+});
+
+export const nameSold = onchainTable(
+  "name_sold",
+  (t) => ({
+    ...sharedEventColumns(t),
+    fromOwnerId: t.hex().notNull(),
+    newOwnerId: t.hex().notNull(),
+    currencyAddress: t.hex().notNull(),
```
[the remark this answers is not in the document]
Zimtente: in regards to the indexes - last we discussed this we said we will move this change into the "sales-endpoint" task as at this point its not clear what indexes may be needed. This is why I "resolved" it.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,283 @@
...
+    console.log("Domain ID not found for", contractAddress, tokenId);
+    return;
+  }
+
+  // Record the sale
+  await context.db.insert(schema.nameSold).values({
+    ...sharedEventValues(context.chain.id, event),
+    fromOwnerId: seller,
+    newOwnerId: buyer,
+    currencyAddress: currencyAddress,
```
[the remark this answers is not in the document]
Zimtente: I think tracking it like this would make it so that we can just fetch currency details we support in our /sales routes. This way we index >all sales< and then just populate these the remaining details ala name, symbol, decimals (hardcoded) - in the future may also be able to fetch currency details from third party provider and cache it @API lvl

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,283 @@
...
+  // Check if this is an offer (payment in offer, NFT in consideration)
+  const paymentItemsInOffer = findPaymentItemsInOffer(offer);
+  const nftInConsideration = consideration.find((item) => isIndexable(item, context));
+
+  if (paymentItemsInOffer.length > 0 && nftInConsideration) {
+    await handleOfferFulfilled(context, event, nftInConsideration, paymentItemsInOffer);
+    return;
+  }
+
+  console.log(
```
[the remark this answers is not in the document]
Zimtente: yeah it does. question is if we need to log it or not. I would say it doenst matter we should not log it.

apps/ensindexer/src/config/validations.ts
```diff
@@ -167,3 +167,31 @@ export function invariant_experimentalResolutionNeedsReverseResolversPlugin(
...
   }
 }
+
+export function invariant_seaportPluginNeedsMinimumOneOtherNamePluginActive(
+  ctx: ZodCheckFnInput<Pick<ENSIndexerConfig, "plugins">>,
+) {
+  const { value: config } = ctx;
+
+  if (config.plugins.includes(PluginName.Seaport)) {
+    const nameProtocols = [
```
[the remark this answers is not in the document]
Zimtente: removed this completely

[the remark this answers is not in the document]
Zimtente: removed this completely

packages/ensnode-schema/src/seaport.schema.ts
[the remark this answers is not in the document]
Zimtente: sharedEventColumns - i copy pasted from subgraph schema as is. but will add comments

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+    fromOwnerId: seller,
+    newOwnerId: buyer,
+    currencyAddress: currencyAddress,
+    chainId: context.chain.id,
+    logIndex: event.log.logIndex,
+    orderHash: orderHash,
+    price: totalAmount,
+    contractAddress: contractAddress,
+    tokenId: tokenId,
+    tokenType: nftItem.itemType === ItemType.ERC721 ? "ERC721" : "ERC1155",
```
lightwalker-eth, line 153: Why do this? We should have already validated the range of values to be correct inside `isSupportedTokenTypeAndContract`

packages/datasources/src/index.ts
```diff
@@ -217,3 +226,145 @@ export function getChainName(chainId: number): string {
...
+  );
+};
+
+/**
+ * Get the domainId by contract address and tokenId
+ * @param namespaceId - The ENSNamespace identifier (e.g. 'mainnet', 'sepolia', 'holesky', 'ens-test-env')
+ * @param contractAddress - contract address of the NFT
+ * @param tokenId - tokenId of the NFT
+ */
+export function getDomainIdByTokenId(
```
lightwalker-eth, line 353: Ah I just noticed an important issue here. We can't trust the result of this function unless we also pass a chainId param too. It's not enough just to look at the contract address by itself.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+  context,
+  event,
+}: {
+  context: Context;
+  event: SeaportOrderFulfilledEvent;
+}) {
+  const { offer, consideration } = event.args;
+
+  // Check if this is a listing (NFT in offer, payment in consideration)
+  const nftInOffer = offer.find((item) => isSupportedTokenTypeAndContract(item, context));
```
lightwalker-eth, line 259: Hmm.. as I understand this will cause us to index offers where at least 1 item is a supported token type and contract.

But what about the case where there's an offer with multiple items? As I understand we don't want to index those. If we did index them (as the code here currently appears to) then we wouldn't represent them honestly in our data model.

Our mindset should be simple: we define a strict set of cases we support. For those cases, we should represent them with perfect accuracy. Anything outside those strict set of supported cases should be completely ignored.

Please ensure this idea is applied everywhere.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+  // In a fulfilled listing, the offerer is selling the NFT, recipient is buying
+  const seller = offerer;
+  const buyer = recipient;
+
+  // Ensure accounts exist
+  await upsertAccount(context, seller);
+  await upsertAccount(context, buyer);
+
+  // Get payment details
+  const { currencyAddress, totalAmount } = validateAndGetPaymentDetails(paymentItems);
```
lightwalker-eth, line 179: This is going to crash the indexer when it throws.

Please ensure you've fully tested everything before the next PR review.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+     * Array of items that the offerer expects to receive in return.
+     * For listings: ETH/ERC20 tokens expected as payment
+     * For offers: NFTs/tokens being requested in exchange
+     */
+    consideration: readonly ConsiderationItem[];
+  }> {}
+
+type PaymentDetails = {
```
lightwalker-eth, line 71: Maybe this can be transformed into just `type Price` where instead of having a `currencyAddress` it has a enum-like value for the set of whitelisted currencies we support.

The `validateAndGetPaymentDetails` function can then simply just become `getPrice` and return a `Price` or throw if a single Price cannot be taken from the paymentItems.

I also think there's a good opportunity to refactor your indexing logic into more clear layers of responsibility. Ex: `handleOrderFulfilled` could handle the full responsibility for validating that the order is something we can index (including getting the price!), and then the `handleListingFulfilled` and `handleOfferFulfilled` functions can be passed "clean" / "validated" params to do their work more simply and with confidence since they can know whatever they are passed is ready for indexing.

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+      "No payment item. Provide at least one payment item to get the payment token address.",
+    );
+  }
+
+  // Get all unique tokens used in payment items
+  const paymentTokens = paymentItems.map((item) => item.token);
+  const uniqueTokens = [...new Set(paymentTokens)];
+
+  // Mixed currencies
+  if (uniqueTokens.length > 1) {
```
lightwalker-eth, line 91: Interesting.. there can be multiple paymentItems with the same token?

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+  identifier: bigint; // token id
+  amount: bigint;
+};
+
+type ConsiderationItem = {
+  itemType: ItemType;
+  token: Address; // contract address
+  identifier: bigint; // token id
+  amount: bigint;
+  recipient: Address;
```
lightwalker-eth, line 23: What is the purpose of this `recipient` field? Can you please document what it means / represents?

packages/datasources/src/index.ts
```diff
@@ -217,3 +226,145 @@ export function getChainName(chainId: number): string {
...
+  namespaceId: ENSNamespaceId,
+  contractAddress: Address,
+  tokenId: string,
+): Hex {
+  const tokenIdHex = `0x${BigInt(tokenId).toString(16).padStart(64, "0")}` as Hex;
```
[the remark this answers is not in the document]
Zimtente: What exactly do you want to be unit tested?

apps/ensindexer/src/handlers/Seaport.ts
```diff
@@ -0,0 +1,276 @@
...
+
+/**
+ * Checks if the item is an ERC721 or ERC1155 item
+ * and if the token contract is a known token contract
+ */
+function isSupportedTokenTypeAndContract(
+  item: OfferItem | ConsiderationItem,
+  context: Context,
+): boolean {
+  if (!item || !item.token) return false;
```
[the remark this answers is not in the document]
Zimtente: May not need to check for undefined or null, agree.

but we may still check item.token not being "" - idk if it can happen (this comes from onchain seaport event - potentially it can?). how would you handle this?

apps/ensindexer/src/handlers/Seaport.ts
Zimtente: itemType is a number - we can use number as well. but string/enum may be easier to read. will change as you want 

image

apps/ensindexer/src/handlers/Seaport.ts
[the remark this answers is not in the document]
lightwalker-eth: Ah thanks for the background info.

Is it true that `item.token` can be an empty string? If so then we should update the type of item.token, because `""` is not a valid `Address`.

We then would need to change the type of item.token from `Address` to just `string` and then perform validation inside our logic here to confirm it is a valid `Address` value before casting it to `Address`. If it isn't then the event should be considered non-indexable.

apps/ensindexer/src/handlers/Seaport.ts
lightwalker-eth: Ah thanks for catching that! The current logic here makes sense then.

I would suggest that we define an enum-like value then for the possible set of `tokenType` values that might be stored in our DB. The best practice for this is not to use a literal Javascript enum, but instead to follow the pattern that was used in other places in our code for enum-like values. For example, please see how we implemented `ENSNamespaceId` here: https://github.com/namehash/ensnode/blob/main/packages/datasources/src/lib/types.ts#L29-L39 This is the best practice.

apps/ensindexer/src/handlers/Seaport.ts
lightwalker-eth: If a sale doesn't meet the criteria for being indexed, we shouldn't crash the whole process. We should simply catch the error and skip the indexing of that sale.

apps/ensindexer/src/handlers/Seaport.ts
lightwalker-eth: Ah interesting! Can you please document this idea in our code? That will be helpful

apps/ensindexer/src/handlers/Seaport.ts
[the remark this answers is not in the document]
lightwalker-eth: We already are not indexing all sales. Only sales meeting our allow-listed criteria are being indexed. This allow-listed criteria should also include that the price of the sale is in an allow-listed currency.

There's a high cost to adding external dependencies and no need for it in this case.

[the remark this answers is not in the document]
Zimtente: I will remove it and replace it with enum values.

packages/ensnode-schema/src/tokenscope.schema.ts
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * The account that received the name
+     */
+    newOwnerId: t.hex().notNull(),
+    /**
+     * Currency address of the payment
+     */
+    currencyAddress: t.hex().notNull(),
+    /**
+     * The amount of the payment
```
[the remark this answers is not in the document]
Zimtente: I dont understand what you want me todo. maybe can give an example. thank you

packages/datasources/src/index.ts
```diff
@@ -19,6 +20,14 @@ export * from "./lib/types";
 // export the shared ResolverABI for consumer convenience
 export { ResolverABI } from "./lib/resolver";
 
+/**
+ * Identifies a specific address on a specific chain.
+ */
+export interface ChainAddress {
+  chainId: number;
```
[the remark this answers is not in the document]
Zimtente: hmm is this in a newer branch I cant seem to find ChainId type anywhere. can you link?

apps/ensindexer/src/handlers/Seaport.ts
lightwalker-eth: It's fine to throw. That makes sense in those utility functions. The key idea is to catch those errors at a higher level inside the indexing function and then skip performing any db write operations.

I don't know how to explain this any more simply. If any question please setup a call with me.

packages/datasources/src/index.ts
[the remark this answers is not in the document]
lightwalker-eth: Did you search for it? I found it right away when searching.

CleanShot 2025-08-06 at 20 37 43


https://github.com/namehash/ensnode/blob/afb994cac1f2bf3aa18ec9ecc1b33e5a58b6525a/packages/ensnode-sdk/src/shared/types.ts#L7

packages/ensnode-schema/src/tokenscope.schema.ts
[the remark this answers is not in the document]
lightwalker-eth: You can see an example in PR 771: https://github.com/namehash/ensnode/pull/771

View the comments in `apps/ensindexer/src/plugins/efp/lib/lsl.ts` for the following fields on `ListStorageLocationContract`:

1. `chainId`
2. `listRecordsAddress`
3. `slot`

The big idea here is simple. When fields have relationships to each other that should be made clear in the documentation. For example, in the tokenscope context, the chainId and contractAddress fields are NOT independent. Likewise for the tokenId field.

I don't know how to explain these ideas in a more simple way. In my mind the idea here is completely obvious and intuitive.

We need to properly document these ideas.

PR #880, opened 2025-07-15
feat(ensnode-sdk): Indexing Status data model

tk-o:
This PR creates a new module inside ENSNode SDK to describe data model for the `IndexingStatus` entity. The data model has been split into two different perspectives:
- Domain (used in business logic across ENSNode applications)
- DTO (used for serialization and deserialization).

The Domain data model can be mapped into the DTO data model, and DTO data model can be serialized into a JSON string. The reverse is also possible, so a JSON string can be deserialized into the DTO data model and then mapped into the Domain data model.

I created the `ENSNode` namespace to capture ENSNode specific, but still general, types:
- `packages/ensnode-sdk/src/ensnode/types.ts`
Also, extended the utils types, which belong beyond the `ENSNode` namespace:
- `packages/ensnode-sdk/src/utils/types.ts`

I used the `ENSNode` namespace to define the Indexing Status data model. Please note how we use `Date` & `DatetimeIso8601` types, and also how we use `Map` & inline `Record` type:
- `packages/ensnode-sdk/src/indexing-status/domain/types.ts` 
- `packages/ensnode-sdk/src/indexing-status/dto/types.ts`

The Domain model can include types that cannot be serialized into JSON string, where the DTO model must use only the types that can be serialized into JSON string.

Lastly, I created functionality to:
- map Domain model into DTO model
    - `packages/ensnode-sdk/src/indexing-status/mappers/domain-to-dto.ts`
- map DTO model into Domain model (strong validation and invariants enforcement)
    - `packages/ensnode-sdk/src/indexing-status/mappers/dto-to-domain.ts`

I've written lots of tests to cover type correctness and conversion capabilities.

> Note: Zod schema I created for parsing the DTO model into the Domain model is quite complex. It uses `z.union` schema definition to join all chain status permutation schemas into a single schema. However, this causes the parsing error message to be quite not helpful. For example, if some chain status validation fails, all the user can see is a somewhat generic message: "Failed to map IndexingStatus DTO: \n✖ Invalid input\n → at 3\n".


packages/ensnode-sdk/src/indexing-status/mappers/domain-to-dto.ts
```diff
@@ -0,0 +1,74 @@
...
+): blockInfo is IndexingStatusDomain.BlockInfo {
+  const blockKeys = Object.keys(blockInfo);
+  // A BlockInfo object has more than one property attached.
+  return blockKeys.length > 1;
+}
+
+/**
+ * Map block info (either partial or full).
+ */
+export function mapBlockInfo(
```
lightwalker-eth, line 19: I'm worried about using namespaces like this. I think it can cause a bunch of confusion in LLMs or when we do things like search all our files for uses of a given type.

Here's an alternative suggestion:
1. Remove any use of namespaces.
2. For our "domain" models, just name them in a "pure" way, such as `BlockInfo`.
3. For our "dto" models, prefix them with "Serialized..." such as "SerializedBlockInfo".

Additionally, instead of prefixing these function names with "map..." we could then prefix these functions names with "serialize..." which all seems to fit together nicely.

Likewise, we can rename "mappers" (the directory this file is in) to "serializers" or "serialization".

What do you think?

packages/ensnode-sdk/src/indexing-status/mappers/dto-to-domain.ts
```diff
@@ -0,0 +1,249 @@
...
+    return indexingStatusDomain;
+  });
+
+/**
+ * Map IndexingStatus DTO object into Indexing Status Domain object.
+ *
+ * @returns {IndexingStatusDomain.IndexingStatus}
+ * @throws {Error} when the DTO object could not be mapped.
+ */
+export function mapIndexingStatusDtoIntoDomain(
```
lightwalker-eth, line 239: Please see my other comment related to the use of the prefix "map...".

For this case, instead of using a prefix of "serialize..." we could use a prefix of "deserialize...".

Ex: `deserializeIndexingStatus(serialized: SerializedIndexingStatus): IndexingStatus`

packages/ensnode-sdk/src/indexing-status/domain/types.ts
```diff
@@ -0,0 +1,40 @@
...
+   * Chain Status
+   *
+   * A value type used in {@link IndexingStatus}.
+   */
+  export type ChainStatus = ENSNode.ChainStatus<BlockInfo>;
+
+  /**
+   * Indexing Status
+   *
+   * Represents key-value pairs of Chain ID and Chain Status.
```
lightwalker-eth, line 37: What do you think about documenting here why we're using Map and not a Record type? Ex: Using Map (rather than a Record) to avoid needing to convert ChainId values to string.

Feel welcome to ignore this comment if you think no benefit.

lightwalker-eth (commented):
@tk-o Really appreciate your special efforts here! Reviewed and shared a few suggestions 👍

tk-o:
Replaced with #896 (which builds on top of #881).

PR #881, opened 2025-07-16
wip: indexing status endpoint

typedarray:
Introduces a new `/indexing-status` endpoint which serves useful progress metrics for each chain.

The custom API endpoint uses data from the Ponder config file, metrics, Status API, and the RPC.

Here's the sample output for a mainnet only instance on my local machine. I also tested it quite a bit on other ponder projects running multiple chains with both multichain and omnichain ordering.

```json
{
  "chains": {
    "1": {
      "status": "backfill",
      "config": {
        "startBlock": {
          "number": 3327417,
          "timestamp": 1489165544
        },
        "endBlock": null
      },
      "latestSyncedBlock": {
        "number": 3884351,
        "timestamp": 1497652888
      },
      "latestIndexedBlock": {
        "number": 3828421,
        "timestamp": 1496734883
      },
      "latestBlock": {
        "number": 22932770,
        "timestamp": 1752681191
      }
    }
  }
}
```

### DIFF
--- a/apps/ensindexer/src/api/lib/indexing-status.ts
+++ b/apps/ensindexer/src/api/lib/indexing-status.ts
@@ -1,0 +1,398 @@
+import { publicClients } from "ponder:api";
+import { createMiddleware } from "hono/factory";
+import config from "../../../ponder.config.js";
+
+type BlockRef = { number: number; timestamp: number };
+
+type ChainIndexingConfig = {
+  // If all contracts, accounts, and block intervals are set to "latest",
+  // startBlock will be null.
+  startBlock: BlockRef | null;
+  // If all contracts, accounts, and block intervals are set to undefined,
+  // endBlock will be null.
+  endBlock: BlockRef | null;
+};
+
+type ChainIndexingStatus =
+  | {
+      status: "not_started";
+      config: ChainIndexingConfig;
+      latestBlock: BlockRef;
+    }
+  | {
+      status: "backfill";
+      config: ChainIndexingConfig;
+      latestSyncedBlock: BlockRef;
+      latestIndexedBlock: BlockRef;
+      latestBlock: BlockRef;
+    }
+  | {
+      status: "realtime";
+      config: ChainIndexingConfig;
+      latestSyncedBlock: BlockRef;
+      latestIndexedBlock: BlockRef;
+      latestBlock: BlockRef;
+    }
+  | {
+      status: "completed";
+      config: ChainIndexingConfig;
+      latestSyncedBlock: BlockRef;
+      latestIndexedBlock: BlockRef;
+      latestBlock: BlockRef;
+    };
+
+type IndexingStatus = { chains: Record<number, ChainIndexingStatus> };
+
+export const indexingStatusMiddleware = createMiddleware(async (c) => {
+  try {
+    const baseUrl = c.req.url.replace(/\/[^\/]*$/, "");
+    const [metricsText, statusJson] = await Promise.all([
+      fetch(`${baseUrl}/metrics`).then((r) => r.text()),
+      fetch(`${baseUrl}/status`).then((r) => r.json()),
+    ]);
+
+    const metrics = parsePrometheusMetrics(metricsText);
+    const status = statusJson as {
+      [chainName: string]: {
+        id: number;
+        block: { number: number; timestamp: number };
+      };
+    };
+
+    const settingsInfo = metrics.getMetricValues("ponder_settings_info");
+    const ordering = settingsInfo[0]?.labels.ordering;
+    if (ordering === undefined) {
+      throw new Error("No ordering metric found");
+    }
+
+    const chainResults = await Promise.all(
+      Object.entries(config.chains).map(async ([chainName, chainConfig]) => {
+        const chainId = (chainConfig as { id: number }).id;
+
+        const chainBlockConfig = await getChainConfig(chainName);
+
+        const latestBlock = await getLatestBlock(chainName);
+
+        const statusBlock = status[chainName]?.block;
+        if (statusBlock === undefined) {
+          throw new Error(`Chain ${chainName} found in config but not in status`);
+        }
+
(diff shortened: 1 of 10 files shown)


apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,398 @@
...
+import { createMiddleware } from "hono/factory";
+import config from "../../../ponder.config.js";
+
+type BlockRef = { number: number; timestamp: number };
+
+type ChainIndexingConfig = {
+  // If all contracts, accounts, and block intervals are set to "latest",
+  // startBlock will be null.
+  startBlock: BlockRef | null;
+  // If all contracts, accounts, and block intervals are set to undefined,
```
lightwalker-eth, line 11:
```suggestion
  // If any contracts, accounts, and block intervals are set to undefined,
```

Is that correct? Appreciate your advice.

lightwalker-eth (commented):
@typedarray Many thanks 👍 Reviewed and shared a few ideas with feedback.

apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,374 @@
...
+
+    return c.json(result);
+  } catch (error) {
+    return c.json({ error: "Internal server error" }, 500);
+  }
+});
+
+const blockCache = new Map<string, Map<number, BlockRef>>();
+
+async function cachedGetBlockByNumber(chainName: string, blockNumber: number) {
```
lightwalker-eth, line 226: Strongly prefer we take a strategy where caching of these results is irrelevant.

Goal: The only RPC calls we should need for this API are to get the block timestamps for the `config` and this should only happen once upon process startup. If this fails upon startup it should throw an error and force the entire process to crash. We must take a strategy where it is guaranteed that zero RPC calls are in the codepath of requests to this API endpoint.

apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,374 @@
...
+    number: Number(block.number),
+    timestamp: Number(block.timestamp),
+  } satisfies BlockRef;
+
+  cache.set(blockNumber, blockRef);
+
+  return blockRef;
+}
+
+async function getChainConfig(chainName: string) {
```
lightwalker-eth, line 251: We should update our strategy such that this is only ever called during process startup. Please see other related comments about when RPC calls related to indexing status are allowed.

apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,374 @@
...
+              config: chainBlockConfig,
+            } satisfies ChainIndexingStatus,
+          };
+        }
+
+        // The backfill end block is equal to the earliest start block
+        // plus the total number of blocks in the backfill.
+        const backfillEndBlockNumber = chainBlockConfig.startBlock.number + historicalTotalBlocks;
+
+        // This will be cached after the first request.
```
lightwalker-eth, line 195: We need to update the strategy here. This value should be calculated once during startup and never dynamically after that. Please see other related comments.

apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,374 @@
...
+            throw new Error(`No ponder_sync_block metric found for realtime chain ${chainName}`);
+          }
+
+          // If the latest sync block number is the same as the latest indexed
+          // block number, use the latest indexed block as an optimization to
+          // avoid the extra RPC request.
+          const latestKnownBlock =
+            latestKnownBlockNumber === latestIndexedBlock.number
+              ? latestIndexedBlock
+              : await cachedGetBlockByNumber(chainName, latestKnownBlockNumber);
```
lightwalker-eth, line 120: We need to update the strategy here. RPC calls are never allowed in the code path of this API endpoint.

We need Ponder to expose through its /metrics endpoint not only the `ponder_sync_block` (block number) for each chain, but also the block time of the associated block on each chain.

Shouldn't Ponder already have this block time? We really need Ponder to expose this data.

apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,374 @@
...
+  startBlock: BlockRef | null;
+  // If all contracts, accounts, and block intervals are set to undefined,
+  // endBlock will be null.
+  endBlock: BlockRef | null;
+};
+
+type ChainIndexingStatus =
+  | {
+      status: "not_started";
+      config: ChainIndexingConfig;
```
lightwalker-eth, line 19: Perhaps we remove this `config` field when status is "not_started"?

Motivation: I'm not sure about possible calls to the API handler before our startup sequence is completed that will build the initial `ChainIndexingConfig` as part of the startup sequence? Please see related comments on this. We need to ensure that ZERO RPC calls are ever in the code path of the indexing status endpoint.

lightwalker-eth (commented):
Shared additional feedback on the important goal of zero RPC calls in the codepath of this API endpoint.

apps/ensindexer/src/api/lib/indexing-status.ts
```diff
@@ -0,0 +1,451 @@
...
+        // the chain has not started yet.
+        if (hasSyncBackfill && historicalCompletedBlocks === 0) {
+          return {
+            chainId,
+            result: {
+              status: "not_started" as const,
+              config: chainConfig,
+            } satisfies ChainIndexingStatus,
+          };
+        }
```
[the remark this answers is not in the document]
tk-o: @typedarray  please confirm that  when
```
ordering === "omnichain" && chainConfig.startBlock.number === statusBlock.number
```
is `true` it is enough to declare the `not_started` indexing status for the chain?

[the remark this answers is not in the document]
typedarray: 1) Confirmed and removed those lines in latest commit
2) Confirmed

PR #894, opened 2025-08-04
feat(ensnode-sdk): shared validations

tk-o:
> Please review on commit-by-commit basis.

This PR refactors ENSNode SDK package by splitting types and functionality from `src/utils` into:
- `src/ens` for ENS definitions,
- `src/shared` for general ENSNode definitions

Also, the PR creates a new way of importing internal modules from ENSNode SDK by using `@ensnode/ensnode-sdk/internal` path. This way, we can share Zod schemas within the monorepo, and avoid having those Zod schemas published into NPM registry.

PR #895, opened 2025-08-04
feat(ensindexer): `/api/config` API endpoint

tk-o:
> Please review on commit-by-commit basis.

This PR builds on top of #894. Two major updates this PR brings:

1. Creates the `src/ensindexer/config` module in ENSNode SDK package which:
    - defines the `ENSIndexerPublicConfig` type and its serialized counterpart, 
    - allows serialization and validated deserialization,
    - includes helpers, such as `isSubgraphCompatible` function useful for validation,
    - defines relevant Zod schemas, such as VersionInfoSchema.

2. Integrates the module from above into the ENSIndexer application:
    - aligns `ENSIndexerConfig`type to enable building `ENSIndexerPublicConfig` type in a straightforward way,
        - reads dependecy info from runtime (Node.js, Ponder), and over the netwrok (ENSRainbow service),
    - derives `indexedChainIds` configuration
    - creates the `GET /api/config` endpoint which returns `SerializedENSIndexerPublicConfig` value,
    - allows printing a redacted `ENSIndexerPublicConfig` into stdout for enhance debugging capabilities.

Other changes made in this PR include fixing tests to work with updated config data model.


apps/ensindexer/src/config/derived-params.ts
```diff
@@ -1,5 +1,39 @@
...
+ * @returns extended configuration
+ */
+export const derive_indexedChainIds = <
+  CONFIG extends Pick<ENSIndexerConfig, "namespace" | "plugins">,
+>(
+  config: CONFIG,
+): CONFIG & { indexedChainIds: ENSIndexerConfig["indexedChainIds"] } => {
+  const indexedChainIds: IndexedChainIds = new Set();
+
+  const datasources = getENSNamespaceAsFullyDefinedAtCompileTime(config.namespace);
```
lightwalker-eth, line 20: Hmm, can you please verify we are still validating that the rpc config for each indexed chain must be defined?

lightwalker-eth (commented):
@tk-o Great work here! 😄 Reviewed and shared a few suggestions

apps/ensindexer/src/config/config.schema.ts
```diff
@@ -160,13 +156,15 @@ const DatabaseUrlSchema = z.union(
...
+const VersionInfoSchema = makeVersionInfoSchema();
+
 const ENSIndexerConfigSchema = z
   .object({
     namespace: ENSNamespaceSchema,
     globalBlockrange: BlockrangeSchema,
     ensNodePublicUrl: EnsNodePublicUrlSchema,
     ensAdminUrl: EnsAdminUrlSchema,
-    ponderDatabaseSchema: PonderDatabaseSchemaSchema,
+    databaseSchemaName: PonderDatabaseSchemaSchema,
```
shrugs, line 167: what do you think about calling this just `databaseSchema`?

apps/ensindexer/src/config/index.ts
```diff
@@ -1,11 +1,17 @@
-import { buildConfigFromEnvironment } from "@/config/config.schema";
+import { EnsRainbowEndpointUrlSchema, buildConfigFromEnvironment } from "@/config/config.schema";
 import { ENSIndexerEnvironment } from "@/config/types";
 import { getRpcConfigsFromEnv } from "@/lib/lib-config";
+import { getENSRainbowVersionInfo, getPackageVersion } from "@/lib/version-info";
+
+// Fetch ENSRainbow Version info
+const ensRainbowEndpointUrl = EnsRainbowEndpointUrlSchema.parse(process.env.ENSRAINBOW_URL);
```
shrugs, line 7: i don't love the top-level fetch call introduced here that creates a synchronous dependency on ENS rainbow service to build a config (even though it's required by the application during indexing).

Could we instead remove the `versionInfo` section from the `ENSIndexerConfig` object and instead include this information at request-time within the `/api/config` api handler? we can still use zod to validate the input like we're doing here, just not including it with the synchronous construction of the `ENSIndexerConfig` used by ENSIndexer itself. so response type of `/api/config` is like `SerializedENSIndexerPublicConfig & { versionInfo: SerializedVersionInfo }`

if going with that approach, we can either make the request at the top level in the API handler file (so it's still a sync dependency at startup), or what's perhaps more clean is to fetch it upon request to `/api/config` and then cache the results for the duration of the process.

apps/ensindexer/src/config/types.ts
```diff
@@ -69,18 +69,18 @@ export interface ENSIndexerConfig {
    * Invariant:
    * - The URL must be a valid URL. localhost urls are allowed (and expected).
    */
-  ensRainbowEndpointUrl: string;
+  ensRainbowEndpointUrl: URL;
```
shrugs, line 72: `EndpointUrl` is redundant, let's use `ensRainbowUrl` to match the existing pattern (`ensAdminUrl`, `ensNodePublicUrl`, etc)

apps/ensindexer/src/lib/lib-config.ts
```diff
@@ -47,20 +48,22 @@ export function getRpcConfigsFromEnv(): Record<number, RpcConfigEnvironment> {
 }
 
 export function prettyPrintConfig(config: ENSIndexerConfig) {
+  const serializedConfig = serializeENSIndexerPublicConfig(config);
```
shrugs, line 51: it looks all of the relevant properties are included in the public config, but it might be better to have an additional serialization function for the internal config (which can of course call `serializeENSIndexerPublicConfig`, but also explicitly specify any additional properties) just so there's no change of future properties being unintentionally not-printed at startup.

then it ight make sense to refactor this obfuscation logic into an `redactENSIndexerConfig` function that is more or less just

```ts
export function redactENSIndexerConfig(config: ENSIndexerConfig) {
  return {
    ...config,
    databaseUrl: '****',
    rpcConfigs: // .... etc
  }
}
```
and finally refactor the replacer into a `ENSIndexerConfigReplacer` function that truncates `abi` entries as already implemented.

shrugs: so this function looks like:

```ts
return JSON.stringify(redactENSIndexerConfig(serializeENSIndexerConfig(config), ENSIndexerConfigReplacer);
```

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { getENSRainbowApiCLient } from "@/lib/ensraibow-api-client";
+
+/**
+ * Get NPM package version.
+ */
+export function getPackageVersion(packageName: string) {
```
shrugs, line 16: correct me if i'm wrong but can this be replaced with `import packageJson from "@/../package.json";` and then reading `packageJson.version`?

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -0,0 +1,68 @@
...
+    console.warn(`Could not find version for ${packageName}:`, errorMessage);
+
+    return "unknown";
+  }
+}
+
+/**
+ * Get Version Info from ENSRainbow service.
+ */
+export async function getENSRainbowVersionInfo(ensRainbowEndpointUrl: URL) {
```
shrugs, line 64: if we end up resolving ensrainbow version info in the config api handler then this helper can implement this fetch-once-and-cache-forever behavior for this value

apps/ensindexer/test/mocks/http-server.ts
```diff
@@ -0,0 +1,17 @@
...
+import { http, HttpResponse } from "msw";
+import { setupServer } from "msw/node";
+
+export const server = setupServer(
+  /**
+   * Mock HTTP response for ENSRainbow.
+   *
+   * It is required for during the ENSIndexer bootstrap phase.
+   */
+  http.get("https://customrainbow.com/v1/version", () => {
```
shrugs, line 11: instead of using `msw`, it may be preferrable to use vitest's built-in mocks, what that approach tried?

see `packages/ensrainbow-sdk/src/client.test.ts` for an example of stubbing `fetch` and replacing responses with the appropriate expectations

shrugs: i do like how msw integrates at the vitest top-level and allows us to more generally mock external services, but i think per-test mocking is likely the correct approach for now, wdyt?

shrugs (commented):
ENSIndexerConfig redefines many properties already implemented by `ENSIndexerPublicConfig`. It seems to me that the goal for that is to add additional internal-focused comments — in this case i support the idea of adding a note to the `ENSIndexerConfig` type that specifies that that's what's going on, because otherwise it seems redundant.

it could also make sense to, instead of extending `ENSIndexerPublicConfig` to just keep them as two separate types that happen to overlap significantly, rather than extending and then re-defining every property

apps/ensindexer/src/lib/version-info.ts
tk-o: @shrugs since we use PNPM's `catalog:` references, reading just read from the `package.json` file will give us `catalog:` values, and not resolved version values.

apps/ensindexer/test/mocks/http-server.ts
tk-o: We could still use MSW to have a precise way to mock network responses on per-test level. Will update that approach 👍

apps/ensindexer/src/lib/graphnode-helpers.ts
```diff
@@ -1,20 +1,9 @@
...
-if (
-  ensRainbowApiClient.getOptions().endpointUrl === EnsRainbowApiClient.defaultOptions().endpointUrl
-) {
-  console.warn(
-    `Using default public ENSRainbow server which may cause increased network latency.
-    For production, use your own ENSRainbow server that runs on the same network
-    as the ENSIndexer server.`,
-  );
-}
+const ensRainbowApiClient = getENSRainbowApiCLient(new URL(config.ensRainbowEndpointUrl));
```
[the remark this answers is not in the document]
tk-o: Good catch 👍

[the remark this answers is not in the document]
tk-o: Will do 👍

apps/ensindexer/src/lib/ensraibow-api-client.ts
```diff
@@ -0,0 +1,46 @@
+import { UrlString } from "@ensnode/ensnode-sdk";
+import { EnsRainbowApiClient } from "@ensnode/ensrainbow-sdk";
+
+// Cache to store client instances by URL
+const clientCache = new Map<UrlString, EnsRainbowApiClient>();
```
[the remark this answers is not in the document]
tk-o: The goal was to cover the scenario someone would use `getENSRainbowApiClient` method with URL value other than `config.ensRainbowEndpointUrl`. In this case, there would be two clients, with two distinct in-memory caches.

I can remove this `clientCache` layer, but also thought it was better to be safe with ensuring the ENSNode operator would see the warning.

packages/ensnode-sdk/src/ensindexer/config/types.ts
```diff
@@ -0,0 +1,167 @@
...
+   * unknown labelhashes.
+   * @see https://ensnode.io/ensrainbow/overview/what-is-ensrainbow
+   *
+   * For best performance, ENSRainbow should be colocated with ENSIndexer and use private/internal
+   * networking to minimize latency.
+   *
+   * Invariant:
+   * - The URL must be a valid URL. localhost urls are allowed (and expected).
+   */
+  ensRainbowEndpointUrl: URLType;
```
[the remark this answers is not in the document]
tk-o: Will do 👍

packages/ensnode-sdk/src/ensindexer/config/types.ts
```diff
@@ -0,0 +1,167 @@
...
+   */
+  databaseSchemaName: string;
+
+  /**
+   * A set of {@link PluginName}s indicating which plugins to activate.
+   *
+   * Invariants:
+   * - A set of valid {@link PluginName}s with at least one value
+   * - For each plugin, its required datasources must be defined within the {@link namespace}
+   * - For each plugin specified, a valid {@link rpcConfigs} entry is required for
```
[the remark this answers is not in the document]
tk-o: Correct, will make the update soon 👍

apps/ensindexer/src/lib/lib-config.ts
```diff
@@ -47,20 +48,22 @@ export function getRpcConfigsFromEnv(): Record<number, RpcConfigEnvironment> {
 }
 
 export function prettyPrintConfig(config: ENSIndexerConfig) {
```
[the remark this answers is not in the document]
tk-o: Yes, the goal was to work with the `ENSIndexerConfig` value to present it in the CLI output.

apps/ensindexer/src/lib/lib-config.ts
```diff
@@ -47,20 +48,22 @@ export function getRpcConfigsFromEnv(): Record<number, RpcConfigEnvironment> {
...
           {
             ...rpcConfig,
-            url: "*******",
+            url: new URL("/*******", rpcConfig.url.href),
           },
         ]),
       ),
-    } as ENSIndexerConfig,
+    },
     (key: string, value: unknown) => (key === "abi" ? `(truncated ABI output)` : value),
```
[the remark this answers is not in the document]
tk-o: Will drop the type cast 👍

apps/ensindexer/src/config/types.ts
```diff
@@ -28,7 +28,7 @@ export interface RpcConfig {
 /**
  * The complete runtime configuration for an ENSIndexer instance.
  */
-export interface ENSIndexerConfig {
+export interface ENSIndexerConfig extends ENSIndexerPublicConfig {
```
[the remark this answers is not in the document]
tk-o: The comments between `ENSIndexerConfig` and `ENSIndexerPublicConfig` differ. 

We need to be able to turn an `ENSIndexerConfig` object into an `ENSIndexerPublicConfig` object, but not vice-versa.

Perhaps I should update:
```
interface ENSIndexerConfig extends ENSIndexerPublicConfig {
```
to be just:
```
interface ENSIndexerConfig {
```

In this way, we don't signal any relationship between those two types, and have freedom on defining docs on each of them separately.

apps/ensindexer/src/config/index.ts
tk-o: If we assume no validations are required to version info, for example, checking if connected ENSRainbow API version matches the version range expected by the ENSIndexer, then we can exclude the `versionInfo` prop from `ENSIndexerConfig`.

In this case, we can asynchronously fetch version info from ENSRainbow once, and then await on the returned promise from inside of the `/api/config` route handler.

apps/ensindexer/src/lib/lib-config.ts
tk-o: Great suggestion :)

apps/ensindexer/src/lib/ensraibow-api-client.ts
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Then suggest removing any possibility of calling `getENSRainbowApiClient` with a custom URL value.

There should be a single ENSRainbowApiClient instance created across the entire lifecycle of an ENSIndexer process. This single instance can then be used where needed across the process.

apps/ensindexer/src/lib/lib-config.ts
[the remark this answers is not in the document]
lightwalker-eth: Ok cool, thanks for the correction 👍

apps/ensindexer/src/config/types.ts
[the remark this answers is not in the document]
lightwalker-eth: I like the idea to make these two separate objects without any inheritance relationship 👍 

@tk-o Then you can create a function that takes as input an `ENSIndexerConfig` and returns back an `ENSIndexerPublicConfig`.

apps/ensindexer/src/lib/version-info.ts
lightwalker-eth: @tk-o Good note here! Could you please document this helpful background info on the `getPackageVersion` function? Thanks 🙌

apps/ensindexer/src/lib/graphnode-helpers.ts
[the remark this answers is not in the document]
tk-o: I've double checked, and the `ENSRAINBOW_URL` has always been required, so no need to make a changeset here.

apps/ensindexer/src/config/types.ts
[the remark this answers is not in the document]
tk-o: I think we can create a `buildENSIndexerPublicConfig` function that does not take any input, just reads directly from `import config from "@/config";`. I don't think there's a use case that would require passing `ENSIndexerConfig` value as input.

apps/ensindexer/src/config/types.ts
```diff
@@ -230,4 +230,10 @@ export interface ENSIndexerEnvironment {
     endBlock: string | undefined;
   };
   rpcConfigs: Record<number, RpcConfigEnvironment>;
+  versionInfo: {
```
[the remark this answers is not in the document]
tk-o: Following feedback from @shrugs, I'll move fetching version info away from the config, and will make a helper to extend `config` value with `versionInfo` prop when serialization is needed (and version info data is needed too).

apps/ensindexer/src/lib/lib-config.ts
```diff
@@ -45,23 +46,3 @@ export function getRpcConfigsFromEnv(): Record<number, RpcConfigEnvironment> {
...
-            ...rpcConfig,
-            url: "*******",
-          },
-        ]),
-      ),
-    } as ENSIndexerConfig,
-    (key: string, value: unknown) => (key === "abi" ? `(truncated ABI output)` : value),
-    2,
-  );
-}
```
tk-o, line 67: Move to `apps/ensindexer/src/config/helpers.ts`

apps/ensindexer/src/lib/ponder-helpers.ts
```diff
@@ -45,9 +45,7 @@ export const constrainBlockrange = (
  *
  * @returns A function that fetches ENSRainbow version information
  */
-export const createEnsRainbowVersionFetcher = (
-  endpointUrl: ENSIndexerConfig["ensRainbowEndpointUrl"],
-) => {
+export const createEnsRainbowVersionFetcher = (endpointUrl: ENSIndexerConfig["ensRainbowUrl"]) => {
   const client = new EnsRainbowApiClient({ endpointUrl: new URL(endpointUrl) });
```
lightwalker-eth, line 49: Was confused reading this:

1. Why do we need to build a function that returns a function to achieve the goal of checking the ENSRainbow version response?
2. Why is this function in a ponder-helpers.ts file? It has nothing to do with Ponder.
3. Is there a special reason we couldn't make use of the EnsRainbowApiClient we use everywhere else (same singleton?)
4. Why is the param that's passed to this function of type `ENSIndexerConfig["ensRainbowUrl"]`? Why not just `URL`?
5. Why are we calling `new URL(endpointUrl)` when the endpointUrl value should already be of type URL?

apps/ensindexer/src/lib/ponder-metadata-provider.ts
```diff
@@ -20,7 +20,7 @@ export const fetchFirstBlockToIndexByChainId = createFirstBlockToIndexByChainIdF
 );
 
 // setup ENSRainbow version fetching
-export const fetchEnsRainbowVersion = createEnsRainbowVersionFetcher(config.ensRainbowEndpointUrl);
+export const fetchEnsRainbowVersion = createEnsRainbowVersionFetcher(config.ensRainbowUrl);
```
lightwalker-eth, line 23: What plan do you suggest for how we ultimately clean up the old strategies for ponder-metadata? Is that something we should do in our existing PRs or something we do in another future PR? Appreciate your advice

apps/ensindexer/src/lib/ensraibow-api-client.ts
```diff
@@ -1,23 +1,12 @@
...
  */
-export function getENSRainbowApiCLient(ensRainbowEndpointUrl: URL) {
-  if (clientCache.has(ensRainbowEndpointUrl.href)) {
-    return clientCache.get(ensRainbowEndpointUrl.href)!;
-  }
-
+export function getENSRainbowApiClient() {
   const ensRainbowApiClient = new EnsRainbowApiClient({
-    endpointUrl: new URL(ensRainbowEndpointUrl),
+    endpointUrl: new URL(config.ensRainbowUrl),
```
lightwalker-eth, line 9: Why are we calling `new URL` here? Isn't `config.ensRainbowUrl` already a URL?

apps/ensindexer/src/lib/dependency-info.ts
```diff
@@ -59,10 +68,25 @@ export function getPackageVersion(packageName: string) {
 }
 
 /**
- * Get Version Info from ENSRainbow service.
+ * Get complete Version info for ENSIndexer app.
```
lightwalker-eth, line 71:
```suggestion
 * Get complete DependencyInfo for ENSIndexer.
```

apps/ensindexer/src/config/types.ts
```diff
@@ -151,9 +156,20 @@ export interface ENSIndexerConfig extends ENSIndexerPublicConfig {
...
+  /**
+   * Indexed Chain IDs
+   *
+   * Includes Chain ID for each chain being indexed.
+   *
+   * Invariants:
+   * - No duplicates
+   * - Only positive integer values matching {@link ChainId}
+   */
+  indexedChainIds: IndexedChainIds;
```
lightwalker-eth, line 172: Is there a special reason to keep the `IndexedChainIds` type definition?

Overall I think it would be more clear to just write `Set<ChainId>`.

I had to open up my IDE to figure out what the type of `IndexedChainIds` actually was. It wasn't convenient.

Additionally, since this is a `Set<ChainId>` we should not be documenting invariants here such as:
   * - No duplicates
   * - Only positive integer values matching {@link ChainId}

These invariants are already self-documenting. We know no duplicates because it is a `Set`. And we know the values must be positive integer values based on the definition of `ChainId`.

apps/ensindexer/src/config/helpers.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/**
+ * Pretty print {@link ENSIndexerConfig} object in a non-blocking way.
+ *
+ * Invariant:
+ * - All sensitive values are redacted before printing a string representation
+ *   of the config into stdout.
+ */
+export async function prettyPrintConfig(config: ENSIndexerConfig): Promise<void> {
```
lightwalker-eth, line 90:
```suggestion
export async function prettyPrintRedactedConfig(config: ENSIndexerConfig): Promise<void> {
```

packages/ensnode-sdk/src/ensindexer/config/types.ts
```diff
@@ -143,7 +151,7 @@ export interface ENSIndexerPublicConfig<URLType = URL, IndexedChainIdsType = Ind
    *
    * Invariants:
    * - No duplicates
-   * - Each key (chain id) must be a number
+   * - Only positive integer values matching {@link ChainId}
    */
   indexedChainIds: IndexedChainIdsType;
```
lightwalker-eth, line 156: Please see my other related comment on this field.

Key suggestions: Change to `Set<ChainId>` and refine documentation of invariants.

apps/ensindexer/src/config/types.ts
```diff
@@ -151,9 +156,20 @@ export interface ENSIndexerConfig extends ENSIndexerPublicConfig {
    * Configuration for each indexable RPC, keyed by chain id.
    *
    * Invariants:
```
lightwalker-eth, line 158: We should also document a new invariant here that each value in `indexedChainIds` is guaranteed to have a value here too.

Can you please also confirm and document the opposite case? As I understand it is possible to have rpcConfig values for a chainId that are not in the set of `indexedChainIds`? If so I think that's ok and maybe has a meaningful purpose based on some of the other goals in our upcoming roadmap. Appreciate your confirmation and advice.

lightwalker-eth (approved):
@tk-o Thanks for updates! Reviewed and shared a few suggestions 👍

apps/ensindexer/src/config/types.ts
```diff
@@ -151,9 +156,20 @@ export interface ENSIndexerConfig extends ENSIndexerPublicConfig {
...
-   * - Each key (chain id) must be a number
+   * - Each key (chain id) must be a {@link ChainIdString} value.
    */
-  rpcConfigs: Record<number, RpcConfig>;
+  rpcConfigs: Record<ChainIdString, RpcConfig>;
```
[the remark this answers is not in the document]
tk-o: Good point, will update the type and its references 👍

apps/ensindexer/src/lib/ponder-helpers.ts
tk-o: This is a previously existing function that powers `/metadata` endpoint. Once we introduce `/api/config` and `/api/indexing-status` endpoints, we should deprecate the `/metadata` one, and later remove it with all its dependencies, such as `apps/ensindexer/src/lib/ponder-metadata-provider.ts` and half of `apps/ensindexer/src/lib/ponder-helpers.ts`.

Before we can remove the old code, I'm happy to update it to use the newly introduced EnsRainbowApiClient singleton 👍 At the same time, I'd like to stay away from refactoring these files:
- `ponder-helpers.ts`,
-  `ponder-metadata-provider.ts`.
There's a risk of regression for `/metadata` endpoint output, so it's wise to not tweak much in those at the moment when we need to focus on other topics.

@lightwalker-eth what do you think?

apps/ensindexer/src/lib/ponder-metadata-provider.ts
tk-o: As for the following files:
- `apps/ensindexer/src/lib/ponder-helpers.ts`,
- `apps/ensindexer/src/lib/ponder-metadata-provider.ts`.

We can either update them (most of updates to be expected in #896 as it has more APIs to offer replacements), or do nothing and plan to 80% of functionality these files host currently. More details in [this comment](https://github.com/namehash/ensnode/pull/895#discussion_r2255986866).

apps/ensindexer/src/lib/ensraibow-api-client.ts
tk-o: Correct, no need to call extra constructor 👍

packages/ensnode-sdk/src/ensindexer/config/types.ts
tk-o: The main reason for using `IndexedChainIdsType` was the idea to share documentation of type properties between `ENSIndexerPublicConfig` and `SerializedENSIndexerPublicConfig`.

If we use concrete native types, such as `Set<ChainId>` or `URL` on `ENSIndexerPublicConfig` type, we have to update the `SerializedENSIndexerPublicConfig` to use serialized representation of the native types.

The update could look like this:
```diff
diff --git a/packages/ensnode-sdk/src/ensindexer/config/serialize.ts b/packages/ensnode-sdk/src/ensindexer/config/serialize.ts
index 3c340543..792fe546 100644
--- a/packages/ensnode-sdk/src/ensindexer/config/serialize.ts
+++ b/packages/ensnode-sdk/src/ensindexer/config/serialize.ts
@@ -1,13 +1,12 @@
+import { ChainId } from "../../shared";
 import { serializeUrl } from "../../shared/serialize";
 import { SerializedENSIndexerPublicConfig, SerializedIndexedChainIds } from "./serialized-types";
-import { ENSIndexerPublicConfig, IndexedChainIds } from "./types";
+import { ENSIndexerPublicConfig } from "./types";
 
 /**
  * Serializes a {@link ChainConfig} object.
  */
-export function serializeIndexedChainIds(
-  indexedChainIds: IndexedChainIds,
-): SerializedIndexedChainIds {
+export function serializeIndexedChainIds(indexedChainIds: Set<ChainId>): SerializedIndexedChainIds {
   return Array.from(indexedChainIds);
 }
 /**
diff --git a/packages/ensnode-sdk/src/ensindexer/config/serialized-types.ts b/packages/ensnode-sdk/src/ensindexer/config/serialized-types.ts
index cbc3ee7d..192ca059 100644
--- a/packages/ensnode-sdk/src/ensindexer/config/serialized-types.ts
+++ b/packages/ensnode-sdk/src/ensindexer/config/serialized-types.ts
@@ -6,7 +6,28 @@ export type SerializedIndexedChainIds = Array<ChainId>;
 /**
  * Serialized representation of {@link ENSIndexerPublicConfig}
  */
-export type SerializedENSIndexerPublicConfig = ENSIndexerPublicConfig<
-  UrlString,
-  SerializedIndexedChainIds
->;
+export interface SerializedENSIndexerPublicConfig
+  extends Omit<
+    ENSIndexerPublicConfig,
+    "ensAdminUrl" | "ensNodePublicUrl" | "ensRainbowUrl" | "indexedChainIds"
+  > {
+  /**
+   * String representation of {@link ENSIndexerPublicConfig.ensAdminUrl}.
+   */
+  ensAdminUrl: UrlString;
+
+  /**
+   * String representation of {@link ENSIndexerPublicConfig.ensNodePublicUrl}.
+   */
+  ensNodePublicUrl: UrlString;
+
+  /**
+   * String representation of {@link ENSIndexerPublicConfig.ensRainbowUrl}.
+   */
+  ensRainbowUrl: UrlString;
+
+  /**
+   * Array representation of {@link ENSIndexerPublicConfig.indexedChainIds}.
+   */
+  indexedChainIds: ChainId[];
+}
diff --git a/packages/ensnode-sdk/src/ensindexer/config/types.ts b/packages/ensnode-sdk/src/ensindexer/config/types.ts
index d088e12b..cd932959 100644
--- a/packages/ensnode-sdk/src/ensindexer/config/types.ts
+++ b/packages/ensnode-sdk/src/ensindexer/config/types.ts
@@ -31,18 +31,13 @@ export interface DependencyInfo {
   ensRainbowSchema: number;
 }
 
-/**
- * Indexed Chain IDs
- */
-export type IndexedChainIds = Set<ChainId>;
-
 /**
  * Complete public configuration object for ENSIndexer.
  *
  * We use parameter types to maintain fields layout and documentation across
  * the domain model and its serialized counterpart.
  */
-export interface ENSIndexerPublicConfig<URLType = URL, IndexedChainIdsType = IndexedChainIds> {
+export interface ENSIndexerPublicConfig {
   /**
    * The ENS namespace that ENSNode operates in the context of.
    *
@@ -58,11 +53,8 @@ export interface ENSIndexerPublicConfig<URLType = URL, IndexedChainIdsType = Ind
    * identified by {@link ensNodePublicUrl}.
    *
    * @see https://ensnode.io/ensadmin/overview/what-is-ensadmin
-   *
-   * Invariants:
-   * - The URL must be a valid URL (localhost urls are allowed)
    */
-  ensAdminUrl: URLType;
+  ensAdminUrl: URL;
 
   /**
    * The publicly accessible endpoint of the ENSNode API
@@ -70,11 +62,8 @@ export interface ENSIndexerPublicConfig<URLType = URL, IndexedChainIdsType = Ind
    *
    * ENSAdmin will use this url to connect to the ENSNode api for querying
    * state about the ENSNode instance.
-   *
-   * Invariants:
-   * - The URL must be a valid URL (localhost urls are allowed)
    */
-  ensNodePublicUrl: URLType;
+  ensNodePublicUrl: URL;
 
   /**
    * An ENSRainbow API Endpoint (ex: http://localhost:3223). ENSIndexer uses
@@ -87,7 +76,7 @@ export interface ENSIndexerPublicConfig<URLType = URL, IndexedChainIdsType = Ind
    * Invariant:
    * - The URL must be a valid URL. localhost urls are allowed (and expected).
    */
-  ensRainbowUrl: URLType;
+  ensRainbowUrl: URL;
 
   /**
    * A Postgres database schema name. This instance of ENSIndexer will write
@@ -150,10 +139,9 @@ export interface ENSIndexerPublicConfig<URLType = URL, IndexedChainIdsType = Ind
    * Includes Chain ID for each chain being indexed.
    *
    * Invariants:
-   * - No duplicates
    * - Only positive integer values matching {@link ChainId}
    */
-  indexedChainIds: IndexedChainIdsType;
+  indexedChainIds: Set<ChainId>;
 
   /**
    * A flag derived from the built config indicating whether ENSIndexer is operating in a
```

apps/ensindexer/src/config/types.ts
tk-o: Right, it's possible to have rpcConfig values for the chain IDs were not included in `indexedChainIds`. Will document that 👍

apps/ensindexer/src/lib/ponder-helpers.ts
tk-o: I just discovered, that `ponder-helpers.ts` file is included while the `ENSIndexerConfig` object is being built. This  requires `ponder-helpers.ts` file and all its dependencies not to import from `@/config` path.

apps/ensindexer/src/lib/ponder-metadata-provider.ts
tk-o: I shared a suggestion on Slack and we agreed to follow it:
>  I suggest we minimize the scope of changes in those files to prevent regression risk that could affect our current `/metadata` endpoint.

apps/ensindexer/src/lib/ponder-metadata-provider.ts
```diff
@@ -20,7 +21,23 @@ export const fetchFirstBlockToIndexByChainId = createFirstBlockToIndexByChainIdF
...
+      schema_version: versionResponse.versionInfo.schema_version,
+    };
+  } catch (error) {
+    console.error("Failed to fetch ENSRainbow version", error);
+    return {
+      version: "unknown",
+      schema_version: 0,
+    };
+  }
+};
```
tk-o, line 40: Moved from `ponder-helpers.ts`.

apps/ensindexer/test/lib-helpers.test.ts
```diff
@@ -1,13 +1,7 @@
-import { bigintMax, hasNullByte, stripNullBytes, uniq } from "@/lib/lib-helpers";
+import { bigintMax, hasNullByte, stripNullBytes } from "@/lib/lib-helpers";
 import { describe, expect, it } from "vitest";
 
 describe("helpers", () => {
-  describe("uniq", () => {
-    it("should return unique elements from an array", () => {
-      expect(uniq([1, 2, 2, 3, 4, 4])).toEqual([1, 2, 3, 4]);
-    });
-  });
```
tk-o, line 9: Covered in ENSNode SDK: `collections.ts`

apps/ensindexer/src/config/types.ts
```diff
@@ -155,21 +149,25 @@ export interface ENSIndexerConfig {
   /**
    * Configuration for each indexable RPC, keyed by chain id.
    *
+   * Note: There may be some {@link RpcConfig} values preset for
+   * the chain IDs that are not included in {@link indexedChainIds}.
+   *
    * Invariants:
-   * - Each key (chain id) must be a {@link ChainIdString} value.
+   * - Each key must be a {@link Chain} value.
```
lightwalker-eth, line 156: We don't need to explicitly document invariants that are implied already by the type system.

lightwalker-eth (approved):
@tk-o Great updates! Shared a few final suggestions. Please feel welcome to merge when you feel ready 👍

PR #904, opened 2025-08-08
deps: address security issues

tk-o:
Resolves:
- https://github.com/namehash/ensnode/security/dependabot/30
- https://github.com/namehash/ensnode/security/dependabot/31


shrugs:
good diff 👍

PR #907, opened 2025-08-08
fix(ponder-subgraph): fix package exports

volkyeth:
tsup config was incompatible with the declared package export

### DIFF
--- a/packages/ponder-subgraph/package.json
+++ b/packages/ponder-subgraph/package.json
@@ -15,7 +15,7 @@
     "Subgraph"
   ],
   "exports": {
-    ".": "./src/index.ts"
+    ".": "./dist/index.js"
   },
   "files": [
     "dist"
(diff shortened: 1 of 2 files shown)


packages/ponder-subgraph/package.json
```diff
@@ -15,7 +15,7 @@
     "Subgraph"
   ],
   "exports": {
-    ".": "./src/index.ts"
+    ".": "./dist/index.js"
```
shrugs, line 18: this change should be reverted, as it defines the esm-specific import, which should be the index.ts file

volkyeth:
I initially thought I would be able to use this package for subgraph compatibility on my own ponder API since I found it on npmjs, but after digging a bit deeper I found out that it's tightly coupled to ensnode specifics.

I might do some work to decouple it for external usage in the future. lmk if you'd be interested in a contribution.

lightwalker-eth:
@volkyeth Sure, such a contribution would be very welcome 👍

PR #915, opened 2025-08-14
feat: replace wagmi and rpc configs with just ENSNode React SDK

shrugs:
also removes dependency on `RPC_URL`s — going forward ENSAdmin should only need a connection to ENSIndexer in order to power its entire featureset


apps/ensadmin/src/components/identity/index.tsx
```diff
@@ -39,32 +39,19 @@ export function Identity({
...
-  const {
-    data: ensName,
-    isLoading,
-    isError,
-  }: UseEnsNameReturnType<string> = useEnsName({
+  // Lookup the primary name for address using ENSNode
+  const { data, status } = usePrimaryName({
     address,
-    chainId: ensRootChainId,
+    chainId: 1,
```
tk-o, line 45: What if we need to display data from Sepolia? I reckon we should still use the `ensRootChainId` as before. Please let me know what you think, @shrugs?

lightwalker-eth (approved):

Issue #936, opened 2025-08-19
Refine Environment Variable Documentation


lightwalker-eth:
# Background

In our v0.33.0 release we included a new environment variable for ENSIndexer that in theory should be required: `ENSINDEXER_URL`.

When we deployed v0.33.0, we accidentally forgot to include configuration of the `ENSINDEXER_URL` environment variable across our hosted instances of ENSIndexer. Therefore the expected outcome would be each ENSIndexer instance failing to start due to a failure to build an `ENSIndexerConfig` that passes all of our required invariants. However, the actual outcome is that this expected invariant check didn't result in any errors.

# Goals

Enforce that ENSIndexer will refuse to start (as part of its config building and config invariant enforcement) if the `ENSINDEXER_URL` environment variable is not defined or if it is not a valid URL.

# Subtasks

- [x] Enforce the goal described above.
- [x] Refine the comments in `apps/ensindexer/.env.local.example` such that we make it explicit which fields are required and which are optional

# Side quests

Suggesting these in this issue because they are small tasks and involve related files.

- [x] Refine documentation for `INDEX_ADDITIONAL_RESOLVER_VALUES` in `apps/ensindexer/.env.local.example` such that we explicitly identify how this must be activated if you wish to use protocol acceleration.
- [x] Refine documentation for `PORT` in the same file. Doesn't this have a default value? This default value should be documented inside the `.env.local.example` file.

PR #947, opened 2025-08-20
Fix/add ensindexer variable to render

BanaSeba:
Added ENSINDEXER_URL to Render Terraform: https://github.com/namehash/ensnode/issues/936
Added a flag for enforcing deployment to Render on service update: https://github.com/namehash/ensnode/issues/936


lightwalker-eth (approved):
@BanaSeba Thanks

PR #951, opened 2025-08-20
status page mocks

notrab:


shrugs (approved):
approved for velocity, pending changes can be opened in a future pr

PR #970, opened 2025-08-25
Feat/tokenscope

lightwalker-eth:
Replaces https://github.com/namehash/ensnode/pull/872 and closes https://github.com/namehash/ensnode/issues/832

- [ ] Testing


packages/ensnode-schema/src/tokenscope.schema.ts
```diff
@@ -0,0 +1,116 @@
...
+     * - ETH/WETH: Amount in wei (1 ETH = 10^18 wei)
+     * - USDC: Amount in micro-units (1 USDC = 10^6 units)
+     * - DAI: Amount in wei-equivalent (1 DAI = 10^18 units)
+     */
+    amount: t.bigint().notNull(),
+
+    /**
+     * Unix timestamp of the block timestamp when the sale occurred.
+     */
+    timestamp: t.integer().notNull(),
```
shrugs, line 107: Both `timestamp` and `blockNumber` are integer here when they could/should be `bigint` as we do elsewhere in the schema — is this important?

packages/ensnode-schema/src/tokenscope.schema.ts
```diff
@@ -0,0 +1,116 @@
...
+     * Combined with `chainId` and `contractAddress`, creates unique NFT identifier
+     * for the domain being transferred.
+     *
+     * Interpretation depends on 'tokenType':
+     * - ERC721: Unique token within contract
+     * - ERC1155: Token type identifier (multiple copies may exist)
+     *
+     * The tokenId is formatted as a hex string representing a uint256.
+     */
+    tokenId: t.text().notNull(),
```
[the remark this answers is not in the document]
lightwalker-eth: Agreed, thanks.

shrugs (approved):

apps/ensindexer/src/lib/currencies.ts
```diff
@@ -169,19 +138,14 @@ const getSupportedCurrencyContractsForNamespace = (
...
-  for (const [currencyId, supportedCurrencyContract] of Object.entries(
-    supportedCurrencyContracts,
-  )) {
-    if (accountIdEqual(supportedCurrencyContract, contract)) {
-      return currencyId as CurrencyId;
-    }
-  }
-
-  return null;
+export const getCurrencyIdForContract = (contract: AccountId): CurrencyId | null => {
```
lightwalker-eth, line 139: For strict correctness it seems we should:
1. Keep a namespace param on `getCurrencyIdForContract`.
2. Update the data model for `KNOWN_CURRENCY_CONTRACTS` so that each record includes its associated namespace.
3. Update `getSupportedCurrencyContractsForChain` so that it also takes a namespace param.
   1. Update logic in this function so that it filters by namespace in addition to the existing chain filtering logic.

lightwalker-eth (commented):
Super updates and fixed 👍 Just 1 point of feedback that I'll add a reference to in https://github.com/namehash/ensnode/issues/972 for the future

PR #997, opened 2025-08-27
Replace Unnormalized

### LINKED ISSUE #846: Replace unnormalized
Official Deliverable Description:

> Ship optional prevention of unnormalized ENS name values from being returned in API requests (returning “encoded labelhashes” instead, improving security and reducing ENS developer complexity).

# Background Info

Partial list of background info:

- Special care is needed for handling null bytes in indexed names / labels. See related issues that have been closed in favor of this issue:
   - https://github.com/namehash/ensnode/issues/45
   - https://github.com/namehash/ensnode/issues/830

# Subtasks

- [x] Add a new optional environment variable to ENSIndexer: `REPLACE_UNNORMALIZED`. Boolean value. Default to `true`.
   - [x] Update `ENSIndexerEnvironment`
   - [x] Update `SerializedENSIndexerConfig`
   - [x] Update `ENSIndexerPublicConfig`
   - [x] Update all related Zod parsing and validation logic.
   - [x] Provide detailed documentation in `apps/ensindexer/.env.local.example`.
      - [x] Identify as optional.
      - [x] Identify default value.
      - [x] Identify that this value must be set to false for full subgraph backwards compatibility.
   - [x] Update the `isSubgraphCompatible` function, such that if this configuration is true then the function returns false.
   - [x] Update the `indexingBehaviorDependencies` as found [here](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/ponder/config.ts#L21-L31) such that Ponder will refuse to start indexing into database_schema X if it was previously indexing into the same database_schema with a different configuration.
  - [x] Update the ENSAdmin UI for displaying an ENSIndexerPublicConfig to include a display of this field.
  - [x] Enforce the "replace unnormalized" configuration in indexing logic, such that the guarantee of `REPLACE_UNNORMALIZED` lives at the ENSDb level for all name and label fields and not just in the responses of our APIs.
     - [x] For the domain table I believe we can achieve this by simply updating the `isLabelIndexable` function so that after the existing checks are completed we also check that the label `isNormalized`.
     - [x] Please suggest a solution for indexed `name` resolver record values that are used in primary name lookups. My first idea is if we encounter a `name` resolver record during indexing where `isNormalized` returns false then we instead update the indexed value of that `name` resolver record value to `null` (simulating that it is not set). I believe a related function is `sanitizeNameRecordValue` but please note how the `sanitizeNameRecordValue` function was merged in [PR 902](https://github.com/namehash/ensnode/pull/902) and that there are a number of open comments and action items in comments on PR 902, including [this related comment](https://github.com/namehash/ensnode/pull/902#discussion_r2279146625).
     - [x] Are there any other places where a name or label value is recorded in our indexed data model? If so, please suggest how we implement our guarantees.
   - [x] Enforce the "replace unnormalized" configuration in our resolution APIs, such that the guarantee of `REPLACE_UNNORMALIZED` also lives at an API level for all name and label fields.
      - [x] I believe this can be fully achieved by updating the logic that reads `name` record values from RPC calls. If the `name` value that is encountered is not normalized then the response should simulate as if it is not set at all.
- [x] Update [our docs on unnormalized labels being returned](https://ensnode.io/docs/usage/querying-best-practices/#unnormalized-labels) to make reference to this capability ENSNode has over the ENS Subgraph.
- [x] Update the docs on our resolution APIs (including our client SDKs and React Hooks).
- [x] Update the docs on our database schemas as defined in our `ensnode-schema` package for each field containing name or label values .
- [x] Across all our docs updates for this feature, we need to be precise in our language. This feature does not prevent unnormalized names from being returned in our APIs. Any name where 1 or more labels are encoded labelhashes is unnormalized. Instead this feature guarantees that any unnormalized names returned by our APIs are strictly for the reason that one or more labels in a name are represented as encoded labelhashes. This guarantees certain nasty edge cases are impossible, including labels containing null bytes, full-stop characters (periods), or exotic unicode characters that may break UIs.
- [ ] Coordinate DevOps actions required for our hosted environments. (suggest to coordinate these at the same time as the updates from [Issue 959: Require pinned ENSRainbow labelset and version in ENSIndexer](https://github.com/namehash/ensnode/issues/959))
   - [ ] Before we release these changes we should proactively add configurations for these environment variables to all our ENSIndexer instances across all our hosted environments. Note how it's fine to add these environment variables before they actually start to be used once we release these enhancements.
      - [ ] Update all of Railway Blue (requires manual update)
      - [ ] Update all of Railway Green (requires manual update)
      - [x] Update all of Render yellow (requires updates to our Terraform scripts)

# Acceptance criteria

When feature enabled:

- [x] All names / labels returned by queries to ENSNode are either normalized or encoded labelhashes.
   - [x] True for our GraphQL API
   - [x] True for our Resolution APIs
   - [x] True for state materialized in ENSDb.
- [x] The `node` for a name is always accurate no matter if it contains unnormalized labels or not according to the implementation of the `namehash` function inside ENSjs as found at `packages/ensjs/src/utils/normalise.ts`. Please note how this implementation of the `namehash` function includes special logic for "encoded labelhashes" and therefore differs from a pure implementation of `namehash`.

shrugs:
closes #846

- [x] settle on naming for 'valid' terminology @lightwalker-eth 
- [ ] need to do a subgraph compat snapshot check


apps/ensindexer/src/config/config.schema.ts
```diff
@@ -116,6 +117,9 @@ const IndexAdditionalResolverRecordsSchema = makeEnvStringBoolSchema(
   "INDEX_ADDITIONAL_RESOLVER_RECORDS",
 ).default(DEFAULT_INDEX_ADDITIONAL_RESOLVER_RECORDS);
 
+const ReplaceUnnormalizedSchema = makeEnvStringBoolSchema("REPLACE_UNNORMALIZED") //
```
lightwalker-eth, line 120:
```suggestion
const ReplaceUnnormalizedSchema = makeEnvStringBoolSchema("REPLACE_UNNORMALIZED")
```

packages/ensnode-sdk/src/ensindexer/config/helpers.ts
```diff
@@ -1,15 +1,19 @@
 import { type ENSIndexerPublicConfig, PluginName } from "./types";
 
 /**
- * Subgraph compatibility
+ * Determines whether the provided `config` should be run in subgraph-compatibility mode (maintains
+ * backwards compatibility with the ENS Subgraph's resulting index/api).
```
lightwalker-eth, line 5:
```suggestion
 * Determines if the provided `config` produces an indexed state representation
 * where each data record has completely equivalent values as the ENS subgraph.
```

docs/ensnode.io/src/content/docs/docs/reference/terminology.mdx
```diff
@@ -58,35 +58,35 @@ For the _normalized_ name `vitalik.eth`, its **node** is `0xee6c4522aab0003e8d14
 https://docs.ens.domains/resolution/names#namehash
```
lightwalker-eth, line 60: Suggest to also document the following definitions somewhere in this file. You may want to link to these definitions from relevant places in the code.

* Encoded LabelHash: A LabelHash value wrapped in square bracket characters.
* Encoded NameHash: A NameHash value wrapped in square bracket characters.
* Normalized Label: a Label that is ENS normalized, does not contain any '.' (full stop) characters, and is not an empty string.
* Unnormalized Label: A Label that is not a Normalized Label.
* Unknown Label: a Label with an unknown value (that may be a Normalized Label or an Unnormalized Label) that is represented in the form of an Encoded LabelHash. Unknown Labels are a subset of Unnormalized Labels due to these square bracket characters that are not normalizable.
* Normalized Name: A Name that contains 0 or more Normalized Labels. Note that the empty string is a Normalized Name that contains no Labels.
* Unnormalized Name: A Name that contains 1 or more Unnormalized Labels.
* Unknown Name: A Name that contains 1 or more Unknown Labels, any number of Normalized Labels, and 0 Unnormalized Labels.


We should also make a distinction between Literal Labels and Interpreted Labels. This is very important for the purpose of not breaking our namehash / Node calculations. This distinction impacts how we should process  Label that is in the format of an Encoded LabelHash or an Encoded NameHash.
* When we read a Label from onchain data it begins as a Literal Label. If it is not a Normalized Label then we must convert it into an Unknown Label before we perform any processing of it. This process transforms the Literal Label into an Interpreted Label.
* When we read a Label from any other source we should treat it as an Interpreted Label.

The exact same pattern of ideas applies for Literal Names and Interpreted Names.

Please ask me if any questions 👍

lightwalker-eth (commented):
Quick partial feedback.

apps/ensindexer/src/handlers/Registry.ts
```diff
@@ -193,21 +194,25 @@ export const handleNewOwner =
         healedLabel = await labelByLabelHash(labelHash);
       }
```
lightwalker-eth, line 196: On this line where I'm adding this comment healedLabel represents a "Literal Label".

In the next statement after this comment is where we transform it into an "Interpreted Label".

Please explicitly document the transition happening here.

(please see my related comment with definitions).

apps/ensindexer/src/config/config.schema.ts
shrugs: that's there on purpose so the wrapping is easier to parse, otherwise it will autoformat as 

```ts
const ReplaceUnnormalizedSchema = makeEnvStringBoolSchema("REPLACE_UNNORMALIZED").default(
  DEFAULT_REPLACE_UNNORMALIZED,
);
```

versus the preferred

```ts
const ReplaceUnnormalizedSchema = makeEnvStringBoolSchema("REPLACE_UNNORMALIZED") //
  .default(DEFAULT_REPLACE_UNNORMALIZED);
```

and this pattern is used elsewhere

packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -9,10 +9,34 @@ import { monkeypatchCollate } from "./collate";
...
+  /**
+   * The human readable name that this Domain represents.
+   *
+   * If REPLACE_UNNORMALIZED is true, this value is guaranteed to be a Name consisting entirely of
+   * Labels that are either normalized or represented as Encoded LabelHashes. Note that the type of
+   * the column will remain string | null, for legacy subgraph compatibility.
+   *
+   * If REPLACE_UNNORMALIZED is false, this value may contain:
+   * a) null (in the case of the root node), or
+   * b) a name that may or may not be normalized and may or may not contain Encoded LabelHashes.
```
lightwalker-eth, line 22: Can we refine this idea?

Given the following conceptual framing: "non-null values may be normalized or unnormalized. Unnormalized names may contain X"

What are the classes of labels that should be listed in X?

1. Normalized labels (isNormalizedLabel is true)
2. Normalizable labels (ens normalize does not throw an exception but returns a different value. Ex: "VITALIK.ETH" is Normalizable, but not normalized.)
3. Unnormalizable labels (ens normalize throws an exception):
   1. Literal labels that are unnormalizable because they are formatted as an Encoded LabelHash but if treated as an interpreted labelhash would NOT contain the true labelhash of the subname and therefore would break the namehash calculation of the full interpreted name.
   2. Interpreted labels that are unnormalizable because they are formatted as an Encoded LabelHash and if treated as an interpreted labelhash, WOULD contain the true labelhash of the subname and therefore would NOT break the namehash calculation of the full interpreted name.
   3. unnormalizable labels that are not formatted as encoded labelhashes.
      1. Labels that pass ens normalize (which interprets its input as a full name, not a label) but that are unnormalizable labels (when interpreted as a label by isNormalizedLabel) because they can break the way ens normalize interprets the total number of labels in a name which ultimately breaks the namehash calculation.
         1. Unnormalized labels that contain a full-stop character.
         2. Unnormalized labels that are empty strings.
      2. Unnormalized labels that contain null byte characters.
      3. Unnormalized labels that are otherwise unnormalized (ex: contain a prohibited character such as "|" or violate  any other normalization rule).

It will be very helpful if for both the `true` and `false` case of `REPLACE_UNNORMALIZED` you could identify exactly which classes of labels can be found here.

Additionally, for the overall name value (not just labels in the name), for both true and false cases, we should identify which of the following cases apply:

1. null
2. empty string
3. normalized (with correct namehash)
4. unnormalized (with correct namehash)
5. unnormalized (with incorrect namehash when using any namehash implementation because of fullstop characters embedded in the middle of a label)
6. unnormalized (with incorrect namehash when using a namehash implementation like the one in ENSjs that assumes all labels formatted as encoded labelhashes should be interpreted and are not literal labels)

packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -9,10 +9,34 @@ import { monkeypatchCollate } from "./collate";
...
+   */
   name: t.text(),
-  // The human readable label name (imported from CSV), if known
+
+  /**
+   * The human-readable Label associated with the Domain.
```
lightwalker-eth, line 27: Suggest caution with using language such as "human-readable", as depending on the classes of label values held in this field, I would say some aren't human readable.

Here's some alternative ideas that could be good to document here:
1. The labelName is the label of the child-most subname associated with each domain record.
   1. How does this work if the domain record is the root (which is an empty string, which should be considered as having no labels and therefore no labelName or labelhash?
2. Please see my comment above that goes into detail requesting to understand the exact classes of label values that could be contained in this field depending on if `REPLACE_UNNORMALIZED` is true or false.
3. What is the possible relationships between the value of the `labelName` field and the `labelhash` field? Is it truly always guaranteed that labelhash is always `keccak256(labelName)` no matter what the value of `REPLACE_UNNORMALIZED` is?

packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -9,10 +9,34 @@ import { monkeypatchCollate } from "./collate";
...
+   *
+   * If REPLACE_UNNORMALIZED is true, this value is guaranteed to be either:
+   * a) null, representing an invalid label (i.e. empty string)
```
lightwalker-eth, line 30: When `REPLACE_UNNORMALIZED` is true the value in this field should never be null (with a possible exception only for the ENS root).

Instead, when `REPLACE_UNNORMALIZED` is true, if the label is an "invalid label" (suggest we completely remove any use of the terminology "invalid label" because it isn't precise enough, please see my other comments that provide a taxonomy of classifications of labels) then the value here should be an Interpreted Encoded LabelHash that guarantees `decodeEncodedLabelHash(labelName) === labelhash`

packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -131,9 +155,13 @@ export const resolver = onchainTable(
     // NOTE: we avoid .notNull.default([]) to match subgraph behavior
     coinTypes: t.bigint().array(),
 
-    // NOTE(resolver-records): include the value of the reverse-resolution name() record
-    // https://docs.ens.domains/ensip/3
-    // NOTE: this is the sanitized name record value with  (see @/lib/sanitize-name-record)
+    /**
+     * Represents the value of the reverse-resolution (ENSIP-3) name() record.
+     *
+     * The value of this field is guaranteed to be either:
```
lightwalker-eth, line 161: Please see my comments above providing a taxonomy of label / name classifications.

What is the set of possible label / name classifications stored in this field? How do those classifications change depending on the value of `REPLACE_UNNORMALIZED`?

packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -177,7 +205,19 @@ export const registration = onchainTable(
...
+     * a) null, representing an invalid label (i.e. empty string)
+     * b) a non-empty-string, normalized label, or
+     * c) an Encoded LabelHash
+     *
+     * If REPLACE_UNNORMALIZED is false, this value could contain an unnormalized Label.
+     *
+     * TODO: can Registration.labelName ever be invalid? likely not, so in practice labelName would
+     * never be null.
+     */
     labelName: t.text(),
```
lightwalker-eth, line 221: For documenting the labelName field in the registration table, could we just say that this labelName field is guaranteed to always be the same value as the labelName field in the related domain record?

packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -218,7 +258,20 @@ export const wrappedDomain = onchainTable(
...
+     * b) a Name consisting entirely of Labels that are either normalized or represented as
+     *    Encoded LabelHashes.
+     *
+     * If REPLACE_UNNORMALIZED is false, this value could contain an unnormalized Name.
+     *
+     * TODO: can NameWrapper even emit invalid dns-encoded names? if so, the type of this must be null
+     * because the name is set in a later transaction than the creation BUT in practice it would
+     * never be null.
+     */
     name: t.text(),
```
lightwalker-eth, line 275: For documenting the name field in the wrapped_domains table, could we just say that this name field is guaranteed to always be the same value as the name field in the related domain record?

lightwalker-eth (commented):
@shrugs Hey thanks for the updates. Here's feedback from a partial-review that focuses on data model invariants and how we document them. I'm thinking we focus on getting these pinned down and then all the other logic in this PR can flow accordingly.

packages/ensnode-schema/src/subgraph.schema.ts
shrugs: no, similar to registration labels, all of the logic for this is too all over the place to say that with any confidence. wrapped domain name/label only comes from namewrapper, but materializes to domain.

apps/ensindexer/src/api/lib/resolution/forward-resolution.ts
```diff
@@ -122,7 +122,7 @@ async function _resolveForward<SELECTION extends ResolverRecordsSelection>(
           //////////////////////////////////////////////////
 
           // Invariant: Name must be normalized
-          if (!isNormalized(name)) {
+          if (!isNormalizedName(name)) {
```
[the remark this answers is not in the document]
shrugs: confirming that this works as expected and this normalization check is required because forward resolution is also its own entry point when we want to enforce that only normalized names are queried.

apps/ensindexer/src/handlers/ThreeDNSToken.ts
```diff
@@ -200,26 +203,26 @@ export async function handleRegistrationCreated({
...
-  if (!isLabelIndexable(label)) {
+  if (!label || !name) {
     console.table({ ...event.args, tx: event.transaction.hash });
-    throw new Error(`Expected indexable label, got "${label}"`);
+    throw new Error(`Invariant: expected valid DNSPacketBytes: "${fqdn}"`);
   }
 
-  // Invariant: >2LDs never emit RegistrationCreated
-  // TODO: is this invariant exactly correct? it seems to be, but unclear
+  // Invariant: ThreeDNSToken only emits RegistrationCreated for TLDs or 2LDs
```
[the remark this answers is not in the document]
shrugs: uses gt `>`

apps/ensindexer/src/lib/resolver-records-helpers.ts
```diff
@@ -1,18 +1,15 @@
...
-export async function handleResolverNameUpdate(
-  context: Context,
-  resolverId: string,
-  name: string | null,
-) {
+export async function handleResolverNameUpdate(context: Context, resolverId: string, name: string) {
   await context.db
     .update(schema.resolver, { id: resolverId })
-    .set({ name: sanitizeNameRecordValue(name) });
+    .set({ name: interpretNameRecord(name) });
```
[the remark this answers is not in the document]
shrugs: this is name record, subgrpah does not use

apps/ensindexer/src/lib/resolver-records-helpers.ts
```diff
@@ -52,26 +49,24 @@ export async function handleResolverTextRecordUpdate(
   value: string | undefined | null,
 ) {
   // if value is undefined, this is a LegacyPublicResolver (DefaultPublicResolver1) event, nothing to do
+  // TODO: fetch the resolver value using ponder's cached publicClient
```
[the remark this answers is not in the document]
shrugs: todo general, already tracked with issue

packages/ensnode-sdk/src/ens/encode-labelhash.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Represents a LabelHash as an Encoded LabelHash.
+ *
+ * @see https://ensnode.io/docs/reference/terminology#encoded-labelhash
+ *
+ * @param labelHash - A 32-byte hash string starting with '0x'
+ * @returns The encoded label hash in format `[hash_without_0x_prefix]`
+ */
+export const encodeLabelHash = (labelHash: LabelHash): EncodedLabelHash =>
+  `[${labelHash.slice(2)}]`;
```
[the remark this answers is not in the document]
shrugs: unnecessary as per previous feedback, LabelHash is already in the correct format

packages/ensnode-sdk/src/ensindexer/api/zod-schemas.ts
```diff
@@ -34,7 +34,7 @@ const stringarray = z
 
 const name = z
```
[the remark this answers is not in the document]
shrugs: disagree, `name` is the name of the argument and it should be that, the fact that it's normalized is appropriately encoded in the validation logic

packages/ensnode-sdk/src/shared/interpretation.ts
```diff
@@ -0,0 +1,80 @@
...
+  // context of a name record value, we want to coerce empty string to null, to represent the
+  // non-existence of a record value.
+  if (value === "") return null;
+
+  // if not normalized, is not valid `name` record value
+  if (!isNormalizedName(value)) return null;
+
+  // otherwise, this is a non-empty-string normalized Name that can be used as a name() record value
+  return value as Name;
+}
```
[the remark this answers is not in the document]
shrugs: i disagree about the usage of the literal/interpreted terminology for the name record value here. i've adjusted the comments and hope that they read simpler now. i've also moved this helper to `apps/ensindexer/src/lib/resolver-records-helpers.ts` because it is an ensindexer implementation detail

apps/ensindexer/src/handlers/ThreeDNSToken.ts
```diff
@@ -200,26 +203,26 @@ export async function handleRegistrationCreated({
...
-  }
+  // Interpret the decoded Literal Label/Name into an Interpreted Label/Name
+  // see https://ensnode.io/docs/reference/terminology#interpreted-label
+  // see https://ensnode.io/docs/reference/terminology#interpreted-name
+  const label = interpretLiteralLabel(_label);
+  const name = interpretLiteralName(_name);
 
-  // Invariant: ThreeDNS validates that labels only use alphanumeric characters and hypens
+  // ThreeDNS always emits a valid DNS Packet and an Interpreted Name/Label but/and we explicitly
+  // enforce this invariant here.
```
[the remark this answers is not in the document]
shrugs: probably does but don't have access to their code to check

lightwalker-eth (commented):
Sharing quick partial feedback

packages/ensnode-sdk/src/shared/interpretation.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @param name - The Literal Name string to validate
+ * @returns An Interpreted Name
+ */
+export function interpretLiteralName(name: Name): Name {
+  // if the name is already normalized (includes empty string), good to go
+  if (isNormalizedName(name)) return name;
+
+  // otherwise ensure the name is composed of Interpreted Labels
+  return name.split(".").map(interpretLiteralLabel).join(".");
```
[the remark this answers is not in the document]
shrugs: i've addressed these concerns by changing the function to simply enforce that the dns-encoded name values emitted by the NameWrapper and ThreeDNSToken (which is the reason this function was created) are normalized (and therefore are interpreted names).

lightwalker-eth (commented):
@shrugs Sharing partial review that finishes a review of all code files. Only 5 files with documentation / comments to review remain 👍

lightwalker-eth (approved):

packages/ensnode-sdk/src/shared/interpretation.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Requires that the provided Literal Name is an Interpreted Name.
+ *
+ * @see https://ensnode.io/docs/reference/terminology#literal-name
+ * @see https://ensnode.io/docs/reference/terminology#interpreted-name
+ *
+ * @param name - The Literal Name to interpret
+ * @returns An Interpreted Name
+ * @throws If the Literal Name is not normalized
+ */
+export function requireInterpretedName(name: Name) {
```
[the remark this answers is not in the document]
shrugs: reverted to interpretLiteralName, see other comments

packages/ensnode-sdk/src/shared/interpretation.ts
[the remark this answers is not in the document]
shrugs: i don't think this ideal of namehash1  === namehash2 is possible to guarantee for any name because of how periods are handled in labels. given a name it's impossible to know whether the periods are 'part' of a label or are separators. the implementation of interpretLiteralName is the best we can do with the information available, and it (perhaps technically correctly) interprets a name like `this..name` as being composed of the labels `this`, `encodeLabelHash(labelhash(''))`, and `name`.

PR #998, opened 2025-08-27
feat/NFTs

lightwalker-eth:
Pending tasks include:

- [ ] Testing
- [x] Changesets
- [x] Refine docs on relationship between plugins and required RPCs
- [x] Explore expanding support to include NameWrapper and 3DNSToken
- [x] Document how a single domain may have any number of concurrent NFTs minted (give example of how NameWrapper works)
- [x] Explore explicit constraints / docs on the ERC1155 tokens we index all being for a single token (not multiple).


apps/ensindexer/src/lib/tokenscope/assets.ts
```diff
@@ -9,6 +12,56 @@ export const AssetNamespaces = {
...
+  }).toLowerCase();
+};
+
+/**
+ * Builds a CAIP-19 AssetId for the provided NFT.
+ *
+ * @param nft - The NFT to build an AssetId for
+ * @returns The CAIP-19 AssetId for the NFT
+ */
+export const buildNftAssetId = (nft: SupportedNFT): AssetId => {
```
shrugs, line 65: `buildSupportedNFTAssetId` seems more explicit, since a 'supported nft' is a distinct concept here. documentation as well could read like `Builds a CAIP-19 AssetId for the SupportedNFT.`

and i think we shoujld match the type's capitalization in our method names: `buildSupportedNFTAssetId`

lightwalker-eth (commented):
Small suggestions

PR #1007, opened 2025-09-01
Render environment fixes

BanaSeba:
- ENSIndexer localhost issue fixed
- replace_unnormalized set to false for all indexer
- ENSDB size set to 250GB
- Removed 'tokenscope' plugin for alpha-sepolia


tk-o (approved):
LGTM

PR #1008, opened 2025-09-01
Fix NameWrapper indexing error when REPLACE_UNNORMIZED=true

shrugs:


apps/ensindexer/src/lib/dns-helpers.ts
```diff
@@ -109,6 +109,56 @@ function concatUint8Arrays(a: Uint8Array, b: Uint8Array): Uint8Array {
...
+  while (offset < bytes.length) {
+    const len = bytes[offset];
+
+    // Invariant: the while conditional enforces that there's always _something_ in bytes at len
+    if (len === undefined) {
+      throw new Error(`Invariant: bytes[offset] is undefined after offset < bytes.length check.`);
+    }
+
+    // stop condition
+    if (len === 0) break;
```
lightwalker-eth, line 143: I note how the legacy / subgraph implementation of this logic seems to have a special case for the length of the first label not being allowed to be 0.

However here the logic seems to allow that.

Is this ok? Double checking 👍

apps/ensindexer/src/lib/dns-helpers.ts
```diff
@@ -56,7 +56,7 @@ import { type Label, type Name, isLabelIndexable } from "@ensnode/ensnode-sdk";
  *
  * @returns [(first) Label, Name] if decoding succeeded, otherwise [null, null]
  */
-export function decodeDNSPacketBytes(buf: Uint8Array): [Label, Name] | [null, null] {
+export function legacy_decodeDNSPacketBytes(buf: Uint8Array): [Label, Name] | [null, null] {
```
lightwalker-eth, line 59: As I understand, perhaps we can completely remove `legacy_decodeDNSPacketBytes` and replace it with a little wrapper around `v1_decodePacketIntoLiteralLabels` that simply checks each literal label with the `isLabelIndexable` function.

That would be nice for several reasons:
1. We retain just a single implementation of decoding encoded dns names.
2. It's more clear what the difference is between the legacy subgraph logic and our new replace unnormalized logic.
3. We more deeply test your new implementation of decoding encoded dns names when we make snapshots against our legacy subgraph logic.

shrugs (commented):
I know you have an aversion to the 'valid' terminology but in this case i believe it to be appropriate, since the terminology used in the legacy ens subgraph is whether the label is 'valid'. I've written the code and documentation as clearly as possible, referencing the concept of 'subgraph-valid' and 'subgraph-invalid' at every point, so it should be clear to readers that this is a subgraph-specific condition of the label.

shrugs:
`ed8220e9` confirmed subgraph compatible with `REPLACE_UNNORMALIZED=false` and `subgraph,threedns` configuration indexes to latest with `REPLACE_UNNORMALIZED=true`

packages/ensnode-sdk/src/ens/subgraph-label-valid.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Check if any characters in `label` are "valid" according to legacy Subgraph logic.
+ *
+ * Implements the following ENS Subgraph `checkValidLabel` function:
+ * @see https://github.com/ensdomains/ens-subgraph/blob/c8447914e8743671fb4b20cffe5a0a97020b3cee/src/utils.ts#L68
+ *
+ * @param label - The label to check. Note: A `null` value for `label` represents an unknown label.
+ * @returns whether the provided Label is subgraph-valid
+ */
+export const isLabelSubgraphValid = (label: Label | null): label is Label => {
```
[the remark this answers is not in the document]
shrugs: which is to say that an unknown label is not subgraph-indexable.

docs/ensnode.io/src/content/docs/docs/usage/querying-best-practices.mdx
```diff
@@ -179,13 +179,13 @@ It's important to note that while onchain ENS contracts do not enforce ENSIP-15
 
 ### Unknown Label Representation
```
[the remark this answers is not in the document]
lightwalker-eth: Agreed, comment here is suggesting not to completely remove any reference to "unknown label" -- just all references outside the cited exception which is the same exception you also noted here.

Currently "unknown label" is being used in too many places where a variation of "interpreted label" would better apply.

apps/ensindexer/src/handlers/ThreeDNSToken.ts
```diff
@@ -161,36 +161,37 @@ export async function handleNewOwner({
...
       healedLabel = await labelByLabelHash(labelHash);
     }
 
     // Interpret the `healedLabel` Literal Label into an Interpreted Label
     // see https://ensnode.io/docs/reference/terminology#literal-label
     // see https://ensnode.io/docs/reference/terminology#interpreted-label
-    const label =
+    const interpretedLabel = (
       healedLabel !== null
         ? literalLabelToInterpretedLabel(healedLabel as LiteralLabel)
```
lightwalker-eth, line 182: Maybe we don't need the "as" here anymore based on other updates that have been made?

If it's still required here, no worries 👍

docs/ensnode.io/src/content/docs/docs/usage/querying-best-practices.mdx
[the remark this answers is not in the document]
lightwalker-eth: Adding a reply here during the review just because I appreciate how GitHub makes it difficult to track comments made outside of PRs.

docs/ensnode.io/src/content/docs/docs/usage/querying-best-practices.mdx
```diff
@@ -164,28 +164,15 @@ According to [ENSIP-1](https://docs.ens.domains/ensip/1#namehash-algorithm), the
...
-While many characters are not supported by ENSIP-15, only the following four characters are specifically classified as subgraph-invalid due to indexing concerns:
-
-1. `\0` (null byte) - PostgreSQL does not allow storing this character in text fields
-2. `.` (period) - Conflicts with ENS label separator logic
-3. `[` (left square bracket) - Conflicts with "unknown label" representations
-4. `]` (right square bracket) - Conflicts with "unknown label" representations
-
-It's important to note that while onchain ENS contracts do not enforce ENSIP-15 normalization (due to factors like gas costs), these specific characters require special handling by indexers to avoid conflicts.
-
 ### Unknown Label Representation
```
lightwalker-eth, line 167: Suggest to preface this whole section of the docs with a note that it only applies if `REPLACE_UNNORMALIZED` is false.

Is that fair?

docs/ensnode.io/src/content/docs/docs/usage/querying-best-practices.mdx
```diff
@@ -164,28 +164,15 @@ According to [ENSIP-1](https://docs.ens.domains/ensip/1#namehash-algorithm), the
...
 
-Therefore, this represents the subgraph-invalid label as an "unknown label", encoding the labelhash of the original unnormalized label (including its square brackets) in square brackets. This demonstrates why square brackets are considered subgraph-invalid - they create ambiguity between literal labels and the representation of unknown labels.
+Therefore, this represents the subgraph-unindexable label as an "unknown label", encoding the labelhash of the original unnormalized label (including its square brackets) in square brackets. This demonstrates why square brackets are considered subgraph-unindexable - they create ambiguity between literal labels and the representation of unknown labels.
 
-When ENSNode encounters an subgraph-invalid label, it will represent it as an unknown label even if the actual label data is available. This is a safety mechanism to ensure data integrity and consistent behavior across the system.
+When ENSNode encounters an subgraph-unindexable label, it will represent it as an unknown label even if the actual label data is available. This is a safety mechanism to ensure data integrity and consistent behavior across the system.
```
lightwalker-eth, line 176:
```suggestion
When ENSNode encounters an subgraph-unindexable label, it will represent it as an unknown label even if the actual label data is available.
```

Goal: I think that sentence is better removed.

docs/ensnode.io/src/content/docs/docs/usage/querying-best-practices.mdx
```diff
@@ -164,28 +164,15 @@ According to [ENSIP-1](https://docs.ens.domains/ensip/1#namehash-algorithm), the
...
 
-For more detailed information about subgraph-invalid labels and their handling, please refer to the [ENSNode SDK implementation](https://github.com/namehash/ensnode/blob/main/packages/@ensnode/ensnode-sdk/src/ens/subgraph-label-valid.ts).
+For more detailed information about subgraph-unindexable labels and their handling, please refer to the [ENSNode SDK implementation](https://github.com/namehash/ensnode/blob/main/packages/@ensnode/ensnode-sdk/src/ens/subgraph-label-valid.ts).
```
lightwalker-eth, line 178: I think we might need to update this link?

docs/ensnode.io/src/content/docs/docs/reference/terminology.mdx
```diff
@@ -165,3 +165,28 @@ A **Literal Name** is composed of **Literal Labels** as they appear in raw oncha
...
 An **Interpreted Name** is composed of **Interpreted Labels** after processing, where each label has been classified and potentially converted to its appropriate display format (using **Encoded LabelHash** for unknown labels).
+
+## Subgraph Indexability & Label/Name Interpretation
+
+### Subgraph-Indexable Labels / Subgraph-Unindexable Labels
+The legacy ENS Subgraph specifies that unknown labels and labels containing certain UTF-8 characters are "invalid". We refer to this concept as `subgraph-indexable` and `subgraph-unindexable`.
+
+A `subgraph-unindexable` Label is either:
+  - an **Unknown Label**, or
+  - a **Literal Label** that contains any of the prohibited UTF-8 characters.
```
lightwalker-eth, line 176:
```suggestion
  - a **Literal Label** that contains any of the following prohibited UTF-8 characters:
     1. ...
     2. ... etc..
```

Goal: Reduce the distance in the docs between related ideas. Please see related comment below.

docs/ensnode.io/src/content/docs/docs/reference/terminology.mdx
```diff
@@ -165,3 +165,28 @@ A **Literal Name** is composed of **Literal Labels** as they appear in raw oncha
...
+
+A `subgraph-indexable` Label is a **Known Literal Label** that does NOT contain any of the prohibited UTF-8 characters.
+
+The `subgraph-unindexable` UTF-8 characters are:
```
lightwalker-eth, line 180: Please see my comment above, suggesting to move this list up 2 paragraphs.

docs/ensnode.io/src/content/docs/docs/reference/terminology.mdx
```diff
@@ -165,3 +165,28 @@ A **Literal Name** is composed of **Literal Labels** as they appear in raw oncha
...
+1. `\0` (null byte) - PostgreSQL does not allow storing this character in text fields
+2. `.` (period) - Conflicts with ENS label separator logic
+3. `[` (left square bracket) - Conflicts with "unknown label" representations
```
lightwalker-eth, line 183: For 3 and 4 in this list, I think better to replace the reference to "unknown label" with a reference to labels formatted as an Encoded LabelHash.

lightwalker-eth (commented):
@shrugs Super updates 💪 I think this is very nearly good to go. Just a few small suggestions remaining.

Issue #989, opened 2025-08-26
Refine date formatting on "Backfill Status" panel


lightwalker-eth:
<img width="2338" height="1160" alt="Image" src="https://github.com/user-attachments/assets/b3a85a1f-02ce-4e26-908c-8c08c5243664" />

Different countries interpret dates in the format "3/10/2017" differently. Is this March 10, 2017 or is it October 3, 2017?

Even if we always localize the formatting of this to the user's computer it can still be confusing. For example, as someone who is originally from the US but has lived so much of their life abroad, when I see a date in this format I have doubts about how to interpret it. Is the locale based on my device or based on the server or based on the location of the devs who wrote the code? Etc..

Therefore, suggest to change the presentation of these dates in the format of: "Mar 10, 2017". It takes up a tiny bit more space but in exchange we avoid confusion.

PR #1011, opened 2025-09-02
refactor(indexing-status): include `omnichainIndexingCursor`

### LINKED ISSUE #946: Refine ENSIndexerOverallIndexingCompletedStatus
Currently reviewing PR 919 and I see it would improve our logic inside the `RecentRegistrations` component if `ENSIndexerOverallIndexingCompletedStatus` included an `omnichainIndexingCursor` field the same way the `ENSIndexerOverallIndexingFollowingStatus` has it.

We could then update our code to guarantee that `currentIndexingDate` is guaranteed to be defined.

tk-o:
This PR extends the `ENSIndexerOverallIndexingCompletedStatus` data model with the `omnichainIndexingCursor` field.
It also aligns date & datetime display format.

Resolves #946, #989.

### Preview

| Backfill timeline | Recent Registrations List |
|--------|--------|
| image | image |

### DIFF
--- a/apps/ensadmin/src/components/datetime-utils/index.tsx
+++ b/apps/ensadmin/src/components/datetime-utils/index.tsx
@@ -1,24 +1,51 @@
-import { formatDistance, formatDistanceStrict, intlFormat } from "date-fns";
+import {
+  IntlFormatFormatOptions,
+  formatDistance,
+  formatDistanceStrict,
+  intlFormat,
+} from "date-fns";
 import { millisecondsInSecond } from "date-fns/constants";
 import { useEffect, useState } from "react";
 
 /**
- * Client-only date formatter component
+ * Format date as a datetime string.
  */
-export function FormattedDate({
-  date,
-  options,
-}: {
-  date: Date;
-  options: Intl.DateTimeFormatOptions;
-}) {
-  const [formattedDate, setFormattedDate] = useState<string>("");
+export function datetimeFormat(date: Date, args: Partial<IntlFormatFormatOptions> = {}): string {
+  const {
+    year = "numeric",
+    month = "short",
+    day = "numeric",
+    hour = "numeric",
+    minute = "numeric",
+    second = "numeric",
+    hour12 = true,
+  } = args;
 
-  useEffect(() => {
-    setFormattedDate(intlFormat(date, options));
-  }, [date, options]);
+  return intlFormat(date, {
+    year,
+    month,
+    day,
+    hour,
+    minute,
+    second,
+    hour12,
+  });
+}
 
-  return <>{formattedDate}</>;
+/**
+ * Format date as a date string.
+ */
+export function dateFormat(
+  date: Date,
+  args: Partial<Pick<IntlFormatFormatOptions, "dateStyle" | "year" | "month" | "day">> = {},
+): string {
+  const { year = "numeric", month = "short", day = "numeric" } = args;
+
+  return intlFormat(date, {
+    year,
+    month,
+    day,
+  });
 }
 
 /**
--- a/apps/ensadmin/src/components/indexing-status/backfill-status.tsx
+++ b/apps/ensadmin/src/components/indexing-status/backfill-status.tsx
@@ -8,12 +8,11 @@ import {
   ENSIndexerOverallIndexingBackfillStatus,
   getTimestampForHighestOmnichainKnownBlock,
   getTimestampForLowestOmnichainStartBlock,
-  sortAscChainStatusesByStartBlock,
+  sortEarliestOmnichainStartBlock,
 } from "@ensnode/ensnode-sdk";
-import { fromUnixTime } from "date-fns";
+import { fromUnixTime, intlFormat } from "date-fns";
 import { Clock } from "lucide-react";
 
-import { FormattedDate } from "@/components/datetime-utils";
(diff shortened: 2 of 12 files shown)


apps/ensadmin/src/components/datetime-utils/index.tsx
```diff
@@ -1,24 +1,51 @@
...
  */
-export function FormattedDate({
-  date,
-  options,
-}: {
-  date: Date;
-  options: Intl.DateTimeFormatOptions;
-}) {
-  const [formattedDate, setFormattedDate] = useState<string>("");
+export function datetimeFormat(date: Date, args: Partial<IntlFormatFormatOptions> = {}): string {
```
lightwalker-eth, line 13:
```suggestion
export function formatDatetime(date: Date, args: Partial<IntlFormatFormatOptions> = {}): string {
```

Goal: For functions that perform some string formatting, it's best to make "format" (the verb) the prefix, not the suffix.

apps/ensadmin/src/components/datetime-utils/index.tsx
```diff
@@ -1,24 +1,51 @@
...
+  });
+}
 
-  useEffect(() => {
-    setFormattedDate(intlFormat(date, options));
-  }, [date, options]);
+/**
+ * Format date as a date string.
+ */
+export function dateFormat(
```
lightwalker-eth, line 38:
```suggestion
export function formatDate(
```

Goal: For functions that perform some string formatting, it's best to make "format" (the verb) the prefix, not the suffix.

apps/ensadmin/src/components/datetime-utils/index.tsx
```diff
@@ -1,24 +1,51 @@
...
+  formatDistance,
+  formatDistanceStrict,
+  intlFormat,
+} from "date-fns";
 import { millisecondsInSecond } from "date-fns/constants";
 import { useEffect, useState } from "react";
 
 /**
- * Client-only date formatter component
+ * Format date as a datetime string.
```
lightwalker-eth, line 11: The JSDoc here should give an example of the output of this function.

apps/ensadmin/src/components/datetime-utils/index.tsx
```diff
@@ -1,24 +1,51 @@
...
+    second,
+    hour12,
+  });
+}
 
-  useEffect(() => {
-    setFormattedDate(intlFormat(date, options));
-  }, [date, options]);
+/**
+ * Format date as a date string.
```
lightwalker-eth, line 36: The JSDoc here should give an example of the output of this function.

apps/ensadmin/src/components/datetime-utils/index.tsx
lightwalker-eth, line 13: In some earlier discussions we made a decision to stop using `Date` in our data models and instead to use `UnixTimestamp`.

Therefore all functions in this file should take as input a `UnixTimestamp` and internally perform whatever operations might be needed to convert that into a `Date` value that is then passed into `intlFormat`.

apps/ensadmin/src/components/indexing-status/backfill-status.tsx
```diff
@@ -72,18 +71,11 @@ export function BackfillStatus({ indexingStatus }: BackfillStatusProps) {
...
-                    year: "numeric",
-                    month: "short",
-                    day: "numeric",
-                    hour: "numeric",
-                    minute: "numeric",
-                    second: "numeric",
-                    hour12: true,
-                  }}
-                />
+                {intlFormat(omnichainIndexingCursorDate, {
```
lightwalker-eth, line 74: Hmm. I don't like that we're directly calling `intlFormat` here. Why aren't we calling one of the functions in `apps/ensadmin/src/components/datetime-utils/index.tsx`?

If we need to add some new function to that file, then fine let's do it.

Goal: Each time we display a `UnixTimestamp` value in our UI, prefer to always pass that through a finite set of formatX functions rather than use totally custom calls to intlFormat each time. This way we have more manageable control over the UX standards for how we display timestamp values in our UIs. If each place in our UIs uses its own custom datetime formatting its difficult to maintain.

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -57,22 +57,23 @@ export function RecentRegistrations({
...
-    indexingStatus.overallStatus === OverallIndexingStatusIds.Following
-      ? fromUnixTime(indexingStatus.omnichainIndexingCursor)
-      : null;
+  const omnichainIndexingCursor = fromUnixTime(indexingStatus.omnichainIndexingCursor);
 
   return (
     <Card className="w-full">
       <CardHeader>
         <CardTitle className="flex justify-between items-center">
           <span>Latest indexed registrations</span>
```
lightwalker-eth, line 66: Let's remove visibility of the `omnichainIndexingCursor` on this Latest indexed registrations panel.

Instead, for all cases where the `omnichainIndexingCursor` field is defined, please display that value in the top-right corner of the Indexing Status panel where we show a breakdown of the indexing status for each indexed chain.

lightwalker-eth (changes requested):
@tk-o Thanks for updates 👍 Reviewed and shared feedback

tk-o:
Replaced with:
- #1015 
- #1093

PR #1012, opened 2025-09-02
ensadmin / feat: Enhance ENSNodeConfig UIs

Y3drk:
Includes:
- [x] New mock page, specifically for this status component
- [x] Reorganization of `/status/mock/` routing
- [x] Several new design proposals for this component (`in progress`)
- [x] Add changeset

Closes Issue #1045

### DIFF
--- a/apps/ensadmin/src/app/mock/config-info/data.json
+++ b/apps/ensadmin/src/app/mock/config-info/data.json
@@ -6,7 +6,7 @@
       "labelSetId": "subgraph",
       "labelSetVersion": 0
     },
-    "indexedChainIds": [1, 8453, 59144, 10, 42161, 534352],
+    "indexedChainIds": [1, 8453, 59144, 10, 42161, 534352, 567],
     "databaseSchemaName": "alphaSchema0.34.0",
     "healReverseAddresses": true,
     "indexAdditionalResolverRecords": true,
--- a/apps/ensadmin/src/app/mock/indexing-stats/data.json
+++ b/apps/ensadmin/src/app/mock/indexing-stats/data.json
@@ -102,6 +102,17 @@
           },
           "endBlock": null
         }
+      },
+      "567": {
+        "status": "unstarted",
+        "config": {
+          "strategy": "indefinite",
+          "startBlock": {
+            "timestamp": 1720768999,
+            "number": 6682889
+          },
+          "endBlock": null
+        }
       }
     },
     "omnichainIndexingCursor": 1580899661


apps/ensadmin/src/app/status/mock/dependency-info/data.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  ChainId,
+  ENSIndexerPublicConfig,
+  ENSNamespaceId,
+  ENSNamespaceIds,
+  PluginName,
+} from "@ensnode/ensnode-sdk";
+
+export const MockENSIndexerPublicConfig: ENSIndexerPublicConfig = {
```
lightwalker-eth, line 9: Please review the strategy we used for defining mock data for "indexing-status": `apps/ensadmin/src/app/status/mock/data.json`

Specifically, please note how:
1. Mock data is defined as pure JSON.
2. ... therefore the type definition we use when reading from the pure JSON data is `SerializedENSIndexerOverallIndexingStatus`. Please see the logic here: https://github.com/namehash/ensnode/blob/main/apps/ensadmin/src/app/status/mock/page.tsx#L14
3. ... we then convert from `SerializedENSIndexerOverallIndexingStatus` (pure JSON) to `ENSIndexerOverallIndexingStatus` (a "rich" object) using `deserializeENSIndexerIndexingStatus`. Please see the logic here: https://github.com/namehash/ensnode/blob/main/apps/ensadmin/src/app/status/mock/page.tsx#L23
4. ... we then pass the `ENSIndexerOverallIndexingStatus` (the "rich" object) into the UI component mock.

We should repeat this exact strategy when building the mock for the "config".
1. Your pure JSON should match `SerializedENSIndexerPublicConfig`.
2. You should deserialize from pure JSON into a "rich" object using `deserializeENSIndexerPublicConfig`

Benefits of this strategy include how it helps us to not only test mocks of the UI component rendering for various cases, it also helps us with more end-to-end testing of how raw JSON values (that are returned in our API responses) are deserialized and rendered into the UI. Therefore we're also testing deserialization logic too which is great. More end-to-end :+1:

Please also note how the deserialization logic can fail. If this fails (ex: we might have written the mock JSON in an incorrect way) then we should gracefully handle that case on the mock page too. I believe the mock indexing-status page should already have some logic for this.

This touches on another benefit of this testing approach with raw JSON: it can also help us with testing scenarios where the JSON returned by our API has values that are unexpected and therefore the deserialization call fails. It's out of scope for us to add these tests now, but just highlighting how it's something we'll explore later.

apps/ensadmin/src/app/status/mock/dependency-info/page.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <h1 className={cn(headerStyles, "text-lg")}>Current production version</h1>
+      <ENSIndexerDependencyInfo ensIndexerConfig={MockENSIndexerPublicConfig} />
+      {contentSeparator}
+
+      <h1 className={cn(headerStyles, "text-lg")}>Version strictly based on Figma</h1>
+      <FigmaBasedDependencyInfo ensIndexerConfig={MockENSIndexerPublicConfig} />
+      {contentSeparator}
+
+      <h1 className={cn(headerStyles, "text-lg")}>Evolution of Figma-based design</h1>
+      <FigmaEvolutionDependencyInfo ensIndexerConfig={MockENSIndexerPublicConfig} />
```
lightwalker-eth, line 28: I'm happy with where this FigmaEvolution is going 👍 

Suggested next steps:
1. Please completely remove `ENSIndexerDependencyInfo` and `FigmaBasedDependencyInfo`.
2. Please rename `FigmaEvolutionDependencyInfo` to `ENSNodeConfigInfo`. The existing (non-mock) `/status` route should also be updated to make use of this new "config" component you're working on in this PR.

Goal: Let's focus our efforts on this new concept you're building 👍

lightwalker-eth (commented):
@Y3drk Hey great to see the updates here 👍 Reviewed and shared feedback

apps/ensadmin/src/app/status/mock/dependency-info/data.ts
Y3drk: Applied

apps/ensadmin/src/components/indexing-status/dependecy-info.tsx
```diff
@@ -10,7 +10,7 @@ import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { InlineSummary } from "@/components/ui/inline-summary";
 import { ENSIndexerPublicConfig } from "@ensnode/ensnode-sdk";
 
-interface ENSIndexerDependencyInfoProps {
+export interface ENSIndexerDependencyInfoProps {
```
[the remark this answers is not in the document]
Y3drk: Applied the change name to props & adjusted the component's name as well.

apps/ensadmin/src/app/@breadcrumbs/mock/config/page.tsx
```diff
@@ -0,0 +1,14 @@
+import { BreadcrumbItem, BreadcrumbPage, BreadcrumbSeparator } from "@/components/ui/breadcrumb";
+
+//TODO: is there any better way to do this?
```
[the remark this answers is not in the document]
Y3drk: @lightwalker-eth 
Just did and got some suggestions from Jamie, will have a closer look on Monday.

> Leaving this comment unresolved as a reminder

apps/ensadmin/src/app/mock/page.tsx
```diff
@@ -0,0 +1,26 @@
+import { Button } from "@/components/ui/button";
+import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
+import Link from "next/link";
+
+export default function Mock() {
```
lightwalker-eth, line 5:
```suggestion
export default function MockList() {
```

apps/ensadmin/src/app/mock/page.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <section className="flex flex-col gap-6 p-4">
+      <Card>
+        <CardHeader>
+          <CardTitle className="text-2xl leading-normal">Status Mocks</CardTitle>
```
lightwalker-eth, line 10:
```suggestion
          <CardTitle className="text-2xl leading-normal">UI Mocks</CardTitle>
```

apps/ensadmin/src/app/mock/page.tsx
```diff
@@ -0,0 +1,26 @@
...
+          <CardDescription>Select a component mock to preview</CardDescription>
```
lightwalker-eth, line 11:
```suggestion
          <CardDescription>Select a UI component</CardDescription>
```

apps/ensadmin/src/app/@breadcrumbs/mock/config-info/page.tsx
```diff
@@ -0,0 +1,21 @@
+import {
+  BreadcrumbItem,
+  BreadcrumbLink,
+  BreadcrumbPage,
+  BreadcrumbSeparator,
+} from "@/components/ui/breadcrumb";
+
+//TODO: is there any better way to do this?
```
lightwalker-eth, line 8: Suggest checking with @notrab

apps/ensadmin/src/app/mock/config-info/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </CardHeader>
+          <CardContent>
+            <pre className="text-sm text-red-700 whitespace-pre-wrap">{validationError}</pre>
+          </CardContent>
+        </Card>
+      )}
+
+      {deserializedConfig && <ENSNodeConfigInfo ensIndexerConfig={deserializedConfig} />}
+
+      {selectedConfig === "Loading" && <IndexingStatusPlaceholder />}
```
lightwalker-eth, line 76: For loading states, I think a nice way to do it is something like the following:

```
<ENSNodeConfigInfo ensIndexerConfig={null} />
```

^^ Please note the special `null` value to symbolically represent the idea that the loading state should be displayed. Please also update the logic and docs related to `ENSNodeConfigInfo` wherever it is used.

Then, inside of the implementation of `ENSNodeConfigInfo` you can check for the special case that the provided `ensIndexerConfig` is `null`. If so, then it internally can output an `<IndexingStatusPlaceholder />`.

However, I'm also concerned by the use of `<IndexingStatusPlaceholder />` inside the loading state of a `ENSNodeConfigInfo`.

Do you see the problem here?

We're mixing up UI components.

There should be a distinct loading state UI component for ConfigInfo vs IndexingStatus.

apps/ensadmin/src/app/mock/config-info/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+      {selectedConfig === "Loading Error" && <ENSNodeConfigInfoError />}
```
lightwalker-eth, line 78: I'm a fan of having a single main UI component that handles the overall responsibility for rendering all possible state variations of that UI component to the UI, including error states and loading states.

One possible way to handle this is to update `ENSNodeConfigInfo` so that you add an optional `error` param.

You can then document and enforce (by throwing an error) an invariant that if the `error` param is defined, `ensIndexerConfig` must be `null`.

Then, inside the logic of `ENSNodeConfigInfo`, if the `error` param is defined and `ensIndexerConfig` is `null` then it renders `<ENSNodeConfigInfoError />`.

lightwalker-eth (commented):
@Y3drk Great updates 👍 Reviewed and shared feedback 😄

apps/ensadmin/src/app/@breadcrumbs/mock/config/page.tsx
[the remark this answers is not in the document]
Y3drk: After reviewing Jamie's suggestion and how he does a very similar thing for the `name detail page`, I came to the conclusion that both approaches are parallel and make sense for their respective use cases. 

Because of that, I want to leave this code as is and focus on other fixes/changes, especially since the current logic works, and the only reason we currently consider changes here is to refactor.

apps/ensadmin/src/app/mock/config-info/page.tsx
Y3drk: @lightwalker-eth
I believe that there already is an issue for this refactor - namely Issue #1023 
I will link this thread there.

apps/ensadmin/src/components/indexing-status/indexing-status.tsx
```diff
@@ -27,14 +27,19 @@ export function IndexingStatus() {
   const indexingStatusQuery = useIndexingStatus();
```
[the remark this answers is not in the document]
Y3drk: Until now, `ENSNodeConfigInfo` and `IndexingStatus` were tightly coupled in this file, hence the confusion. 

As per [my comment above](https://github.com/namehash/ensnode/pull/1012#discussion_r2332736350), I'll try to remedy that in a refactor still in this PR

apps/ensadmin/src/app/mock/config-info/page.tsx
```diff
@@ -60,7 +65,7 @@ export default function MockConfigPage() {
         </CardContent>
       </Card>
 
-      {validationError && (
+      {validationError ? (
         <Card className="border-red-200 bg-red-50">
```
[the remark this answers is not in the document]
Y3drk: I think we can take this even further and create a generic `<ErrorInfo />` component that would take props described [in your other comment](https://github.com/namehash/ensnode/pull/1012#discussion_r2331573425) and display them in a unified way across the whole app

apps/ensadmin/src/app/mock/config-info/page.tsx
```diff
@@ -17,25 +19,29 @@ type ConfigVariant = keyof typeof mockConfigData | LoadingVariant;
 
 export default function MockConfigPage() {
   const [selectedConfig, setSelectedConfig] = useState<ConfigVariant>("Alpha Mainnet");
-  const { deserializedConfig, validationError, configError } = useMemo(() => {
+  const { props, deserializationError } = useMemo(() => {
     switch (selectedConfig) {
```
lightwalker-eth, line 23: Please make it so there's another mock variant called "Serialization Error". When selected it should provide an invalid `SerializedENSIndexerPublicConfig` such that it triggers the serialization error case so that we can also test it on this mock UI page.

Suggest adding this "Serialization Error" case directly to `mockConfigData` in `data.json`.

apps/ensadmin/src/app/mock/config-info/page.tsx
```diff
@@ -17,25 +19,29 @@ type ConfigVariant = keyof typeof mockConfigData | LoadingVariant;
 
 export default function MockConfigPage() {
   const [selectedConfig, setSelectedConfig] = useState<ConfigVariant>("Alpha Mainnet");
```
lightwalker-eth, line 21: suggest defining a named constant such as `DEFAULT_VARIANT` for "Alpha Mainnet" and then referencing that here.

apps/ensadmin/src/components/indexing-status/indexing-status.tsx
```diff
@@ -27,14 +27,19 @@ export function IndexingStatus() {
...
 
   if (ensIndexerConfigQuery.isError) {
-    return <ENSNodeConfigInfoError />;
+    return <ENSNodeConfigInfo ensIndexerConfig={null} error={true} />;
   }
   if (indexingStatusQuery.isError) {
     return <p className="p-6">Failed to fetch Indexing Status.</p>;
```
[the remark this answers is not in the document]
lightwalker-eth: Cool I see your 👍 here. I understand this update is pending 👍

lightwalker-eth (commented):
@Y3drk Great updates. They inspired me with a few other little ideas 😄

apps/ensadmin/src/components/indexing-status/config-info/config-info.tsx
```diff
@@ -9,46 +9,45 @@ import { ENSDbIcon } from "@/components/ensdb-icon";
...
+ * Error - ensIndexerConfig: null, error: string
  *
- * Invariant:
- * If the error parameter is defined then the ensIndexerConfig parameter must be null
+ * @throws If the error parameter is defined and the ensIndexerConfig is not null
  */
 export interface ENSNodeConfigProps {
   ensIndexerConfig: ENSIndexerPublicConfig | null;
-  error?: boolean;
+  error?: string;
```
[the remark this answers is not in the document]
Y3drk: Makes total sense, will apply.

apps/ensadmin/src/components/indexing-status/indexing-status.tsx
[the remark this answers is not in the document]
Y3drk: I feel like this update should be a part of the `IndexingStatus` refactor discussed in Issue #1023 .

In my opinion, to decouple that now would essentially require said refactor.

I probably "thumbed" it initially, then tried to make it as a "small refactor" change and when that didn't work, forgot to comment on it, sorry for the confusion

apps/ensadmin/src/components/indexing-status/indexing-status.tsx
```diff
@@ -27,14 +27,19 @@ export function IndexingStatus() {
...
   }
 
   if (!ensIndexerConfigQuery.isSuccess || !indexingStatusQuery.isSuccess) {
-    return <IndexingStatusPlaceholder />;
+    return (
+      <section className="flex flex-col gap-6 p-6">
+        <ENSNodeConfigInfo ensIndexerConfig={null} />
```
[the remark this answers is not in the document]
Y3drk: Update on that:
* The separation of `ENSNodeConfigInfo` and the exact `IndexingStatus` is not hard, but then we have components like `RecentRegistrations` that use the data from both hooks, and things get harder.
* This is yet another item that I think would be best addressed by the refactor discussed in Issue #1023

apps/ensadmin/src/app/mock/config-info/page.tsx
Y3drk: Applied.

The cause of the error is a `null` value for `ensNodePublicURL` field in the serialized response 
(for context see: `/app/mock/config-info/data.json`)

PR #1029, opened 2025-09-08
feat: hardcode new devnet addresses

### LINKED ISSUE #837: Refine ens-test-env Namespace definition
1. Refine https://github.com/namehash/ensnode/blob/main/packages/datasources/src/ens-test-env.ts#L47-L48
   1. Instead of using (a customized version of) [anvil](https://github.com/wevm/viem/blob/d0eb2ec7bbd6f5bccfb11af6a3bbf972d9688455/src/chains/definitions/anvil.ts), seems we should use (the non-customized version of) [localhost](https://github.com/wevm/viem/blob/d0eb2ec7bbd6f5bccfb11af6a3bbf972d9688455/src/chains/definitions/localhost.ts) as defined by viem?
2. I'm worried about how [getENSTestEnvDeploymentAddresses](https://github.com/namehash/ensnode/blob/main/packages/datasources/src/lib/ens-test-env-deployment-addresses.ts#L12-L31) is defined. Specifically it's calling `process.env`. However the datasources package is also used in purely frontend contexts, such as within the ENSAdmin frontend.
   1. Should this logic that reads from environment variables move into `ensindexer`? For the `datasources` package, maybe there's some way that [ENSTestEnvDeploymentAddresses](https://github.com/namehash/ensnode/blob/main/packages/datasources/src/lib/ens-test-env-deployment-addresses.ts#L3-L10) could be optionally passed into the datasources package from ensindexer?
3. I'm worried about the typesystem violation created by [EMPTY_ADDRESS](https://github.com/namehash/ensnode/blob/main/packages/datasources/src/ens-test-env.ts#L17). What if we used a zero address for a fallback instead? At least that way the fallback values would be valid address values.
4. Because the contract addresses for the `ens-test-env` are dynamic (not hardcoded) it means that when ENSIndexer sends its config to ENSAdmin, the config it sends should include all the Namespaces / Datasources / etc.. actually in use by the connected ENSIndexer instance.
   1. Consider the following case as one of many possible examples: See the work that Jedrzej is advancing in PR 812 for [getEffectiveOwner](https://github.com/namehash/ensnode/blob/y3drk/chore/ensadmin/fix-ensnamespace-handling-in-registrations-panel/apps/ensadmin/src/components/recent-registrations/hooks.ts#L30-L52). This logic runs in ENSAdmin. When ENSAdmin connects to an ENSIndexer running the ens-test-env, the current logic will fail because ENSAdmin and the connected ENSIndexer instance will both have different value returned by [getNameWrapperAddress](https://github.com/namehash/ensnode/blob/y3drk/chore/ensadmin/fix-ensnamespace-handling-in-registrations-panel/packages/datasources/src/index.ts#L56-L62).
   2. I believe this might further support the opportunity I previously flagged in our recent work on refining the plugin system? Currently we've "smashed together" the data model exposed by our plugins with the data model expected by Ponder. In my thinking, it would be great if we could separate these ideas into distinct layers. The goal would be to make it easier for the `ENSIndexerConfig` to include in its data model all the details of the Datasources that are activated (ex: exact contract addresses that might be determined at runtime), where no Ponder-specific types are included in that data model.

One alternative option to (4) could be for us to push harder into the switch away from GraphQL and into our own REST APIs. For example, we could expose a new REST API in ENSIndexer to get registrations, and then move the logic that calculates the effective owner of a registration out of ENSAdmin and into ENSIndexer instead.

If any questions please let me know. Advice appreciated 👍

### LINKED ISSUE #1021: Refine contract addresses for `ens-test-env` datasource
# Background

1. The `ens-test-env` was previously understood to need dynamically-defined contract deployment addresses (rather than predictable contract deployment addresses as we define for other ENS namespaces).
2. ... therefore, the `datasources` package currently includes logic that reads from the `NEXT_PUBLIC_DEPLOYMENT_ADDRESSES` or `DEPLOYMENT_ADDRESSES` environment variables to dynamically set contract deployment addresses for datasources in the `ens-test-env` namespace.
   1. Note: These environment variables are currently undocumented in https://github.com/namehash/ensnode/blob/main/apps/ensindexer/.env.local.example
4. The `datasources` package is a dependency of the `ensnode-sdk` and `ensnode-react` packages. Both of these packages are used in a frontend context where any environment variable logic should not exist.
5. Existing related issue: https://github.com/namehash/ensnode/issues/837 where we identify how this issue also breaks logic in ENSAdmin, among other points.
# Context

See [related Slack thread](https://namehash.slack.com/archives/C08BMPFUUDD/p1757074758817229) from the day this issue was created.

# Action Items

1. Update logic for the `ens-test-env` datasource such that all all contract addresses are defined explicitly within the `datasources` package.
   1. I note how there's quite a number of TODOs in this `ens-test-env` namespace definition. Perhaps we can also resolve some of those as we work on this?
2. Remove any logic related to reading environment variables from the `datasources` package.
   1. Therefore, all remaining environment variables used by ENSIndexer should always be documented in https://github.com/namehash/ensnode/blob/main/apps/ensindexer/.env.local.example
4. Also fully resolve all goals in https://github.com/namehash/ensnode/issues/837

# Bonus Points

1. While working on this issue, suggest we also revisit this goal: https://github.com/namehash/ensnode/issues/922

shrugs:
closes #1021
closes #837 — no longer relevant

ok so terminology, as i understand it:
- `devnet` : the docker image that runs the l1, l2, etc instances necessary to simulate ensv2 and includes a deterministic deployment of the ens protocol
- `ens-test-env` the developer-facing project that coordinates running the `devnet` docker image along with `ensindexer`

so the namespace could be renamed to `devnet` or stay as `ens-test-env`. there's a lot of nomenclature around `ens-test-env` in the documentation/code at the moment, so i'm not inclined to update it.

### DIFF
--- a/packages/datasources/src/ens-test-env.ts
+++ b/packages/datasources/src/ens-test-env.ts
@@ -20,16 +20,17 @@ import { ResolverABI, ResolverFilter } from "./lib/resolver";
  * for development and testing.
  *
  * @see https://github.com/ensdomains/ens-test-env
+ * @see https://github.com/ensdomains/namechain
  *
  * NOTE: The ens-test-env ENS namespace does not support Basenames, Lineanames, or 3DNS.
- * NOTE: The ens-test-env ENS namespace does not support ENSIP-19 Reverse Resolvers.
+ * NOTE: The ens-test-env ENS namespace does not (yet) support ENSIP-19 Reverse Resolvers.
  */
 export default {
   /**
    * ENSRoot Datasource
    *
-   * Addresses and Start Blocks from ens-test-env
-   * https://github.com/ensdomains/ens-test-env/
+   * Addresses and Start Blocks from Namechain devnet
+   * https://github.com/ensdomains/namechain
    */
   [DatasourceNames.ENSRoot]: {
     chain: ensTestEnvL1Chain,
--- a/packages/datasources/src/lib/chains.ts
+++ b/packages/datasources/src/lib/chains.ts
@@ -1,6 +1,15 @@
 import { Chain, localhost } from "viem/chains";
 
+/**
+ * The Namechain Devnet Default L2 Chain Id
+ * @see https://github.com/ensdomains/namechain/blob/26ad1550f6a119728d56f96b70326e42d6a4bdde/contracts/script/setup.ts#L47
+ */
 const l2ChainId = 0xeeeeee;
+
+/**
+ * The Namechain Devnet Default L1 Chain Id
+ * @see https://github.com/ensdomains/namechain/blob/26ad1550f6a119728d56f96b70326e42d6a4bdde/contracts/script/setup.ts#L48
+ */
 const l1ChainId = l2ChainId - 1;
 
 export const ensTestEnvL1Chain = {


packages/datasources/src/ens-test-env.ts
```diff
@@ -16,29 +13,13 @@ import { WrappedEthRegistrarController as root_WrappedEthRegistrarController } f
...
+ * for development and testing.
  *
- * If the addresses are not available in the environment, we use empty string as a mock to ensure
- * type-correctness: consumers of these ens-test-env Datasources, if using outside of the context
- * of the ens-test-env tool, should validate that an Address is provided, or they may experience
- * undefined runtime behavior.
+ * @see https://github.com/ensdomains/ens-test-env
  *
  * NOTE: The ens-test-env ENS namespace does not support Basenames, Lineanames, or 3DNS.
  * NOTE: The ens-test-env ENS namespace does not support ENSIP-19 Reverse Resolvers.
```
lightwalker-eth, line 25: Could you check if this is still true? I imagine it should support ENSIP-19 Reverse Resolvers now?

packages/datasources/src/ens-test-env.ts
```diff
@@ -16,29 +13,13 @@ import { WrappedEthRegistrarController as root_WrappedEthRegistrarController } f
...
- * lists the addresses of each contract after deployment. These addresses are different in each
- * app and may change over time.
+ * 'ens-test-env' represents a deterministic deployment of the ENS protocol to a local Anvil chain
+ * for development and testing.
  *
- * If the addresses are not available in the environment, we use empty string as a mock to ensure
- * type-correctness: consumers of these ens-test-env Datasources, if using outside of the context
- * of the ens-test-env tool, should validate that an Address is provided, or they may experience
- * undefined runtime behavior.
+ * @see https://github.com/ensdomains/ens-test-env
```
lightwalker-eth, line 22: Should we also include a link here to https://github.com/ensdomains/namechain and include a little description of the relationship between ens-test-env and the devnet in the namechain repo? Seems that would be nice here.

packages/datasources/src/ens-test-env.ts
```diff
@@ -51,17 +32,16 @@ export default {
    * https://github.com/ensdomains/ens-test-env/
```
lightwalker-eth, line 32: I assume we should update this comment now since as I understand these addresses and start blocks actually come from the namechain repo now? https://github.com/ensdomains/namechain

lightwalker-eth, line 32: I note how all the `startBlock` values remain as `0`.

I assume that's perfectly fine for this case, but then perhaps we should refine the comment here? The addresses come from the namechain repo (as I understand it) but the startBlocks we configure here are all being set to 0 because for the ens-test-env there's no meaningful benefit to skipping the indexing of blocks because there's not many blocks to index?

packages/datasources/src/lib/chains.ts
```diff
@@ -0,0 +1,16 @@
+import { Chain, localhost } from "viem/chains";
+
+const l2ChainId = 0xeeeeee;
```
lightwalker-eth, line 3: Could we document where these values come from?

lightwalker-eth (approved):
@shrugs Great to see these updates. Reviewed and shared a few suggestions. Please merge when ready 👍

Issue #1053, opened 2025-09-12
Refine Indexing Status UI


lightwalker-eth:
<img width="2356" height="1482" alt="Image" src="https://github.com/user-attachments/assets/f6913ed0-599a-40bf-b92c-a39b6bd064b5" />

1. Please remove the "#" prefix from all the block numbers we display.
2. There seems to be a special bug in this case. Here we need to both remove the "#" prefix and also the "$" prefix.
3. Currently we put the chain name before the chain icon. Let's please swap this so that we put the chain icon before the chain name.

<img width="2344" height="900" alt="Image" src="https://github.com/user-attachments/assets/05eb7f73-8972-4f07-acb4-d091d03653f2" />

1. Please update each of these so we stop showing the chain name here. Instead, we should show the chain icon for each indexed chain. We also need to make it so that hovering over any of these icons displays the chain name as a tooltip.

PR #1062, opened 2025-09-15
chore (ensadmin): Refine indexing status UI

Y3drk:
Closes Issue #1053 &rarr; its description provides all necessary details

### DIFF
--- a/apps/ensadmin/src/components/indexing-status/block-refs.tsx
+++ b/apps/ensadmin/src/components/indexing-status/block-refs.tsx
@@ -6,7 +6,7 @@ import { RelativeTime } from "@/components/datetime-utils";
 import { getBlockExplorerUrlForBlock } from "@/lib/namespace-utils";
 import { BlockRef, ChainId } from "@ensnode/ensnode-sdk";
 import { fromUnixTime } from "date-fns";
-import { ExternalLink } from "lucide-react";
+import { ExternalLink as ExternalLinkIcon } from "lucide-react";
 
 export interface BlockRefViewModel extends BlockRef {
   get date(): Date;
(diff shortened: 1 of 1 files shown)


apps/ensadmin/src/components/indexing-status/block-refs.tsx
```diff
@@ -43,13 +43,13 @@ function BlockNumber({ chainId, block }: BlockNumberProps) {
         rel="noreferrer noopener"
         className="w-fit text-lg font-semibold flex items-center gap-1 text-blue-600 hover:underline cursor-pointer"
       >
-        #{block.number}
+        {block.number}
         <ExternalLink size={16} className="inline-block flex-shrink-0" />
```
lightwalker-eth, line 47: Suggest renaming (as an import) to `ExternalLinkIcon`. Can you please apply this update everywhere across our code? I find it very confusing to be reviewing code where I read <ExternalLink> and that's actually just an icon. Please see how Jamie actioned this idea in PR 1024.

I'm not sure if the `<a...` here should make use of the new `ExternalLinkWithIcon` from https://github.com/namehash/ensnode/pull/1024 that Jamie created? Advice appreciated.

If so, is there anywhere else in ENSAdmin that we should update for this too?

lightwalker-eth (approved):
@Y3drk Looks good 👍 Just 1 suggestion. Please merge when ready 😄

apps/ensadmin/src/components/indexing-status/block-refs.tsx
Y3drk: I think the best way to apply these changes would be to include them in PR #1024

Issue #1058, opened 2025-09-15
Use Dynamic ENS Namespace on Name Details Page


lightwalker-eth:
<img width="1488" height="1096" alt="Image" src="https://github.com/user-attachments/assets/fcc7d650-c5f9-42f1-85e5-bee9f6228d1f" />

We need to update this logic to always use the ENS Namespace from the active ENSNode connection. Otherwise results will be wrong if the connected ENSNode is not on the mainnet namespace.

# Subtasks

- [ ] This logic also needs to be fixed on `apps/ensadmin/src/app/@actions/name/[name]/page.tsx`

PR #1073, opened 2025-09-16
use namespace id and avatar url hooks

notrab:
I found when building on ENSAdmin that we were calling a lot of functions and having to pass props down (to `ProfileHeader` for the avatarUrl), and then for that same function, get the `namespaceId` which was already available elsewhere in the provider...

Instead, we should provide these values via React Hooks so we can use the data inside the provider. So if this value changes, it changes everywhere, and we have a much better DX around fetching these values.

This PR introduces two new React Hooks (changeset not yet added until I'm happy with them):

* `useAvatarUrl()`
* `useNamespaceId()`

You can see inside `useAvatarUrl()` that we use `useNamespaceId()` 😄  Composition!

Then inside `ProfileHeader` I was able to get the avatar there, and remove a prop type!

```ts
const { data: avatarUrl } = useAvatarUrl({ name });
```

This is in draft because there is still an outstanding issue to fix the original issue #1058 but since `@actions` is at the same level as the `layout.tsx` where the Provider is mounted, we don't yet have that context. I'll need to move some things around to solve that.

Closes https://github.com/namehash/ensnode/issues/1058


apps/ensadmin/src/app/@actions/name/[name]/page.tsx
Comment (notrab) [inline apps/ensadmin/src/app/@actions/name/[name]/page.tsx:1]: @lightwalker-eth ignore all changes in this file. You'll see the fix is implemented, but it doesn't yet work due to another issue I will resolve.

apps/ensadmin/src/app/name/[name]/page.tsx
```diff
@@ -17,29 +16,25 @@ export default function NameDetailPage() {
   const params = useParams();
   const name = decodeURIComponent(params.name as string);
 
-  // TODO: Get the namespace from the active ENSNode connection
-  const namespaceId = ENSNamespaceIds.Mainnet;
+  const { data: namespaceId, isLoading: namespaceLoading } = useNamespaceId();
```
Comment (notrab) [inline apps/ensadmin/src/app/name/[name]/page.tsx:19]: We have the `namespaceId` in the context already, let's use it!

apps/ensadmin/src/app/name/[name]/page.tsx
```diff
@@ -17,29 +16,25 @@ export default function NameDetailPage() {
...
   } = useRecords({
     name,
-    selection: DefaultRecordsSelection.mainnet,
+    selection: DefaultRecordsSelection[namespaceId],
   });
 
-  const avatarUrl = getNameAvatarUrl(name, namespaceId);
-
-  if (isLoading) {
+  if (namespaceLoading || isLoading) {
```
Comment (notrab) [inline apps/ensadmin/src/app/name/[name]/page.tsx:30]: We can make this more explicit by naming `isLoading` to `isRecordsLoading`

apps/ensadmin/src/app/name/[name]/page.tsx
```diff
@@ -17,29 +16,25 @@ export default function NameDetailPage() {
...
     return <NameDetailPageSkeleton />;
   }
 
   return (
     <div className="container mx-auto p-6 max-w-4xl">
       <ProfileHeader
         name={name}
-        avatarUrl={avatarUrl}
```
Comment (notrab) [inline apps/ensadmin/src/app/name/[name]/page.tsx:42]: Yay! One less prop to throw around.

apps/ensadmin/src/hooks/use-avatar-url.ts
```diff
@@ -0,0 +1,27 @@
+"use client";
+
+import { getNameAvatarUrl } from "@/lib/namespace-utils";
+import { Name } from "@ensnode/ensnode-sdk";
+import { useNamespaceId } from "./use-namespace-id";
+
+export interface UseAvatarUrlParameters {
+  name: Name;
+}
+
```
notrab, line 10: Missing jsdoc. Will add if the concept is approved.

notrab:
@shrugs rhats right that there are efforts to go into that direction. This hook would still be useful in those cases I believe where you want to get the values from context regardless. The loading state we could remove for sure, but I kept it in because i think it still serves a purpose.

apps/ensadmin/src/hooks/use-avatar-url.ts
```diff
@@ -0,0 +1,27 @@
...
+    return {
+      data: null,
+      isLoading: true,
+    };
+  }
+
+  const avatarUrl = getNameAvatarUrl(name, namespaceId);
+
+  return {
+    data: avatarUrl,
```
lightwalker-eth, line 24: Would it be against convention to rename `data` to `avatarUrl`?

apps/ensadmin/src/hooks/use-namespace-id.ts
```diff
@@ -0,0 +1,14 @@
...
+import { useENSIndexerConfig } from "@ensnode/ensnode-react";
+
+/**
+ * Hook to get the namespace ID from the active ENSNode connection
+ */
+export function useNamespaceId() {
+  const { data: config, ...rest } = useENSIndexerConfig();
+
+  return {
+    data: config?.namespace ?? ENSNamespaceIds.Mainnet,
```
lightwalker-eth, line 11: Would it be against convention to rename `data` to `namespaceId`?

Also, I believe any UI component that calls `useNamespaceId` should have a dependency that it is a child of somewhere in the DOM where the ENSNode connection is guaranteed to be active and therefore this hook should be guaranteed to never return undefined / null.

I appreciate we might not be ready for this yet, but suggest to at least document a few TODOs here for this idea.

apps/ensadmin/src/hooks/use-namespace-id.ts
```diff
@@ -0,0 +1,14 @@
+import { ENSNamespaceIds } from "@ensnode/datasources";
+import { useENSIndexerConfig } from "@ensnode/ensnode-react";
+
+/**
+ * Hook to get the namespace ID from the active ENSNode connection
+ */
+export function useNamespaceId() {
```
lightwalker-eth, line 7: What do you think about explicitly defining the return type? I think this is nice if it doesn't cause some issue.

apps/ensadmin/src/hooks/use-avatar-url.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { getNameAvatarUrl } from "@/lib/namespace-utils";
+import { Name } from "@ensnode/ensnode-sdk";
+import { useNamespaceId } from "./use-namespace-id";
+
+export interface UseAvatarUrlParameters {
+  name: Name;
+}
+
+export function useAvatarUrl({ name }: UseAvatarUrlParameters) {
```
lightwalker-eth, line 11: What do you think about explicitly defining the return type? I think this is nice if it doesn't cause some issue.

apps/ensadmin/src/hooks/use-avatar-url.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { data: namespaceId, isLoading } = useNamespaceId();
+
+  if (isLoading || !namespaceId) {
+    return {
+      data: null,
+      isLoading: true,
+    };
+  }
+
+  const avatarUrl = getNameAvatarUrl(name, namespaceId);
```
lightwalker-eth, line 21: On both `getNameAvatarUrl` and `useAvatarUrl` suggest to document the following ideas as otherwise it can be confusing depending on someone's familiarity with ENS and the overall ENS ecosystem.

Both of these functions make use of the ENS Metadata Service (https://metadata.ens.domains/docs) to fetch the avatar image for the given name in the given namespace. The ENS Metadata Service acts as proxy that offers deterministic routes for these fetch requests, and then works internally to lookup the true avatar record of the provided name and namespace and then to load the image associated with that true avatar record.

lightwalker-eth (commented):
@notrab This looks like a cool direction. I like it 👍 Shared a few suggestions

apps/ensadmin/src/hooks/use-avatar-url.ts
notrab: Yep, the convention is to have this as `data`. It lends itself from the same conventions in API specs where you return items inside the `data`, whether it be an object, or array.

apps/ensadmin/src/hooks/use-avatar-url.ts
notrab: @lightwalker-eth yeah I can look at that, although we don't on any of the other hooks, which seem to have good type inference. But, I can look at adding this.

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
+import { ENSNamespaceIds } from "@ensnode/datasources";
+import { useENSIndexerConfig } from "@ensnode/ensnode-react";
+import { useQuery } from "@tanstack/react-query";
+
+/**
+ * Hook to get the namespace ID from the active ENSNode connection.
+ *
+ * Returns the ENS namespace identifier for the currently configured ENSNode client.
+ * This determines which ENS network (Mainnet, Sepolia, Holesky, etc.) the application
```
lightwalker-eth, line 9:
```suggestion
 * This determines which ENS namespace (Mainnet, Sepolia, Holesky, etc.) the connected ENSNode
```

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
...
+ * is connected to. Falls back to Mainnet if no configuration is available.
```
lightwalker-eth, line 10:
```suggestion
 * is associated with. Falls back to Mainnet if no ENSNode is actively connected.
```

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @example
+ * ```typescript
+ * import { useNamespaceId } from "@/hooks/useNamespaceId";
+ * import { ENSNamespaceIds } from "@ensnode/datasources";
+ *
+ * function NetworkIndicator() {
+ *   const { data: namespaceId, isLoading, error } = useNamespaceId();
+ *
+ *   if (isLoading) return <div>Loading network...</div>;
```
lightwalker-eth, line 22:
```suggestion
 *   if (isLoading) return <div>Connecting to ENSNode...</div>;
```

- Namespaces are a different concept than networks.
- My understanding is this `isLoading` state should only happen during the process of connecting to an ENSNode?

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
...
+ *   if (error) return <div>Failed to detect network</div>;
```
lightwalker-eth, line 23:
```suggestion
 *   if (error) return <div>Error connecting to ENSNode</div>;
```

Please see related comment above

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ *   const networkName = namespaceId === ENSNamespaceIds.Mainnet ? "Mainnet" :
```
lightwalker-eth, line 25: These are not network names. They are ENS Namespaces

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
...
+ *                      namespaceId === ENSNamespaceIds.Sepolia ? "Sepolia" : "Unknown";
+ *
+ *   return <div>Connected to: {networkName}</div>;
```
lightwalker-eth, line 28: These are not network names. They are ENS Namespaces

apps/ensadmin/src/hooks/useNamespaceId.ts
```diff
@@ -0,0 +1,41 @@
...
+ * is connected to. Falls back to Mainnet if no configuration is available.
+ *
+ * @returns Query result with namespace ID, loading state, and error handling
+ *
+ * @example
+ * ```typescript
+ * import { useNamespaceId } from "@/hooks/useNamespaceId";
+ * import { ENSNamespaceIds } from "@ensnode/datasources";
+ *
+ * function NetworkIndicator() {
```
lightwalker-eth, line 19: Namespaces are not Networks.

apps/ensadmin/src/hooks/useNamespaceId.ts
lightwalker-eth, line 10: I'm not confident I understand this idea of "... Falls back to Mainnet if .." or how the `initialData` works in the logic below.

From first principles: The ENSNode connection is either disconnected, connecting, or connected.

If the ENSNode connection is disconnected or connecting: We can't know the ENS Namespace of the ENSNode yet.

If the ENSNode connection is connected, we are guaranteed to know the ENS Namespace of the connected ENSNode.

Please also note: This is an important idea! ENSAdmin is a special app where it needs to support connecting to any ENSNode in any ENS namespace. For most apps, they will know in advance exactly what ENS namespace is expected from the connected ENSNode. Therefore, most apps have a need to ensure the ENSNode they connect to is returning data for the expected namespace (and not a different namespace for some reason!). There is also an additional goal / benefit that if an app already knows in advance what namespace it expects, more of our logic that requires specifying an ENSNamespace can then work even when an ENSNode connection isn't established yet.

I believe we want to make the following happen:
1. When someone is building an app with ensnode-react, they should be able to _optionally_ customize an `expectedENSNamspaceId`. By default this should be set to Mainnet, but it should also be possible to set it to `null` instead if someone wishes such that no specific ENS namespace is expected from the ENSNode connection once it is established. ENSAdmin is a special app that would need to set this to `null`.
2. If ENSNode is disconnected or connecting (is not actively connected) then `useNamespaceId` should always return the `expectedENSNamspaceId` (which can now be customized based on point 1 above, and may also be null!)
3. When ENSNode is in the process of validating a connection (transitioning from "connecting" to "connected") it must first compare the ENS namespace of the ENSNode server with the `expectedENSNamspaceId`. If the `expectedENSNamspaceId` is non-null then both ENS namespaces must be equal, otherwise the connection should be refused.

apps/ensadmin/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,52 @@
...
+import { getNameAvatarUrl } from "@/lib/namespace-utils";
+import { Name } from "@ensnode/ensnode-sdk";
+import { useQuery } from "@tanstack/react-query";
+
+export interface UseAvatarUrlParameters {
+  name: Name;
+}
+
+/**
+ * Hook to get the avatar image URL for an ENS name using the ENS Metadata Service.
```
lightwalker-eth, line 13:
```suggestion
 * Hook to get the URL for loading the avatar image for an ENS name through the ENS Metadata Service.
```

Important: We are _not_ loading the image URL from the ENS Metadata Service. That sounds like we send an API call to the ENS Metadata Service asking for the URL. That's not true.

Instead, we are simply formatting a string to a URL of the ENS Metadata Service. There are zero API calls here. We are just formatting a string.

All the language here needs to be reviewed and refined to clarify this.

Please study the logic inside `getNameAvatarUrl`.

apps/ensadmin/src/hooks/use-avatar-url.ts
notrab: @shrugs do you have a preference with how return types are defined with the `useQuery` hook? Perhaps we pass a generic for the data response as string | null?

apps/ensadmin/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Makes use of the ENS Metadata Service (https://metadata.ens.domains/docs) to fetch
```
lightwalker-eth, line 15: The ideas here need to be carefully reviewed and refined. The current language here is not correct. Please see related comment above.

This does absolutely no fetching.

This simply builds a URL string that can then be used elsewhere in a fetch operation.

apps/ensadmin/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @example
+ * ```typescript
+ * import { useAvatarUrl } from "@/hooks/useAvatarUrl";
+ *
+ * function ProfileAvatar() {
+ *   const { data: avatarUrl, isLoading, error } = useAvatarUrl({
+ *     name: "vitalik.eth"
+ *   });
+ *
+ *   if (isLoading) return <div>Loading avatar...</div>;
```
lightwalker-eth, line 33: All the messages here for `isLoading`, `error`, and the case that no avatarUrl is defined are incorrect and need to be updated.

notrab: @lightwalker-eth how so? This is just generic behaviour of `useQuery`. Whether or not we have code underneath that always returns a URL shouldn't be tied to this hook in my opinion. We should treat this hook like any other `useQuery` powered hook.

apps/ensadmin/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,52 @@
...
+export function useAvatarUrl({ name }: UseAvatarUrlParameters) {
+  const { data: namespaceId, isLoading: namespaceLoading } = useNamespaceId();
+
+  return useQuery({
+    queryKey: ["avatarUrl", name, namespaceId],
+    queryFn: () => {
+      if (!namespaceId) return null;
+      return getNameAvatarUrl(name, namespaceId);
+    },
+    enabled: !namespaceLoading && !!namespaceId,
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth we'll run into some "rules of react hooks" if we do that I believe. We can't assume another hook is deterministic.

apps/ensadmin/src/app/@actions/name/[name]/page.tsx
```diff
@@ -6,22 +6,18 @@ import { useNamespaceId } from "@/hooks/useNamespaceId";
...
-  return null;
-  // const params = useParams();
-  // const name = decodeURIComponent(params.name as string);
+  const params = useParams();
+  const name = decodeURIComponent(params.name as string);
+  const { data: namespaceId, isLoading } = useNamespaceId();
 
-  // const { data: namespaceId, isLoading: namespaceLoading } = useNamespaceId();
-  // const ensAppUrl = getExternalEnsAppNameUrl(name, namespaceId);
+  const ensAppUrl = namespaceId ? getExternalEnsAppNameUrl(name, namespaceId) : null;
```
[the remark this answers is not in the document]
Comment (lightwalker-eth) [reply apps/ensadmin/src/app/@actions/name/[name]/page.tsx:14]: This comment is a reply to a thread with an open / pending question.

apps/ensadmin/src/hooks/useEnsMetadataServiceAvatarUrl.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @example
+ * ```typescript
+ * import { useEnsMetadataServiceAvatarUrl } from "@/hooks/useEnsMetadataServiceAvatarUrl";
+ *
+ * function ProfileAvatar() {
+ *   const { data: avatarUrl, isLoading, error } = useEnsMetadataServiceAvatarUrl({
+ *     name: "vitalik.eth"
+ *   });
+ *
+ *   if (isLoading) return <div>Constructing avatar URL...</div>;
```
lightwalker-eth, line 33: This comment is on the JSDoc and therefore is raising a concern about the JSDoc.

Given how `useEnsMetadataServiceAvatarUrl` is actually implemented in reality, does this `isLoading` or `error` check ever make any sense? Ex: Can this ever actually happen in reality? If so, please help me understand.

lightwalker-eth (commented):
@notrab Appreciate your help with these questions.

apps/ensadmin/src/hooks/useEnsMetadataServiceAvatarUrl.ts
lightwalker-eth: Thank you. I see now that my question above was not clear. All of that was already well understood. What I'm really trying to ask about is: What will the value of `namespaceId` be if `isLoading` or `error` are truthy. I suppose this isn't so important at the moment though so I'll just mark this as resolved.

apps/ensadmin/src/app/@actions/name/[name]/page.tsx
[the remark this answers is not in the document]
Comment (lightwalker-eth) [reply apps/ensadmin/src/app/@actions/name/[name]/page.tsx:14]: The avatar URL hook is causing no controversy at all. All controversy is coming exclusively from JSDoc accuracy issues and (in this thread) a desire for consistency.

I am not suggesting to put async hooks around all our sync functions. That would be retarded.

The issue is that `getExternalEnsAppNameUrl` is a function of the namespace of the actively connected ENSNode, just in the same way that other hooks in this PR are. Therefore it seems to me we should apply our ideas consistently across our work.

apps/ensadmin/src/app/@actions/name/[name]/page.tsx
```diff
@@ -2,18 +2,15 @@
...
-import { useNamespaceId } from "@/hooks/useNamespaceId";
-import { getExternalEnsAppNameUrl } from "@/lib/namespace-utils";
+import { useENSAppNameUrl } from "@/hooks/use-ens-app-name-url";
 import { useParams } from "next/navigation";
 
 export default function ActionsNamePage() {
-  const params = useParams();
-  const name = decodeURIComponent(params.name as string);
-  const { data: namespaceId, isLoading } = useNamespaceId();
+  const { name } = useParams<{ name: string }>();
```
Comment (lightwalker-eth) [inline apps/ensadmin/src/app/@actions/name/[name]/page.tsx:9]: I assume all is good here, just wanted to confirm it's ok that we removed the line that previously called `decodeURIComponent`. Feel welcome to resolve this comment if all is ok.

apps/ensadmin/src/app/name/[name]/page.tsx
```diff
@@ -1,49 +1,49 @@
...
-  // see: https://github.com/namehash/ensnode/issues/1083
-  const AdditionalTextRecords = ["status", "eth.ens.delegate"];
-  const AllRequestedTextRecords = [
-    ...HeaderPanelTextRecords,
-    ...ProfilePanelTextRecords,
-    ...SocialLinksTextRecords,
-    ...AdditionalTextRecords,
-  ];
+export default function NameDetailPage() {
+  const { name } = useParams<{ name: string }>();
```
Comment (lightwalker-eth) [inline apps/ensadmin/src/app/name/[name]/page.tsx:42]: Duplicate comment:

I assume all is good here, just wanted to confirm it's ok that we removed the line that previously called `decodeURIComponent`. Feel welcome to resolve this comment if all is ok.

apps/ensadmin/src/app/name/[name]/page.tsx
```diff
@@ -1,49 +1,49 @@
...
+  const { namespace } = useActiveENSNodeConfig();
```
Comment (lightwalker-eth) [inline apps/ensadmin/src/app/name/[name]/page.tsx:43]: I understand there's now several strategies available to get the namespace of the actively connected ENSNode instance.

How should it be decided which to use where? It seems confusing and inconsistent to me currently.

apps/ensadmin/src/hooks/use-ens-app-name-url.ts
```diff
@@ -0,0 +1,12 @@
+import { getExternalEnsAppNameUrl } from "@/lib/namespace-utils";
+import { useENSIndexerConfig } from "@ensnode/ensnode-react";
+import { Name } from "@ensnode/ensnode-sdk";
+
+export function useENSAppNameUrl(name: Name) {
+  const { data, ...query } = useENSIndexerConfig();
+
+  return {
+    ...query,
+    data: data?.namespace ? getExternalEnsAppNameUrl(name, data.namespace) : null,
```
lightwalker-eth, line 10: This comment is about `getExternalEnsAppNameUrl`.

Suggest we rename this function from `get...` to `build...` and align the rest of its name with the name of this hook.

Please see my other related comment suggesting an improved name for this hook.

Goal: Cleanup and align the names for deeply connected ideas.

apps/ensadmin/src/hooks/use-ens-app-name-url.ts
```diff
@@ -0,0 +1,12 @@
+import { getExternalEnsAppNameUrl } from "@/lib/namespace-utils";
+import { useENSIndexerConfig } from "@ensnode/ensnode-react";
+import { Name } from "@ensnode/ensnode-sdk";
+
```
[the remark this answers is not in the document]
shrugs: modern preference, especially in a nextjs project, is using kebab-case for more or less every file

 i'll update the files in this pr to use that pattern but perhaps the larger refactor can be jettied to a separate issue

shrugs:
ok i have:
1. fixed the merge conflicts
2. re-named the files in this pr to kebab-case to match best practices (see https://github.com/namehash/ensnode/issues/1094)
3. re-named the `useNamespaceId` hook to `useNamespace`, because it reads better. if that's not desired we can move it back, but this mirrors the `ENSIndexerConfig#namespace` terminology. if `namespaceId` is preferred then we should refactor the ENSIndexerConfig to use `namespaceId` instead of `namespace`
4. re-organized the hooks into `active` and `async` folders
    i. the `active` hooks are ensadmin-specific and provide synchronous access to components that are namespace-contextual
    ii. the `async` hooks are more generic, and propagate the true async nature of the ENSNode's Config to their consumers. these hooks are theoretically going to be used within the ensnode react library or something
5. refactored usage of `useNamespace` — if the component is an ensadmin-specific component under the `RequireActiveENSNodeConfig` parent, it can use the `useActiveNamespace` method and avoid additional checks on status/error
  - so `useENSAppProfileUrl` uses `useNamespace` because it is more generic and not ensadmin specific
  - likewise for `useEnsMetadataServiceAvatarUrl`
6. renamed and documented the `ActiveENSNodeProvider` concept to better communicate its purpose
7. added exhaustive docstrings to every hook/component involved in this pr.

by the way `useEnsMetadataServiceAvatarUrl` isn't used anywhere in the codebase — should it be ? while fixing the numerous merge conflicts due to drift, its original usage may have been lost

open to alternative naming choices for the various decisions made here

lightwalker-eth (approved):
@notrab @shrugs Hey guys, happy to approve all the progress made on this PR. I have some follow up questions that are logged in https://github.com/namehash/ensnode/issues/1096 but that can be handled outside of this PR.

So happy to merge this PR whenever you're ready 👍

apps/ensadmin/src/components/ui/avatar.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-import { getNameAvatarUrl } from "@/lib/namespace-utils";
+import { buildEnsMetadataServiceAvatarUrl } from "@/lib/namespace-utils";
```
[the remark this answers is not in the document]
Y3drk: @notrab 
Understood. Just had a look at it on main, will keep that in mind. Thanks!

Issue #921, opened 2025-08-15
Refine ChainId type definitions


lightwalker-eth:
- Background: https://github.com/namehash/ensnode/pull/902#discussion_r2262750203
- Background: packages/ensnode-sdk/src/shared/types.ts where we define `ChainId` and identify how it is guaranteed to be a positive integer. 
- Seems we should define a new type alias that's something like `DefaultableChainId` which identifies how it also supports `0` as a special value to represent the default chainId. Then, across all our code, wherever 0 is a valid possible chainid value we should change the type to `DefaultableChainId`. This will be very helpful in our code to make it more clear where 0 can and cannot be used. Currently I find it quite confusing when reading the code.

PR #1074, opened 2025-09-17
feat: assume all addresses lowercase

tk-o:
This PR updates Ponder to version 0.12 where all address values provided by Ponder are guaranteed to be lowercase.

We now assume all address values we use in ENSNode to be lowercase. We ensure to create only lowercase addresses from relevant user input and network data.


apps/ensindexer/src/lib/ids.ts
```diff
@@ -48,15 +47,15 @@ export const parseResolverId = (
   if (parts.length === 2) {
     return {
       chainId: null,
-      address: getAddress(parts[0] as Address),
+      address: parts[0] as Address,
```
lightwalker-eth, line 50: Please review anywhere we might document ideas about the resolver id field, including the Drizzle schemas, and explicitly document how the address portion of this field is guaranteed to be in lowercase.

apps/ensindexer/src/lib/ids.ts
```diff
@@ -48,15 +47,15 @@ export const parseResolverId = (
...
       node: parts[1] as Node,
     };
   }
 
   if (parts.length === 3) {
     return {
       chainId: parseInt(parts[0]!),
```
lightwalker-eth, line 57: Maybe we don't need the `!` here anymore? Please ignore if not relevant.

apps/ensindexer/src/lib/interpret-record-values.ts
```diff
@@ -37,7 +37,7 @@ export function interpretNameRecordValue(value: string): NormalizedName | null {
  *   iv. zeroAddress
  * b) an address record value that
  *   i. does not contain null bytes
- *   ii. (if is an EVM address) is checksummed
+ *   ii. (if is an EVM address) is lowercase
```
lightwalker-eth, line 40: Hmm, we need to separate the ideas of "accepting" (and possibly converting) vs "validating" (that a specific format is already true).

apps/ensindexer/src/lib/interpret-record-values.ts
```diff
@@ -58,8 +58,8 @@ export function interpretAddressRecordValue(value: string): string | null {
   // interpret zeroAddress as deletion
   if (isAddressEqual(value, zeroAddress)) return null;
 
-  // otherwise ensure checksummed
-  return getAddress(value);
+  // otherwise ensure lowercase
```
lightwalker-eth, line 61: Hmm, we need to separate the ideas of "accepting" (and possibly converting) vs "validating" (that a specific format is already true).

packages/ensnode-sdk/src/shared/address.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    expect(result).toBe("0x6bd421b6e762d6ad89780eb54b9255f9ab5840bf");
+  });
+
+  it("should throw an error for an invalid EVM address", () => {
+    // arrange
+    const input = "0xinvalid";
+
+    // act & assert
+    expect(() => asLowerCaseAddress(input)).toThrowError(
+      `Cannot make an invalid EVM address "${input}" lowercase.`,
```
lightwalker-eth, line 22:
```suggestion
      `Cannot make an invalid EVM address "${input}" a lowercase address.`,
```

apps/ensindexer/src/lib/interpret-record-values.ts
shrugs: in this case the word ensure means "in this case the word ensure means if it is not already in our preferred format, make it so", not "require"

if this terminology was the cause for confusion, perhaps we can just update that comment to "interpret" or "convert" or "normalize"

packages/ensnode-sdk/src/shared/zod-schemas.ts
```diff
@@ -79,6 +89,72 @@ export const makeChainIdStringSchema = (valueLabel: string = "Chain ID String")
...
+ */
+export const makeDefaultableChainIdStringSchema = (
+  valueLabel: string = "Defaultable Chain ID String",
+) =>
+  z
+    .string({ error: `${valueLabel} must be a string representing a chain ID.` })
+    .pipe(z.coerce.number({ error: `${valueLabel} must represent a non-negative integer (>=0).` }))
+    .pipe(makeDefaultableChainIdSchema(`The numeric value represented by ${valueLabel}`));
+
+const knownCoinTypes = new Set<number>(Object.values(coinNameToTypeMap));
```
lightwalker-eth, line 111: I don't feel comfortable with adding this idea in this PR.

Let's please constrain the validation in this PR just to the more abstract range of values that should be acceptable without validating if a specific value within the theoretically valid range is specifically recognized.

I like the goal here, but only when we make a full dedicated project out of it. I don't like just adding it into our zod schemas in one place in a completely undocumented way.

packages/ensnode-sdk/src/shared/types.ts
```diff
@@ -8,6 +9,18 @@ import { Address } from "viem";
...
+ * Defaultable Chain ID
+ *
+ * Represents a unique identifier for a chain, including
+ * the default chain as defined by ENSIP-19.
+ *
+ * @see https://docs.ens.domains/ensip/19/#annex-supported-chains
+ *
+ * Guaranteed to be a non-negative integer.
+ **/
+export type DefaultableChainId = typeof DEFAULT_EVM_CHAIN_ID | ChainId;
```
lightwalker-eth, line 22: I expected to see the `DefaultableChainId` type being used across our code wherever the chainId value might legitimately be 0?

Goal: Across all our code the reader should explicitly and immediately have 100% confidence on the range of possible values of a chainId based on its type. Is it a `ChainId` or a `DefaultableChainId`?

The reader should not need to think or guess at all with any uncertainty.

lightwalker-eth (commented):
@tk-o Thanks for updates. Reviewed and shared feedback

packages/ensnode-sdk/src/shared/zod-schemas.ts
tk-o: Ok, I'll drop that constraint from the validation schema 👍

packages/ensnode-sdk/src/shared/types.ts
tk-o: This task should be completed in another PR with focus on _resolving_ #921.

PR #1093, opened 2025-09-19
refactor(ensadmin): replace `Date` type with `UnixTimestamp`

### LINKED ISSUE #989: Refine date formatting on "Backfill Status" panel

tk-o:
This PR replaces `Date` type with `UnixTimestamp` for ENSAdmin app.

What's important, date operations performed by `date-fns` require either `Date` object or a number representing unix timestamp in milliseconds. However, our server returns unix timestamp values expressed in seconds. That's why there's internal conversion from unix timestamp value expressed in seconds to a `Date` object.

Also, the PR sets the correct formatting for chain indexing start date.

### Preview

| Backfill | Following |
|--------|--------|
| image | image |

Resolves #989.


shrugs (approved):

lightwalker-eth (approved):
@tk-o Happy to see this! Thank you :+1:

PR #1095, opened 2025-09-20
refactor: refine Indexing Status API response rchema

### LINKED ISSUE #942: Refine Indexing Status API Response Schema
I'm currently reviewing `packages/ensnode-sdk/src/client.ts` in PR 919. Here the logic in `indexingStatus` should be improved. The key issue is that the API response data model needs to include an optional boolean field to identify if the requested max distance was achieved. It's fine to also use the HTTP status code to identify if this is true or false, but in addition to the HTTP status code, this boolean field needs to be explicitly included in the response data model.

As an acceptance criteria for resolving this issue, none of the logic inside `indexingStatus` in client.ts should check the HTTP status code. All the goals of this function should be achieved through the explicit API response data model.

Another acceptance criteria is that an app making use of this `indexingStatus` function should be able to use the response data model to identify if the optionally requested max distance was achieved.

tk-o:
Resolves #942.


packages/ensnode-sdk/src/ensindexer/indexing-status/types.ts
```diff
@@ -239,6 +254,14 @@ export interface ENSIndexerOverallIndexingUnstartedStatus {
...
    */
   chains: Map<ChainId, ChainIndexingQueuedStatus>;
+
+  /**
+   * Max realtime indexing distance
+   *
+   * Optional, if provided, describes the requested max realtime distance,
+   * and how it was not satisfied for the 'unstarted' overall status.
+   */
+  maxRealtimeDistance?: MaxRealtimeIndexingDistance<false>;
```
lightwalker-eth, line 264: I don't feel good about adding this field directly on all of the `ENSIndexerOverallIndexingStatus` types. These types should be a pure function of ENSIndexer and have nothing to do with an optional param sent in the request.

Instead: Suggest making it so the overall indexing status API response is a new "...Response" type with 2 fields:

1. An indexing status field of type `ENSIndexerOverallIndexingStatus` (which is for ENSIndexer and not specific to any request / response).
2. An optional `SatisfiableMaxRealtimeIndexingDistance` field (which is specific to the response for a specific request).

A key goal motivating this suggestion is how I'm thinking (later, not now!) we might start having ENSIndexer store an `ENSIndexerOverallIndexingStatus` in the database every few seconds. Goals for that would include being able to get an idea of the (approximate) indexing status even in the error case and possibly (not sure about this) to work towards a future where our API server never makes any requests directly to the main indexer but instead reads all of its state from ENSDb.

packages/ensnode-sdk/src/ensindexer/indexing-status/zod-schemas.ts
```diff
@@ -7,188 +7,145 @@
...
  * Makes Zod schema for {@link ChainIndexingConfig} type.
  */
 const makeChainIndexingConfigSchema = (valueLabel: string = "Value") =>
-  z.discriminatedUnion("strategy", [
+  z.discriminatedUnion("configType", [
     z.strictObject({
-      strategy: z.literal(ChainIndexingStrategyIds.Indefinite),
+      configType: z.literal(ChainIndexingConfigTypeIds.Indefinite),
       startBlock: makeBlockRefSchema(valueLabel),
       endBlock: z.null(),
```
lightwalker-eth, line 63: Please see related comments asking to completely remove this field.

packages/ensnode-sdk/src/ensindexer/indexing-status/zod-schemas.ts
```diff
@@ -7,188 +7,145 @@
...
 
 /**
- * Makes Zod schema for {@link ChainIndexingStatus} per chain.
+ * Makes Zod schema for {@link ChainIndexingStatusSnapshot} per chain.
  */
 export const makeChainIndexingStatusesSchema = (valueLabel: string = "Value") =>
   z
-    .record(makeChainIdStringSchema(), makeChainIndexingStatusSchema(valueLabel), {
+    .record(makeChainIdStringSchema(), makeChainIndexingStatusSnapshotSchema(valueLabel), {
       error: "Chains configuration must be an object mapping valid chain IDs to their configs.",
```
lightwalker-eth, line 145: Maybe can refine this error message?

packages/ensnode-sdk/src/ensindexer/indexing-status/zod-schemas.test.ts
```diff
@@ -2,313 +2,318 @@ import { describe, expect, it } from "vitest";
...
         // act
-        const notParsed = formatParseError(makeChainIndexingStatusSchema().safeParse(serialized));
+        const notParsed = formatParseError(
+          makeChainIndexingStatusSnapshotSchema().safeParse(serialized),
+        );
 
         // assert
-        expect(notParsed).toBe(`✖ config.startBlock must be before or same as config.endBlock.`);
+        expect(notParsed).toBe(
+          "✖ `config.startBlock` must be before or same as `config.endBlock`.",
```
lightwalker-eth, line 64: Suggest to not match on error messages in such a complete / exact way.

Please apply this idea across this file.

lightwalker-eth (commented):
@tk-o Hey super job on all the big updates here 💪 Overall very happy. Reviewed and shared some suggestions.

Issue #1000, opened 2025-08-28
ENSAdmin Cross-Tab Connection Management


lightwalker-eth:
Reported by @tk-o 🙌 

If you have multiple tabs of ENSAdmin opened in your browser, changing connection in one of them results in changed connection in all of them.

Previously, when the connection was managed by the ensnode URL search param, you could have observe progress across multiple ENSAdmin windows, each connected to a different ENSNode.

Related discussion: https://namehash.slack.com/archives/C086Z6FNBHN/p1756392785023969

PR #1098, opened 2025-09-21
feat(apps/ensadmin): store connection state in URL params

notrab:
Fixes https://github.com/namehash/ensnode/issues/1000

Todo
- [x] Add changesets
- [x] Note follow ups in https://github.com/namehash/ensnode/issues/1123
   - [x] Rename `NEXT_PUBLIC_SERVER_ENSNODE_URLS` to `NEXT_PUBLIC_SERVER_CONNECTION_LIBRARY`

URL Params
- **Active connection tracking**: The URL's `connection` query parameter always reflects the currently active ENSNode connection
- **Automatic URL updates**: When users select connections (existing or new), the URL parameter updates accordingly
- **Deep linking support**: Users can visit pages with `?connection=<url>` to automatically connect
- Removes the logic for storing selected connection in local storage completely

Validation
- **Input sanitization**: All URLs from localStorage are now validated and normalized before use
- **Graceful error handling**: Invalid URLs are silently filtered out and cleaned up from storage
- **Type safety improvements**: Replaced generic `string` types with `UrlString` for better type safety and consistency

State
- **Variable naming**: Renamed variables for clarity (`urls` → `customConnections`, `setUrls` → `storeCustomConnections`) except in the new connection modal, since the field is a `url`. We should probably refactor this into a separate file/component later.
- **Validation on read**: localStorage data is validated and normalized when retrieved, ensuring data integrity
- **Automatic cleanup**: Invalid stored connections are automatically removed to prevent accumulation

Connections
- **Existing connections**: If URL parameter matches an existing connection, it's selected automatically
- **New connections**: If URL parameter is a new valid connection, it's validated, added to custom connections, and selected
- **Invalid connections**: Bad URLs show error toasts and are removed from the URL to maintain clean state
- **Infinite loop prevention**: Failed connection attempts are tracked to prevent retry loops

Toasts
- **Connection toasts**: Users see "Connected to X" for all connection changes (switching or new)
- **Validation feedback**: Clear success/error messages when connections are added or fail validation
- **Dual notifications**: New custom connections show both connection and save confirmation toasts

### DIFF
--- a/apps/ensadmin/src/app/@actions/api/subgraph/page.tsx
+++ b/apps/ensadmin/src/app/@actions/api/subgraph/page.tsx
@@ -4,10 +4,13 @@ import { CopyButton } from "@/components/copy-button";
 import { useSelectedConnection } from "@/hooks/active/use-selected-connection";
 
 export default function ActionsSubgraphCompatPage() {
+  const { validatedSelectedConnection } = useSelectedConnection();
+
   // TODO: we need a broader refactor to recognize the difference between
   // a selected connection being in a valid format or not.
-  const { rawSelectedConnection: baseUrl } = useSelectedConnection();
-  const url = new URL(`/subgraph`, baseUrl).toString();
+  if (!validatedSelectedConnection.isValid) return null;
+
+  const url = new URL(`/subgraph`, validatedSelectedConnection.url).toString();
 
   return (
     <div className="flex w-full max-w-md items-center space-x-2">
--- a/apps/ensadmin/src/app/api/subgraph/page.tsx
+++ b/apps/ensadmin/src/app/api/subgraph/page.tsx
@@ -4,10 +4,21 @@ import { SubgraphGraphiQLEditor } from "@/components/graphiql-editor";
 import { useSelectedConnection } from "@/hooks/active/use-selected-connection";
 
 export default function SubgraphGraphQLPage() {
+  const { validatedSelectedConnection } = useSelectedConnection();
+
   // TODO: we need a broader refactor to recognize the difference between
   // a selected connection being in a valid format or not.
-  const { rawSelectedConnection: baseUrl } = useSelectedConnection();
-  const url = new URL(`/subgraph`, baseUrl).toString();
+  if (!validatedSelectedConnection.isValid) {
+    return (
+      <div className="flex w-full max-w-md items-center space-x-2">
+        <span className="font-mono text-xs select-none text-red-500">
+          Invalid connection URL: {validatedSelectedConnection.error}
+        </span>
+      </div>
+    );
+  }
+
+  const url = new URL(`/subgraph`, validatedSelectedConnection.url).toString();
 
   return <SubgraphGraphiQLEditor url={url} />;
 }


lightwalker-eth (commented):
@notrab Hey good to see the refinements here. Reviewed and shared feedback 👍

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -95,26 +117,35 @@ function _useENSNodeConnections() {
...
     active,
     addConnection,
     addAndSelectConnection,
     removeConnection,
-    selectConnection,
   };
 }
 
-export const [ENSNodeConnectionsProvider, useENSNodeConnections] = constate(_useENSNodeConnections);
+function ENSNodeConnectionsProviderWithSuspense({
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth my plan is to remove this as it was a workaround because of our server initiated loading.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -95,26 +117,35 @@ function _useENSNodeConnections() {
...
+    <Suspense fallback={null}>
+      <ENSNodeConnectionsProviderInner>{children}</ENSNodeConnectionsProviderInner>
+    </Suspense>
+  );
+}
+
+const [ENSNodeConnectionsProviderInner, useENSNodeConnections] = constate(_useENSNodeConnections);
+
+export {
+  ENSNodeConnectionsProviderWithSuspense as ENSNodeConnectionsProvider,
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth this is type aliasing so the consumer side of the component doesn't leak the internal Suspense handling. It's cleaner and hides the implementation detail of using Suspense. This is quite common, and means if we remove the suspense stuff (we likely will when we move away from Next.js) that we don't have to update the public consumer of this, just anything with Suspense in it.

apps/ensindexer/src/api/index.ts
```diff
@@ -46,8 +46,8 @@ app.onError((error, ctx) => {
 
 // use root to redirect to the environment's ENSAdmin URL configured to connect back to the environment's ENSNode Public URL
 app.use("/", async (ctx) => {
-  const ensAdminRedirectUrl = new URL("connect", config.ensAdminUrl);
-  ensAdminRedirectUrl.searchParams.set("ensnode", config.ensNodePublicUrl.href);
+  const ensAdminRedirectUrl = new URL(config.ensAdminUrl);
+  ensAdminRedirectUrl.searchParams.set("connecton", config.ensNodePublicUrl.href);
```
shrugs, line 50: `connection`*

apps/ensadmin/src/components/connections/connection-selector.tsx
```diff
@@ -37,23 +38,103 @@ import { useENSNodeConnections } from "@/hooks/ensnode-connections";
...
+      params.set(CONNECTION_PARAM_KEY, url);
+      router.replace(`?${params.toString()}`);
+    },
+    [router, searchParams],
+  );
+
+  useEffect(() => {
+    const connectionParam = searchParams.get(CONNECTION_PARAM_KEY);
+
+    // If no connection parameter exists and we have an active connection, update URL
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth mostly belt and braces that if this component is mounted, it still can recover somewhat.

apps/ensadmin/src/components/connections/connection-selector.tsx
```diff
@@ -37,23 +38,103 @@ import { useENSNodeConnections } from "@/hooks/ensnode-connections";
...
+    if (!connectionParam && activeENSNodeUrl && connections.length > 0) {
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth mostly belt and braces that if this component is mounted, it still can recover somewhat.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -1,92 +1,114 @@
...
+
+    // Clean up localStorage if we found invalid URLs
+    if (validatedUrls.length !== rawCustomConnectionUrls.length) {
+      storeCustomConnections(validatedUrls);
+    }
+
+    return validatedUrls;
+  }, [rawCustomConnectionUrls, storeCustomConnections]);
+
   const connections = useMemo(
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth I didn't really look at this part since I didn't change anything, but I can see the confusion myself. I've updated it.

apps/ensadmin/src/components/require-active-ensnode-connection.tsx
```diff
@@ -1,14 +1,106 @@
...
+    availableConnections,
+    active,
+    addAndSelectCustomConnection: _addAndSelectConnection,
+  } = useAvailableENSNodeConnections();
+
+  const addConnectionFromUrl = useMutation({
+    mutationFn: _addAndSelectConnection,
+  });
+
+  const [existingConnectionUrl, setExistingConnectionUrl] = useState<string | null>(null);
```
lightwalker-eth, line 27: I believe there's many places where we might make use of `UrlString`

apps/ensadmin/src/components/connections/default-connections-list.tsx
```diff
@@ -0,0 +1,55 @@
+"use client";
+
+import { CopyButton } from "@/components/copy-button";
+import { DropdownMenuItem, DropdownMenuLabel } from "@/components/ui/dropdown-menu";
+import { cn } from "@/lib/utils";
+
+interface Connection {
+  url: string;
+  isDefault: boolean;
```
lightwalker-eth, line 9:
```suggestion
  fromServerLibrary: boolean;
```

For background, please see related comments

apps/ensadmin/src/components/connections/default-connections-list.tsx
```diff
@@ -0,0 +1,55 @@
+"use client";
+
+import { CopyButton } from "@/components/copy-button";
+import { DropdownMenuItem, DropdownMenuLabel } from "@/components/ui/dropdown-menu";
+import { cn } from "@/lib/utils";
+
+interface Connection {
+  url: string;
```
lightwalker-eth, line 8: Suggest we make all of these urls a `URL` everywhere at a data-model layer and then when we display them in the UI we can make use of the beautify url utility function you already created for the name detail page.

apps/ensadmin/src/components/connections/default-connections-list.tsx
```diff
@@ -0,0 +1,55 @@
+"use client";
+
+import { CopyButton } from "@/components/copy-button";
+import { DropdownMenuItem, DropdownMenuLabel } from "@/components/ui/dropdown-menu";
+import { cn } from "@/lib/utils";
+
+interface Connection {
```
lightwalker-eth, line 7:
```suggestion
interface ConnectionOption {
```

Goal: we have so many "connection" -- this makes it more clear it's only a connection option.

lightwalker-eth, line 7: I note how we are defining this interface in multiple files. Both here and in `apps/ensadmin/src/components/connections/custom-connections-list.tsx`.

It seems to me that it should be defined in `apps/ensadmin/src/hooks/ensnode-connections.tsx` and exported.

It also seems to me that the data model of some of the objects returned by `useAvailableENSNodeConnections` should match this interface.

apps/ensadmin/src/components/connections/default-connections-list.tsx
```diff
@@ -0,0 +1,55 @@
...
+  isDefault: boolean;
+}
+
+interface DefaultConnectionsListProps {
+  availableConnections: Connection[];
+  activeConnectionUrl: string;
+  onSelectDefaultConnection: (url: string) => void;
+}
+
+export function DefaultConnectionsList({
```
lightwalker-eth, line 18: Suggest to rename all the references to terminology such as "default" or "active" based on feedback in other comments.

apps/ensadmin/src/lib/env.ts
```diff
@@ -119,7 +119,16 @@ export function defaultEnsNodeUrls(): Array<URL> {
   }
 
   try {
-    const urlList = envVarValue.split(",").map((maybeUrl) => parseUrl(maybeUrl));
+    const urlList = envVarValue
+      .split(",")
+      .map((maybeUrl) => maybeUrl.trim()) // Remove whitespace
+      .filter((maybeUrl) => maybeUrl.length > 0) // Remove empty strings
+      .map((maybeUrl) => parseUrl(maybeUrl))
```
lightwalker-eth, line 126: Suggest replacing all references to `parseUrl` in this file (there are many) with calls to `normalizeUrl`. It seems we should put all incoming URLs through the same input processing logic.

lightwalker-eth (commented):
@notrab Hey excited to see the progress and direction this is taking. Here's some additional feedback. I appreciate this feedback involves renaming many things but I think this language is an important and valuable step to invest in 👍 Thanks

apps/ensindexer/src/api/index.ts
```diff
@@ -47,7 +47,7 @@ app.onError((error, ctx) => {
 // use root to redirect to the environment's ENSAdmin URL configured to connect back to the environment's ENSNode Public URL
 app.use("/", async (ctx) => {
   const ensAdminRedirectUrl = new URL(config.ensAdminUrl);
-  ensAdminRedirectUrl.searchParams.set("connecton", config.ensNodePublicUrl.href);
+  ensAdminRedirectUrl.searchParams.set("connection", config.ensNodePublicUrl.href);
```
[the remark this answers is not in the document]
notrab: The const is part of ENSAdmin and this is a separate package. I'd didn't do it for that reason. 

I think it's ok not to in this instance.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -99,53 +123,67 @@ function _useENSNodeConnections() {
...
     if (!currentConnection) return new URL(first);
     if (!isInConnections(currentConnection)) return new URL(first);
     return new URL(currentConnection);
-  }, [hydrated, connections, currentConnection, isInConnections]);
+  }, [hydrated, availableConnections, currentConnection, isInConnections]);
 
   return {
-    connections,
+    availableConnections,
     active,
```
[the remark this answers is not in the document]
notrab: It's just the selected URL. All of these are that since fetching statuses and config happen elsewhere once it knows the selected URL - which is output here.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -28,49 +29,72 @@ const validateAndNormalizeUrls = (urls: UrlString[]): UrlString[] => {
     .filter((url, index, array) => array.indexOf(url) === index); // remove duplicates
```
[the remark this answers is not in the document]
notrab: Good idea

apps/ensadmin/src/components/connections/connection-selector.tsx
```diff
@@ -1,80 +1,74 @@
...
-    removeConnection,
-    selectConnection,
-  } = useENSNodeConnections();
+    availableConnections,
+    addAndSelectCustomConnection: _addAndSelectCustomConnection,
+    removeCustomConnection,
+  } = useAvailableENSNodeConnections();
   const activeENSNodeUrl = useActiveENSNodeUrl().toString();
-  const addAndSelectConnection = useMutation({ mutationFn: _addAndSelectConnection });
+  const addAndSelectConnection = useMutation({
```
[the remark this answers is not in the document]
notrab: @shrugs done, thanks!

apps/ensadmin/src/components/require-active-ensnode-connection.tsx
```diff
@@ -1,14 +1,106 @@
...
+
+    setExistingConnectionUrl(currentConnection);
+  }, [searchParams, existingConnectionUrl]);
+
+  // Handle URL parameter synchronization and connection loading
+  useEffect(() => {
+    const connectionParam = searchParams.get(CONNECTION_PARAM_KEY);
+
+    // If no connection parameter exists and we have an active connection, update URL
+    if (!connectionParam && active && availableConnections.length > 0) {
```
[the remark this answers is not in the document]
notrab: thanks @shrugs 🤝

apps/ensadmin/src/lib/env.ts
```diff
@@ -119,7 +119,16 @@ export function defaultEnsNodeUrls(): Array<URL> {
...
+      .filter(
+        (url, index, array) =>
```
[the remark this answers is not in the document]
notrab: @shrugs updated. Let me know if this looks better to you.

apps/ensadmin/src/hooks/ensnode-connections.tsx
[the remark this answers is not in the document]
notrab: @lightwalker-eth following up:

The current `selectedConnection` is just step 1 of that step 2 process. The naming change `selectedConnection` better reflects this reality - it's a selected URL, not an active/working connection yet.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -28,49 +29,72 @@ const validateAndNormalizeUrls = (urls: UrlString[]): UrlString[] => {
...
+ * - Each URL is in normalizeUrl form
+ * - All URLs are unique (no duplicates)
+ * - Contains at least 1 URL
+ *
+ * These invariants are maintained by:
+ * 1. defaultEnsNodeUrls() already validates and ensures at least 1 URL
+ * 2. Converting to string then normalizing ensures consistent format
+ * 3. validateAndNormalizeUrls removes any potential duplicates and invalid URLs
+ */
+const DEFAULT_CONNECTION_URLS = (() => {
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth that makes sense. I think that's clearer, and I've updated the Vercel environment variables in prep for merging this. I'll clean up the old name once we do in fact do a release.

apps/ensadmin/src/components/require-active-ensnode-connection.tsx
```diff
@@ -1,14 +1,106 @@
...
+
+    if (failedConnectionUrls.has(connectionParam)) return;
+
+    // Check if connection URL already exists in available connections (both default and custom)
+    const existingConnection = availableConnections.find((conn) => conn.url === connectionParam);
+    if (existingConnection) {
+      return;
+    }
+
+    // Try to add connection from URL parameter
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth yeah, that's correct. If someone goes to a URL with the `connection` param in the URL, we check if it exists in the "connection library". If it doesn't exist, we add it. We could have somewhere the concept of a "Share" since this would add the connection to another users connection library.

apps/ensadmin/src/hooks/active/use-active-ensnode-url.tsx
```diff
@@ -16,7 +16,7 @@ import { useENSNodeConnections } from "@/hooks/ensnode-connections";
  * @throws Error if no active ENSNode Connection is available
  */
 export function useActiveENSNodeUrl() {
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth there were a bunch of changes made by @shrugs that moved some of the files into an `active` folder. What shall we do there?

apps/ensadmin/src/components/connections/ensnode-url-validator.ts
```diff
@@ -1,9 +1,12 @@
+import { normalizeUrl } from "@/lib/url-utils";
+
 type ValidationResult = { isValid: true; error?: never } | { isValid: false; error: string };
 
 // TODO: more advanced validation (i.e. confirm ENSNode status response, version numbers...)
 export async function validateENSNodeUrl(url: string): Promise<ValidationResult> {
```
lightwalker-eth, line 6: Feels like the logic for `validateENSNodeUrl` should:
1. Move out of this file and into `@/lib/url-utils` where other similar functions live.
2. Be renamed to something like `isValidENSNodeConnectionUrl`.
3. Should be synchronous, not async.
4. The same logic that we use for validating ENSNode connection URLs from user input should be used to validate ENSNode connection URLs everywhere else we validate one. In other words, I believe the pipeline should always be: (some input source, including environment variables, local storage, user input fields, params in the URL, etc) -> normalizeUrl -> `isValidENSNodeConnectionUrl`.

In other words:
1. Only a subset of `UrlString` pass `normalizeUrl` without throwing an error.
2. Only a subset of the `UrlString` that pass `normalizeUrl` will also pass `isValidENSNodeConnectionUrl`.
3. Everywhere in our data model that has a connection option url or the selected connection url should guarantee that url passes `isValidENSNodeConnectionUrl`.
4. All the fields where this guarantee applies should have their docs updated to reflect this invariant.

Goal: Align invariant enforcement in a consistent way from all input sources and across related data models.

apps/ensadmin/src/hooks/active/use-selected-ensnode-url.tsx
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * This hook provides synchronous access to the URL of the currently selected ENSNode.
+ * If the selected ENSNode's Config is not synchronously available, components using
+ * this hook will throw. Components that use this hook should be a child of
+ * RequireActiveENSNodeConfig such that the selected ENSNode's config is synchronously
+ * available during render. This simplifies state in components that only make sense
+ * within the context of a selected ENSNode.
+ *
+ * @returns The selected ENSNode connection URL
+ * @throws Error if no selected ENSNode Connection is available
```
lightwalker-eth, line 16:
```suggestion
 * @throws Error if no ENSNode connection is selected
```

apps/ensadmin/src/hooks/active/use-selected-ensnode-url.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { useAvailableENSNodeConnections } from "@/hooks/ensnode-connections";
+
+/**
+ * Hook to get the currently selected ENSNode connection URL synchronously.
+ *
+ * This hook provides synchronous access to the URL of the currently selected ENSNode.
+ * If the selected ENSNode's Config is not synchronously available, components using
+ * this hook will throw. Components that use this hook should be a child of
+ * RequireActiveENSNodeConfig such that the selected ENSNode's config is synchronously
```
lightwalker-eth, line 11: Please see related comment suggesting to rename `RequireActiveENSNodeConfig`

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -41,32 +34,34 @@ const validateAndNormalizeUrls = (urls: UrlString[]): UrlString[] => {
  * 2. Converting to string then normalizing ensures consistent format
  * 3. validateAndNormalizeUrls removes any potential duplicates and invalid URLs
  */
-const DEFAULT_CONNECTION_URLS = (() => {
+const serverConnectionLibrary = (() => {
   const rawUrls = defaultEnsNodeUrls().map((url) => url.toString());
   const validatedUrls = validateAndNormalizeUrls(rawUrls);
```
lightwalker-eth, line 39: Can you please double check our new logic for how we build `defaultEnsNodeUrls`? I believe there should be no need to call `validateAndNormalizeUrls` or to validate `validatedUrls.length === 0` a few lines below this anymore.

In my understanding, the validation each of these responsibilities should live inside `defaultEnsNodeUrls`.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -1,36 +1,29 @@
...
-  return urls
-    .filter(isValidUrl)
-    .map(normalizeUrl)
-    .filter((url, index, array) => array.indexOf(url) === index); // remove duplicates
+  return uniq(urls.filter(isValidUrl).map(normalizeUrl));
 };
 
 /**
- * Default ENSNode connection URLs with guaranteed invariants:
+ * Server connection library - ENSNode connection URLs provided by the server with guaranteed invariants:
```
lightwalker-eth, line 26: I believe the ideas in the JSDoc for `serverConnectionLibrary` here should be merged into the JSDoc for `defaultEnsNodeUrls`.

Ultimately I believe the `serverConnectionLibrary` function here should be removed and replaced with a call directly to `defaultEnsNodeUrls`.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -41,32 +34,34 @@ const validateAndNormalizeUrls = (urls: UrlString[]): UrlString[] => {
...
   return validatedUrls;
 })();
 
-const CONNECTION_PARAM_KEY = "connection";
-
 function _useAvailableENSNodeConnections() {
   const hydrated = useHydrated();
+  const router = useRouter();
   const searchParams = useSearchParams();
   const currentConnection = searchParams.get(CONNECTION_PARAM_KEY);
```
lightwalker-eth, line 54:
```suggestion
  const rawSelectedConnection = searchParams.get(CONNECTION_PARAM_KEY);
```

I assume that's fair?

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
+  const selectedConnection = useMemo<URL | null>(() => {
+    // no selected ensnode connection in server environments
     if (!hydrated) return null;
 
-    // NOTE: guaranteed to have a valid set of `availableConnections` here, on the client
-    // NOTE: guaranteed to have at least 1 connection because defaults must have length > 0
-    const first = availableConnections[0].url;
+    // NOTE: guaranteed to have a valid set of `connectionLibrary` here, on the client
+    // NOTE: guaranteed to have at least 1 connection because server library must have length > 0
+    const first = connectionLibrary[0].url;
```
lightwalker-eth, line 151:
```suggestion
    const defaultSelectedConnection = connectionLibrary[0].url;
```

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
 
     if (!currentConnection) return new URL(first);
     if (!isInConnections(currentConnection)) return new URL(first);
     return new URL(currentConnection);
```
lightwalker-eth, line 155: It seems we haven't normalized / validated the `rawSelectedConnection`.

Additionally, in the case that the `rawSelectedConnection` is not the same as the normalized `rawSelectedConnection` then I believe we should automatically redirect to the normalized version of the `rawSelectedConnection`?

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
-  }, [hydrated, availableConnections, currentConnection, isInConnections]);
+  }, [hydrated, connectionLibrary, currentConnection, isInConnections]);
+
+  const updateCurrentConnectionParam = useCallback(
```
lightwalker-eth, line 158: I see logic for updating the `CONNECTION_PARAM_KEY` in multiple places. Please see my other related comment about this. It would be nice if we could move all the logic for updating this in a single place.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
+    if (!currentConnection) return;
+    if (failedConnectionUrls.has(currentConnection)) return;
+
+    // Check if connection URL already exists in connection library
+    const existingConnection = connectionLibrary.find((conn) => conn.url === currentConnection);
+    if (existingConnection) {
+      return;
+    }
+
+    // Automatically add connection from URL parameter to library (enables shareable connection links)
```
lightwalker-eth, line 192: I don't feel good about automatically adding URLs to the custom connection library.

It should be possible to connect to an ENSNode instance (via a hyperlink) and not have it be added to your custom connection library.

Please also note how this means the `selectedConnection` is not necessarily inside the connection library.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
+    addCustomConnection(currentConnection)
+      .then((addedUrl) => {
+        updateCurrentConnectionParam(addedUrl);
+        toast.success(`URL saved to connection library`);
+        toast.success(`Connected to ${addedUrl}`);
+      })
+      .catch((error) => {
```
lightwalker-eth, line 199: I don't feel confident about the error catching logic here.

We should note how we are not even attempting to connect to anywhere here, so I don't understand things such as:
* "Failed to connect ..."
* `setFailedConnections` (what failed about the connection exactly?)

Can you please review the related code to identify exactly what cases can actually cause an error to be caught here and then fully align the logic here with those possible cases?

Thanks

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
+        toast.error(`Failed to connect: ${error.message}`);
+
+        // Track this as a failed connection to prevent retry loop
+        setFailedConnections((prev) => new Set(prev).add(currentConnection));
+
+        // Remove invalid connection param from URL
```
lightwalker-eth, line 205: A few ideas here:
1. If we remove the invalid connection param from the URL, doesn't that just mean that we will trigger the logic that updates the connection param again to be the default connection? Therefore, it seems best not to send the user through that whole process flow and instead automatically redirect them directly to the default connection here? Appreciate your advice.
2. Is there a special reason we can't use `updateCurrentConnectionParam` here?

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -99,53 +123,67 @@ function _useENSNodeConnections() {
...
 }
 
-function ENSNodeConnectionsProviderWithSuspense({
+/**
+ * Provider for available ENSNode connections (both default and custom).
+ *
+ * Wraps the inner provider with Suspense boundary to handle the async nature
+ * of useSearchParams() which can suspend during SSR/hydration.
+ *
+ * Provides access to:
```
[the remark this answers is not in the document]
lightwalker-eth: This is exported and I understand that this is destined for use by others.

Let's build a culture of strong docs.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -123,52 +141,121 @@ function _useAvailableENSNodeConnections() {
...
-    addAndSelectCustomConnection,
     removeCustomConnection,
   };
 }
 
+const [AvailableENSNodeConnectionsProviderInner, useAvailableENSNodeConnections] = constate(
+  _useAvailableENSNodeConnections,
+);
+
+export { useAvailableENSNodeConnections };
```
lightwalker-eth, line 244: Let's please create strong JSDocs on this export.

apps/ensadmin/src/hooks/ensnode-connections.tsx
```diff
@@ -41,32 +34,34 @@ const validateAndNormalizeUrls = (urls: UrlString[]): UrlString[] => {
...
   const currentConnection = searchParams.get(CONNECTION_PARAM_KEY);
   const [rawCustomConnectionUrls, storeCustomConnections] = useLocalstorageState<UrlString[]>(
     CUSTOM_CONNECTIONS_LOCAL_STORAGE_KEY,
     [],
   );
 
-  // Validate and normalize URLs from localStorage
-  const customConnections = useMemo(() => {
+  const [existingConnectionUrl, setExistingConnectionUrl] = useState<UrlString | null>(null);
+  const [failedConnectionUrls, setFailedConnections] = useState<Set<UrlString>>(new Set());
```
lightwalker-eth, line 61: Why do we need to track failedConnectionUrls at all? Couldn't this idea be completely removed from our code and to solve the same goal just through the "existing" connection Url concept?

apps/ensadmin/src/lib/env.ts
```diff
@@ -102,33 +105,31 @@ const DEFAULT_ENSNODE_URL =
...
- * Get list of URLs for default ENSNode instances.
+ * Get list of URLs for server ENSNode instances (ServerConnectionLibrary).
  *
- * @returns a list (with at least one element) of URLs for default ENSNode instances
+ * @returns a list (with at least one element) of URLs for server ENSNode instances
  */
 export function defaultEnsNodeUrls(): Array<URL> {
-  const envVarName = "NEXT_PUBLIC_DEFAULT_ENSNODE_URLS";
-  let envVarValue = process.env.NEXT_PUBLIC_DEFAULT_ENSNODE_URLS;
+  const envVarName = "NEXT_PUBLIC_SERVER_ENSNODE_URLS";
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth I've already updated Vercel with the same values but with the same key so we can transition then remove the old ones. Updated that issue too.

apps/ensadmin/src/lib/env.ts
```diff
@@ -102,33 +105,31 @@ const DEFAULT_ENSNODE_URL =
   "https://api.alpha.ensnode.io,https://api.alpha-sepolia.ensnode.io,https://api.mainnet.ensnode.io,https://api.sepolia.ensnode.io,https://api.holesky.ensnode.io";
 
 /**
- * Get list of URLs for default ENSNode instances.
+ * Get list of URLs for server ENSNode instances (ServerConnectionLibrary).
  *
- * @returns a list (with at least one element) of URLs for default ENSNode instances
+ * @returns a list (with at least one element) of URLs for server ENSNode instances
  */
 export function defaultEnsNodeUrls(): Array<URL> {
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth it started off me trying to be consistent with others contributing to the repo. We should have some linter rules 😿 

CleanShot 2025-09-27 at 09 56 43@2x

apps/ensadmin/src/hooks/ensnode-connections.tsx
notrab: @lightwalker-eth ok, seems fair. Perhaps we can add a button inside the toast as a follow-up which asks if a user wants to save this to their connections?

apps/ensadmin/src/hooks/ensnode-connections.tsx
notrab: @lightwalker-eth this has been removed as part of the request to remove the logic which automatically adds to the connection library on share.

apps/ensadmin/src/hooks/ensnode-connections.tsx
notrab: @lightwalker-eth this is all removed now.

apps/ensadmin/src/hooks/ensnode-connections.tsx
notrab: @lightwalker-eth I've moved the use of this const into a single hook, and removed the constants file so things are better scoped.

apps/ensadmin/src/app/@actions/api/subgraph/page.tsx
```diff
@@ -1,10 +1,12 @@
...
 
 import { CopyButton } from "@/components/copy-button";
-import { useActiveENSNodeUrl } from "@/hooks/active/use-active-ensnode-url";
+import { useSelectedENSNodeConnection } from "@/hooks/active/use-selected-ensnode-connection";
 
 export default function ActionsSubgraphCompatPage() {
-  const baseUrl = useActiveENSNodeUrl();
+  // TODO: we need a broader refactor to recognize the difference between
+  // a selected connection being in a valid format or not.
+  const baseUrl = useSelectedENSNodeConnection().rawSelectedConnection;
```
notrab, line 9:
```suggestion
  const { rawSelectedConnection : baseUrl} = useSelectedENSNodeConnection();
```

@lightwalker-eth is there any reason we can't format it like this how we do other hook usage?

lightwalker-eth (approved):
@notrab Excited for all the improvements here 🚀

Issue #1104, opened 2025-09-22
Simplify ENSIndexerConfig using SUBGRAPH_COMPAT


shrugs:
1. parse the SUBGRAPH_COMPAT env variable with zod
2. if `true`, provide the following 'default' ENSIndexerConfig:

```
PLUGINS=subgraph
LABEL_SET=subgraph
LABEL_SET_VERSION=0
```

3. if `false`, provide the following 'default' ENSIndexerConfig:

```
PLUGINS=subgraph
LABEL_SET=whatever the latest is
LABEL_SET_VERSION=whatever the latest is
```

this defaulting could be structured as a zod-schema-factory that accepts a set of keyed defaults

```ts
export function makeENSIndexerConfigSchemaWithDefaults(defaults: { plugins: string }) {
  const PluginsSchema = z.whatever().default(defaults.plugins);
}
```

or with some simple logic when constructing the input like so:

```ts
const environment =  {
  plugins: process.env.PLUGINS || defaults.plugins,
  // ...
}
```

4. the `config.subgraphCompatible` property tracks `SUBGRAPH_COMPAT`
5. add invariant that checks that if `config.subgraphCompatible` is `true`, the rest of the config is subgraph-compatible (use existing `isSubgraphCompatible` derivation, just inverted logic)

PR #1119, opened 2025-09-24
Introduce SUBGRAPH_COMPAT behavior

shrugs:
closes https://github.com/namehash/ensnode/issues/1104

`SUBGRAPH_COMPAT` is now user-configurable, reflected in the internal docs

- [ ] need to do a pass over the terraform
- [ ] need to do a pass over the docs site


lightwalker-eth (approved):
Approving on the assumption it will be easier to review PRs 1118 and 1119 together.

PR #1126, opened 2025-09-26
hotfix: correct boolean for namewrapper labelname materialization & fix TextChanged decoding

shrugs:
- [x] must have lost the `!` during a previous refactor
- [x] resolver entities out-of-order?
- [x] resolver entities sometimes have `texts` when they shouldn't

```gql
{
  textChangeds(where:{resolver_:{address:"0x226159d592e2b063810a10ebf6dcbada94ed68b8"}}) {
    id
    key
    value
  }
}
```

subgraph
```json
{
  "data": {
    "textChangeds": []
  }
}
```

ensnode
```json

```

so basically the ens subgraph has NO `TextChanged` events from this old public resolver `0x226159d592e2b063810a10ebf6dcbada94ed68b8` but ensnode is seeing them. this is the older TextChanged event, without the value. subgraph literally isn't seeing them (no textChanged entities) but _IS_ seeing the other events (Multicoin, Address, etc)

ah i think it's because despite being called `indexedKey` and the LegacyPublicResolver ABI having `indexed` in its signature, the actual event emitted doesn't have that param indexed:

from https://etherscan.io/address/0x226159d592e2b063810a10ebf6dcbada94ed68b8#code
`event TextChanged(bytes32 indexed node, string indexedKey, string key);`

but both subgraph and ensnode specify the indexed abi

https://github.com/ensdomains/ens-subgraph/blob/master/subgraph.yaml#L109C19-L109C69
```yaml
        - event: "TextChanged(indexed bytes32,indexed string,string)"
          handler: handleTextChanged
```

```ts
  ponder.on(
    "Resolver:TextChanged(bytes32 indexed node, string indexed indexedKey, string key)",
    handleTextChanged,
  );
```

and perhaps ponder is indexing the unindexed version of that event anyway? perhaps a change in 0.12? extra events would also cause the 'out of order' resolver issue, if there are additional resolver entities perceived by the indexer.


shrugs:
merging this as hotfix

pnpm-workspace.yaml
```diff
@@ -14,7 +14,7 @@ catalog:
   typescript: ^5.7.3
   viem: ^2.22.13
   vitest: ^3.2.4
-  ponder: 0.12.24
+  ponder: 0.13.8
```
lightwalker-eth, line 17: Let's add a changeset for any Ponder version upgrades to ensindexer 👍

Issue #1106, opened 2025-09-22
Refactor `resolution` schema/plugin


shrugs:
- make core-plugin-independent
  - add entity to track ResolverNameRecord by node by resolver AccountId
  - remove any drizzle relationships back to core plugins
- re-implement all Resolver handlers to store records, remove any conditional `INDEX_ADDITIONAL_RESOLVER_RECORDS` branches
- update terraform config
- queue up environment variable changes for railway environments

PR #1128, opened 2025-09-26
Protocol Acceleration plugin

shrugs:
closes https://github.com/namehash/ensnode/issues/1106

- [x] chose `protocol-acceleration` (`ProtocolAcceleration`) over `resolution`
- [x] makes the `protocol-acceleration.schema.ts` independent of core schema, merges
  - primary names
  - resolver records (inc. name)
  - resolver records relations
- [x] need to abstract registryold migration status to a core-plugin-independent schema
  - can make this specific to mainnet, since it's the only chain that has the registry migration
  - or embed it in protocol-acceleration plugin so it's appropriately isolated?
- [x] is the `LegacyReverseResolver` handler necessary? presumably it's handled by the more general Resolver filter, right?
- [x] consider getting rid of all of the `make*Id` functions in protocol-acceleration schema and just using composite primary keys
- [x] need to move domain-resolver relationships to protocol-acceleration plugin and extract from subgraph plugin
- [x] need to do a mental pass over forward resolution & protocol acceleration touchpoints
  - maybe some of the protocol acceleration logic could be simplified now that it's centralized to ProtocolAcceleration plugin
- [x] rename domain-resolver relationships to node-resolver relationships and update entity
- [x] need to update `resolver-records-indexed-on-chain.ts` to use simplified logic, need to abstract "which chains have Resolvers actively indexed" so that both `resolver-records-indexed-on-chain` can reference the same set of chainIds  `protocol-acceleration/plugin.ts`
- [x] need to do a docs pass

---

## Description

- isolated all protocol acceleration logic into new `protocol-acceleration` plugin
  - core-plugin independent, can be run in isolation
  - indexing of Resolver Records for all Resolver contracts on ENS Root, Base, Linea, and Optimism
    - in order to accelerate Forward Resolution
  - indexing of Node-Resolver Relationships for ENS Root, Basenames, Lineanames, and ThreeDNS
    - in order to accelerate UniversalResolver#findResolver
  - indexing of ENSIP-19 StandaloneReverseRegistrars for Base, Linea, Optimism, Arbitrum, and Scroll
    - in order to accelerate ENSIP-19 Reverse Resolvers
- protocol-acceleration schema refactored from concatenated primary keys to cleaner and likely more efficient composite primary keys
- reorganized some files & folders
  - `/api/lib` — organized the monolithic lib folder
    - `api/lib/handlers/` - Request handling utilities (validation, error responses)
    - `api/lib/protocol-acceleration/` - Protocol acceleration logic (resolver finding, record caching, known resolvers)
    - `api/lib/tracing/` - OpenTelemetry instrumentation and protocol tracing
    - `api/lib/resolution/` - Resolution logic (forward, reverse, primary name)
  - `lib/protocol-acceleration` — moved protocol-acceleration-specific libs to this subdir
- renamed `DomainResolverRelationship` to `NodeResolverRelationship`
- slightly refactored and cleaned-up the protocol acceleration logic during resolution now that everything has been moved to the protocol-acceleration plugin


packages/ensnode-schema/src/subgraph.schema.ts
```diff
@@ -177,21 +177,6 @@ export const resolver = onchainTable(
     // The set of observed SLIP-44 coin types for this resolver
     // NOTE: we avoid .notNull.default([]) to match subgraph behavior
     coinTypes: t.bigint().array(),
-
-    /**
-     * Represents the value of the reverse-resolution (ENSIP-3) name() record.
```
shrugs, line 182: we finally get to remove this extension to the subgraph schema!

packages/ensnode-sdk/src/ensindexer/config/zod-schemas.test.ts
```diff
@@ -45,8 +49,8 @@ describe("ENSIndexer: Config", () => {
           ),
         ).toContain("Plugins cannot contain duplicate values");
 
-        expect(formatParseError(makePluginsListSchema().safeParse([]))).toContain(
-          "Plugins must be a list with at least one valid plugin name. Valid plugins are: subgraph, basenames, lineanames, threedns, reverse-resolvers, referrals, tokenscope",
+        expect(formatParseError(makePluginsListSchema().safeParse([]))).toMatch(
```
shrugs, line 52: made this error message check more general so the test is less fragile but still accomplishes its goal

lightwalker-eth (approved):
@shrugs This is awesome 🚀 Super work here and love the comments 💪 Shared a few small suggestions. Please merge when ready 👍

PR #1131, opened 2025-09-28
Refine pr 1117

lightwalker-eth:
Suggests refinements to https://github.com/namehash/ensnode/pull/1117

### DIFF
--- a/packages/ensnode-sdk/src/ensindexer/indexing-status/types.ts
+++ b/packages/ensnode-sdk/src/ensindexer/indexing-status/types.ts
@@ -44,7 +44,7 @@ export interface ChainIndexingConfigIndefinite {
    * A {@link BlockRef} to the block where indexing of the chain should end.
    */
   // TODO: this field should be removed.
-  endBlock: null;
+  endBlock?: null;
 }
 
 /**
@@ -678,7 +678,11 @@ export type CurrentIndexingProjectionOmnichain = {
 /**
  * A status code for generating an indexing status response.
  */
-export const IndexingStatusResponseCodes = {
+// TODO: rename to `IndexingStatusResponseCodes` (this was temporarily named as
+// `IndexingStatusResponseCodesTemp`) because there's already an existing
+// `IndexingStatusResponseCodes` type in the `api/types.ts` file. This idea
+// perhaps should move into the `api/types.ts` file.
+export const IndexingStatusResponseCodesTemp = {
   /**
    * Represents that the indexing status is available.
    */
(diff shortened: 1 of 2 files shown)


packages/ensnode-sdk/src/ensindexer/indexing-status/types.ts
```diff
@@ -1,514 +1,741 @@
...
+ *
+ * Use the `strategy` field to determine the specific type interpretation
+ * at runtime.
+ *
+ * Currently, only omnichain indexing is supported. This type could theoretically
+ * be extended to support other multichain indexing strategies in the future,
+ * such as Ponder's "multichain" indexing strategy that indexes each chain
+ * independently without deterministic ordering.
+ */
+export type MultichainIndexingStatusSnapshot = MultichainIndexingStatusSnapshotOmnichain;
```
tk-o, line 637: Ponder indexing strategies are called _mulitchain_ and _omnichain_. If we consider that ENSIndex might support the _multichain_ strategy, the type name for it would be `MultichainIndexingStatusSnapshotMultichain`. It does not read well.

Perhaps we should name the type as _crosschain_, and allow the following naming in the future:
```ts
export type CrosschainIndexingStatusSnapshot = CrosschainIndexingStatusSnapshotOmnichain | CrosschainIndexingStatusSnapshotMultichain;
```

packages/ensnode-sdk/src/ensindexer/indexing-status/types.ts
```diff
@@ -1,514 +1,741 @@
...
+ *     - the `config.startBlock.timestamp` for all indexed chains.
+ *     - the `config.endBlock.timestamp` for all indexed chains with a `config.configType` of
+ *       {@link ChainIndexingConfigTypeIds.Definite}.
+ *     - the `backfillEndBlock.timestamp` for all chains with `chainStatus` of
+ *       {@link ChainIndexingStatusIds.Backfill}.
+ *     - the `latestKnownBlock.timestamp` for all chains with `chainStatus` of
+ *       {@link ChainIndexingStatusIds.Following}.
  */
-export type CurrentIndexingProjectionOmnichain = {
+export interface MultichainIndexingStatusSnapshotOmnichain {
```
tk-o, line 597: I'm not sure if we really need to wrap `OmnichainIndexingSnapshot` with `MultichainIndexingStatusSnapshotOmnichain`. Why not compose the _strategy_ and _slowestChainIndexingCursor_ field into to `OmnichainIndexingSnapshot` type?

In other words, I'd recommend not go build deep hierarchies (wrapping a type with another type), but rather focus on composition (having a type to implement relevant interfaces).

PR #1133, opened 2025-09-29
feat(basenames-plugin): index the new upgradeable registrar controller contract

### LINKED ISSUE #1089: Basenames: Add support for new contract deployments
Image

Perhaps Basenames is using a new registrar / registrar controller that we haven't indexed yet? This might be related to the recent upgrades on Basenames to support ENSIP-19.

Edit: It seems so. I just made a quick investigation of this recent transaction to register a base name that appeared as unknown in ENSAdmin: https://basescan.org/tx/0x87c1879a82efdb43b3c4edd356855a5c0f2baab8968f4399c8f236af60a457d6

That registration occurred through this proxy contract that was deployed only 15 days ago: https://basescan.org/address/0xa7d2607c6bd39ae9521e514026cbb078405ab322#code

The above proxy contract routes to https://basescan.org/address/0x9ad14968093c5e8c2a8cc86f6868cfee8c659717#code

Reviewing the commit history to the main branch in the basenames repo shows a number of PRs over the past months with the relevant changes: https://github.com/base/basenames/commits/main/

tk-o:
Resolves #1089


packages/datasources/src/mainnet.ts
```diff
@@ -145,15 +146,37 @@ export default {
...
         abi: ResolverABI,
         address: "0xc6d566a56a1aff6508b41f6c90ff131615583bcd",
         startBlock: 17575714,
       },
+      // QUESTION: do we need to track this one?
+      // UpgradeableL2Resolver: {
+      //   abi: ResolverABI,
+      //   address: "0x426fA03fB86E510d0Dd9F70335Cf102a98b10875",
+      //   startBlock: 35286620,
+      // },
```
tk-o, line 179: @shrugs, please let me know what you think about tracking L2Resolver contracts configuration in this file?

I was not able to find a single spot where we point to `L2Resolver` contract. Perhaps we should just drop config for it?

packages/datasources/src/mainnet.ts
```diff
@@ -145,15 +146,37 @@ export default {
...
       },
       RegistrarController: {
+        /**
+         * This controller was removed from BaseRegistrar contract
+         * at block 35936564 with the following tx:
+         * https://basescan.org/tx/0x88a3cc03291bb1a4b2bd9ccfe6b770988470001905d96c32bd41b866797b684b
+         */
         abi: base_RegistrarController,
         address: "0x4ccb0bb02fcaba27e82a56646e81d8c5bc4119a5",
         startBlock: 18619035,
```
shrugs, line 156: should we consider adding an `endBlock: 35936564` to this contract since it was removed? i don't think any transactions would be successful but maybe worth adding? i don't think it would be, but if this causes an issue with indexing or our handling of the generated ponder config, let's not bother

packages/datasources/src/mainnet.ts
shrugs: actually i think we need to include them and add them to the `apps/ensindexer/src/api/lib/protocol-acceleration/known-onchain-static-resolver.ts` logic — perhaps i just forgot to do this long ago when writing this protocol acceleration logic?

they're both onchain static resolvers so add them to `isKnownOnchainStaticResolver`. the original L2Resolver does _not_ implement defaulting, so don't add that to `onchainStaticResolverImplementsDefaultAddress` but the second _does_ implement defaulting so it should be included there. 

and maybe add a comment that these contracts are indexed by the more general Resolver filter but/and we list them here specifically because they are needed for protocol acceleration logic

shrugs (approved):
lgtm

Issue #1088, opened 2025-09-18
Enhance multichain awareness in latest registrations panel


lightwalker-eth:
# Goals

- Extend the "dogfooding" of our resolution APIs and our end-to-end support for ENSIP-19.

# Background

- In [PR 1072](https://github.com/namehash/ensnode/pull/1072) we are introducing a feature in the `Identity` component to support passing in a chainId.

# Idea

- [ ] Hack: Create a utility function: `guessChainIdFromRegisteredName` that takes as input a namespace and a name and then returns the relevant chainId depending on if the registration is for a direct subname of base.eth or linea.eth.
   - [ ] Carefully document this dirty hack, why we are doing it, it's limitations, why it is safe for our current context, and how we will solve this issue in a non-hack way in the future.
- [ ] For each registration we display on the latest indexed registrations panel, pass a chain id into the`Identity` component based on `guessChainIdFromRegisteredName`.
- [ ] Refine the `Identity` component such that whenever we display a primary name it is explicit which chainId that primary name is for.

Issue #1048, opened 2025-09-11
Update NameDisplay to show beautified names


lightwalker-eth:
... wait until PR 1024 is merged first

PR #1137, opened 2025-09-30
Feat/Identity

lightwalker-eth:
Closes https://github.com/namehash/ensnode/issues/1088
Closes https://github.com/namehash/ensnode/issues/1048

CleanShot 2025-09-30 at 17 33 13

CleanShot 2025-09-30 at 17 34 50

CleanShot 2025-09-30 at 17 37 07

CleanShot 2025-09-30 at 17 39 01

CleanShot 2025-09-30 at 17 42 42

CleanShot 2025-09-30 at 19 30 40

CleanShot 2025-09-30 at 19 31 15

CleanShot 2025-09-30 at 19 28 27

CleanShot 2025-09-30 at 19 29 43

CleanShot 2025-09-30 at 19 42 05

### DIFF
--- a/apps/ensadmin/src/app/mock/display-identity/page.tsx
+++ b/apps/ensadmin/src/app/mock/display-identity/page.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { DisplayIdentity } from "@/components/identity";
-import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
+import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
 import { Checkbox } from "@/components/ui/checkbox";
 import { Input } from "@/components/ui/input";
 import {


apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -143,10 +150,52 @@ interface RegistrationRowProps {
...
+/**
+ * NOTE: This function is a "dirty-hack".
+ *
+ * TODO: Update our indexed data model for registrations so that this dirty hack
+ *       is no longer needed.
+ */
+const guessChainIdFromRegisteredName = (name: Name, namespaceId: ENSNamespaceId): ChainId => {
+  const labels = name.split(".");
+
+  if (labels.length === 3) {
```
shrugs, line 162: do we need the hierarchy depth check at all? even `example.example.linea.eth` would be registered on linea, right?

apps/ensadmin/src/components/recent-registrations/components.tsx
```diff
@@ -143,10 +150,52 @@ interface RegistrationRowProps {
...
+
+      if (linea) return linea.chain.id;
+
+      // basenames is undefined for the namespace, therefore fallback
+      // to the ENS root chain
+    }
+  }
+
+  // fallback to assuming the the registration occurred on the ENS root chain
+  return getENSRootChainId(namespaceId);
```
shrugs, line 191: I would make this function

```ts
function guessChainIdFromRegisteredName(name: Name): ChainId | null {
   // ...
}
```

and implement the defaulting within the component like

```ts
const chainId = guessChainIdFromRegisteredName(registration.name) ?? getENSRootChainId(namespaceId)
```

so that the 'guess' verbiage reads better, and we don't need to pass the namespace to the helper

also that helper could/should live in an ensadmin/lib file somewhere, perhaps `guess-registration-chain-id.ts`

packages/ensnode-sdk/src/resolution/ensip19-chainid.ts
```diff
@@ -0,0 +1,46 @@
...
+ * "chainId param" is always set to chainId: 1 (mainnet), even if the
+ * `chainId` where the primary name lookup is actually happening
+ * on a non-mainnet ENS Root Chain, such as on a testnet or
+ * the ens-test-env.
+ *
+ * @param namespaceId The namespace id for the primary name lookup.
+ * @param chainId The chain id where the primary name lookup will actually happen.
+ * @returns The "chainId param" that should be used for the primary name lookup.
+ */
+export const getResolvePrimaryNameChainIdParam = (
```
shrugs, line 24: i see how this api surface is causing confusion, and perhaps we should restrict the api to accepting coinTypes, but that can be done later.

shrugs: i also think that this is specifically about the Resolve Primary Name api surface and not specifically ensip-19, right?

packages/ensnode-sdk/src/resolution/ensip19-chainid.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Translates a `DefaultableChainId` a `ChainId`
+ * such that if the provided `chainId` is `DEFAULT_EVM_CHAIN_ID`,
+ * the `ChainId` of the ENS Root Chain for the provided `namespaceId` is returned.
+ *
+ * @param chainId The `DefaultableChainId` to translate.
+ * @param namespaceId The namespace id for the translation.
+ * @returns the translated `ChainId`.
+ */
+export const translateDefaultableChainIdToChainId = (
```
shrugs, line 41: maybe `resolveDefaultableChainId` over 'translate'?

apps/ensadmin/src/components/identity/types.ts
```diff
@@ -0,0 +1,28 @@
+import { DefaultableChainId, ENSNamespaceId, Name } from "@ensnode/ensnode-sdk";
+import { Address } from "viem";
+
+/**
+ * Represents the result of an ENSIP-19 identity resolution.
+ */
+export interface ResolvedIdentity {
```
shrugs, line 7: i would consider having two types:

`UnnamedIdentity` and `NamedIdentity` and making `Identity` the union of them.

```ts
interface UnnamedIdentity {
  namespaceId: ENSNamespaceId,
  chainId: ChainId,
  address: Address
}

interface NamedIdentity implements UnnamedIdentity {
  name: Name;
}

type Identity = UnnamedIdentity | NamedIdentity;
// results in identity.name being Name | undefined

const isNamedIdentity = (identity: Identity): identity is NamedIdentity => identity.name !== undefined;

// and then when rendering you can do 
if (isNamedIdentity(identity)) {
  // access identity.name as Name
}
```

could also try 

```ts
interface UnresolvedIdentity {
  namespaceId: ENSNamespaceId,
  chainId: ChainId,
  address: Address
}

interface UnnamedIdentity implements UnresolvedIdentity {
  name: null;
}

interface NamedIdentity implements UnresolvedIdentity {
  name: Name;
}

type ResolvedIdentity = UnnamedIdentity | NamedIdentity;
type Identity = UnresolvedIdentity | ResolvedIdentity;
```

apps/ensadmin/src/components/identity/index.tsx
```diff
@@ -33,44 +42,61 @@ export function Identity({
...
-  // Establish chainId, preferring user-supplied and defaulting to the ENS Root Chain Id.
-  const definedChainId = chainId ?? ensRootChainId;
+  if (chainId === undefined) {
+    // default to resolving the identity for `address` on the ENS Root Chain
+    // for the provided `namespaceId`.
+    chainId = getENSRootChainId(namespaceId);
+  }
 
-  // Lookup the primary name for address using ENSNode
+  // resolve the primary name for `address` on `chainId` using ENSNode
```
shrugs, line 51: if using the various `Identity` types i proposed, there could be a `useResolvedIdentity` hook like so:

```ts
const { identity, status } = useResolvedIdentity({ namespaceId, chainId, address });


function useResolvedIdentity({namespaceId, chainId, address}: UnresolvedIdentity) {
  const {data: name, ...query} = usePrimaryName({ address: identity.address, chainId: getResolvePrimaryNameChainIdParam(chainId, namespaceId) })

  return {
   ...query,
   identity: name !== undefined ? { namespaceId, chainId, address, name } : undefined
  }
}
```

and maybe that's clean?

apps/ensadmin/src/components/identity/types.ts
tk-o: Also, we could use `AccountId` for `UnnamedIdentity` definition:

```ts
interface UnnamedIdentity extends AccountId {
  namespaceId: ENSNamespaceId,
}

apps/ensadmin/src/components/recent-registrations/components.tsx
lightwalker-eth: Thanks. Changed to >=

PR #1141, opened 2025-10-02
fix config deserialization

lightwalker-eth:
This PR fixes the following change management issue with plugin names changing between client vs server.

image


Y3drk (commented):
The code changes themselves look fine to me, but I am not sure how this affects ENSIndexer/ENSAdmin on a deeper level that @tk-o mentioned. 

I assume this would mostly result in making the deserialization functions less strict, which is beneficial for ENSAdmin, but I don't have enough knowledge about ENSIndexer to comment on that.

Also, it seems like the unit tests in `ensnode-sdk/src/ensindexer/config/zod-schemas.test.ts` need some tweaks.

tk-o (approved):

Issue #1085, opened 2025-09-18
Attempt avatar image loading through useRecords


lightwalker-eth:
# Background

- Slack thread: https://namehash.slack.com/archives/C086Z6FNBHN/p1758215732488819?thread_ts=1758190597.058309&cid=C086Z6FNBHN
- ENSAdmin currently loads avatar images through the ENS Metadata Service (rather than through useRecords) for a few reasons:
   - When we first built ENSAdmin our `useRecords` hook and related API didn't exist yet.
   - Even with the new `useRecords` hook and related API, there's complexity to correctly handling all possible ENS avatar edge cases. For example, consider URI types other than http or https: https://docs.ens.domains/ensip/12#uri-types
   - Solving all ENS avatar edge cases is not a simple project. For details see https://github.com/ensdomains/ens-avatar which is the library embedded within https://github.com/ensdomains/ens-metadata-service that the ENS Metadata Service is understood to use.
   - Building a "native" replacement for the avatar image loading capabilities of the ENS Metadata Service is a goal of ENSNode and is on our roadmap. But it is expected to be a meaningful project and is a very low priority now. We need to prioritize other goals. This issue suggests what can be a nice incremental improvement.
- In https://github.com/namehash/ensnode/pull/1073 we are introducing a `useEnsMetadataServiceAvatarUrl` hook.
- The `useEnsMetadataServiceAvatarUrl` has some shortcomings:
   - It adds an external dependency: the ENS Metadata Service.
   - It is not currently supported in the ens-test-env, but theoretically could once https://github.com/namehash/ensnode/issues/1078 is solved.
   - It does not support holesky since there is no ENS Metadata Service deployment for holesky.
- ... therefore, ENSAdmin currently cannot load any avatar images for holesky or the ens-test-env. That's a bummer.

# Goals

- Support more avatar image scenarios, to include:
   - Avatar images using http or https URIs in the holesky and ens-test-env namespaces
- Grow our "dogfooding" of our own resolution APIs
- Reduce our dependence on the ENS Metadata Service (we shouldn't fully remove use of it yet)

# Suggested Strategy

Define a new hook: `useAvatarUrl` that takes as input a `Name` and returns back a `URL` which can then be fetched to load the avatar image for the name.

At first, `useAvatarUrl` should attempt to get the URL of the avatar for the given name through a `useRecords` call where the `avatar` text record is requested.

If `useRecords` has an error, return null.

[...]

PR #1142, opened 2025-10-02
feat(ensnode-react): useAvatarUrl

notrab:
Closes https://github.com/namehash/ensnode/issues/1085

# New Notes (from lightwalker)

- Consider: https://viem.sh/docs/ens/actions/getEnsAvatar#assetgatewayurls-optional
- Consider: transitioning these responsibilities into ENSApi.
   - Return "raw" avatar URL
   - For the case of images referenced through NFTs: maybe return the "nftRaw" avatar URL based on a lookup of the data for the associated NFT
   - Return "http-safe" avatar URL
   - Return cached / size-optimized / etc.. avatar URL (?)
- Consider: full support for cover images as well

### DIFF
--- a/packages/ensnode-react/src/hooks/useAvatarUrl.ts
+++ b/packages/ensnode-react/src/hooks/useAvatarUrl.ts
@@ -183,7 +183,6 @@ export function useAvatarUrl(
     queryKey: readonly unknown[];
     queryFn: () => Promise<UseAvatarUrlResult>;
     retry: boolean;
-    placeholderData: UseAvatarUrlResult;
   } = {
     queryKey: [
       "avatarUrl",
@@ -221,11 +220,6 @@ export function useAvatarUrl(
       };
     },
     retry: false,
-    placeholderData: {
-      rawAvatarTextRecord: null,
-      browserSupportedAvatarUrl: null,
-      usesProxy: false,
-    } as const,
   };
 
   const options = {
--- a/apps/ensadmin/src/app/name/_components/ProfileHeader.tsx
+++ b/apps/ensadmin/src/app/name/_components/ProfileHeader.tsx
@@ -67,7 +67,7 @@ export function ProfileHeader({ name, namespaceId, headerImage, websiteUrl }: Pr
       <CardContent className="pt-6">
         <div className="flex items-start justify-between">
           <div className="flex items-center gap-4">
-            <EnsAvatar className="-mt-16 h-20 w-20 ring-4 ring-white" name={name} />
+            <EnsAvatar className="-mt-16 h-20 w-20 ring-4 ring-white bg-white" name={name} />
             <div className="flex-1">
               <h1>
                 <NameDisplay className="text-3xl font-bold" name={name} />


packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @example
+ * ```typescript
+ * import { useAvatarUrl } from "@ensnode/ensnode-react";
+ *
+ * function ProfileAvatar() {
+ *   const { data: avatarUrl, isLoading, error } = useAvatarUrl({
+ *     name: "vitalik.eth"
+ *   });
+ *
+ *   if (isLoading) return <div>Loading...</div>;
```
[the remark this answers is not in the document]
notrab: I've improved the example to show a placeholder. In my mind, something like:

```tsx

<div className="avatar">
        {isLoading || !avatarUrl ? (
          <div className="avatar-placeholder" />
        ) : (
          <img src={avatarUrl} alt={`${name} avatar`} />
        )}
      </div>
```

is what I would do in a real world app.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+ *
+ * @example
+ * ```typescript
+ * normalizeAvatarUrl("example.com/avatar.png") // Returns URL with https://example.com/avatar.png
+ * normalizeAvatarUrl("http://example.com/avatar.png") // Returns URL with http://example.com/avatar.png
+ * normalizeAvatarUrl("ipfs://QmHash") // Returns URL with ipfs://QmHash (requires fallback)
+ * normalizeAvatarUrl("invalid url") // Returns null
+ * ```
+ */
+function normalizeAvatarUrl(url: string | null | undefined): AvatarUrl | null {
```
lightwalker-eth, line 64: Suggest moving this utility function out of this hook and into `packages/ensnode-sdk/src/shared/url.ts`

Hmm.. wait. This doesn't make sense.

Why is this function returning our special type alias without checking if it the URL is actually supported by browsers?

Need to rely on us getting our types right and strongly enforcing our constraints. Please pay more attention to this.

Also suggest to enforce that this function exclusively takes `string` values as input.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+  /**
+   * Optional custom fallback function to get avatar URL when the avatar text record
+   * uses a non-http/https protocol (e.g., ipfs://, ar://, eip155:/).
+   *
+   * If not provided, defaults to using the ENS Metadata Service.
+   *
+   * @param name - The ENS name to get the avatar URL for
+   * @returns Promise resolving to the avatar URL, or null if unavailable
+   */
+  browserUnsupportedProtocolFallback?: (name: Name) => Promise<string | null>;
```
lightwalker-eth, line 43: What do you think about this callback function also taking as input the normalized avatar URL, in theory it might implement different logic depending on the attributes of this normalized avatar URL.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+ * are preserved and can be handled by fallback mechanisms.
+ *
+ * @param url - The URL string to normalize
+ * @returns A URL object if the input is valid, null otherwise
+ *
+ * @example
+ * ```typescript
+ * normalizeAvatarUrl("example.com/avatar.png") // Returns URL with https://example.com/avatar.png
+ * normalizeAvatarUrl("http://example.com/avatar.png") // Returns URL with http://example.com/avatar.png
+ * normalizeAvatarUrl("ipfs://QmHash") // Returns URL with ipfs://QmHash (requires fallback)
```
lightwalker-eth, line 60: Please see related comments for how I believe we want to define this function more generally and not only for this hook.

Therefore, the more general version of this function shouldn't make any reference to avatar-specific concepts such as "fallback", etc. The docs would need to be appropriately updated.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+  }
+}
+
+/**
+ * Result returned by the useAvatarUrl hook.
+ */
+export interface UseAvatarUrlResult {
+  /**
+   * The original avatar text record value from ENS, before any normalization or fallback processing.
+   * Null if no avatar record exists.
```
lightwalker-eth, line 80: We need to be careful saying it is Null if no avatar record exists. That doesn't seem strictly true. Right?

Please refine how we describe what the null case represents.

We also need to document the invariant: if `rawAvatarUrl` is `null` then `browserSupportedAvatar` must be `null` as well.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+   */
+  rawAvatarUrl: string | null;
+  /**
+   * A browser-supported (http/https) avatar URL ready for use in <img> tags.
+   * Populated when the avatar uses http/https protocol or when a fallback successfully resolves.
+   * Null if no avatar exists or resolution fails.
```
lightwalker-eth, line 86: Please see related comment above about how we need to be more careful with the accuracy of our language.

packages/ensnode-sdk/src/ens/metadata-service.ts
```diff
@@ -0,0 +1,37 @@
+import type { ENSNamespaceId } from "@ensnode/datasources";
+import { ENSNamespaceIds } from "@ensnode/datasources";
+
+import type { Name } from "./types";
+
+/**
```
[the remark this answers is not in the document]
notrab: Updated

packages/ensnode-react/src/hooks/useAvatarUrl.ts
notrab: @lightwalker-eth so here's what I think we need to do. Rename the SDK function from `normalizeAvatarUrl` to something like `toBrowserSupportedUrl`, or `validateBrowserSupportedUrl` to clarify it's a strict type validator, not a normalizer.

The hook shouldn't import the SDK version because it needs the lenient behaviour. However, the SDK version's name is misleading because it's not really about "avatars", it's about validating any URL as browser-supported.

lightwalker-eth: @notrab Sounds about right 👍

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+    selection: { texts: [AVATAR_TEXT_RECORD_KEY] },
+    config: _config,
+    query: { enabled: canEnable },
+  });
+
+  // Get namespace from config
+  const configQuery = useENSIndexerConfig({ config: _config });
+  const namespaceId = configQuery.data?.namespace ?? null;
+
+  // Create default fallback using ENS Metadata Service if namespaceId is available
```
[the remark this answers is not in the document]
notrab: This is no longer the case with the changes to `buildEnsMetadataServiceAvatarUrl` and just calling that.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
notrab: The `normalizeAvatarUrl` is only used in one place now, and the type enforcer is its own component so this function can remain inside the file, with the `toBrowserSupportedUrl` function moved to the shared url library file.

apps/ensadmin/src/components/ens-avatar.tsx
```diff
@@ -1,27 +1,30 @@
...
 
-  if (avatarUrl === null) {
+  const { data: avatarData } = useAvatarUrl({
+    name,
+  });
+
+  const avatarUrl = avatarData?.browserSupportedAvatarUrl;
+
+  if (avatarUrl === null || avatarUrl === undefined) {
     return (
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth I think we could make a change here so that we are explicit in the loading:

```tsx
 const { data: avatarData, isLoading: isAvatarUrlLoading } = useAvatarUrl({
    name,
  });

if (isAvatarUrlLoading) {
    return (
      <Avatar className={className}>
        <AvatarLoading />
      </Avatar>
    );
  }
```

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * Type alias for avatar URLs. Avatar URLs must be URLs to an image asset
+ * accessible via the http or https protocol.
+ */
+export type AvatarUrl = URL;
+
+/**
+ * Parameters for the useAvatarUrl hook.
+ *
+ * If `name` is null, the query will not be executed.
```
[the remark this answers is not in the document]
notrab: @lightwalker-eth let's go with your initial hunch. @shrugs also doesn't see any reason not to either, so I'll make sure to apply your original suggestion to this hook, and the other functions for consistency.

apps/ensadmin/src/components/ens-avatar.tsx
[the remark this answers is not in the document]
notrab: @lightwalker-eth I've added a new JSDoc for our implementation `ens-avatar`, and also updated the `useAvatarUrl` JSDoc. Both i'm not yet happy with, so will keep this marked as unresolved and iterate.

packages/ensnode-sdk/src/ens/metadata-service.ts
```diff
@@ -1,31 +1,41 @@
...
 export function buildEnsMetadataServiceAvatarUrl(
   name: Name,
   namespaceId: ENSNamespaceId,
-): URL | null {
+): BrowserSupportedAssetUrl | null {
   switch (namespaceId) {
     case ENSNamespaceIds.Mainnet:
-      return new URL(name, `https://metadata.ens.domains/mainnet/avatar/`);
+      return toBrowserSupportedUrl(
+        new URL(name, `https://metadata.ens.domains/mainnet/avatar/`).toString(),
```
lightwalker-eth, line 33:
```suggestion
      return toBrowserSupportedUrl(`https://metadata.ens.domains/mainnet/avatar/{$encodeURIComponent(name)}`),
```

Seems more clean. wdyt?

lightwalker-eth (commented):
@notrab Thanks for updates. Reviewed and shared feedback 👍

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -0,0 +1,287 @@
...
+
+  const options = {
+    ...baseQueryOptions,
+    ...queryOptions,
+    enabled:
+      canEnable &&
+      recordsQuery.isSuccess &&
+      configQuery.isSuccess &&
+      (queryOptions?.enabled ?? true),
+  } as typeof baseQueryOptions;
```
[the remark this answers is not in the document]
notrab: @shrugs good idea. I've moved some of the code into its own function so it's easier to test. I'll continue to iterate to get closer to the above suggestion.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -20,6 +21,111 @@ import { useRecords } from "./useRecords";
...
+ * This is the core resolution logic extracted for testing.
+ *
+ * @param avatarTextRecord - The raw avatar text record value from ENS
+ * @param name - The ENS name
+ * @param namespaceId - The ENS namespace ID
+ * @param browserSupportedAvatarUrlProxy - Optional custom proxy function
+ * @returns The resolved avatar URL result
+ * @internal
+ */
+export function resolveAvatarUrl(
```
lightwalker-eth:
```suggestion
export function buildBrowserSupportedAvatarUrl(
```

Goal: We shouldn't use "resolve" terminology here. In ENS, "resolve" has special meaning. For example, you can say "resolve the avatar text record for the name X". But note how the signature of this function requires that you already have completed a resolution of the avatar text record.

Please also note in the suggested function name I wrote "...AvatarUrl" and not "AssetUrl". That was intentional.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -20,6 +21,111 @@ import { useRecords } from "./useRecords";
...
 const AVATAR_TEXT_RECORD_KEY = "avatar" as const;
 
+/**
+ * Resolves an avatar text record to a browser-supported URL.
+ * This is the core resolution logic extracted for testing.
+ *
+ * @param avatarTextRecord - The raw avatar text record value from ENS
+ * @param name - The ENS name
+ * @param namespaceId - The ENS namespace ID
+ * @param browserSupportedAvatarUrlProxy - Optional custom proxy function
```
lightwalker-eth:
```suggestion
 * @param browserSupportedAvatarUrlProxy - Optional function for generating browser support asset urls that route through a custom proxy.
```

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -20,6 +21,111 @@ import { useRecords } from "./useRecords";
...
+          rawAvatarUrl: avatarTextRecord,
+          browserSupportedAvatarUrl: null,
+          usesProxy: false,
+        };
+      }
+    }
+  }
+
+  // Default proxy is to use the ENS Metadata Service
+  const defaultProxy = (name: Name, avatarUrl: URL): BrowserSupportedAssetUrl | null => {
```
lightwalker-eth: Please see related feedback for how the callback function should have a namespaceId param added.

Once this change is made, we should also move the definition of the default callback function out of the internal logic of this function. The default callback function can be defined at the root file level.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -20,6 +21,111 @@ import { useRecords } from "./useRecords";
...
+    return buildEnsMetadataServiceAvatarUrl(name, namespaceId);
+  };
+
+  // Use custom proxy if provided, otherwise use default
+  const activeProxy: (name: Name, avatarUrl: URL) => BrowserSupportedAssetUrl | null =
```
lightwalker-eth: Suggest to separately define the type for this callback. Then we can reference that type in all the places that need it.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -114,18 +217,32 @@ export interface UseAvatarUrlResult {
...
+ *         // return toBrowserSupportedUrl(`https://my-gateway.example.com/ipfs/${ipfsPath}`);
  *       }
+ *
+ *       // Handle Arweave protocol
+ *       if (avatarUrl.protocol === 'ar:') {
+ *         const arweaveId = avatarUrl.href.replace('ar://', '');
+ *         return toBrowserSupportedUrl(`https://arweave.net/${arweaveId}`);
+ *       }
+ *
+ *       // Fall back to ENS Metadata Service for other protocols
```
lightwalker-eth: Hmm. Can you double check that this really falls back to the ENS Metadata Service by returning null. I don't believe that's what happens here. We replaced the default URL builder with a custom one here.

apps/ensadmin/src/app/mock/ens-avatar/page.tsx
```diff
@@ -0,0 +1,187 @@
...
+              )}
+            </div>
+          </div>
+        </div>
+      </CardContent>
+    </Card>
+  );
+}
+
+export default function MockAvatarUrlPage() {
```
lightwalker-eth: We need a UI where I can input the following:
* An ENSNamespace
* A Name
* A raw avatar text record

And then view the result of this. Note how this functionality I'm requesting should make 0 requests for the real avatar text record of the name and therefore does not need to care at all about the active namespace of the connected ENSNode instance.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
lightwalker-eth: A few key enhancements needed:
1. This function should be removed from this file and instead move into `ensnode-sdk`.
2. The data type returned by this function should be more generic and not tied specifically to the idea of "useAvatarUrl".
3. Good unit tests should be defined for this function. I want to see that craftsmanship was invested into planning each of the unit tests and that it's not just AI.

packages/ensnode-react/src/hooks/useAvatarUrl.ts
```diff
@@ -20,6 +21,111 @@ import { useRecords } from "./useRecords";
...
+  const activeProxy: (name: Name, avatarUrl: URL) => BrowserSupportedAssetUrl | null =
+    browserSupportedAvatarUrlProxy ?? defaultProxy;
+
+  // For other protocols (ipfs, data, NFT URIs, etc.), use proxy if available
+  if (activeProxy) {
+    try {
+      const proxyUrl = activeProxy(name, buildUrl(avatarTextRecord));
+
+      // Invariant: BrowserSupportedAssetUrl must pass isHttpProtocol check
+      if (proxyUrl !== null && !isHttpProtocol(proxyUrl)) {
```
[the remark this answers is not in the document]
notrab: Yeah, it should use `isBrowserSupportedProtocol`. Fixed.

packages/ensnode-sdk/src/ens/metadata-service.ts
```diff
@@ -0,0 +1,273 @@
...
+   * Indicates whether the browserSupportedAssetUrl uses a proxy service.
+   * True if the url uses a proxy (either the default ENS Metadata Service or a custom proxy).
+   * False if the URL comes directly from the asset text record, or if there's no asset text record,
+   * or if the asset text record has an invalid format, or if no url is known for loading the asset using a proxy.
+   */
+  usesProxy: boolean;
+}
+
+/**
+ * Builds a browser-supported asset URL for a name's asset image from the name's raw asset text record value.
```
notrab:
```suggestion
* Builds a browser‑supported avatar image URL from the ENS avatar text record (key: "avatar").
* Accepts non‑browser protocols (ipfs://, ar://) and NFT URIs (CAIP‑22 ERC‑721 / CAIP‑29 ERC‑1155, eip155 chain namespace),
* and converts them via a proxy/gateway (default: ENS Metadata Service) to http/https/data URLs.
```

This keeps the "asset" generic type for reuse, but adjusts the JSDoc to emphasize the use of the "avatar text record" in this context.

Issue #621, opened 2025-04-17
Display App Versions in ENSAdmin


lightwalker-eth:
# Background

![Image](https://github.com/user-attachments/assets/36e047ac-827e-4c06-98d1-efee291be1f4)

We're currently displaying a "Version" for ENSRainbow, but not for ENSIndexer.

# Goal

Provide easy visibility of all ENSNode app versions within ENSAdmin.

# Subtasks

Both ENSIndexer and ENSRainbow should display their version number in the ENSAdmin UI. Both should display the version number based on their respective package.json file which is automatically incremented and managed by changesets.

![Image](https://github.com/user-attachments/assets/158f333f-9206-4bfd-a0fc-002f8bc9e8a5)

Additionally, ENSAdmin should display its version number at the bottom of the main nav menu as highlighted in the screenshot above.

tk-o:
PR #895 introduced ENSIndexer Config API, which includes dependency info with an accurate information about the integrated ENSRainbow service.

tk-o:
PR #919 includes server-driven dependency info which solves the issue of hardcoded dependency information:

<img width="2880" height="1014" alt="Image" src="https://github.com/user-attachments/assets/ad06543d-c2af-4b14-990c-74e39a3ab5e4" />

lightwalker-eth:
Progress is being made on this issue inside of https://github.com/namehash/ensnode/pull/1012 -- here we're formally adding fields into this UI showing the version of ENSRainbow, ENSIndexer, and ENSAdmin.

However, there remain open issues before we can resolve this issue (issue 621):
1. The config object needs to formally include distinct version numbers for each of the following:
   1. ENSIndexer
   2. ENSDb (for now this should always be the same version as ENSIndexer)
   3. ENSRainbow (this value should be passed from ENSRainbow -> ENSIndexer -> ENSAdmin). Note how it's possible that the ENSRainbow version can be different than the ENSIndexer version. We need to remove / update any code that incorrectly assumes these must always be the same.
2. We need to add the ENSAdmin version to the bottom-left corner of the ENSAdmin nav menu.

lightwalker-eth:
This issue will be closed as part of PR 1145.

Other goals described in this issue have been moved to a new issue: https://github.com/namehash/ensnode/issues/1153

PR #1145, opened 2025-10-02
Connection page updates

### LINKED ISSUE #1144: Priority "Connection" page UI updates
Image

1. The currently selected connection is almost always too big to fit in such a small place. Let's please stop displaying this value here. Suggest to either completely remove any field here or to put some simple text here such as "Select ENSNode" or "Select ENSNode connection" (if that fits).
2. The value we're currently showing here is coming from the server, which as you can see from the URL bar in the screenshot above, can be distinct from the URL the user selected to connect to. This can be confusing. Therefore we should stop showing the URL value we're getting from the server here and instead display the URL of the currently selected connection here instead. This also gives a lot more space for displaying the currently selected connection URL.
3. We need to completely remove this "ENSAdmin" panel from the UI here. I previously suggested we add it here based on ideas I have sense realized were incorrect. ENSAdmin should remain an app that lives outside of an ENSNode instance and not be an app inside of an ENSNode instance.

notrab:
Closes #1144 
Closes https://github.com/namehash/ensnode/issues/621

@Y3drk I made some changes to `ConfigInfoAppCard` that made it more reusable. It now wont' render a `CardHeader` if there isn't `name`, `docsLink`, `version`. This means we can reuse the card and its layout properties for other things, as shown below:
 
CleanShot 2025-10-02 at 20 06 44@2x

The implemented changes include a `ConnectionLine` between the elements @lightwalker-eth . Feel free to adjust the colours here. I just used the branded blue one we have done elsewhere.

CleanShot 2025-10-02 at 20 07 12@2x

I also changed the `icon` in the CopyButton component (used in connection list too) so we're consistent, which also means we can skip passing the icon component.


lightwalker-eth (commented):
@notrab Cool to see this! Reviewed and shared feedback 👍

lightwalker-eth (approved):

PR #1150, opened 2025-10-03
fix: protocol acceleration node-resolver relationship deferrals to registries

shrugs:
closes https://github.com/namehash/ensnode/issues/1148

- node-resolver relations need to be registry-specific in order to differentiate between multiple registries on the same chain (basenames, threedns)
- offchain lookup resolvers don't defer to a chain, they defer to a specific registry.
- forward resolution as implement is always in the context of a specific registry, which owns the node-resolver relationhips. the top-level resolution must always begin with the ENS Root Registry, but subsequent accelerated calls can provide their own L2-deployed (shadow)Registry

no changelog because all internal details at the moment and bug wasn't occurring


tk-o (approved):
LGTM

lightwalker-eth (commented):
@shrugs Hey super cool to see this. This PR is related to some of our priority work on ENSv2 so have a number of special follow up questions for you. Appreciate your advice, thanks 👍

Issue #1111, opened 2025-09-22
Refactor ext_ prefixing into subgraph_ prefixing for core entities


shrugs:
- rename all subgraph entities with `subgraph_` prefixes, and remove the `ext_` prefix from all other entities in our drizzle schema
- actually maybe just add a `prefixEntities` helper and a`filterEntitiesAndStripPrefix` helper that allows us to tag and filter the `subgraph.schema.ts` 

```ts
import * as subgraphSchema from './subgraph.schema';

export prefixEntities('subgraph_', subgraphSchema);
```

configure the subgraph api with the filtered entities and confirm that it still works even after we've played god with the tsName of the drizzle entities

- [ ] mainnet index (tests indexing w/ new schema names)
- [ ] snapshot comparison (tests api layer)
- [ ] ensjs/ens-app-v3 tests (tests api layer)

PR #1152, opened 2025-10-03
refactor: prefix all subgraph entities with subgraph_

shrugs:
closes https://github.com/namehash/ensnode/issues/1111

review with hide whitespace on

- [ ] need to test subgraph api working
  - need to handle issue with dataloader accessing `drizzle.query[table.tsName]` because the passed in drizzle context (ponder's `db`) is still prefixed. maybe can solve that by constructing drizzle from provided schema instead of importing from ponder
- [ ] might as well also test subgraph compat


packages/ensnode-schema/tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
   "extends": "@ensnode/shared-configs/tsconfig.ponder.json",
-  "include": ["./**/*.ts"],
+  "include": [
+    "./**/*.ts",
+    "../../apps/ensindexer/src/lib/filter-schema-by-prefix.ts",
```
lightwalker-eth, line 5: Hmm. Should we move these operations into the `ensnode-schema` package? Assume it will cause an unfortunate surprise sometime in the future if changing a file in `ensindexer` results in a change to the `ensnode-schema` package.

Or alternatively, why does the `ensnode-schema` package need to export these operations? Couldn't they just live in `ensindexer`?

apps/ensindexer/src/api/index.ts
```diff
@@ -18,12 +19,19 @@ import {
...
 
 import ensNodeApi from "@/api/handlers/ensnode-api";
+import { makeDrizzle } from "@/api/lib/handlers/drizzle";
+
+// generate a subgraph-specific subset of the schema
+const subgraphSchema = filterSchemaByPrefix("subgraph_", schema);
+// and a drizzle db object that accesses it
+const subgaphDrizzle = makeDrizzle({
+  schema: subgraphSchema,
+  databaseUrl: config.databaseUrl!,
```
lightwalker-eth, line 32: Interesting, I assumed `config.databaseUrl` would always be guaranteed to be defined.

lightwalker-eth (approved):
@shrugs Nice updates 👍

Issue #1153, opened 2025-10-06
configs: Improve app version number tracking & visibility


lightwalker-eth:
# Background

This issue is created as a more focused spinoff of https://github.com/namehash/ensnode/issues/621

# Goals

The config object needs to formally include distinct version numbers for each of the following:
1. ENSIndexer
2. ENSDb (for now ENSIndexer should always set this version number to be the same as the version number of ENSIndexer)
3. ENSRainbow (this value should be passed from ENSRainbow -> ENSIndexer). Note how it's possible that the ENSRainbow version can be different than the ENSIndexer version. We need to remove / update any code that incorrectly assumes these must always be the same.

Issue #1102, opened 2025-09-22
Add ens-normalize version to ENSIndexerConfig


djstrong:
New versions of ens-normalize library may change data returned by ENSIndexer.
- ens-normalize should be returned in ENSIndexerConfig
- all packages using ens-normalize and connecting to ENSNode should
  - have peerDependency of ens-normalize
  - warn if they operate with the same version of ens-normalize  as in ENSIndexerConfig

lightwalker-eth:
Related issue: https://github.com/namehash/ensnode/issues/1035

tk-o:
As just agreed with @lightwalker-eth, we all our `apps/*` to be guaranteed to use a specific version of the `@adraffy/ens-normalize` package.

Right now, what happens is that we have multiple versions of the `@adraffy/ens-normalize` package imported by different packages. For example, there are external packages we depend on that include `@adraffy/ens-normalize` on their own list of dependencies:
- `viem` uses `ox/ens` which includes `"@adraffy/ens-normalize": "^1.11.0",`
- `@ensdomains/ensjs` includes `"@adraffy/ens-normalize": "1.10.1",`

Issue #1035, opened 2025-09-09
Create visibility for the ens-normalize package version in the ENSIndexer config dependencies


lightwalker-eth:
Background: ens-normalize changes over time. See https://ensnode.io/docs/usage/querying-best-practices/#ens-normalization-standard-changes

Goal: give visibility in the public config object and in ENSAdmin for exactly what version of the ens-normalize package ENSIndexer is using.

lightwalker-eth:
Related issue: https://github.com/namehash/ensnode/issues/1102

PR #1157, opened 2025-10-06
feat(ensindexer): extend dependecy info

tk-o:
This PR addresses:

- #1102 by
  - [x] enforcing the same version of `@adraffy/ens-normalize` package being used across all `apps/*` in the monorepo, and including `ensNormalize` version in `/api/config` response.
- #1035 by
  - [x] including  `ensNormalize` version in `/api/config` response.
  - [ ] presenting `ensNormalize` version in ENSAdmin.
-  #1153 by
  - [x] including  `ensDb` version, `ensIndexer` version in `/api/config` response.
  - [ ] ensuring that  `ensRainbow` version is not tied to `ensIndexer` version.

### DIFF
--- a/apps/ensindexer/src/lib/dependency-info.ts
+++ b/apps/ensindexer/src/lib/dependency-info.ts
@@ -1,6 +1,7 @@
 import { existsSync, readFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
+import packageJson from "@/../package.json";
 import { getENSRainbowApiClient } from "@/lib/ensraibow-api-client";
 import { DependencyInfo } from "@ensnode/ensnode-sdk";
 import { makeDependencyInfoSchema } from "@ensnode/ensnode-sdk/internal";
@@ -67,17 +68,30 @@ export async function getDependencyInfo(): Promise<DependencyInfo> {
   const ensRainbowApiClient = getENSRainbowApiClient();
   const { versionInfo: ensRainbowDependencyInfo } = await ensRainbowApiClient.version();
 
+  // ENSRainbow version (fetched from the actual ENSRainbow service instance)
   // use a fallback for backwards compatibility
   const ensRainbowSchema =
     ensRainbowDependencyInfo.dbSchemaVersion ||
     // @ts-ignore
     ensRainbowDependencyInfo.schema_version;
 
+  // ENSIndexer version
+  const ensIndexerVersion = packageJson.version;
+
+  // ENSDb version
+  //
+  // For now ENSIndexer should always set this version number to be the same as
+  // the version number of ENSIndexer
+  const ensDbVersion = ensIndexerVersion;
+
   const schema = makeDependencyInfoSchema();
   const data = {
     ensRainbow: ensRainbowDependencyInfo.version,
     nodejs: process.versions.node,
     ponder: getPackageVersion("ponder"),
+    ensDb: ensDbVersion,
+    ensIndexer: ensIndexerVersion,
+    ensNormalize: getPackageVersion("@adraffy/ens-normalize"),
     ensRainbowSchema,
   } satisfies DependencyInfo;
 
--- a/apps/ensadmin/package.json
+++ b/apps/ensadmin/package.json
@@ -22,6 +22,7 @@
     "test": "vitest"
   },
   "dependencies": {
+    "@adraffy/ens-normalize": "catalog:",
     "@anthropic-ai/sdk": "^0.60.0",
     "@ensnode/datasources": "workspace:*",
     "@ensnode/ensnode-react": "workspace:*",
--- a/apps/ensadmin/src/app/mock/config-info/data.json
+++ b/apps/ensadmin/src/app/mock/config-info/data.json
@@ -22,6 +22,9 @@
     "dependencyInfo": {
       "nodejs": "22.18.0",
       "ponder": "0.11.43",
+      "ensDb": "0.35.0",
+      "ensIndexer": "0.35.0",
+      "ensNormalize": "1.11.1",
       "ensRainbow": "0.34.0",
       "ensRainbowSchema": 3
     }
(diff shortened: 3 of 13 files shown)


apps/ensindexer/src/lib/dependency-info.ts
```diff
@@ -67,17 +68,30 @@ export async function getDependencyInfo(): Promise<DependencyInfo> {
   const ensRainbowApiClient = getENSRainbowApiClient();
   const { versionInfo: ensRainbowDependencyInfo } = await ensRainbowApiClient.version();
 
+  // ENSRainbow version (fetched from the actual ENSRainbow service instance)
   // use a fallback for backwards compatibility
```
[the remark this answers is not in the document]
tk-o: Great catch! We don't have to use this fallback anymore. I'll drop it from the code 👍

apps/ensadmin/src/components/connection/config-info/config-info.tsx
```diff
@@ -145,6 +146,25 @@ export function ENSNodeConfigInfo({
                     <p className={cardItemValueStyles}>{ensIndexerConfig.dependencyInfo.nodejs}</p>
                   ),
                 },
+                {
+                  label: "ENS Normalize",
```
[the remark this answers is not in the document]
tk-o: @lightwalker-eth I've just deployed the most recent updates, you can check them out at:
https://adminensnode-62wfvv39p-namehash.vercel.app/connection?connection=https%3A%2F%2Fensnode.run.tko.box%2F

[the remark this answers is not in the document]
lightwalker-eth: @tk-o Thanks. Maybe server is down now?

CleanShot 2025-10-08 at 18 10 16

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -62,18 +67,43 @@ export function getPackageVersion(packageName: string) {
...
   // ENSRainbow version (fetched from the actual ENSRainbow service instance)
-  // use a fallback for backwards compatibility
-  const ensRainbowSchema =
-    ensRainbowDependencyInfo.dbSchemaVersion ||
-    // @ts-ignore
-    ensRainbowDependencyInfo.schema_version;
+  const ensRainbowSchema = ensRainbowVersionInfo.dbSchemaVersion;
 
   // ENSIndexer version
   const ensIndexerVersion = packageJson.version;
```
lightwalker-eth, line 109: Suggest moving the logic to get the ENSIndexer version into a separate helper function.

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -84,21 +114,21 @@ export async function getDependencyInfo(): Promise<DependencyInfo> {
   // the version number of ENSIndexer
   const ensDbVersion = ensIndexerVersion;
 
-  const schema = makeDependencyInfoSchema();
+  const schema = makeVersionInfoSchema();
   const data = {
```
lightwalker-eth, line 118: Hmm. I'm not so clear on what's going on here.

Are we creating what is supposed to be a `SerializedVersionInfo` here and then attempting to deserialize it through `makeVersionInfoSchema` and then returning the deserialized `VersionInfo`?

Or are we just attempting to verify that the `VersionInfo` we return passes our validation checks?

If so, then suggest to rename `data` to something meaningful. Ex: `unvalidatedVersionInfo` and then rename `parsed` to something like `validatedVersionInfo`.

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -62,18 +67,43 @@ export function getPackageVersion(packageName: string) {
 }
 
 /**
- * Get complete {@link DependencyInfo} for ENSIndexer app.
+ * Search PNPM's .pnpm directory for a package
+ * PNPM stores packages like: .pnpm/@adraffy+ens-normalize@1.10.1/node_modules/@adraffy/ens-normalize
+ */
+function searchPnpmStore(pnpmDir: string, packageName: string): string | null {
```
lightwalker-eth, line 73: It's not clear what this function returns.

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -62,18 +67,43 @@ export function getPackageVersion(packageName: string) {
...
+  }
+
+  return null;
+}
+
+/**
+ * Get complete {@link VersionInfo} for ENSIndexer app.
  */
-export async function getDependencyInfo(): Promise<DependencyInfo> {
+export async function getVersionInfo(): Promise<VersionInfo> {
```
lightwalker-eth, line 101:
```suggestion
export async function getEnsIndexerVersionInfo(): Promise<EnsIndexerVersionInfo> {
```

Goal: We should note how we're going to have nested "VersionInfo" objects for different apps in our overall architecture. Each app's "VersionInfo" should have a name that makes it explicit exactly which app's "VersionInfo" it is related to.

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -84,21 +114,21 @@ export async function getDependencyInfo(): Promise<DependencyInfo> {
   // the version number of ENSIndexer
   const ensDbVersion = ensIndexerVersion;
 
-  const schema = makeDependencyInfoSchema();
+  const schema = makeVersionInfoSchema();
   const data = {
-    ensRainbow: ensRainbowDependencyInfo.version,
+    ensRainbow: ensRainbowVersionInfo.version,
```
lightwalker-eth, line 119: I believe we should be building nested schemas.

In other words:
1. The type of `ensRainbow` here should be the type of an `EnsRainbowVersionInfo` and contain nested fields inside of itself.
2. The `ensRainbowSchema` field should be removed at this level. It already lives inside `EnsRainbowVersionInfo`, right?

lightwalker-eth (commented):
@tk-o Thanks for updates 👍 Reviewed and shared feedback

Issue #1112, opened 2025-09-22
RPC Config QoL Refactor


shrugs:
allow the passing of `ALCHEMY_API_KEY` and `DRPC_KEY` (and infura, etc) to auto-generate a comprehensive set of RPCConfig defaults. allow the existing chain-id-specific env variable parsing to override this auto-generated.

if multiple are specified, use a const ordering `['alchemy', 'drpc', 'infura', ...]` to construct the environment object that gets passed to zod.

also generate and provide the ws:// urls at the end of the generated environment

PR #1161, opened 2025-10-07
RPC Shortcuts & Environment Refactor

shrugs:
closes https://github.com/namehash/ensnode/issues/1112

- refactors `ENSIndexerEnvironment` to match 1:1 with the actual environment object that we access through `process.env`
- types the `process.env` as `ENSIndexerEnvironment` — now access of `process.env` is typechecked
- refactor `buildConfigFromEnvironment` to accept any `ENSIndexerEnvironment` and produce an ENSIndexerConfig
- confirmed that process.env isn't used anywhere else in ensindexer (except for OTel configuration which is unimportant here)


apps/ensindexer/types/env.d.ts
```diff
@@ -0,0 +1,7 @@
+import type { ENSIndexerEnvironment } from "@/config/types";
+
+declare global {
+  namespace NodeJS {
+    interface ProcessEnv extends ENSIndexerEnvironment {}
```
shrugs, line 5: here's where we extend process.env with our new ensindexerenvironment type

apps/ensindexer/src/config/rpc-configs-from-env.ts
```diff
@@ -0,0 +1,51 @@
...
+  ).map(([, datasource]) => datasource.chain);
+
+  const alchemyApiKey = env["ALCHEMY_API_KEY"];
+  const drpcKey = env["DRPC_API_KEY"];
+
+  const rpcConfigs: Record<ChainIdString, RpcConfigEnvironment> = {};
+
+  for (const chain of chainsInNamespace) {
+    const urls = [
+      // specific env value, if specified
```
lightwalker-eth, line 34: I think if a specific env value is specified for a chainId then we don't add alchemyApiKey or drpcKey for that chainId?

In other words, if you manually set an env value for a chainId, then you get total control over how that chainId's rpc config works. The alchemy / drpc urls would not be autogenerated for that chainId, even if an alchemyApiKey / drpcKey were defined.

wdyt?

apps/ensindexer/src/config/rpc-configs-from-env.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Constructs dynamic chain configuration from environment variables, scoped to chain IDs that appear
+ * in the specified `namespace`.
+ *
+ * This function provides the following RPC URLs in the following order:
+ * 1. RPC_URL_*, if available in the env
+ * 2. Alchemy, if ALCHEMY_API_KEY is available in the env
+ * 3. DRPC, if DRPC_API_KEY is available in the env
+ *
+ * TODO: also inject wss:// urls for alchemy, drpc keys
```
lightwalker-eth, line 15: Feel welcome to move this idea into a separate follow up issue if you want.

One quick idea for this might be to introduce another environment variable: `WSS_RPC_AUTOGEN` (or something like that).

Options for this variable might then be:
* `none` (default)
* `alchemy`
* `drpc`

An invariant might then be defined for building a config:
* if `WSS_RPC_AUTOGEN` is `alchemy` then `ALCHEMY_API_KEY` must be defined.
* if `WSS_RPC_AUTOGEN` is `drpc` then `DRPC_API_KEY ` must be defined.

Then, if `WSS_RPC_AUTOGEN` is not `none` (default) then each chainId in the namespace would have an autogenerated `wss` RPC included in its config.

Or if you don't like that feel welcome to scrap it. Just a quick idea.

.changeset/early-pants-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"ensindexer": minor
+---
+
+ALCHEMY*API_KEY and DRPC_API_KEY may now be set in lieu of RPC_URL*_ environment variables. If specified, they will be used in the following order: RPC*URL*_ > Alchemy > DRPC.
```
lightwalker-eth, line 5:
```suggestion
ALCHEMY_API_KEY and DRPC_API_KEY may now be set in lieu of RPC_URL*_ environment variables. If specified, they will be used in the following order: RPC_URL*_ > Alchemy > DRPC.
```

lightwalker-eth (approved):
@shrugs Looks great. Shared some minor feedback. Please merge when ready 👍

PR #1179, opened 2025-10-15
refactor: simplify TS types

tk-o:
Drops  from  type and updates all consumers of  type to know when they can access the  property. Also, this PR drops  type from ENSAdmin, as we can now rely on the  one.

### DIFF
--- a/apps/ensadmin/src/components/indexing-status/indexing-stats.tsx
+++ b/apps/ensadmin/src/components/indexing-status/indexing-stats.tsx
@@ -183,7 +183,10 @@ export function IndexingStatsForSnapshotCompleted({
   const chainEntries = sortChainStatusesByStartBlockAsc([...omnichainSnapshot.chains.entries()]);
 
   return chainEntries.map(([chainId, chain]) => {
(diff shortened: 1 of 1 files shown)


apps/ensadmin/src/components/indexing-status/indexing-stats.tsx
```diff
@@ -176,7 +183,7 @@ export function IndexingStatsForSnapshotCompleted({
   const chainEntries = sortChainStatusesByStartBlockAsc([...omnichainSnapshot.chains.entries()]);
 
   return chainEntries.map(([chainId, chain]) => {
-    const endBlock = chain.config.endBlock ? blockViewModel(chain.config.endBlock) : null;
+    const endBlock = chain.config.endBlock ? chain.config.endBlock : null;
```
lightwalker-eth: Suggest to update this logic to check `chain.config.configType` and then decide the endBlock based on that.

tk-o: Nice catch, will update.

PR #1191, opened 2025-10-20
feat(apps/ensadmin): add splash page

notrab:
This handles the root `/` route for ENSAdmin.


lightwalker-eth (approved):
@notrab I like the strategy you took here. Thank you. Looks nice 👍

PR #1194, opened 2025-10-20
ENSAPI Service

### LINKED ISSUE #945: Remove /metadata endpoint
Currently reviewing https://github.com/namehash/ensnode/pull/919

Here it looks like we're removing any use of the `/metadata` endpoint. Is that right? If so let's fully remove all code related to that endpoint.

As a follow-up task, please also cleanup any related old code we no longer need. For example, can we remove?
- `@/lib/ponder-metadata-provider`
- the whole `@ensnode/ponder-metadata` package

### LINKED ISSUE #1114: Create ENS API Service
- [x] add another app to the monorepo `ensapi`, it is a hono app
- [x] ~port behavior from ponder's built-in api to shared lib (`ensapi-private`?) and expose via both ponder api and ensapi app~
- [x] configured with `DATABASE_*`, and Root Chain RPC
- [x] refactor shared behaviors from `ENSIndexerConfig` such that they can be re-used by `ENSApiConfig`
- [x] establish connected ensindexer's (public) config & status from database
- [x] always serve all of the relevant api endpoints
  - [x] add middleware to relevant apis that conditionally enables/disables the API depending on connected config
- [x] log warning on startup if connected ensindexer does not have `resolution` plugin, mention that certain protocol acceleration pathways will be disabled
- [ ] ~indexer service doesn't need `ENSINDEXER_URL` because it can just use `localhost:{process.envPORT}` but maybe that's a bad default and it needs to be overridden.~

all of the api needs to live in a shared internal library (`ensapi-private`?) that exports a hono app and the public `ensapi` app is just a thin hono server that includes the ENSNode API implementation

- [x] add new app to CI workflows
- [x] update terraform config to deploy this service
- [ ] queue up documentation to adjust railway environments to use this new service — instead of `ponder serve`
- [ ] update ens-test-env config to use the new ensapi service? maybe ens-test-env is a good example of a deployment that _doesn't_ want to have a separate api service and just wants to use ponder's built-in.
- [x] ensindexer should be able to serve its own healthcheck based on indexing status, so perhaps it _does_ need to be able to self-fetch with `ENSINDEXER_URL`...
- [x] docker config for ensapi depends on ensindexer healthy to start (?)
- [x] ensapi's healthcheck maybe depends on ensindexer when starting up but remains healthy even if ensindexer goes down?

### LINKED ISSUE #1174: Optimise `canAccelerateResolution` function
PR #1095 introduced concepts of projections and snapshots.

This issue requires ENSIndexer to store cached projection in memory and update it asynchronously at an interval
so that new snapshots are never needed to be built synchronously with handling a resolution request.

In other words, the `canAccelerateResolution()` function implemenation:
https://github.com/namehash/ensnode/blob/637fcc0e23b4cdd1e083a044b1e2f3d33a6bcc7c/apps/ensindexer/src/api/lib/indexing-status/can-accelerate-resolution.ts#L22
needs to be altered to ensure the `realtimeProjection` is loaded from cache.

The cache needs to be updated asynchronously at an interval, and by `GET /api/indexing-status` request handler each time it gets called.

Edit: See related issue: https://github.com/namehash/ensnode/issues/966
Edit: See related discussion: https://github.com/namehash/ensnode/pull/1095/files/d22bac87153c06a1c47e3a0ae4202d969e19e9e9..24a9e89172440e441d7f9ee94a91c6d24b125023#r2408088555

shrugs:
closes #945 
closes #1114
closes #1174

- [x] obfuscate/pretty-print the ensindexerpublicconfig/ensapiconfig
- [x] docs pass
- [x] docstrings pass
- [x] changesets
- [x] fill out Railway update spec https://github.com/namehash/ensnode/issues/1198

Followup Issue: https://github.com/namehash/ensnode/issues/1195

### DIFF
--- a/docs/ensnode.io/config/integrations/starlight.ts
+++ b/docs/ensnode.io/config/integrations/starlight.ts
@@ -110,6 +110,32 @@ export function starlight(): AstroIntegration {
           ],
         },
         {
+          label: "ENSApi",
+          link: "/ensapi",
+          icon: "star",
+          items: [
+            {
+              label: "Overview",
+              items: [
+                {
+                  label: "What is ENSApi?",
+                  link: "/ensapi",
+                },
+              ],
+            },
+            {
+              label: "Using ENSApi",
+              collapsed: false,
+              autogenerate: { directory: "ensapi/usage" },
+            },
+            {
+              label: "Contributing",
+              collapsed: false,
+              autogenerate: { directory: "ensapi/contributing" },
+            },
+          ],
+        },
+        {
           label: "ENSRainbow",
           link: "/ensrainbow",
           icon: "star",


apps/ensindexer/src/config/config.schema.ts
```diff
@@ -114,62 +82,22 @@ const EnsRainbowUrlSchema = makeUrlSchema("ENSRAINBOW_URL");
...
-);
-
 const IsSubgraphCompatibleSchema =
   makeEnvStringBoolSchema("SUBGRAPH_COMPAT").default(DEFAULT_SUBGRAPH_COMPAT);
 
 const ENSIndexerConfigSchema = z
   .object({
-    namespace: ENSNamespaceSchema,
-    globalBlockrange: BlockrangeSchema,
+    databaseUrl: DatabaseUrlSchema,
```
shrugs: using same schemas as previous, now shared between ensindexer/ensapi via ensnode-sdk/internal

apps/ensindexer/src/lib/lib-helpers.ts
```diff
@@ -1,9 +1,5 @@
 export const bigintMax = (...args: bigint[]): bigint => args.reduce((a, b) => (a > b ? a : b));
 
-export const hasNullByte = (value: string) => value.indexOf("\u0000") !== -1;
```
shrugs: moved to shared lib

apps/ensindexer/src/plugins/protocol-acceleration/plugin.ts
```diff
@@ -113,11 +119,19 @@ export default createPlugin({
...
       contracts: {
         // a multi-chain Resolver ContractConfig
-        [namespaceContract(pluginName, "Resolver")]: resolverContractConfig(
-          config.namespace,
-          DATASOURCES_WITH_RESOLVERS,
-          config.globalBlockrange,
-        ),
+        [namespaceContract(pluginName, "Resolver")]: {
+          abi: ResolverABI,
+          chain: getDatasourcesWithResolvers(config.namespace).reduce(
```
shrugs: using the shared `getDatasourcesWithResolvers` to generate the Resolver ContractConfig for the protocol acceleration plugin

apps/ensindexer/test/lib-helpers.test.ts
```diff
@@ -1,28 +1,11 @@
...
+import { bigintMax } from "@/lib/lib-helpers";
+
 describe("helpers", () => {
   describe("bigintMax", () => {
     it("should return the maximum bigint value", () => {
       expect(bigintMax(1n, 2n, 3n)).toBe(3n);
     });
   });
-
-  describe("hasNullByte", () => {
```
shrugs: tests moved to shared as well

apps/ensadmin/package.json
```diff
@@ -56,6 +56,7 @@
     "date-fns": "catalog:",
     "gqlpt": "0.0.0-alpha.33",
     "graphiql": "^3.8.3",
+    "graphql": "^16.10.0",
```
shrugs: This fixes that really annoying peer dependency log when installing.

apps/ensapi/src/handlers/resolution-api.ts
```diff
@@ -37,21 +30,36 @@ const getCanAccelerateResolution = simpleMemoized(canAccelerateResolution, 30_00
...
       const { result, trace } = await captureTrace(() =>
-        resolveForward(name, selection, { accelerate }),
+        resolveForward(name, selection, { accelerate, canAccelerate }),
       );
 
       const response = {
         records: result,
-        accelerationAttempted: accelerate,
+
+        accelerationRequested: accelerate,
```
shrugs: i added this acceleration requested and acceleration attempted metadata to the resolution api, we can integrate it into ensadmin's tracing page later (in followup pr)

shrugs (commented):
followup issue: https://github.com/namehash/ensnode/issues/1195

the metadata endpoint has removed.

the anthropic api key spec has been moved to a separate issue. https://github.com/namehash/ensnode/issues/1197

regarding config: i've decided at the moment to JUST proxy the `EnsIndexerPublicConfig` rather than setting up the `EnsApiPublicConfig` — this way, nothing, including ENSAdmin, needs any updating in this PR. this note has beed deferred to the followup issue

re: ensapiconfig & namespace/rpcConfigs:
you can see how the namespace and rpc configs are downstream of the connected ensindexer. the rpc configs are downstream because we materialize them based on namespace. i think in the future ensapi's rpc config need not be materialized and should be derived on demand (i.e. in `getPublicClient`), but i see why materialization also has its benefits. in any case, it requires knowing the namespace at runtime for ENS API in order to generate the rpc configs. but that requires some minor updates to the libraries that depend on synchronous access to `config.rpcConfigs`.

i'm aware that there could be more docstrings — i didn't have the time — please feel free to open a pr with suggested docstring changes (`gh pr checkout 1194`) or i can get to them when i'm back

the docs also need a review, which i will do when i'm back

docker-compose.yml
```diff
@@ -58,30 +68,26 @@ services:
...
   ensadmin:
     container_name: ensadmin
     image: ghcr.io/namehash/ensnode/ensadmin:latest
     ports:
       - "4173:4173"
     environment:
-      # Override ENSADMIN_PUBLIC_URL to point to docker compose ensnode-api
-      # Note: it must be a "public" URL accessible from a web browser (i.e. it cannot be a hostname exclusive to the internal docker network)
+      # Override environment variables to point to docker instances
       ENSADMIN_PUBLIC_URL: http://localhost:4173
```
lightwalker-eth: Are we still using this environment variable for something?

Please ignore if not relevant.

docker-compose.yml
```diff
@@ -58,30 +68,26 @@ services:
...
-      # Override NEXT_PUBLIC_DEFAULT_ENSNODE_URLS to the docker compose ensnode-api
-      # Note: it must be a comma-separated list of URLs that are accessible from a web browser
-      # (i.e. it cannot be a hostname in the docker network)
-      NEXT_PUBLIC_DEFAULT_ENSNODE_URLS: http://localhost:42069
+      NEXT_PUBLIC_SERVER_CONNECTION_LIBRARY: http://localhost:4334
```
lightwalker-eth:
```suggestion
      NEXT_PUBLIC_SERVER_CONNECTION_LIBRARY: http://ensapi:4334
```

Is that ok? Please ignore if not relevant

apps/ensindexer/src/config/environment.ts
```diff
@@ -0,0 +1,28 @@
+import {
+  DatabaseEnvironment,
+  EnsIndexerUrlEnvironment,
+  EnsNamespaceEnvironment,
+  RpcEnvironment,
+} from "@ensnode/ensnode-sdk/internal";
```
lightwalker-eth: Perhaps instead of creating this `/internal` subdirectory within `ensnode-sdk`, would it be better to create a new package in our monorepo called `ensnode-internal` that is not published to NPM?

Please feel welcome to transfer this idea into a separate follow up issue if you think it is valid.

apps/ensapi/src/config/environment.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  DatabaseEnvironment,
+  EnsIndexerUrlEnvironment,
+  EnsNamespaceEnvironment,
+  PortEnvironment,
+  RpcEnvironment,
+} from "@ensnode/ensnode-sdk/internal";
+
+/**
+ * Represents the raw, unvalidated environment variables for the ENSAPI application.
```
lightwalker-eth:
```suggestion
 * Represents the raw, unvalidated environment variables for the ENSApi application.
```

lightwalker-eth (approved):
@shrugs Great to see your work here 🚀 Shared a few comments with suggestions. Please merge when ready 👍

docker-compose.yml
```diff
@@ -2,43 +2,53 @@ services:
...
     image: ghcr.io/namehash/ensnode/ensindexer:latest
+    ports:
+      - "42069:42069"
     environment:
-      # Override DATABASE_URL to point to docker compose postgres
+      # Override environment variables to point to docker instances
       DATABASE_URL: postgresql://postgres:password@postgres:5432/postgres
-      # Override ENSRAINBOW_URL to point to docker compose ensrainbow
       ENSRAINBOW_URL: http://ensrainbow:3223
+      ENSINDEXER_URL: http://ensindexer:42069
```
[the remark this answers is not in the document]
shrugs: it does seem like it's required at the moment — the indexing status api served by ensindexer requires the ability to self-fetch the ponder metrics and ponder status pages. i don't think there's a better way to configure this at the moment

docker-compose.yml
shrugs: unfortunately not possible here — because these urls must be accessible on the host (it's the browser that uses that `http://localhost:4334` url to talk to ensapi, not the docker container) it has to be host-accessible, and i don't think the docker hostname's work on the host by default

apps/ensindexer/src/config/environment.ts
shrugs: https://github.com/namehash/ensnode/issues/1206

apps/ensapi/.env.local.example
```diff
@@ -0,0 +1,105 @@
+# The port ENSApi listens on.
+# Optional. If this is not set, the default value is 4334.
+# PORT=4334
+
+# RPC configuration
```
lightwalker-eth: Suggest we refine the RPC configuration docs for ENSApi:

1. This should only need an RPC config for the ENS root chain.
2. Suggest we document how ENSIndexer builds its config on startup to determine its ENS root chain, and then how ENSApi connects to ENSIndexer via its `/api/config` endpoint to determine what root chain ENSApi must support and therefore which RPC must be configured for ENSApi.
3. Shouldn't there be only a single RPC endpoint defined for the ENS root chain for ENSApi, not 1 or more?
4. Shouldn't there be 0 WSS RPC endpoints defined for ENSApi?

apps/ensapi/.env.local.example
```diff
@@ -0,0 +1,105 @@
...
+# Required. When ENSIndexer starts up it verifies RPCs are defined for each of the chains
+# that are to be indexed based on the configured NAMESPACE and PLUGINS, along with the ENS Root Chain.
+#
+# NOTE: You must configure your own private RPC endpoints.
```
lightwalker-eth: Suggest we refine the docs here about ENSIndexer making millions of RPC requests during operation. The RPC endpoints configured for ENSApi should be distinct from the RPC endpoints configured for ENSIndexer. The volume of RPC calls performed by ENSApi is a function of the API requests sent to ENSApi and other factors such as if those requests are accelerated or not.

apps/ensapi/.env.local.example
```diff
@@ -0,0 +1,105 @@
...
+# To optimize performance, Ponder automatically adapts to the rate limits (429 errors) of each
+# provided HTTP/HTTPS RPC. More details at: https://ponder.sh/docs/config/chains#rpc-endpoints
+#
+# The Resolution API, as necessary, issues RPC requests to the _primary_ HTTP RPC URL for a given
+# chain.
+
+# === ENS Namespace: Mainnet ===
+# Ethereum Mainnet
+# - required if the configured namespace is mainnet
+# - required by plugins: subgraph, protocol-acceleration, referrals, tokenscope
```
lightwalker-eth: For each of the ENS Namespaces documented here, suggest we remove any reference of "required by plugins".

As I understand, the RPC configuration of ENSApi has nothing to do with ENSIndexer plugin configurations and is exclusively a function of the configured namespace.

PR #1201, opened 2025-10-22
feat(apps/ensadmin): UX loading improvements

### LINKED ISSUE #1199: ENSAdmin UI Refinements
Image

~1. Please remove the `select-none` or whatever class is causing this text not to be treated as regular text by the browser.~ Done
~2. Let's update this button to read: "View My Connection".~ Done

After I click this button and navigate to `/connection` I see this loading state for several seconds:

Image

Which makes the app feel broken as this loading state is too empty. You might not see this loading state for as long as I do when making a request from Dubai, but it's definitely a UX issue we need to optimize.

For example, shouldn't everything I've highlighted in red in the screenshot below be something we can display during the loading state and we only need loading skeletons for the content I highlighted in blue?

Image

notrab:
This PR enhances the loading state for the connection page:

- If `ENSNodeConfigInfo` is fetching you see a loading spinner on the page (any page in fact)
- If `ensIndexerConfig` is loading, we show the ENSAdmin Version/Connection URL (because we have it from above valid connection)
  - Loading state shows for the ENSNode as outlined in #1199

CleanShot 2025-10-26 at 13 04 56@2x

There was also excess padding for the error states, that is now cleared up:

CleanShot 2025-10-26 at 13 09 52@2x

Other padding issues resolved, and consistent now too:

CleanShot 2025-10-26 at 14 37 47@2x

CleanShot 2025-10-26 at 14 36 10@2x

CleanShot 2025-10-26 at 14 35 44@2x

- [x] Add Changeset
- [x] Refactor ENSNodeConfig loading block

Closes #1199

### DIFF
--- a/apps/ensadmin/src/components/connection/config-info/config-info.tsx
+++ b/apps/ensadmin/src/components/connection/config-info/config-info.tsx
@@ -5,26 +5,28 @@
 
 "use client";
 
+import { Replace } from "lucide-react";
+import { ReactNode } from "react";
+
+import { useENSIndexerConfig } from "@ensnode/ensnode-react";
+import { ENSIndexerPublicConfig } from "@ensnode/ensnode-sdk";
+
 import { ChainIcon } from "@/components/chains/ChainIcon";
 import { ConfigInfoAppCard } from "@/components/connection/config-info/app-card";
 import { ErrorInfo, type ErrorInfoProps } from "@/components/error-info";
-import { HealIcon } from "@/components/icons/HealIcon";
-import { IndexAdditionalRecordsIcon } from "@/components/icons/IndexAdditionalRecordsIcon";
 import { ENSDbIcon } from "@/components/icons/ensnode-apps/ensdb-icon";
 import { ENSIndexerIcon } from "@/components/icons/ensnode-apps/ensindexer-icon";
 import { ENSNodeIcon } from "@/components/icons/ensnode-apps/ensnode-icon";
 import { ENSRainbowIcon } from "@/components/icons/ensnode-apps/ensrainbow-icon";
 import { IconGraphNetwork } from "@/components/icons/graph-network";
+import { HealIcon } from "@/components/icons/HealIcon";
+import { IndexAdditionalRecordsIcon } from "@/components/icons/IndexAdditionalRecordsIcon";
 import { ExternalLinkWithIcon } from "@/components/link";
 import { Card, CardContent, CardHeader, CardTitle } from "@/components/ui/card";
 import { Skeleton } from "@/components/ui/skeleton";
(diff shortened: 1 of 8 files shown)


apps/ensadmin/src/components/ensadmin-version.tsx
```diff
@@ -0,0 +1,26 @@
+"use client";
+
+import { Skeleton } from "@/components/ui/skeleton";
+import { ensAdminVersion } from "@/lib/env";
+import { Suspense, use } from "react";
+
+const versionPromise = ensAdminVersion();
```
notrab: We use `async` to fetch the version from the `package.json` file inside `@/lib/env` so we need to wrap this in a promise and pass it around in React land as children/ReactElement so it can be suspended.

apps/ensadmin/src/components/connection/config-info/config-info.tsx
```diff
@@ -7,61 +7,86 @@
...
- *
- * Standard - ensIndexerConfig: ENSIndexerPublicConfig, error: undefined
- * Loading - ensIndexerConfig: undefined, error: undefined
- * Error - ensIndexerConfig: undefined, error: ErrorInfoProps
- *
- * @throws If both ensIndexerConfig and error are defined
+ * Props for ENSNodeConfigInfoView - internal component that accepts props for testing/mocking
  */
-export interface ENSNodeConfigProps {
+export interface ENSNodeConfigInfoViewProps {
```
notrab: We probably want to rename this to `Display` so it's inline with other display pure-like components.

apps/ensadmin/src/components/connection/config-info/config-info.tsx
```diff
@@ -7,61 +7,86 @@
...
+    description: string;
+  };
+  isLoading?: boolean;
 }
 
-export function ENSNodeConfigInfo({
+/**
+ * Internal view component that accepts props - used by both the main component and mock pages
+ */
+export function ENSNodeConfigInfoView({
```
notrab: Rename to `ENSNodeConfigInfoDIsplay`

apps/ensadmin/src/components/connection/config-info/config-info.tsx
```diff
@@ -116,263 +141,251 @@ export function ENSNodeConfigInfo({
...
-              ]}
-              version={ensIndexerConfig.versionInfo.ensRainbow}
-              docsLink={new URL("https://ensnode.io/ensrainbow/")}
-            />
-          </div>
-        </CardContent>
-      </Card>
-    </div>
+  return (
+    <Card className="w-full">
```
notrab: Refactor this so we fallback to the loading existing state.

PR #1209, opened 2025-10-27
ENSApi Docs + Biome v2

shrugs:
- updates ENSApi .env.example + some docs
- update biome to v2 (explicit monorepo support) and format all files with new import grouping
  - I've manually verified every diff in this pr, so if you want to avoid browsing all of the import and linting changes, simply see the self-review notes below.

### DIFF
--- a/apps/ensapi/src/lib/tracing/treeify-trace.ts
+++ b/apps/ensapi/src/lib/tracing/treeify-trace.ts
@@ -51,7 +51,8 @@ export function treeifySpans(trace: ReadableSpan[]): ProtocolTrace {
   for (const node of idToNode.values()) {
     const parentId = node.parentSpanContext?.spanId;
     if (parentId && idToNode.has(parentId)) {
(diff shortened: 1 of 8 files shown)


apps/ensadmin/package.json
```diff
@@ -69,7 +69,6 @@
     "viem": "catalog:"
   },
   "devDependencies": {
-    "@biomejs/biome": "catalog:",
```
shrugs: The per-package dependency is no longer required due to monorepo usage.

docs/ensrainbow.io/biome.jsonc
```diff
@@ -0,0 +1,21 @@
+{
+  "$schema": "https://biomejs.dev/schemas/2.3.1/schema.json",
+  "extends": "//",
+  "assist": {
+    "actions": {
+      "source": {
+        "organizeImports": "off"
```
shrugs: Also disabling import organization for ensrainbow site

PR #1216, opened 2025-10-28
fix(apps/ensadmin): resolved stability issue that could cause the browser to crash

notrab:
Fixed browser crashes on hard refresh by replacing unsupported `use()` hook with standard React state management in version display component for ENSAdmin.

### DIFF
--- a/.changeset/angry-papers-dig.md
+++ b/.changeset/angry-papers-dig.md
@@ -2,4 +2,4 @@
 "ensadmin": minor
 ---
 
-Resolved stability issue that could cause the browser to crash when refreshing pages in the app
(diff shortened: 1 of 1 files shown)


.changeset/angry-papers-dig.md
```diff
@@ -0,0 +1,5 @@
+---
+"ensadmin": minor
+---
+
+Resolved stability issue that could cause the browser to crash when refreshing pages in the app
```
notrab:
```suggestion
Resolved stability issue that could cause the browser to crash when refreshing pages in ENSAdmin
```

PR #1218, opened 2025-10-28
hotfix: hydration for connections library

notrab:


apps/ensadmin/src/hooks/use-connections-library.tsx
```diff
@@ -197,5 +197,23 @@ export { useConnectionsLibrary };
...
+  return (
+    <Suspense fallback={null}>
+      <ConnectionsLibraryProviderInner>{children}</ConnectionsLibraryProviderInner>
+    </Suspense>
+  );
+}
+
+/**
+ * Wrapper component that delays rendering children until after hydration.
+ * This prevents crashes from useSearchParams during static export hard refreshes.
```
notrab: We'll remove this when we leave Next.js!

apps/ensadmin/src/hooks/use-connection-url-param.tsx
```diff
@@ -59,7 +59,11 @@ export function useRawConnectionUrlParam(): UseRawConnectionUrlParamResult {
   const searchParams = useSearchParams();
 
   // Get the current raw connection URL param
-  const rawConnectionUrlParam = searchParams.get(RAW_CONNECTION_PARAM_KEY);
+  // Use useMemo to prevent unnecessary re-renders
+  const rawConnectionUrlParam = useMemo(
```
shrugs: not sure the usememo is doing anything here?

notrab: Reverted. There was a moment during hydration where the param wasn't present, and then was added, and then readded during my testing which was causing an infinite loop. But I can't seem to recreate this now so reverting the change.

PR #1219, opened 2025-10-28
ENSAdmin Runtime Environment Variables

shrugs:
Turns out this is a pretty normal thing to do. This works by modifying the docker entry point to inject specific environment variables into a javascript file that's placed in the output directory and then loaded by the application before next.js executes hydration. Then by conditionally reading from the runtime variables before reading from the next.js replaced process.env variables, we support runtime overrides of the build time environment. 

also updated pnpm while i was at it

closes https://github.com/namehash/ensnode/issues/1215

### DIFF
--- a/apps/ensadmin/src/hooks/use-connections-library.tsx
+++ b/apps/ensadmin/src/hooks/use-connections-library.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import constate from "constate";
-import { useCallback, useEffect, useMemo, useState } from "react";
+import { Suspense, useCallback, useEffect, useMemo, useState } from "react";
 import { useLocalstorageState } from "rooks";
 import { toast } from "sonner";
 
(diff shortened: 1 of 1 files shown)


apps/ensadmin/src/lib/env.ts
```diff
@@ -60,40 +56,40 @@ function isAppOnVercelPlatform(): boolean {
...
+ * Get ENSAdmin service public HttpHostname.
+ *
+ * Note: a Vercel fallback HttpHostname will be used if application runs on Vercel platform.
+ * If the Vercel fallback HttpHostname cannot be applied, then default HttpHostname will be used.
+ *
+ * @returns application public HttpHostname for ENSAdmin
+ * @throws when Vercel platform was detected but could not determine HttpHostname
+ */
+export function ensAdminPublicUrl(): HttpHostname {
+  const envVarValue =
```
shrugs: Simplified the defaulting logic here such that ENSadmin, vercel, and localhost all return string, which is then parsed by buildHttpHostname — DRYs some of this code up and reads much simpler now

PR #1226, opened 2025-10-30
Refactor/ensindexer src

### LINKED ISSUE #1103: ENSv2 Indexing Prep
Parent issue for preparing the repo/codebase for ENSv2 implementation

### LINKED ISSUE #1147: Reorganize top-level ensindexer lib/config etc folders
rn they're in src which should be exclusively for ponder-source-able handler files — if we move them out of that directory and into a top level lib/config/etc then we'll be able to colocate tests. big refactor though, so needs to wait until after indexing status has gone through

it might be nice to have a src/[plugin-name].ts file instead of the [plugin-name]/event-handlers.ts approach — that way it's extremely clear that these files are being included/executed by ponder in order to register handlers

### LINKED ISSUE #1188: Move identity types/lib to idenity folder and out of resolution

shrugs:
closes #1147
closes #1188
closes #1103

- moves identity-realted lib/types into `ensnode-sdk/src/identity`
- moves ponder root to `./ponder`
  - this allows us to finally colocate test files within ensindexer without ponder complaining about importing vitest code
- adds a `ponder/src/register-handlers.ts` file that centralizes the conditional registration of handlers, better aligning with ponder's execution model. each plugin's `event-handlers.ts` is now just a barrel of individual contract handlers, abstracting that concept away from the top-level registration


apps/ensindexer/Dockerfile
```diff
@@ -16,4 +16,4 @@ WORKDIR /app/apps/ensindexer
 ENV NODE_ENV=production
 EXPOSE 42069
 
-CMD ["pnpm", "exec", "ponder", "start"]
```
shrugs: pnpm start now includes the custom root

apps/ensindexer/ponder/src/register-handlers.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * This file is loaded by ponder as part of ponder dev & ponder start. In here we conditionally
```
shrugs: here is the new register-handlers.ts that aggregates the individual event-handlers.ts files, simplifying the understanding of how ponder expects handlers to be registered within the `(root)/src` directory

apps/ensindexer/src/lib/graphnode-helpers.test.ts
```diff
@@ -2,14 +2,14 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import type { LabelHash } from "@ensnode/ensnode-sdk";
 
-import { setupConfigMock } from "./utils/mockConfig";
+import { setupConfigMock } from "@/lib/__test__/mockConfig";
```
shrugs: i've moved test-related meta files to `@/lib/__test__` which seems to be relatively well-established pattern

apps/ensindexer/src/lib/version-info.ts
```diff
@@ -9,13 +11,6 @@ import { makeENSIndexerVersionInfoSchema } from "@ensnode/ensnode-sdk/internal";
 
 import { getENSRainbowApiClient } from "@/lib/ensraibow-api-client";
 
-/**
- * Get version of ENSIndexer application.
- */
-export async function getENSIndexerVersion(): Promise<string> {
```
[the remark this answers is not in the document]
tk-o: Ah, because of the inline import. Not sure for why I it was inline import in the first place, but I remember @lightwalker-eth wanted a separate `getENSIndexerVersion` function to better highlight the idea getting ENSIndexer version.

Issue #1185, opened 2025-10-17
Create ENS Referral Program leaderboard API as (temporary new) Hono app backend to be consumed by ENSAwards frontend


djstrong:
# Background

- We are launching ENS Holiday Awards as an ENS Referral Incentive Program starting Dec 1. Details are here: https://discuss.ens.domains/t/rfc-ens-holiday-awards/21359
- The ENSAwards site is planned to serve as the main landing page for the ENS Holiday Awards program. Expected functionality on the ENSAwards site includes the creation of a Leaderboard, showing which referrers are currently the top referrers.
- @lightwalker-eth and @tk-o Are working on a refined data model for ENS registrations and referrals. Once this work is complete the exact SQL query and response data model will need to also be refined, but it's important not to be blocked on this. We should proceed now with a demonstration of the overall proof of concept here using the existing registrations and referrals data model.
- @shrugs Is working to build a new app called ENSApi. Eventually we may want to transition the API discussed in this GitHub issue into ENSApi, but it's also important for the goals in this GitHub issue not to be blocked by the work on ENSApi. Therefore, the work on this issue requires the creation of an all new [Hono app](https://hono.dev/) dedicated exclusively for this 1 new API that is needed at the moment.
- The database query to calculate the leaderboard of top referrers is expected to be large and slow to execute dynamically for each visitor to the ENSAwards leaderboards. Therefore, here we propose the creation of ENSAnalytics which will implement a caching strategy as described below.

# Subtasks

- [x] Create a new Hono app under a new GitHub repo for "ENSAnalytics". We can later explore adding this new Hono app to the ENSNode monorepo, but this is not suggested for initial steps.
- [x] Implement a REST API into ENSAnalytics to query the "top referrers".
   - [x] Implement an in-memory caching strategy into the ENSAnalytics app such that when it starts up it stores the "top referrers" in memory and then asynchronously works to refresh the "top referrers" every 5 minutes.
   - [x] The API to get the "top referrers" must only ever return the "top referrers" that ENSAnalytics has already stored into memory. No API request to get the "top referrers" may ever be allowed to trigger a database query. If ENSAnalytics receives an API request to get the "top referrers" and this data hasn't been successfully stored in memory yet, ENSAnalytics should return a HTTP 500 error.
   - [x] ENSAnalytics need an environment variable to configure both the ENSDb URL (Postgres connection string) and the ENSDb schema name (Postgres schema name) to read data from.
[...]

PR #1239, opened 2025-11-01
feat: introduce ENSAnalytics API into ENSApi

Goader:
advances: https://github.com/namehash/ensnode/issues/1185

This PR introduces ENSAnalytics API as part of ENSApi. ENSAnalytics implements a **Stale-While-Revalidate (SWR)** caching strategy via middleware to provide sub-millisecond response times while keeping data fresh with automatic background updates every 5 minutes. All required ENV variables are already available in ENSApi, so nothing was added there.

**Architecture:**
- **Middleware-based caching**: `aggregatedReferrerSnapshotCacheMiddleware` uses SWR pattern to serve cached data immediately (even if stale) while asynchronously revalidating in the background
- **Cache warmup**: On server startup, the cache is pre-populated to ensure immediate availability
- **Date filtering**: Currently hardcoded to ENS Holiday Awards period using `ENS_HOLIDAY_AWARDS_START_DATE` and `ENS_HOLIDAY_AWARDS_END_DATE` from `@namehash/ens-referrals`

**SDK additions:**
- TypeScript types and interfaces (`AggregatedReferrerMetrics`, `AggregatedReferrerMetricsContribution`, etc.)
- Zod schemas for runtime validation
- Serialization/deserialization functions
- Constants: `ITEMS_PER_PAGE_DEFAULT` (25), `ITEMS_PER_PAGE_MAX` (100)

**Endpoint implemented:**

* `/ensanalytics/aggregated-referrers`: Main endpoint for paginated aggregated referrer leaderboard
    * **Input parameters:**
        - `page` (optional): Page number, must be >= 1 (default: 1)
        - `itemsPerPage` (optional): Items per page, must be 1-100 (default: 25)
    * **Validation:**
        - `page >= 1` (error: 400)
        - `1 <= itemsPerPage <= 100` (error: 400)
        - `page` must not exceed total pages (error: 400)
    * **Error handling:**
        - Returns 500 if cache failed to load
    * **Response:**
    ```typescript
    {
      responseCode: "ok",
      data: {
        referrers: [
          {
            referrer: Address;
            totalReferrals: number;
            totalIncrementalDuration: number; // in seconds
            totalReferralsContribution: number; // 0-1 (percentage of grand total)
            totalIncrementalDurationContribution: number; // 0-1 (percentage of grand total)
          },
          ...
        ];
        total: number; // total number of referrers across all pages
        paginationParams: {
          page: number;
          itemsPerPage: number;
        };
        hasNext: boolean;
        hasPrev: boolean;
        updatedAt: UnixTimestamp; // when cache was last updated
      }
    }
    ```


apps/ensapi/src/lib/ensanalytics/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface ReferrerData {
+  referrer: string;
+  total_referrals: number;
+}
+
+export interface PaginatedReferrers {
```
lightwalker-eth: It will be great if you could create JSDocs for each of the fields in this object. Thanks

apps/ensapi/src/lib/ensanalytics/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  page: number;
+  limit: number;
+  hasNext: boolean;
+  hasPrev: boolean;
+}
+
+export interface CacheStats {
+  totalReferrers: number;
+  isInitialized: boolean;
+  lastRefresh: Date;
```
lightwalker-eth: A few suggestions:

1. Rename to `updatedAt`.
2. Change the type of this field from `Date` to `UnixTimestamp` as can be imported from `ensnode-sdk`. Here's a relevant code example for how to work with this type: https://github.com/namehash/ensnode/blob/main/apps/ensapi/src/middleware/can-accelerate.middleware.ts#L117

apps/ensapi/src/index.ts
```diff
@@ -45,9 +47,29 @@ app.route("/api", ensNodeApi);
...
+    ensanalytics: {
+      ok: ensanalyticsOk,
+      cache: cache
+        ? {
+            totalReferrers: cache.referrers.length,
+            updatedAt: cache.updatedAt,
+          }
+        : null,
+    },
+  });
```
Goader: @lightwalker-eth Should we just leave it as it was, with a single "ok" field, but affected by the cache status? Or is adding this separate field for the ensanalytics health ok?

apps/ensapi/src/middleware/referrers-cache.middleware.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { TtlCache } from "@ensnode/ensnode-sdk";
+
+import { getTopReferrers } from "@/lib/ensanalytics/database";
+import type { ReferrerCache } from "@/lib/ensanalytics/types";
+import { factory } from "@/lib/hono-factory";
+
+const TTL_MS = 5 * 60 * 1000; // 5 minutes
+
+// memoizes the getTopReferrers promise across TTL_MS
```
lightwalker-eth: Could you please refine and expand on exactly how the TTL works on this cache and exactly how "across" should be interpreted in this sentence?

I'm especially seeking full confidence in how this will work in the following situations:
1. The cache has expired and an update to the cache hasn't been requested yet.
2. The cache has expired and an update to the cache is currently being fetched / built, but isn't fetched / built yet.
3. The cache has expired and the last attempt to update the cache failed (for any reason).
   1. What is the retry logic?

apps/ensapi/src/lib/ensanalytics/types.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import type { Duration, UnixTimestamp } from "@ensnode/ensnode-sdk";
+
+/**
+ * Represents data for a single referrer and their referral metrics
+ */
+export interface ReferrerData {
+  /** The Ethereum address of the referrer */
+  referrer: Address;
+  /** The total number of qualified referrals made by this referrer */
```
lightwalker-eth: Please include in JSDocs all special invariants of each field.

For example: Guaranteed to be a positive integer.

Same can apply to the `totalIncrementalDuration` field.

apps/ensapi/src/lib/ensanalytics/database.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import { DatasourceNames, maybeGetDatasource } from "@ensnode/datasources";
+import * as schema from "@ensnode/ensnode-schema";
+import { type AccountId, deserializeDuration, serializeAccountId } from "@ensnode/ensnode-sdk";
+
+import { db } from "@/lib/db";
+import type { ReferrerData } from "@/lib/ensanalytics/types";
+import logger from "@/lib/logger";
+
+const START_DATE = getUnixTime(new Date("2025-01-01T00:00:00.000Z"));
```
lightwalker-eth: Suggest moving the start and end date constants out of this file and into https://github.com/namehash/ensnode/tree/main/packages/ens-referrals

We can export them from this package. That will be a nice touch.

apps/ensapi/src/lib/ensanalytics/database.ts
```diff
@@ -0,0 +1,101 @@
...
+const END_DATE = getUnixTime(new Date("2026-01-01T00:00:00.000Z"));
+
+/**
+ * Gets the BaseRegistrar contract AccountId for the configured namespace.
+ *
+ * @returns The AccountId for the BaseRegistrar contract
+ * @throws Error if the contract is not found
+ */
+function getBaseRegistrarSubregistryId(): AccountId {
```
lightwalker-eth:
```suggestion
/**
 * Gets the SubregistryId (an AccountId) of the Ethnames Subregistry contract (this is the "BaseRegistrar" contract for direct subnames of .eth) for the provided namespace.
 *
 * @returns The AccountId for the Ethnames Subregistry contract for the provided namespace.
 * @throws Error if the contract is not found for the given namespace.
 */
function getEthnamesSubregistryId(): AccountId {
```

Background info: There's multiple contracts within an ENS namespace that call themselves "BaseRegistrar", but only one is for direct subnames of .eth. Here we're returning the "BaseRegistrar" that is for subnames of .eth. Additionally, direct subnames of ".eth" have only 1 "BaseRegistrar". Therefore we can simply call this: `getEthnamesSubregistryId`.

Additionally we should change this function so that it takes an ENSNamespaceId param and doesn't make any use of `config` directly. The caller of this function can pass in `config.namespace` as a param as desired. This is helpful for several reasons, including:
1. Moving this function outside of the ENSApi app and into the `ensnode-sdk` package (please do this 👍)
2. Making this function easier to write unit tests for.

lightwalker-eth (commented):
@Goader Hey great job with the enhancements here 💪 This is definitely moving in the right direction and getting close. Reviewed and shared feedback 👍

apps/ensapi/src/lib/ensanalytics/database.ts
lightwalker-eth: @Goader It will be a nice touch to put these definitions in the `ens-referrals` package and import them as needed elsewhere in our work.

Later, in separate goals, we'll be further refining the work in the `ens-referrals` package and it will be useful to have these constants defined there.

apps/ensapi/src/lib/ensanalytics/types.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Represents a snapshot of top referrers.
+ */
+export interface TopReferrersSnapshot {
+  /**
+   * Array of top referrers sorted by total incremental duration (descending)
+   * @invariant Array may be empty if there are no qualified referrers.
+   */
+  topReferrers: AggregatedReferrerMetrics[];
+  /** Unix timestamp of when the top referrers were last updated */
```
lightwalker-eth:
```suggestion
  /** Unix timestamp identifying when this `TopReferrersSnapshot` was generated */
```

apps/ensapi/src/lib/ensanalytics/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  updatedAt: UnixTimestamp;
+  /**
+   * The sum of totalReferrals across all referrers
```
lightwalker-eth:
```suggestion
   * @invariant The sum of totalReferrals across all referrers
```

I see this as a nice invariant too :)

apps/ensapi/src/lib/ensanalytics/types.ts
```diff
@@ -0,0 +1,24 @@
...
+   * @invariant Guaranteed to be a non-negative integer (>= 0)
+   */
+  grandTotalReferrals: number;
+  /**
+   * The sum of totalIncrementalDuration across all referrers
```
lightwalker-eth:
```suggestion
   * @invariant The sum of totalIncrementalDuration across all referrers
```

packages/ensnode-sdk/src/ensanalytics/types.ts
```diff
@@ -0,0 +1,144 @@
...
+   */
+  totalIncrementalDuration: Duration;
+}
+
+/**
+ * Request parameters for paginated top referrers query.
+ */
+export interface PaginatedTopReferrersRequest {
+  /**
+   * Current page number (1-indexed)
```
lightwalker-eth:
```suggestion
   * Requested page number (1-indexed)
```

apps/ensapi/src/middleware/top-referrers-cache.middleware.ts
```diff
@@ -0,0 +1,38 @@
...
+import { getTopReferrers } from "@/lib/ensanalytics/database";
+import { factory } from "@/lib/hono-factory";
+
+const TTL_MS = 5 * 60 * 1000; // 5 minutes
+
+// memoizes the getTopReferrers promise across TTL_MS
+export const fetcher = pMemoize(
+  async () => {
+    const subregistryId = getEthnamesSubregistryId(config.namespace);
+    return getTopReferrers(ENSAWARDS_START_DATE, ENSAWARDS_END_DATE, subregistryId);
```
lightwalker-eth: It seems nice for us to emit some little message into logs each time:
1. We begin the process of building a new snapshot.
2. We successfully complete the process of building a new snapshot.
3. We fail (for any reason) to complete the process of building a new snapshot.

lightwalker-eth (commented):
@Goader Hey good updates 👍 Reviewed and shared feedback. Appreciate your help with the next steps!

packages/ensnode-sdk/src/ensanalytics/types.ts
```diff
@@ -0,0 +1,144 @@
...
+   * @note The response may contain an empty array for the first page if there are no qualified referrers.
+   * When the array is empty, total will be 0, page will be 1, and both hasNext and hasPrev will be false.
+   */
+  Ok: "ok",
+
+  /**
+   * Represents that the request failed due to pagination parameters being out of range.
+   * @note This error is only returned when there is data available but the requested page exceeds the total pages.
+   */
+  PageOutOfRange: "page_out_of_range",
```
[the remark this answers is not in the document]
Goader: @lightwalker-eth I consider it to be a 400 HTTP error, so it felt like mixing 400 and 500 errors to have the same response code could be a little misleading. 

I actually thought that it would be probably best to remove this `PageOutOfRange`, and instead handle that case the way Zod would handle an invalid request. It is defined [here](https://github.com/namehash/ensnode/blob/703a12ed307e98ec1c1ce5911b45905ebcad58d8/apps/ensapi/src/lib/handlers/error-response.ts#L17), and using this would make it consistent with the response we return in a situation when, for example, page is negative, or limit is 1000.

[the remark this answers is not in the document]
lightwalker-eth: @Goader Yes I agree it's a 400 error.

And sure happy to follow your suggestion 👍

packages/ensnode-sdk/src/ensanalytics/types.ts
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Total number of top referrers across all pages
+   * @invariant Guaranteed to be a non-negative integer (>= 0)
+   */
+  total: number;
+  /**
+   * Current page number (1-indexed)
+   * @invariant Guaranteed to be a positive integer (>= 1)
+   */
+  page: number;
```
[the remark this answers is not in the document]
lightwalker-eth: @Goader I like your ideas :)

apps/ensapi/src/lib/ensanalytics/database.ts
```diff
@@ -0,0 +1,112 @@
...
+  type UnixTimestamp,
+} from "@ensnode/ensnode-sdk";
+
+import { db } from "@/lib/db";
+import type { AggregatedReferrerSnapshot } from "@/lib/ensanalytics/types";
+import { ireduce } from "@/lib/itertools";
+import logger from "@/lib/logger";
+
+/**
+ * Fetches all referrers with 1 or more qualified referrals from the `registrar_actions` table and builds an `AggregatedReferrerSnapshot`.
```
lightwalker-eth:
```suggestion
 * Fetches all referrers with 1 or more qualified referrals from the `registrar_actions` table
 * and builds an `AggregatedReferrerSnapshot`.
```

Goal: optimize max line widths

apps/ensapi/src/lib/ensanalytics/database.ts
```diff
@@ -0,0 +1,112 @@
...
+        totalIncrementalDuration: sum(schema.registrarActions.incrementalDuration).as(
+          "total_incremental_duration",
+        ),
+      })
+      .from(schema.registrarActions)
+      .where(
+        and(
+          // Filter by timestamp range
+          gte(schema.registrarActions.timestamp, BigInt(startDate)),
+          lt(schema.registrarActions.timestamp, BigInt(endDate)),
```
lightwalker-eth: I assume this is < and not <=?

Should we make this <=? I believe so.

Likewise, we should refine the docs for the `startDate` and `endDate` param to communicate how they are both inclusive.

packages/ensnode-sdk/src/ensanalytics/types.ts
```diff
@@ -0,0 +1,184 @@
+import type { Address } from "viem";
+
+import type { Duration, UnixTimestamp } from "../shared";
+
+/**
+ * The default number of items per page for paginated aggregated referrer queries.
+ */
+export const PAGINATION_DEFAULT_LIMIT = 25;
```
lightwalker-eth:
```suggestion
export const ITEMS_PER_PAGE_DEFAULT = 25;
```

Goal: Align with the renaming of `PaginatedAggregatedReferrersRequest.itemsPerPage`.

apps/ensapi/src/middleware/aggregated-referrer-snapshot-cache.middleware.ts
```diff
@@ -4,41 +4,35 @@ import {
...
+  promise
+    .then(() => {
+      logger.info("Successfully built aggregated referrer snapshot");
+    })
+    .catch((error) => {
+      logger.error({ error }, "Failed to build aggregated referrer snapshot");
+    });
+
+  return promise;
+}, TTL_MS);
```
lightwalker-eth: This comment is not on the correct line. Goal was to put it on `aggregatedReferrerSnapshotCacheMiddleware` several lines below.

It will be nice to update the JSDocs for that function to update the description of how the cache strategy here works 👍

packages/ensnode-sdk/src/shared/cache.ts
```diff
@@ -195,3 +195,95 @@ export class TtlCache<KeyType extends string, ValueType> implements Cache<KeyTyp
     return this._cache.delete(key);
   }
 }
+
+interface SWRCache<T> {
```
lightwalker-eth: It will be nice if you could add some JSDocs on each of the fields in this interface 👍

packages/ensnode-sdk/src/shared/cache.ts
```diff
@@ -195,3 +195,95 @@ export class TtlCache<KeyType extends string, ValueType> implements Cache<KeyTyp
...
+  value: T;
+  timestamp: number;
```
lightwalker-eth: Would be nice to use one of our `UnixTimestamp` type aliases as that implicitly adds more invariants above a raw `number` which can be values like NaN, -Infinity, etc..

lightwalker-eth:
```suggestion
  updatedAt: number;
```

lightwalker-eth (commented):
@Goader Hey I know you're still working on some updates to this PR but here's some quick feedback on the latest updates 👍

packages/ensnode-sdk/src/shared/cache.ts
```diff
@@ -195,3 +195,95 @@ export class TtlCache<KeyType extends string, ValueType> implements Cache<KeyTyp
...
+  if (!Number.isInteger(ttl) || ttl <= 0) {
+    throw new Error(
+      `staleWhileRevalidate requires ttl to be a positive integer but a ttl of ${ttl} was provided.`,
+    );
+  }
+
+  let cache: SWRCache<T> | null = null;
+
+  return async (): Promise<T> => {
+    const now = Date.now();
```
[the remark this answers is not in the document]
lightwalker-eth: Continue to suggest this :+1:

packages/ensnode-sdk/src/shared/cache.ts
```diff
@@ -232,31 +255,51 @@ interface SWRCache<T> {
...
-      return value;
+      // If initial build already in progress, wait for it
+      if (initialBuild) {
+        return initialBuild;
+      }
+
+      // Start initial build
+      initialBuild = fn()
+        .then((value) => {
+          cache = { value, updatedAt: now };
```
lightwalker-eth: Suggest calculating a fresh value for now here.

apps/ensapi/src/middleware/aggregated-referrer-snapshot-cache.middleware.ts
```diff
@@ -4,41 +4,35 @@ import {
...
-import pMemoize from "p-memoize";
 
-import { getEthnamesSubregistryId, TtlCache } from "@ensnode/ensnode-sdk";
+import { getEthnamesSubregistryId, staleWhileRevalidate } from "@ensnode/ensnode-sdk";
 
 import { getAggregatedReferrerSnapshot } from "@/lib/ensanalytics/database";
 import { factory } from "@/lib/hono-factory";
 import logger from "@/lib/logger";
 
 const TTL_MS = 5 * 60 * 1000; // 5 minutes
```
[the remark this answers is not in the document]
lightwalker-eth: Continue to think this would be a nice step 👍

apps/ensapi/src/index.ts
```diff
@@ -78,7 +78,11 @@ const server = serve(
     logger.info("Warming up ENSAnalytics aggregated referrer snapshot cache...");
     try {
       const cache = await referrersCacheFetcher();
-      logger.info(`ENSAnalytics cache warmed up with ${cache.referrers.size} referrers`);
+      if (cache) {
+        logger.info(`ENSAnalytics cache warmed up with ${cache.referrers.size} referrers`);
+      } else {
+        logger.warn("ENSAnalytics cache returned null - no cached data available yet");
```
lightwalker-eth: Could you have a look at `referrersCacheFetcher` to see if it is still possible that it might throw an error?

If it can't throw an error anymore, we can:
- remove the try / catch logic here
- Update this log statement so it's an error not a warning

However, if my assumption here is wrong then please feel welcome to ignore. Appreciate your advice 👍

packages/ensnode-sdk/src/shared/cache.ts
lightwalker-eth: Appreciate your questions 😄 

Can confirm it's good to switch these to use `UnixTimestamp` and `Duration`. As you noted, there's tradeoffs either way. Optimizing our type usage to be more standardized across our code is worth the tradeoffs. If MS resolution becomes a goal in the future we can create new special MS-resolution types but for now no need 👍

apps/ensapi/src/middleware/aggregated-referrer-snapshot-cache.middleware.ts
```diff
@@ -5,45 +5,53 @@ import {
...
+    const result = await getAggregatedReferrerSnapshot(
+      ENS_HOLIDAY_AWARDS_START_DATE,
+      ENS_HOLIDAY_AWARDS_END_DATE,
+      subregistryId,
+    );
+    logger.info("Successfully built aggregated referrer snapshot");
+    return result;
+  } catch (error) {
+    logger.error({ error }, "Failed to build aggregated referrer snapshot");
+    throw error;
```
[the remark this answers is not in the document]
Goader: Sorry, missed it. The responsibility for handling the error is inside the SWR logic, so this function can throw errors without disrupting the flow. And also this way SWR will know that revalidation failed and will return stale data, instead of treating null as the updated value.

PR #1244, opened 2025-11-03
feat: add additional subgraph compatibility check for thegraph fallback

shrugs:


lightwalker-eth (approved):
@shrugs Looks nice, thanks 👍

Issue #1228, opened 2025-10-31
Create a new plugin to track registrations and renewals


tk-o:
This plugin goal is to isolate event handlers which will index data using updated data model, which includes tracking ENS referrals data.

The event handlers must index the following registrar events:
- `NameRegistered`
- `NameRenewed`

across subregistries managing:

- Ethnames 
- Basenames
- Lineanames

PR #1250, opened 2025-11-05
feat(ensnode-sdk): currencies module

tk-o:
Move and extend the `currencies` module from ENSIndexer into ENSNode SDK in order to support goals related to the new `registrars` plugin.

Related to #1228.

### DIFF
--- a/packages/ensnode-sdk/src/shared/currencies.ts
+++ b/packages/ensnode-sdk/src/shared/currencies.ts
@@ -117,7 +117,7 @@ export function isPriceCurrencyEqual(priceA: Price, priceB: Price): boolean {
 }
 
 /**
- * Check if two {@link Price} values are equal.
+ * Check if two {@link Price} values have the same currency and amount.
  */
 export function isPriceEqual(priceA: Price, priceB: Price): boolean {
   return isPriceCurrencyEqual(priceA, priceB) && priceA.amount === priceB.amount;


packages/ensnode-sdk/src/shared/currencies.ts
```diff
@@ -0,0 +1,153 @@
...
+  amount: CurrencyAmount;
+}
+
+export type Price = PriceEth | PriceDai | PriceUsdc;
+
+/**
+ * Serialized representation of {@link Price}.
+ */
+export interface SerializedPrice extends Omit<Price, "amount"> {
+  currency: CurrencyId;
```
lightwalker-eth: Can you double check this noting how this interface is defined to extend Price?

packages/ensnode-sdk/src/shared/currencies.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Check if two prices have the same currency.
+ */
+export function isPriceCurrencyEqual(priceA: Price, priceB: Price): boolean {
+  return priceA.currency === priceB.currency;
+}
+
+/**
+ * Check if two {@link Price} values are equal.
```
lightwalker-eth:
```suggestion
 * Check if two {@link Price} values have the same currency and amount.
```

PR #1277, opened 2025-11-18
fix(ci): update dependency version overrides

tk-o:
This is a temporary fix, the permanent one will be to update  after this PR:
- https://github.com/changesets/changesets/pull/1772

is released.


Y3drk (approved):
Didn't check for backwards compatibility, so I'll trust you on that.

Issue #1264, opened 2025-11-07
ENSAwards date range environment variables


Goader:
# Background:

* ENS Holiday Awards needs a "Referrer Detail" page, that will display the referrals statistics for a given address.
* Follows #1239 
* ENS Holiday Awards launch is Dec 1st, and we need to be able to test it properly before that date.

# Subtasks:

* [ ] Introduce two new environment variables to ENSApi (`ENSAWARDS_START` and `ENSAWARDS_END`), defining the date range for the ENSAnalytics
* [ ] If no environment variables are defined, then the default value for `ENSAWARDS_START` is "Dec 1, 2025 00:00:00 UTC", and for `ENSAWARDS_END` - "Dec 31, 23:59:59 UTC".

PR #1279, opened 2025-11-18
ENS Holiday Awards Date Range Environment Variables

Goader:
closes: https://github.com/namehash/ensnode/issues/1264

This PR introduces configurable date range environment variables for ENSAnalytics ENS Holiday Awards campaign, allowing the date range to be customized for testing purposes before the Dec 1st launch.

**Changes:**
- **Environment variables**: Added `ENS_HOLIDAY_AWARDS_START` and `ENS_HOLIDAY_AWARDS_END` to ENSApi
- **Default values**: If not set, defaults to hardcoded values from `@namehash/ens-referrals` package ("Dec 1, 2025 00:00:00 UTC" and "Dec 31, 2025 23:59:59 UTC")
- **Validation**: Added `invariant_ensHolidayAwardsEndAfterStart` check to ensure end date is after or equal to start date
- **Date parsing**: Implemented `DateStringToUnixTimestampSchema` to convert date strings to Unix timestamps with validation
- **Cache integration**: Updated `aggregatedReferrerSnapshotCacheMiddleware` to use configurable dates instead of hardcoded constants

**SDK additions:**
- `EnsHolidayAwardsEnvironment` interface for environment type definitions

### DIFF
--- a/packages/ensnode-sdk/src/shared/cache.test.ts
+++ b/packages/ensnode-sdk/src/shared/cache.test.ts
@@ -171,17 +171,17 @@ describe("TtlCache", () => {
   });
 
   it("automatically cleans up expired entries on size access", () => {
-    const ttl = new TtlCache<string, string>(2); // 2 seconds
+    const ttl = new TtlCache<string, string>(20); // 20 seconds
 
     ttl.set("key1", "value1");
     ttl.set("key2", "value2");
     expect(ttl.size).toBe(2);
 
-    vi.advanceTimersByTime(1000); // Advance by 1000ms (1 second)
+    vi.advanceTimersByTime(10000); // Advance by 10000ms (10 seconds)
     ttl.set("key3", "value3");
     expect(ttl.size).toBe(3);
 
-    vi.advanceTimersByTime(1100); // Advance by 1100ms (1.1 seconds) - total 2.1 seconds
+    vi.advanceTimersByTime(11000); // Advance by 11000ms (11 seconds) - total 21 seconds
 
     expect(ttl.size).toBe(1);
     expect(ttl.get("key1")).toBeUndefined();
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -36,6 +36,9 @@ services:
       # Override environment variables to point to docker instances
       DATABASE_URL: postgresql://postgres:password@postgres:5432/postgres
       ENSINDEXER_URL: http://ensindexer:42069
+      # Optional: Define ENSAnalytics date range for ENS Holiday Awards
```
lightwalker-eth: I think we can revert the updates to this file as these can be configured through the .env.local file

apps/ensapi/src/config/config.schema.ts
```diff
@@ -42,6 +51,26 @@ export const DatabaseUrlSchema = z.string().refine(
...
+      }
+    },
+    {
+      error: "Invalid date string. Expected a valid date format (e.g., 'Dec 1, 2025 00:00:00 UTC')",
+    },
+  )
+  .transform((dateStr) => {
+    const timestamp = getUnixTime(new Date(dateStr));
+    return makeUnixTimestampSchema().parse(timestamp);
+  });
```
tk-o: Feel free to use ISO 8601 format for defining datetime values in configuration. You can use an existing schema to parse such format:

```suggestion
// you can import `makeDatetimeSchema` from `@ensnode/ensnode-sdk/internal`

const DateStringToUnixTimestampSchema =  z.coerce
    .string()
    .pipe(makeDatetimeSchema())
    .transform((date) => getUnixTime(date));
```

tk-o (changes requested):
Nice updates, thank you 🙃 I think we are pretty close here to be ready to ship. One change request is about the particular timestamp values we use for start date and end date of ENS Holiday Awards.

Issue #1200, opened 2025-10-21
Enhance handling of a failure to reload a more recent indexing status snapshot


lightwalker-eth:
# Issue Report

Multiple times today while testing ENSAdmin I encountered the following issue:

1. I open the "Latest Indexed Registrations" and they all load successfully. I watch it auto-update every few seconds.
2. Then I see this error appear:

<img width="2880" height="1557" alt="Image" src="https://github.com/user-attachments/assets/1b53ba72-333e-455e-bf3b-a182aa363e6c" />

3. Then, a few seconds later, I see the "Latest Indexed Registrations" automatically appear again.

# Expected Behavior

We recently made big updates to the indexing status API and its associated data model. Please deeply study this well-documented data model here: https://github.com/namehash/ensnode/blob/main/packages/ensnode-sdk/src/ensindexer/indexing-status/types.ts

And deeply study the key function named `createRealtimeIndexingStatusProjection` that's defined in https://github.com/namehash/ensnode/blob/main/packages/ensnode-sdk/src/ensindexer/indexing-status/projection.ts

In particular, please deeply study the following:
* `RealtimeIndexingStatusProjection`
* `CrossChainIndexingStatusSnapshotOmnichain`
* ... and how these relate via `createRealtimeIndexingStatusProjection`.

The big idea here is that once an ENSNode client (such as ENSAdmin) manages to successfully fetch the indexing status once, the client can store that indexing status snapshot in memory (cached) as a `CrossChainIndexingStatusSnapshotOmnichain` and then continuously build new `RealtimeIndexingStatusProjection` on demand as much as the client wants using `createRealtimeIndexingStatusProjection`.

Therefore, if ENSAdmin successfully fetches the realtime indexing status once, it should be storing the associated `CrossChainIndexingStatusSnapshotOmnichain` in memory and then asynchronously in the background working to fetch an updated indexing status snapshot from the connected ENSNode. When this asynchronous background fetching is successful at fetching an updated indexing status snapshot it should update the snapshot it has cached. But if the asynchronous background fetching fails for any reason, whatever indexing status snapshot it might have successfully cached previously should be indefinitely retained in its cache until a subsequent indexing status snapshot is successfully fetched.

Please ask if any questions. This is quite an important and priority enhancement to ENSAdmin.

PR #1282, opened 2025-11-19
feat(ensadmin): introduce `useIndexingStatusWithSwr ` hook

tk-o:
Live preview: https://adminensnode-p5chjdg6o-namehash.vercel.app/registrar-actions?connection=https%3A%2F%2Fapi.alpha-sepolia.green.ensnode.io%2F

This PR replaces `useIndexingStatus` hook in ENSAdmin with a proxy hook called `useIndexingStatusWithSwr`. The API of both hooks is exactly the same, but behaviour differs. 

| `useIndexingStatus` | `useIndexingStatusWithSwr` |
|--------|--------|
| Resolves to either `IndexingStatusResponseOk` or `IndexingStatusResponseError`. | Resolves to `IndexingStatusResponseOk`. |
| Resolves to an error when a subsequent data refetch failed | Almost always resolves to a success, event if subsequent data refetch failed. The only case when it resolves to an error is when there was no previously successful fetch. |
| Makes ENSAdmin to present the error. | Makes ENSAdmin to present the latest successfully resolved, even when the current refetch failed. |

The `useIndexingStatusWithSwr` hook is based on the new `useQuerySwr` proxy hook which allows to create a React Query that will always serve previously cached result, and allow refetching the current result in the background.

This hook will suport goals of #1200.

### DIFF
--- a/apps/ensadmin/src/app/mock/indexing-stats/page.tsx
+++ b/apps/ensadmin/src/app/mock/indexing-stats/page.tsx
@@ -20,7 +20,7 @@ import {
 
 type LoadingVariant = "Loading" | "Loading Error";
 type ResponseOkVariant = keyof typeof indexingStatusResponseOkOmnichain;
-type ResponseErrorVariant = "Response Error";
+type ResponseErrorVariant = "Error ResponseCode";
 type Variant = ResponseOkVariant | ResponseErrorVariant | LoadingVariant;
 
 const variants = [
@@ -30,7 +30,7 @@ const variants = [
   OmnichainIndexingStatusIds.Completed,
   "Loading",
   "Loading Error",
-  "Response Error",
+  "Error ResponseCode",
 ] as const;
 
 let loadingTimeoutId: number;
@@ -50,7 +50,7 @@ async function fetchMockedIndexingStatus(
     case OmnichainIndexingStatusIds.Following:
     case OmnichainIndexingStatusIds.Completed:
       return indexingStatusResponseOkOmnichain[selectedVariant] as IndexingStatusResponseOk;
-    case "Response Error":
+    case "Error ResponseCode":
       throw new Error(
         "Received Indexing Status response with responseCode other than 'ok' which will not be cached.",
       );
--- a/apps/ensadmin/src/components/indexing-status/indexing-stats.tsx
+++ b/apps/ensadmin/src/components/indexing-status/indexing-stats.tsx
@@ -439,7 +439,11 @@ export function IndexingStats(props: IndexingStatsProps) {
   const indexingStatusQuery = props;
 
   if (indexingStatusQuery.isError) {
-    return <p>Failed to fetch Indexing Status.</p>;
+    return (
+      <IndexingStatsShell>
+        <IndexingStatsForUnavailableSnapshot />
+      </IndexingStatsShell>
+    );
   }
 
   if (indexingStatusQuery.isPending) {


shrugs:
or perhaps something like this?

```
const last = useRef<DATA | null>();

const { data, status, ...query } = useQuery({ ... })

if (status === 'success') last.value = data;

return { ...query, status, data: last.value }
```

apps/ensadmin/src/components/indexing-status/use-indexing-status-with-swr.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const DEFAULT_REFETCH_INTERVAL = 3 * 1000;
+
+interface UseIndexingStatusParameters
+  extends IndexingStatusRequest,
+    QueryParameter<IndexingStatusResponseOk> {}
+
+/**
+ * A proxy hook for {@link useIndexingStatus} which applies
+ * stale-while-revalidate cache for successfully resolved responses.
```
lightwalker-eth:
```suggestion
 * stale-while-revalidate cache for successful responses.
```

packages/ensnode-react/src/hooks/useSwrQuery.ts
```diff
@@ -0,0 +1,108 @@
...
+  const cachedResult = derivedQueryClient.getQueryData<TData>(options.queryKey);
+
+  const queryResult = useQuery(
+    {
+      ...options,
+      // cached result can never be stale
+      staleTime: cachedResult ? Infinity : undefined,
+      // cached result can never be removed by garbage collector
+      gcTime: cachedResult ? Infinity : undefined,
+      placeholderData: keepPreviousData, // Keep showing previous data during refetch (does not work for errors)
```
lightwalker-eth: Can you please expand upon and clarify in more details what it means when you say: "does not work for errors"?

lightwalker-eth (approved):
@tk-o Hey great to see this enhancement! Shared a few comments with suggestions. Please feel welcome to merge this when ready 👍

apps/ensadmin/src/components/indexing-status/indexing-stats.tsx
```diff
@@ -429,32 +430,24 @@ export function IndexingStatsForRealtimeStatusProjection({
...
 /**
- * Display indexing stats based on query results from {@link useIndexingStatus}.
+ * Display indexing stats based on query results from {@link useIndexingStatusWithSwr}.
  */
 export function IndexingStats(props: IndexingStatsProps) {
   const indexingStatusQuery = props;
 
-  if (indexingStatusQuery.status === "error") {
+  if (indexingStatusQuery.isError) {
     return <p>Failed to fetch Indexing Status.</p>;
```
lightwalker-eth: It seems we should move the the following logic here that this PR is removing a few lines down in this file?
```
  if (indexingStatus.responseCode === IndexingStatusResponseCodes.Error) {
    return (
      <IndexingStatsShell>
        <IndexingStatsForUnavailableSnapshot />
      </IndexingStatsShell>
    );
  }
```

Of course, I'm not referring to the logic for `IndexingStatusResponseCodes.Error`. I'm talking about the HTML that's generated for the case of an error.

lightwalker-eth (approved):
@tk-o Great updates. Thank you! Just 2 small questions. Please take the lead to merge when ready 🚀 😄

apps/ensadmin/src/app/mock/indexing-stats/page.tsx
```diff
@@ -45,11 +49,13 @@ async function fetchMockedIndexingStatus(
     case OmnichainIndexingStatusIds.Backfill:
     case OmnichainIndexingStatusIds.Following:
     case OmnichainIndexingStatusIds.Completed:
-      return indexingStatusResponseOkOmnichain[selectedVariant];
+      return indexingStatusResponseOkOmnichain[selectedVariant] as IndexingStatusResponseOk;
     case "Response Error":
```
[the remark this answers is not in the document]
tk-o: Nice suggestion, thank you. It's exactly as you described 👍

PR #1301, opened 2025-11-24
fix(ensapi): use native promises with swr cache

tk-o:
Please review this PR with _Hide whitespaces_ option on.

Suggested review order:
1. ENSNode SDK (changed the `staleWhileRevalidate` function input type).
  - `packages/ensnode-sdk/src/shared/cache.ts`
2. ENSApi
  - `apps/ensapi/src/middleware/indexing-status.middleware.ts` documented ideas and guarantees around `c.var.indexingStatus` value, and how `null` being the cached value represents the state when indexing status has never been fetched successfully. Also, when the cached indexing status is available, the `c.var.indexingStatus` value represents the realtime indexing status projection that was created at the time of the request.
  - `apps/ensapi/src/middleware/aggregated-referrer-snapshot-cache.middleware.ts` updated the `fn` fetcher function to log a message for each time the aggregated referrer snapshot is built successfully.
  - `apps/ensapi/src/middleware/is-realtime.middleware.ts` dropped the obsolete logic as `c.var.indexingStatus` can never represent a response with `responseCode: IndexingStatusResponseCodes.Error`.
  - `apps/ensapi/src/lib/subgraph/indexing-status-to-subgraph-meta.ts` simplified the code according to the new guarantees of `c.var.indexingStatus` value, the `c.var.indexingStatus` can never represent a response with `responseCode: IndexingStatusResponseCodes.Error`.
  - `apps/ensapi/src/lib/middleware/require-registrar-actions-plugins..middleware.ts` simplified the code according to the new guarantees of `c.var.indexingStatus` value, the `c.var.indexingStatus` can never represent a response with `responseCode: IndexingStatusResponseCodes.Error`.
  - `apps/ensapi/src/handlers/ensnode-api.ts` updated code docs, added extra log message.

### Context

This PR fixes an issue where the SWR cache query function would return `pReflect` result type, and not native promise. This caused the rejected promise to be seen as a correctly resolved one on the SWR cache level:
image

The SWR cache requires the query `fn` to return a native Promise object, so we cannot wrap the returned value with `pReflect`. On the other hand, in order to maintain the downstream logic which relies on `pReflect` result type, we still use `pReflect`, but this time on the middleware level, and not on the SWR cache query `fn` level.

This PR updates the `c.var.IndexingStatus` object type from `IndexingStatusResponse` to `RealtimeIndexingStatusProjection`. It means that all downstream request handlers won't need to check the `responseCode` of the cached response. Only the OK responses can get cached. The Error ones are discarded (with an appropriate error log).

### DIFF
--- a/.changeset/shaggy-dodos-write.md
+++ b/.changeset/shaggy-dodos-write.md
@@ -2,4 +2,4 @@
 "ensapi": minor
 ---
 
-Indexing Status cache can store only responses with `responseCode: IndexingStatusResponseCodes.Ok`.
+Indexing Status cache only stores responses with `responseCode: IndexingStatusResponseCodes.Ok`.


packages/ensnode-sdk/src/shared/cache.ts
```diff
@@ -257,15 +279,17 @@ interface SWRCache<T> {
  *
  * @param fn The async function to wrap with SWR caching
  * @param ttl Time-to-live duration in seconds. After this duration, data is considered stale
+ * @param onResolved Optional callback invoked when the wrapped function resolves successfully
+ * @param onRejected Optional callback invoked when the wrapped function throws an error
```
lightwalker-eth:
```suggestion
 * @param onError Optional callback invoked when `fn` throws an error
```

Is that fair?

apps/ensapi/src/middleware/indexing-status.middleware.ts
```diff
@@ -49,21 +52,15 @@ export type IndexingStatusVariables = {
...
-    );
-  }
+  const cachedIndexingStatus = await fetcher();
+  // use p-reflect to wrap the cached indexing status or rejection
+  // into a PromiseResult, so the downstream request handlers can
+  // handle both success and failure cases, including error details.
+  const indexingStatus = await pReflect(
+    cachedIndexingStatus !== null
+      ? Promise.resolve(cachedIndexingStatus)
+      : Promise.reject(new Error("Unable to fetch current indexing status.")),
```
lightwalker-eth: Where do we generate the JSON in the API response for the case that we need to return an API response error because the indexing status cache has never been able to be successfully built?

tk-o: The JSON response is generated inside `app.get("/indexing-status", async (c) => {})` handler in `apps/ensapi/src/handlers/ensnode-api.ts` file.

packages/ensnode-sdk/src/shared/cache.ts
```diff
@@ -222,6 +222,18 @@ interface SWRCache<T> {
...
 }
 
+interface StaleWhileRevalidateOptions<T> {
+  /**
+   * The async function to wrap with SWR caching
+   */
+  fn: () => Promise<T>;
+
+  /**
+   * Time-to-live duration in seconds. After this duration, data is considered stale
```
lightwalker-eth:
```suggestion
   * Time-to-live duration in seconds. After this duration, cached data is considered stale but is still retained in the cache until replaced with a new value.
```

apps/ensapi/src/middleware/indexing-status.middleware.ts
```diff
@@ -19,26 +20,49 @@ const client = new ENSNodeClient({ url: config.ensIndexerUrl });
...
+      .then((response) => {
+        // reject response with responseCode other than 'ok'
+        if (response.responseCode !== IndexingStatusResponseCodes.Ok) {
           throw new Error(
-            "Received Indexing Status response with 'error' responseCode which will not be cached.",
+            "Received Indexing Status response with responseCode other than 'ok'. The cached indexing status snapshot (if any) will not be updated.",
           );
         }
 
         // resolve response to be cached
```
lightwalker-eth:
```suggestion
        // new indexing status snapshot has successfully been fetched and validated as safe to be cached
```

PR #1306, opened 2025-11-25
Add release-npm-rc workflow

lightwalker-eth:
Goal: Enable us to publish release candidates of packages to NPM so that work on our latest prerelease packages can be done outside of the ENSNode monorepo. Ex: inside ENSAwards.

### DIFF
--- a/.github/workflows/release-npm-rc.yml
+++ b/.github/workflows/release-npm-rc.yml
@@ -1,0 +1,51 @@
+name: Release NPM RC
+
+on:
+  workflow_dispatch:
+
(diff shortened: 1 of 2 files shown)


tk-o:
The workflow introduces with this PR does not work properly (there is issue about how we applied the `NPM_TOKEN`).

On the other hand, [the official guideline for making pre-releases with changesets](https://github.com/changesets/changesets/blob/cc28222ee892b3a078fa02ee26e1cef98c171532/docs/prereleases.md) advises against doing so from the default branch. We have to put more consideration into this feature.

PR #1319, opened 2025-11-27
Using slowest chain indexing cursor timestamp as `updatedAt` for our leaderboard

Goader:
No related issue.

Sets the aggregated referrer metrics cache `updatedAt` timestamp to `slowestChainIndexingCursor` from the indexing status instead of current system time. This ensures consistency with the indexer state and prevents clock skew issues between systems.


tk-o (approved):
LGTM

PR #1322, opened 2025-11-27
feat(ensapi): proactive cache revalidation for Referrer Leaderboard

### LINKED ISSUE #1293: Eager asynchronous updates of SWR caches
Background info: https://namehash.slack.com/archives/C086Z6FNBHN/p1763927698153179

Our current caching strategy for aggregated referrers has an issue. Specifically: outside of the initial startup of ENSApi it is 100% lazy in refreshing its cache of aggregated referrers.

Therefore, if there is a long duration of time between the last update to the aggregated referrer cache and a new visitor to the leaderboard which then triggers an async operation to update the cache once again, this visitor might see aggregated metrics that are very out of date.

To fix this, we should update the cache logic so that ENSApi will trigger an attempt to update its cache every X seconds (perhaps once every 10 minutes) even if there's no incoming API requests to refresh the cached leaderboard. It's important that if these cache updates attempts fail for any reason that it doesn't destroy whatever the existing cached value might be.

Therefore, in general assuming all cache updates are successful, users should never receive a result that's more than 10 minutes out of date.

### LINKED ISSUE #1323: Proactive warmup strategy for SWR caches
Currently reviewing [PR 1307](https://github.com/namehash/ensnode/pull/1307) which transfers some existing logic into a new `triggerEnsAnalyticsAggregatedReferrerSnapshotCacheWarmup` helper function.

It would be nice if we built a more general and mature mechanism for this goal directly into our core SWR logic. For example, to include capabilities such as auto-retry logic in the case that the first attempt to warm up the cache failed for any reason. Perhaps with some exponential backoff.

tk-o:
This PR introduces background revalidation scheduler and implements it for `staleWhileRevalidate` cached. This is an opt-in feature enabled by passing `revalidationInterval` duration value.

Also, this PR allows warming up SWR cache right after gets created. Opt-in by passing `fetchImmediately: true`.

### ENSNode SDK
- Cache implementation have been split into separate files.
- Implemented Background Revalidation Scheduler and integrated it with the SWR cache.

### ENSApi
- background revalidation was enabled for Indexing Status cache, and Referrer Leaderboard cache (see `revalidationInterval` option added)
- `triggerEnsAnalyticsAggregatedReferrerSnapshotCacheWarmup` was replaced by using `fetchImmediately` option with `swrReferrerLeaderboardFetcher` cache.

TODO:
- [ ] add relevant changesets

Resolves #1293.
Resolves #1323.

### DIFF
--- a/apps/ensapi/src/handlers/ensanalytics-api.test.ts
+++ b/apps/ensapi/src/handlers/ensanalytics-api.test.ts
@@ -42,8 +42,8 @@ import app from "./ensanalytics-api";
 
 describe("/ensanalytics", () => {
   describe("/referrers", () => {
-    it("returns requested records when cached referrer leaderboard has multiple pages of data", async () => {
-      // Arrange: set `referrerLeaderboardCache` context var
+    it("returns requested records when referrer leaderboard has multiple pages of data", async () => {
+      // Arrange: set `referrerLeaderboard` context var
       vi.mocked(middleware.referrerLeaderboardMiddleware).mockImplementation(async (c, next) => {
         const mockedReferrerLeaderboard = await pReflect(
           Promise.resolve(populatedReferrerLeaderboard),


packages/ensnode-sdk/src/shared/cache/lru-cache.ts
```diff
@@ -0,0 +1,67 @@
+import type { Cache } from "./cache";
```
tk-o: Moved as is from deleted `../cache.ts` file.

Goader (approved):

PR #1339, opened 2025-12-02
fix(ens-referrals): Fix `hasNext` parameter calculation

Y3drk:
All details are available [in this Slack thread](https://namehash.slack.com/archives/C086Z6FNBHN/p1764681070988089). Below I present a short summary

# TL:DR

When the `itemsPerPage` parameter is equal to the `totalRecords` parameter (when the relevant parameters are as follows):

```ts
{
  page: 1,
  itemsPerPage: X,
  totalRecords: X,
}
```

Requesting the first page of the leaderboard results in the following error: 
``Invalid ReferrerLeaderboardPaginationContext: if hasNext is true, endIndex (${endIndex}) must be less than total (${context.totalRecords}).``

image

The check itself is valid, but **the calculation of `hasNext` is wrong**, and here is why:

```ts
const hasNext = maxTheoreticalIndexOnPage < totalRecords;
```

The formula above doesn't consider the fact that indexes (including the calculation for `maxTheoreticalIndexOnPage`) are "zero-indexed" (first index is 0 then 1,2,...) and the records are "one-indexed" (first record is "indexed" as 1 then 2,3,4...). 

For example when `X=6` we would have:

```
totalRecords = 6;

startIndex = (1 - 1) * 6 = 0
maxTheoreticalIndexOnPage = 0 + (6 - 1) = 5
hasNext = 5 < 6 = true
```

But that is incorrect as our referrers would be indexed as follows: `0,1,2,3,4,5`, so there shouldn't be a next page.

The fix in that PR is my proposed solution to this problem.

### DIFF
--- a/packages/ens-referrals/src/leaderboard-page.test.ts
+++ b/packages/ens-referrals/src/leaderboard-page.test.ts
@@ -1,0 +1,99 @@
+import type { Address } from "viem";
+import { describe, expect, it, vi } from "vitest";
+
+import type { ReferrerLeaderboard } from "./leaderboard.ts";
+import {
+  buildReferrerLeaderboardPaginationContext,
+  type ReferrerLeaderboardPaginationParams,
+} from "./leaderboard-page.ts";
+import type { AwardedReferrerMetrics } from "./referrer-metrics.ts";
+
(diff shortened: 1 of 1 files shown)


tk-o (approved):
LGTM

PR #1340, opened 2025-12-02
fix(ensadmin): registrar actions relative time

### LINKED ISSUE #1290: Fix UX of sorted relative timestamps
Image

Note how the ordering of the relative timestamps in the screenshot above don't make sense.

In the UI, the following must be guaranteed:
* The relative timestamp of record N must be <= the relative timestamp of record N+1

... otherwise, it's very confusing that this is the "latest indexed registrar actions"

... here's another screenshot of this issue. Here I switched back to the tab with ENSAdmin in it after spending a few minutes writing this issue:

Image

I assume that the relative sorting of the absolute timestamps of each registrar action in the UI is sorted correctly.

I assume the problem here is due to how we're generating the relative timestamps that are displayed in the UI, including how these relative timestamps are not automatically updating in the UI as time advances. Instead, they are only true as of the moment we first generate a relative timestamp in the UI, but then as time advances and new latest registrar actions are prepended to the top of the list, the expected invariant of the ordering of these relative timestamp values is breaking.

There's probably multiple ways we could fix this, including having these relative timestamps auto-update each time we prepend a new data record to the list. Alternatively if we auto-update the relative timestamps on a timer, that can be nice, but it wouldn't fully guarantee our goal here.

I believe the ideal solution is to combine both strategies described in the paragraph above.

tk-o:
Use the current system time to power `useNow()` hook which returns value to be used as `relativeTo` parm for `<RelativeTime />` calls.

This PR updates the _Latest indexed registrar actions_ view to refresh the relative time each second. You can observe the effect on records which display value below 1 minute.

Resolves #1290


apps/ensadmin/src/lib/synced-clock.ts
```diff
@@ -1,106 +1,117 @@
...
+   *
+   * For example, updating clock 60 times in a second requires
+   * tick rate of around 16ms.
+   */
+  private tickRate: number = 16;
 
   /**
    * Clock's tick handler.
    *
    * Runs multiple times a second, and notifies all listeners about
```
lightwalker-eth:
```suggestion
   * Runs at `tickRate`, and notifies all listeners about
```

Is that fair?

apps/ensadmin/src/lib/synced-clock.ts
```diff
@@ -1,106 +1,117 @@
...
    */
-  function addListener(callback: SyncedClockListener) {
-    listeners.add(callback);
+  public addListener(callback: SyncedClockListener) {
+    this.listeners.add(callback);
 
     // start when someone is listening
-    if (listeners.size > 0) {
-      start();
+    if (this.listeners.size > 0) {
```
lightwalker-eth: Hmm.. should we only start when transitioning from 0 -> 1 listener, rather than start again for each incremental listener added after the first?

apps/ensadmin/src/lib/synced-clock.ts
```diff
@@ -1,106 +1,117 @@
...
 /**
- * Create high-precision clock with updates on every frame.
- * @returns {SyncedClock} to be used with {@link useSyncExternalStore}.
+ * High-precision clock with multiple updates over each second.
  */
-export function createSyncedClock(): SyncedClock {
-  const listeners = new Set<SyncedClockListener>();
-  let _currentTime = Date.now();
-  let timerId: ReturnType<typeof setTimeout> | null = null;
+export class HighPrecisionSyncedClock implements SyncedClock {
```
lightwalker-eth: "High precision" doesn't feel like accurate terminology. Given how the JS runtime works I don't think we can say the ticks for this clock are guaranteed to happen with "high precision".

Maybe "High resolution" is better.

lightwalker-eth (approved):
@tk-o Shared a few more suggestions. Please take the lead to merge when ready 👍

PR #1348, opened 2025-12-03
fix: dependency security issues

tk-o:
Update Next.js and React dependencies.


lightwalker-eth (approved):
@tk-o Thanks for your fast action 🫡

Issue #1354, opened 2025-12-04
Fix docs site search bar


lightwalker-eth:
<img width="1264" height="854" alt="Image" src="https://github.com/user-attachments/assets/245b5f3e-a42b-467e-b3f4-c71fa325abcb" />

lightwalker-eth:
Solved via #1356

PR #1356, opened 2025-12-04
fix: fixed styles for index page search

stevedylandev:
## Summary 
- PR satifies issue #1354 

## Problem

Since `ensnode.io` is using a custom index page the search does not have the same defaults seen through the rest of the app. In order to style our landing page the way we wanted to, we had to break out of the standard layout that is used in Astro Starlight. Because of this the search component is taken out of that context and breaks. 

We had already handled the appearance and placement of the search component on the index page, but not what happens when someone clicks on it. 

## Solution

I was able to fix this with a few changes 
- Grabbed the CSS selectors for the search modal (aka `pagefind`)
- Added more custom styles to handle dialog when opened
- Added data attribute to the index page to prevent the CSS changes overriding search on the rest of the docs where it's already working 

Before: 

CleanShot 2025-12-04 at 15 54 38@2x

After:

CleanShot 2025-12-04 at 17 09 57@2x

## Future Updates

If we want to update this some more and update the styles further we certainly can, but I vote that we put that work off for now in case we switch to a new docs platform like [Fumadocs](https://fumadocs.dev) for other robust features like OpenAPI support.

### DIFF
--- a/docs/ensnode.io/src/components/atoms/GenericTooltip.tsx
+++ b/docs/ensnode.io/src/components/atoms/GenericTooltip.tsx
@@ -80,7 +80,7 @@ export default function GenericTooltip({
           onMouseEnter={handleOpen}
           onMouseLeave={handleClose}
           style={{ maxWidth: maxTooltipWidth }}
-          className="rounded-md bg-black focus:outline-none"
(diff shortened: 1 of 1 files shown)


stevedylandev:
Ah good catch! I'll fix that today

PR #1357, opened 2025-12-05
feat(ensnode-sdk): `tokenscope` module

tk-o:
This PR refactors Assets ideas to be part of ENSNode SDK. Also, this PR updates:
- `AssetId` type to be an object type, and  creates the `SerializedAssetId` type to be a CAIP-19 string representation of `AssetId`.
- `SupportedNFT` type to become `DomainAssetId` type.

PR #1358, opened 2025-12-05
feat(ensapi): introduce Name Tokens API

tk-o:
Suggested review order: commit-by-commit.

This PR introduces new Name Tokens API route.

<details>
<summary>Example: `finalroll.eth` (minted, wrapped)</summary>

<div>
Example request:

```
GET /api/name-tokens?name=finalroll.eth
GET /api/name-tokens?domainId=0x57bda742b7d14316ce9f4f90a4a6035b741057e90ffa1d078ac6e24b1c1028df
```

Example response:
```
{
  "responseCode": "ok",
  "registeredNameTokens": {
    "domainId": "0x57bda742b7d14316ce9f4f90a4a6035b741057e90ffa1d078ac6e24b1c1028df",
    "name": "finalroll.eth",
    "tokens": [
      {
        "domainAsset": {
          "assetId": {
            "assetNamespace": "erc721",
            "contract": {
              "chainId": 1,
              "address": "0x57f1887a8bf19b14fc0df6fd9b2acc9af147ea85"
            },
            "tokenId": "0x5dc22117a784fc74bc3d4df34423ac78eabc534a3d233be07950b82fa68fcac7"
          },
          "domainId": "0x57bda742b7d14316ce9f4f90a4a6035b741057e90ffa1d078ac6e24b1c1028df"
        },
        "ownership": {
          "ownershipType": "proxy",
          "owner": {
            "chainId": 1,
            "address": "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401"
          }
        },
        "mintStatus": "minted"
      },
      {
        "domainAsset": {
          "assetId": {
            "assetNamespace": "erc1155",
            "contract": {
              "chainId": 1,
              "address": "0xd4416b13d2b3a9abae7acd5d6c2bbdbe25686401"
            },
            "tokenId": "0x57bda742b7d14316ce9f4f90a4a6035b741057e90ffa1d078ac6e24b1c1028df"
          },
          "domainId": "0x57bda742b7d14316ce9f4f90a4a6035b741057e90ffa1d078ac6e24b1c1028df"
        },
        "ownership": {
          "ownershipType": "effective",
          "owner": {
            "chainId": 1,
            "address": "0x27c9b34eb43523447d3e1bcf26f009d814522687"
          }
        },
        "mintStatus": "minted"
      }
    ],
    "expiresAt": 1796731799,
    "accurateAsOf": 1765280254
  }
}

```
</div>
</details> 

<details><summary>Example: `tko.eth` (unwrapped, expired)</summary>

<div>
Example request:

```
GET /api/name-tokens?name=tko.eth
GET /api/name-tokens?domainId=0xaf20e732cb67b5895d3a9516990611765a4ee138ebdadb60410539beb1dc42bb
```

Example response:
```
{
  "responseCode": "ok",
  "registeredNameTokens": {
    "domainId": "0xaf20e732cb67b5895d3a9516990611765a4ee138ebdadb60410539beb1dc42bb",
    "name": "tko.eth",
    "tokens": [
      {
        "domainAsset": {
          "assetId": {
            "assetNamespace": "erc721",
            "contract": {
              "chainId": 1,
              "address": "0x57f1887a8bf19b14fc0df6fd9b2acc9af147ea85"
            },
            "tokenId": "0x2376b3952dc704bc0cbebbdab63e4b510d63a1234f5910f12362f5b34bdcc704"
          },
          "domainId": "0xaf20e732cb67b5895d3a9516990611765a4ee138ebdadb60410539beb1dc42bb"
        },
        "ownership": {
          "ownershipType": "effective",
          "owner": {
            "chainId": 1,
            "address": "0xd0ed779af7e3063c9121590c1f6a90396a22b824"
          }
        },
        "mintStatus": "minted"
      }
    ],
    "expiresAt": 1758748175,
    "accurateAsOf": 1765280339
  }
}
```
</div>
</details> 

Builds on top of #1357 

### TODOs
- [x] ~Pagination support? Might not be required, though. Saw up to two name token records while testing. Most of the names will have just a single associated name token record.~
  - Not required
- [x] only merge this PR if #1374 is approved to be merged. The #1374 must be merged right after #1358.

### Followup on
- [ ] When there were no `tokens` found for the requested `name`, the response shoul be a 200 success where we just return an empty array of name tokens.
- [ ] Think of better name (if any) for `RegisteredNameTokens` entity that's returned by Name Tokens API.
- [ ] Set explicit order of name tokens in the response.
- [ ] Handle [the edge-case](https://github.com/namehash/ensnode/pull/1358#discussion_r2607118588) found by @lightwalker-eth in ENS app.

### DIFF
--- a/.changeset/thick-ads-nail.md
+++ b/.changeset/thick-ads-nail.md
@@ -1,0 +1,7 @@
+---
+"@namehash/ens-referrals": minor
+"@ensnode/ensnode-sdk": minor
+"ensapi": minor
+---
+
+Renamed `itemsPerPage` to `recordsPerPage` and `paginationContext` to `pageContext` in referrer leaderboard APIs to align with registrar actions terminology.
--- a/apps/ensapi/src/handlers/ensanalytics-api.test.ts
+++ b/apps/ensapi/src/handlers/ensanalytics-api.test.ts
@@ -62,23 +62,23 @@ describe("/ensanalytics", () => {
 
       // Arrange: create the test client from the app instance
       const client = testClient(app);
-      const itemsPerPage = 10;
+      const recordsPerPage = 10;
 
       // Act: send test request to fetch 1st page
       const responsePage1 = await client.referrers
-        .$get({ query: { itemsPerPage: `${itemsPerPage}`, page: "1" } }, {})
+        .$get({ query: { recordsPerPage: `${recordsPerPage}`, page: "1" } }, {})
         .then((r) => r.json())
         .then(deserializeReferrerLeaderboardPageResponse);
 
       // Act: send test request to fetch 2nd page
       const responsePage2 = await client.referrers
-        .$get({ query: { itemsPerPage: `${itemsPerPage}`, page: "2" } }, {})
+        .$get({ query: { recordsPerPage: `${recordsPerPage}`, page: "2" } }, {})
         .then((r) => r.json())
         .then(deserializeReferrerLeaderboardPageResponse);
 
       // Act: send test request to fetch 3rd page
       const responsePage3 = await client.referrers
-        .$get({ query: { itemsPerPage: `${itemsPerPage}`, page: "3" } }, {})
+        .$get({ query: { recordsPerPage: `${recordsPerPage}`, page: "3" } }, {})
         .then((r) => r.json())
         .then(deserializeReferrerLeaderboardPageResponse);
 
@@ -87,17 +87,17 @@ describe("/ensanalytics", () => {
         responseCode: ReferrerLeaderboardPageResponseCodes.Ok,
         data: {
           ...populatedReferrerLeaderboard,
-          paginationContext: {
+          pageContext: {
             endIndex: 9,
             hasNext: true,
             hasPrev: false,
-            itemsPerPage: 10,
+            recordsPerPage: 10,
             page: 1,
             startIndex: 0,
             totalPages: 3,
             totalRecords: 29,
           },
-          referrers: allPossibleReferrersIterator.take(itemsPerPage).toArray(),
+          referrers: allPossibleReferrersIterator.take(recordsPerPage).toArray(),
         },
       } satisfies ReferrerLeaderboardPageResponseOk;
 
@@ -108,17 +108,17 @@ describe("/ensanalytics", () => {
         responseCode: ReferrerLeaderboardPageResponseCodes.Ok,
         data: {
           ...populatedReferrerLeaderboard,
-          paginationContext: {
+          pageContext: {
             endIndex: 19,
             hasNext: true,
             hasPrev: true,
-            itemsPerPage: 10,
+            recordsPerPage: 10,
             page: 2,
             startIndex: 10,
             totalPages: 3,
             totalRecords: 29,
           },
-          referrers: allPossibleReferrersIterator.take(itemsPerPage).toArray(),
+          referrers: allPossibleReferrersIterator.take(recordsPerPage).toArray(),
         },
       } satisfies ReferrerLeaderboardPageResponseOk;
       expect(responsePage2).toMatchObject(expectedResponsePage2);
@@ -128,17 +128,17 @@ describe("/ensanalytics", () => {
         responseCode: ReferrerLeaderboardPageResponseCodes.Ok,
         data: {
           ...populatedReferrerLeaderboard,
-          paginationContext: {
+          pageContext: {
             endIndex: 28,
             hasNext: false,
             hasPrev: true,
-            itemsPerPage: 10,
+            recordsPerPage: 10,
             page: 3,
             startIndex: 20,
             totalPages: 3,
             totalRecords: 29,
           },
-          referrers: allPossibleReferrersIterator.take(itemsPerPage).toArray(),
+          referrers: allPossibleReferrersIterator.take(recordsPerPage).toArray(),
         },
       } satisfies ReferrerLeaderboardPageResponseOk;
       expect(responsePage3).toMatchObject(expectedResponsePage3);
@@ -155,11 +155,11 @@ describe("/ensanalytics", () => {
 
       // Arrange: create the test client from the app instance
       const client = testClient(app);
-      const itemsPerPage = 10;
+      const recordsPerPage = 10;
 
       // Act: send test request to fetch 1st page
       const response = await client.referrers
-        .$get({ query: { itemsPerPage: `${itemsPerPage}`, page: "1" } }, {})
+        .$get({ query: { recordsPerPage: `${recordsPerPage}`, page: "1" } }, {})
         .then((r) => r.json())
         .then(deserializeReferrerLeaderboardPageResponse);
 
@@ -168,10 +168,10 @@ describe("/ensanalytics", () => {
         responseCode: ReferrerLeaderboardPageResponseCodes.Ok,
         data: {
           ...emptyReferralLeaderboard,
-          paginationContext: {
+          pageContext: {
             hasNext: false,
             hasPrev: false,
-            itemsPerPage: 10,
+            recordsPerPage: 10,
             page: 1,
             totalPages: 1,
             totalRecords: 0,
--- a/apps/ensapi/src/handlers/ensanalytics-api.ts
+++ b/apps/ensapi/src/handlers/ensanalytics-api.ts
@@ -26,14 +26,14 @@ const logger = makeLogger("ensanalytics-api");
 // Pagination query parameters schema (mirrors ReferrerLeaderboardPageRequest)
 const paginationQuerySchema = z.object({
   page: z.optional(z.coerce.number().int().min(1, "Page must be a positive integer")),
-  itemsPerPage: z.optional(
+  recordsPerPage: z.optional(
     z.coerce
       .number()
       .int()
-      .min(1, "Items per page must be at least 1")
+      .min(1, "Records per page must be at least 1")
       .max(
         REFERRERS_PER_LEADERBOARD_PAGE_MAX,
-        `Items per page must not exceed ${REFERRERS_PER_LEADERBOARD_PAGE_MAX}`,
+        `Records per page must not exceed ${REFERRERS_PER_LEADERBOARD_PAGE_MAX}`,
       ),
   ),
 }) satisfies z.ZodType<ReferrerLeaderboardPageRequest>;
@@ -65,9 +65,9 @@ const app = factory
         );
       }
 
-      const { page, itemsPerPage } = c.req.valid("query");
+      const { page, recordsPerPage } = c.req.valid("query");
       const leaderboardPage = getReferrerLeaderboardPage(
-        { page, itemsPerPage },
+        { page, recordsPerPage },
         referrerLeaderboard.value,
       );
 
--- a/apps/ensapi/src/lib/ensanalytics/referrer-leaderboard/mocks.ts
+++ b/apps/ensapi/src/lib/ensanalytics/referrer-leaderboard/mocks.ts
@@ -990,9 +990,9 @@ export const referrerLeaderboardPageResponseOk: ReferrerLeaderboardPageResponseO
       grandTotalQualifiedReferrersFinalScore: 16.55216891669386,
       minFinalScoreToQualify: 0,
     },
-    paginationContext: {
+    pageContext: {
       page: 1,
-      itemsPerPage: 100,
+      recordsPerPage: 100,
       totalRecords: 29,
       totalPages: 1,
       hasNext: false,
--- a/packages/ens-referrals/src/leaderboard-page.test.ts
+++ b/packages/ens-referrals/src/leaderboard-page.test.ts
@@ -10,12 +10,12 @@ import {
 import type { AwardedReferrerMetrics } from "./referrer-metrics.ts";
 
 describe("buildReferrerLeaderboardPageContext", () => {
-  const paginationParams: ReferrerLeaderboardPageParams = {
+  const pageParams: ReferrerLeaderboardPageParams = {
     page: 1,
-    itemsPerPage: 3,
+    recordsPerPage: 3,
   };
 
-  it("correctly evaluates `hasNext` when `leaderboard.referrers.size` and `itemsPerPage` are equal", () => {
+  it("correctly evaluates `hasNext` when `leaderboard.referrers.size` and `recordsPerPage` are equal", () => {
     const leaderboard: ReferrerLeaderboard = {
       rules: {
         totalAwardPoolValue: 10000,
@@ -84,15 +84,15 @@ describe("buildReferrerLeaderboardPageContext", () => {
     };
 
     const buildReferrerLeaderboardPageContextSpy = vi.fn(buildReferrerLeaderboardPageContext);
-    const result = buildReferrerLeaderboardPageContextSpy(paginationParams, leaderboard);
+    const result = buildReferrerLeaderboardPageContextSpy(pageParams, leaderboard);
 
     expect(
       buildReferrerLeaderboardPageContextSpy,
-      "buildReferrerLeaderboardPageContext should successfully complete for itemsPerPage=3, leaderboard.referrers.size=3",
+      "buildReferrerLeaderboardPageContext should successfully complete for recordsPerPage=3, leaderboard.referrers.size=3",
     ).toHaveReturned();
     expect(
       result.hasNext,
-      `Leaderboard should only have one page for itemsPerPage=3, leaderboard.referrers.size=3 (expected hasNext to be false, is ${result.hasNext})`,
+      `Leaderboard should only have one page for recordsPerPage=3, leaderboard.referrers.size=3 (expected hasNext to be false, is ${result.hasNext})`,
     ).toStrictEqual(false);
   });
 
@@ -125,16 +125,16 @@ describe("buildReferrerLeaderboardPageContext", () => {
       hasPrev: false,
       startIndex: undefined,
       endIndex: undefined,
-      itemsPerPage: 3,
+      recordsPerPage: 3,
       page: 1,
     };
 
     const buildReferrerLeaderboardPageContextSpy = vi.fn(buildReferrerLeaderboardPageContext);
-    const result = buildReferrerLeaderboardPageContextSpy(paginationParams, leaderboard);
+    const result = buildReferrerLeaderboardPageContextSpy(pageParams, leaderboard);
 
     expect(
       buildReferrerLeaderboardPageContextSpy,
-      "buildReferrerLeaderboardPageContext should successfully complete for itemsPerPage=3, leaderboard.referrers.size=0",
+      "buildReferrerLeaderboardPageContext should successfully complete for recordsPerPage=3, leaderboard.referrers.size=0",
     ).toHaveReturned();
 
     expect(
--- a/packages/ens-referrals/src/leaderboard-page.ts
+++ b/packages/ens-referrals/src/leaderboard-page.ts
@@ -32,7 +32,7 @@ export interface ReferrerLeaderboardPageParams {
    * @invariant Must be a positive integer (>= 1) and less than or equal to {@link REFERRERS_PER_LEADERBOARD_PAGE_MAX}
    * @default {@link REFERRERS_PER_LEADERBOARD_PAGE_DEFAULT}
    */
-  itemsPerPage?: number;
+  recordsPerPage?: number;
 }
 
 const validateReferrerLeaderboardPageParams = (params: ReferrerLeaderboardPageParams): void => {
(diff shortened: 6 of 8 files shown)


apps/ensindexer/src/plugins/registrars/ethnames/lib/registrar-helpers.ts
```diff
@@ -23,11 +23,5 @@ export function tokenIdToLabelHash(tokenId: bigint): LabelHash {
...
  */
 export function getRegistrarManagedName(namespaceId: ENSNamespaceId): RegistrarManagedName {
-  switch (namespaceId) {
-    case "mainnet":
-    case "sepolia":
-    case "holesky":
-    case "ens-test-env":
-      return "eth";
-  }
+  return getRegistrarManagedName(namespaceId);
```
tk-o: This created an endless loop, but was fixed in one of the following commits.

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,124 @@
...
+import { findNameTokens } from "@/lib/name-tokens/find-name-tokens";
+import { getIndexedSubregistries } from "@/lib/name-tokens/get-indexed-subregistries";
+import { nameTokensApiMiddleware } from "@/middleware/name-tokens.middleware";
+
+const app = factory.createApp();
+
+const logger = makeLogger("name-tokens-api");
+
+const RESPONSE_RECORDS_PER_PAGE_DEFAULT = 25;
+const RESPONSE_RECORDS_PER_PAGE_MAX = 25;
```
lightwalker-eth: We shouldn't have any concept of records per page for this API.

Each lookup returns all the tokens for the provided name, no matter how many tokens that is. In practice it's an exception to ever have more than 1 token but such cases can exist.

packages/ensnode-sdk/src/registrars/basenames-subregistry.ts
```diff
@@ -0,0 +1,55 @@
...
+  ENSNamespaceIds,
+  maybeGetDatasource,
+} from "@ensnode/datasources";
+
+import type { Name } from "../ens";
+import type { AccountId } from "../shared";
+
+/**
+ * Gets the SubregistryId (an AccountId) of the Basenames Subregistry contract (this is the
+ * "BaseRegistrar" contract for direct subnames of .eth) for the provided namespace.
```
lightwalker-eth:
```suggestion
 * "BaseRegistrar" contract for Basenames) for the provided namespace.
```

packages/ensnode-sdk/src/ens/names.test.ts
```diff
@@ -30,6 +30,24 @@ describe("names", () => {
     });
   });
 
+  describe("getParentNameFQDN", () => {
+    it("returns FQDN for an empty name", () => {
+      expect(getParentNameFQDN("")).toStrictEqual("");
```
lightwalker-eth: Seems that this case should throw an error, since there's no parent of the ENS root

packages/ensnode-sdk/src/ens/names.test.ts
```diff
@@ -30,6 +30,24 @@ describe("names", () => {
...
+    });
+
+    it("returns FQDN for top-level name", () => {
+      expect(getParentNameFQDN("eth")).toStrictEqual("");
```
lightwalker-eth: We should explicitly document on `getParentNameFQDN` that if a single label name is passed it will return the ENS root ("")

packages/ensnode-sdk/src/api/name-tokens/request.ts
```diff
@@ -0,0 +1,42 @@
+import type { Name, Node } from "../../ens";
+
+/**
+ * Records Filters: Filter Types
```
lightwalker-eth: I don't see this API as having a filter.

It's a key -> value lookup.

It just so happens that the value might be an array of more than one item, but we should still think of it as a key -> value lookup.

I can imagine this API potentially having different ways to provide a key.
1. by domainId
2. by name (which ENSApi then internally converts into a domainId to perform the rest of the lookup)

So thinking the API is setup so that you need to pass exactly one of the following params: `?name=X` or `?domainId=X`

packages/ensnode-sdk/src/api/name-tokens/request.ts
```diff
@@ -0,0 +1,42 @@
...
+  filterType: typeof NameTokensFilterTypes.ByDomainId;
+  domainId: Node;
+};
+
+export type NameTokensFilter = NameTokensFilterByDomainId;
+
+/**
+ * Records Orders
+ */
+export const NameTokensOrders = {
```
lightwalker-eth: There should be no sort operation options for this API.

The order of the tokens returned in the array should be undefined, or perhaps we give it some defined ordering but it should not be something you can change via the API call.

Please note this API should not be thought of as returning a list that you're paginating through. It's fundamentally a key -> value lookup.

apps/ensapi/src/middleware/name-tokens.middleware.ts
```diff
@@ -0,0 +1,91 @@
...
+      );
+    }
+
+    if (c.var.indexingStatus.isRejected) {
+      // no indexing status available in context
+      logger.error(
+        {
+          error: c.var.indexingStatus.reason,
+        },
+        `Name Tokens API requested but indexing status is not available in context.`,
```
lightwalker-eth:
```suggestion
        `Name Tokens API requested but indexing status is not available in context yet.`,
```

Details like these are important as this failure hopefully is temporary, not permanent the way the required plugin error is.

apps/ensapi/src/middleware/name-tokens.middleware.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    const { omnichainSnapshot } = c.var.indexingStatus.value.snapshot;
+
+    if (!nameTokensPrerequisites.hasIndexingStatusSupport(omnichainSnapshot.omnichainStatus))
+      return c.json(
+        serializeNameTokensResponse({
+          responseCode: NameTokensResponseCodes.Error,
+          errorCode: NameTokensResponseErrorCodes.Generic,
+          error: {
+            message: `Name Tokens API is not available`,
```
lightwalker-eth:
```suggestion
            message: `Name Tokens API is not available yet`,
```

apps/ensapi/src/middleware/name-tokens.middleware.ts
```diff
@@ -0,0 +1,91 @@
...
+            details: `The cached omnichain indexing status of the Connected ENSIndexer must be one of the following ${nameTokensPrerequisites.supportedIndexingStatusIds.map((statusId) => `"${statusId}"`).join(", ")}.`,
```
lightwalker-eth: ... but is currently X

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,124 @@
...
+
+const indexedSubregistries = getIndexedSubregistries(config.namespace);
+
+// Middleware managing access to Name Tokens API routes.
+// It makes the routes available if all prerequisites are met.
```
lightwalker-eth:
```suggestion
// It makes the route available if all prerequisites are met, and if not returns the appropriate HTTP 503 (Service Unavailable) error
```

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (c.var.indexingStatus === undefined) {
+      throw new Error(`Invariant(name-tokens-api): indexingStatusMiddleware required`);
+    }
+
+    // Invariant: Indexing Status has been resolved successfully.
+    if (c.var.indexingStatus.isRejected) {
+      throw new Error(
+        `Invariant(name-tokens-api): Indexing Status has to be resolved successfully`,
+      );
+    }
```
lightwalker-eth: Should the middleware pass the validated indexingStatus object in some new context param so that we don't need to repeat these sanity checks?

For example maybe before the middleware is called we call it `unvalidatedIndexingStatus` and then after the middleware we call it `indexingStatus`.

Appreciate your advice and also happy to create a separate follow-up issue for this.

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const { name } = c.req.valid("param");
+
+    const parentNode = namehash(getParentNameFQDN(name));
+    const subregistry = indexedSubregistries.find((subregistry) => subregistry.node === parentNode);
+
+    if (!subregistry) {
+      logger.error(`The requested '${name}' name is unknown to ENSNode.`);
```
lightwalker-eth: We need to invest more into our terminology and language. Optics really matter. Please give these details more attention.

```suggestion
      logger.error(`This ENSNode instance has not been configured to index tokens for the requested name: '${name}'.`);
```

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,124 @@
...
+            message: "Internal Server Error",
+            details: `The requested '${name}' name is unknown to ENSNode.`,
+          },
+        } satisfies NameTokensResponseErrorUnknownNameContext),
+      );
+    }
+
+    const { snapshot } = c.var.indexingStatus.value;
+    const { subregistryId } = subregistry;
+    const latestIndexedBlockRef = getLatestIndexedBlockRef(snapshot, subregistryId.chainId);
```
lightwalker-eth: We should prepare this API endpoint for the complexity of having indexed tokens for the name across multiple subregistryIds on multiple chains. This can already happen with 3DNS names and will happen with .eth names upon the launch of ENSv2.

Therefore we should build `accurateAsOf` off of the omnichain indexing status cursor and not off the last indexed block ref of the specific subregistry being referenced here.

packages/ensnode-sdk/src/api/name-tokens/response.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * A response when Name Tokens API can respond with requested data.
+ */
+export type NameTokensResponseOk = {
+  responseCode: typeof NameTokensResponseCodes.Ok;
+
+  /**
+   * Name Tokens for the requested name.
+   */
+  nameTokens: RegisteredNameToken[];
```
lightwalker-eth: It will be high value to document some ideas here. We need to explicitly document the potential state space of responses. Suggest to look for names with each of the state permutations I cite below and then consider creating some integration test for the API responses we make for them. Advice appreciated.

1. What does this array contain? -> It contains EVERY tokenized representation of the name that has ever been indexed for the given name as of `accurateAsOf`, even if the given token has been burned or expired.
2. What does it mean if this array is empty? -> It means the name has never been registered / minted before.
3. What does it mean if this array contains more than 1 item? ->
   1. It means that more than 1 distinct tokenized representation of the ownership of the name has been indexed for the given name as of `accurateAsOf`.
   2. All possible permutations of mint statuses of these tokens is possible. Multiple could be actively minted. Multiple could be burnt. Some could be burnt, others could be minted.
   3. We should give specific guidance for how direct subnames of .eth that are wrapped by the NameWrapper will be represented.
      1. Case A: A direct subname of .eth that has never been registered.
      2. Case B: A direct subname of .eth that has been registered but has never been wrapped by the NameWrapper, and...:
         1. Case B.1: Is still actively minted (independent of its expiry state)
         2. Case B.2: Has been burned by sending it to the null address.
      5. Case C: A direct subname of .eth that has been registered and has been wrapped by the NameWrapper, and...
         1. Case C.1: Is still actively wrapped by the NameWrapper (independent of its expiry state)
         2. Case C.2: Is no longer wrapped by the NameWrapper, but is still actively minted by the BaseRegistrar (independent of its expiry state)
         3. Case C.3: Is no longer wrapped by the NameWrapper, and is also no longer minted by the BaseRegistrar (both tokens now burned by sending to the null address)
4. In what ordering will the values in the array be returned? Please suggest what we should use.
5. Please cite what is guaranteed to be distinct about each value in the array.
6. Explicitly document why it can be useful to get tokenized representations of the name that are now burnt: This can be helpful for looking up historical activity for the name, including past buy orders, sell orders, and sales.

packages/ensnode-sdk/src/tokenscope/name-token.ts
```diff
@@ -0,0 +1,52 @@
...
+   */
+  owner: Address;
+
+  /**
+   * The mint status of the token.
+   *
+   * After an NFT was indexed it can never be deleted. Instead, when an
+   * indexed NFT is burned onchain its record is retained and its mint
+   * status is updated as `burned`. If the NFT is minted again after
+   * it was burned, its mint status is updated to `minted`.
```
lightwalker-eth:
```suggestion
   * After ENSNode indexes the token for a name, even if that token is burned, ENSNode will never forget how the token once represented the name. When the token for a name is burned, ENSNode remembers this token but updates its mint status to `burned`. If this token becomes minted again after it was burned, its mint status is updated to `minted` again.
```

lightwalker-eth (commented):
@tk-o Really happy to see this! Shared some suggestions and feedback. Appreciate your efforts to help make this API amazing 🫡

packages/ensnode-sdk/src/api/name-tokens/response.ts
tk-o: @lightwalker-eth I suggest we don't include the case when `nameTokens` array is empty. We simply should return a 404 error with explanation how no name tokens were never indexed for the requested name / domainId. It will simplify the data model (we won't have to wonder what to do with `expiresAt` field for a name that was never indexed 😅).

packages/ensnode-sdk/src/api/name-tokens/response.ts
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * The derived string union of possible {@link NameTokensResponseCodes}.
+ */
+export type NameTokensResponseCode =
+  (typeof NameTokensResponseCodes)[keyof typeof NameTokensResponseCodes];
+
+/**
+ * Error codes for Name Tokens API responses with 'error' response code.
+ */
+export const NameTokensResponseErrorCodes = {
```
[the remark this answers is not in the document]
tk-o: Updated 👍

packages/ensnode-sdk/src/tokenscope/name-token.ts
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * The mint status of the token.
+   *
+   * After ENSNode indexes the token for a name, even if that token is burned,
+   * ENSNode will never forget how the token once represented the name.
+   * When the token for a name is burned, ENSNode remembers this token but
+   * updates its mint status to `burned`. If this token becomes minted again
+   * after it was burned, its mint status is updated to `minted` again.
+   */
+  mintStatus: NFTMintStatus;
```
lightwalker-eth: Suggest to document the invariant here that `mintStatus` will be burned if and only if `ownership.ownershipType` is `NameTokenOwnershipTypes.Burned`.

packages/ensnode-sdk/src/tokenscope/name-token.ts
```diff
@@ -0,0 +1,129 @@
...
+  | NameTokenOwnershipEffective
+  | NameTokenOwnershipBurned;
+
+export interface NameToken {
+  /**
+   * Domain Asset
+   *
+   * References NFT that tokenizes the ownership of a domain.
+   */
+  domainAsset: DomainAssetId;
```
lightwalker-eth:
```suggestion
  token: AssetId;
```

Goals:
1. Won't it be redundant to keep repeating the same `domainId` value in each `NameToken` that's returned for the same name? Therefore suggesting to change this field from a `DomainAssetId` to just an `AssetId`. I also note how `RegisteredNameTokens` already includes both a `domainId` and `name` field already so seems really no need to duplicate `domainId` further for each `NameToken` in the response.
2. I imagine code that uses this object will read nicer if we rename this field from `domainAsset` to `token`.

apps/ensapi/src/lib/name-tokens/find-name-tokens-for-domain.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Find all Name Tokens for the registered domain by domain ID,
+ * accurate as of `accurateAsOf`.
+ *
+ * @returns {RegisteredNameTokens} if some tokens associated with
+ * domain ID were found. Otherwise returns null.
+ *
+ * Note: if no tokens were found, it means that the subregistry managing
+ * the name associated with the domainId is not an actively indexed subregistry.
+ */
+export async function findRegisteredNameTokensForDomain(
```
lightwalker-eth:
```suggestion
export async function findIndexedNameTokensForDomain(
```

apps/ensapi/src/lib/name-tokens/find-name-tokens-for-domain.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * Internal function which executes a single query to get all data required to
+ * build a list of {@link RegisteredNameToken} objects.
+ */
+async function _findRegisteredNameTokensForDomain(
+  domainId: Node,
+): Promise<FindRegisteredNameTokensForDomainRecord[]> {
+  const query = db
```
lightwalker-eth: I don't see anything here related to the ordering of the tokens that will ultimately be returned

apps/ensapi/src/lib/name-tokens/get-namewrapper-addresses.ts
```diff
@@ -0,0 +1,40 @@
...
+  type ENSNamespaceId,
+  getDatasource,
+  maybeGetDatasource,
+} from "@ensnode/datasources";
+import { PluginName } from "@ensnode/ensnode-sdk";
+
+/**
+ * Get all NameWrapper contract addresses withing the ENS Namespace for the active plugins.
+ */
+export function getNameWrapperAddresses(
```
lightwalker-eth: Considering how Lineanames is involved here I don't feel good about this function returning `Address` values.

Whenever we're in a potential multichain context we should return `Account` values.

This function should be appropriate renamed / docs updated / etc. Thanks

packages/ensnode-sdk/src/api/name-tokens/serialize.ts
```diff
@@ -0,0 +1,42 @@
...
+  NameTokensResponseCodes,
+  type RegisteredNameTokens,
+} from "./response";
+import type {
+  SerializedNameTokensResponse,
+  SerializedNameTokensResponseOk,
+  SerializedRegisteredNameTokens,
+} from "./serialized-response";
+
+export function serializeRegisteredNameTokens({
```
lightwalker-eth:
```suggestion
export function serializeIndexedNameTokens({
```

packages/ensnode-sdk/src/api/name-tokens/response.ts
```diff
@@ -0,0 +1,196 @@
...
+   *   1) More than 1 distinct tokenized representation of the ownership of
+   *      the `name` has been indexed as of `accurateAsOf`.
+   *   2) All possible permutations of mint statuses of these tokens are
+   *      possible:
+   *      a) Multiple could be actively minted.
+   *      b) Multiple could be burned.
+   *      c) Some could be burned, others could be minted.
+   *  - Order of name tokens follows the order of onchain events that were
+   *    indexed when a token was minted, or burned.
+   *  - Each name token has a distinct `domainAsset` value which references
```
lightwalker-eth:
```suggestion
   *  - Each name token has a distinct `token` value which references
```

packages/ensnode-sdk/src/api/name-tokens/response.ts
```diff
@@ -0,0 +1,196 @@
...
+   *    NFT that tokenizes the ownership of a domain.
```
lightwalker-eth:
```suggestion
   *    the NFT that currently or previously tokenized ownership of `name`.
```

packages/ensnode-sdk/src/api/name-tokens/response.ts
```diff
@@ -0,0 +1,196 @@
...
+   * 2) A direct subname of .eth that has been registered and
+   *    has been wrapped by the NameWrapper, and:
+   *    a) Is still actively wrapped by the NameWrapper (independent of its
+   *       expiry state).
+   *    b) Is no longer wrapped by the NameWrapper, but is still actively
+   *       minted by the BaseRegistrar (independent of its expiry state).
+   *    c) Is no longer wrapped by the NameWrapper, and is also no longer
+   *       minted by the BaseRegistrar (both tokens now burned by sending to
+   *       the null address).
+   */
```
lightwalker-eth: There's another edge case to consider which is defined here: https://namehash.slack.com/archives/C08BMPFUUDD/p1765288628066849

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,139 @@
...
+            details: `The requested '${name}' name is unknown to ENSNode. No Name Tokens were indexed for it.`,
+          },
+        } satisfies NameTokensResponseErrorNameNotIndexed),
+        404,
+      );
+    }
+
+    domainId = namehash(name);
+  } else {
+    domainId = request.domainId;
```
lightwalker-eth: What `name` value is returned in this case?

apps/ensapi/src/handlers/name-tokens-api.ts
```diff
@@ -0,0 +1,150 @@
...
+const logger = makeLogger("name-tokens-api");
+
+const indexedSubregistries = getIndexedSubregistries(
+  config.namespace,
+  config.ensIndexerPublicConfig.plugins,
+);
+
+const nameWrapperAddresses = getNameWrapperAddresses(
+  config.namespace,
+  config.ensIndexerPublicConfig.plugins as PluginName[],
```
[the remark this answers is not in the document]
tk-o: Here it is:
https://github.com/namehash/ensnode/blob/81c027b022ec717867c00242ff1e0614efa5e4a6/packages/ensnode-sdk/src/ensindexer/config/types.ts#L103-L112

packages/ensnode-sdk/src/tokenscope/name-token.ts
```diff
@@ -16,9 +16,26 @@ import {
  * An enum representing the possible Name Token Ownership types.
  */
 export const NameTokenOwnershipTypes = {
+  /**
+   * Name Token is owned by NameWrapper account.
```
lightwalker-eth:
```suggestion
   * The owner of this Name Token is a NameWrapper account.
```

packages/ensnode-sdk/src/tokenscope/name-token.ts
```diff
@@ -16,9 +16,26 @@ import {
...
+   */
   NameWrapper: "namewrapper",
+
+  /**
+   * Name Token is owned fully onchain.
```
lightwalker-eth:
```suggestion
   * The owner's ultimate control over the Name Token is known to be fully onchain.
```

packages/ensnode-sdk/src/api/name-tokens/response.ts
```diff
@@ -0,0 +1,196 @@
...
+   * - When it includes more than one name token, it means that:
+   *   1) More than 1 distinct tokenized representation of the ownership of
+   *      the `name` has been indexed as of `accurateAsOf`.
+   *   2) All possible permutations of mint statuses of these tokens are
+   *      possible:
+   *      a) Multiple could be actively minted.
+   *      b) Multiple could be burned.
+   *      c) Some could be burned, others could be minted.
+   *  - Order of name tokens follows the order of onchain events that were
+   *    indexed when a token was minted, or burned.
```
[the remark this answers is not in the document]
lightwalker-eth: @tk-o The more I think about it, the more I think we should just simplify this and say that the order of tokens returned in the response is undefined and not add any sort rules to it. There's no special need or goal for sort rules here.

Issue #1381, opened 2025-12-10
ENSAnalytics: refactor `..items..` to `..records..`


Goader:
We want to align how pagination works across different APIs. Look at [this](https://github.com/namehash/ensnode/pull/1379#pullrequestreview-3562510315) for more details.

PR #1382, opened 2025-12-10
refactor(ENSAnalytics): Renaming `..items..` to `..records..`

Goader:
closes: #1381 

Renames the occurrences of `paginationContext` to `pageContext`, and `itemsPerPage` to `recordsPerPage`

### DIFF
--- a/.changeset/thick-ads-nail.md
+++ b/.changeset/thick-ads-nail.md
@@ -1,2 +1,7 @@
 ---
+"@namehash/ens-referrals": minor
+"@ensnode/ensnode-sdk": minor
+"ensapi": minor
 ---
+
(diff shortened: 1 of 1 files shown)


packages/ens-referrals/src/leaderboard-page.ts
```diff
@@ -32,7 +32,7 @@ export interface ReferrerLeaderboardPageParams {
    * @invariant Must be a positive integer (>= 1) and less than or equal to {@link REFERRERS_PER_LEADERBOARD_PAGE_MAX}
    * @default {@link REFERRERS_PER_LEADERBOARD_PAGE_DEFAULT}
```
lightwalker-eth: It would be nice if we could standardize the use of this constant too. I think no need for a distinct constant for the referrer's leaderboard page.

PR #1391, opened 2025-12-11
fix memory leakes (probably)

shrugs:
- modifies `useResolvedIdentity` to accept query options like other hooks, so enswards can mark its usage as `ASSUME_IMMUTABLE` and avoid the extra requests.
- caches public client instances by chain id for the resolution api, which should cut down on the number of `net_listening` rpc calls
- removes p-reflect because the current usage of it is incorrect and non-sensical
  1. by awaiting the swrcache.readCache within the middleware, the entire point of pReflect was defeated — every request in the app ends up running `await swrcache.readCache` which on first run must await the indexing status promise.
  2. The original purpose of this was to kick off the promise in the background and consumers could await it as necessary. By awaiting it in the middleware, we should just provide the value itself to consumers
  3. this pattern was only relevant for the indexing status middleware because it's used at the top level but _not every route_ needs synchronous access to the current indexing status: some apis should continue to function even if ensindexer is down. for example, the refer a leaderboard middleware should never have used this pattern, as the referrer leaderboard is a hard requirement for that endpoint.
  4. but because we're awaiting the first indexing status _anyway_ and delaying any initial requests until it resolves, i've just removed that feature entirely, which dramatically improves the readability of everything related to that
- SWRCache was likely the culprit for the memory issues:  each background revalidation created unresolved promises that accumulated indefinitely. The background scheduler and SWR cache were also needlessly verbose and their complexity almost certainly contributed to the existence of this bug 
  - I've rewritten the swr cache and the existing test suite continues to pass. I don't see how this implementation could have a memory leak, so it's clearly preferable. 

We haven't found a good way to test that this is actually the solution, but all of these changes seem clearly preferable, and there's no harm in making them. Because the refer leaderboard used the SWR cache, it's likely that it was the frequency of refer leaderboard requests over time that created the stack of unresolved leaderboard-building promises. This would explain the memory and CPU usage depending on how complex that function is. 

I'm also not exactly sure if this explains the public networking ingress and egress metrics that we're seeing. But the networking throughput is clearly related to the linear increase in cpu and memory over time. If Railway were incorrectly counting database.


packages/ensnode-sdk/src/shared/cache/swr-cache.ts
```diff
@@ -26,48 +22,27 @@ export interface CachedValue<ValueType> {
...
-   * considered stale but is still retained in the cache until successfully replaced with a new value.
+   * considered stale but is still retained in the cache until successfully replaced.
    */
   ttl: Duration;
 
   /**
-   * Optional time-to-proactively-revalidate duration in seconds. After this duration, automated attempts
-   * to asynchronously revalidate the cached value will be made in the background.
-   *
-   * If defined:
```
lightwalker-eth: The "If defined:" and "If undefined:" comments seem useful to keep here?

lightwalker-eth (approved):
@shrugs Looks good. Thanks for the improvements and cleanup 👍 I flagged a few places where it would be good to update the comments more based on the changes.

Approved. Please take the lead to merge when ready 👍 

@tk-o I understand you're working on this now too? Please feel welcome to take the lead on next steps.

lightwalker-eth (approved):
@shrugs Thanks. This PR looks good and is ready to merge 👍

PR #1397, opened 2025-12-12
Extract Protocol Tracing from OTel

### LINKED ISSUE #909: Separate Protocol Tracing SDK
the protocol tracing spans are annoying internally, so remove that into separate sdk with no exporting to otel collector and then our internal traces will be cleaner

shrugs:
- separates the public-facing protocol-tracing from otel (internal tracing)
- removes any possibility of a memory leak in the hacky original version
- api identical
- output is identical, so ensadmin doesn't need an update.
- 
closes #909


tk-o (approved):
LGTM, left couple of minor comments with suggestions for improving code readability 👍

lightwalker-eth (approved):
@shrugs  Looks good to me. Appreciate your advice on the open comments. Please take the lead to merge 👍

PR #1399, opened 2025-12-12
feat/add package prereleases

### LINKED ISSUE #1359: Add prerelease option to release workflows
## Problem

Currently the ENSNode monorepo uses a unified versioning approach to releases, meaning `ensnode-sdk` will only release and publish to NPM when the GitHub changeset PR is merged. A unified versioning is desired to help simplify interdependency and prevent the chaos that comes with mixed versions within packages. 

However we also would like the ability to have prerelease publishing for some of our packages for both internal and external parties. We need to find a way to introduce prerelease packages to our workflows that doesn't break the existing flow. 

## Current Setup

Main Release Process (`.github/workflows/release.yml`)

**Trigger:** Pushes to `main` branch or manual workflow dispatch

**Flow:**
1. **Changesets Job** - Uses [Changesets](https://github.com/changesets/changesets) to manage versions
   - Creates a "Release New Version" PR when changesets are detected
   - When PR is merged, publishes packages to NPM
   - Command: `pnpm changeset-publish`
   - Version bumps are determined by changeset files in `.changeset/` directory

2. **Build & Push Docker Images** 
   - Only runs if packages were published
   - Filters for "apps": `ensindexer`, `ensadmin`, `ensapi`, `ensrainbow`
   - Tags images with semantic version from changeset output
   - Pushes to `ghcr.io/namehash/ensnode/{app-name}`

3. **Create GitHub Release**
   - Extracts version from the release PR body
   - Generates release notes from PR description
   - Creates GitHub release with tag `v{version}`
   - Sends Slack notification

## Possible Solutions

>[!WARNING]
> Section still under construction as research continues. Feel free to add suggestions in the comments! 

### Option 1: Snapshot Releases

Changesets has a built-in snapshot feature which creates temporary prerelease versions without affecting the main reelease flow. 

```bash
pnpm changeset version --snapshot rc
pnpm changeset publish --tag rc
```

This creates versions like `1.1.1-rc-20231205120000` and publishes them with the `rc` dist-tag. 

[Changesets docs on snapshots](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md)

### Option 2: Prelease Mode

This would be more complicated as it requires adding a new branch where a prerelease workflow would run. Right now our release workflow runs on every commit to `main`, and in this option a prerelease workflow would run on a designated branch for each commit:

```yaml
on:
  push:
    branches:
      - main
      - next  # prerelease branch

```

Then we would use changesets' prerelease mode:

- Enter prerelease: `pnpm changeset pre enter rc`
- Create changesets and version as normal
- Exit prerelease: `pnpm changeset pre exit

In essence this would result in a workflow where all active work is on `next` and when official releases happen we would merge to `main`

stevedylandev:
This PR closes #1359 by adding the following:
- `snapshot-release.yml` - Automatically publishes `@next` packages and docker images for commits to main
- `preview-release.yml` - A manual workflow that can be run to create `@preview` packages for a branch, helpful for testing a PR before merging to `main`
- Updates `package.json` to support scripting in new workflows
- Updates docs at `ensnode.io/docs/contributing/releases` with new flow

### DIFF
--- a/docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
+++ b/docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
@@ -104,16 +104,36 @@ When using ENSNode artifacts, you have several release types to choose from.
 ## Choosing the Right Release Type
 
 ### Pinned Full Release Versions
-:::note
-In the installation examples below, `[version]` is a placeholder that must be replaced with a specific version number from the [Releases Page](https://github.com/namehash/ensnode/releases). Pinning to specific versions is critical for all published assets because the `latest` tag is a mutable pointer. Each time you pull or redeploy, you may receive a different version, introducing unexpected breaking changes or bugs into your running deployment without warning.
-:::
+
+When deploying ENSNode to production environments, it is advisable to use a Pinned Full Release. Pinned full releases are required for those who want to use any published ENSNode artifacts. By using a pinned version you can maintain full control over features or patches that might be included. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
 
 ```bash
 npm install @ensnode/[package-name]@[version]
 docker run ghcr.io/namehash/ensnode/[app-name]:[version]
 ```
 
-Pinned full releases are required for those who want to use any published ENSNode artifacts. By using a pinned version you can maintain full control over features or patches that might be included. This is specifically important for ENSNode as a new version might require updates to environment variables such as `DATABASE_SCHEMA` or change how ENSNode apps work together. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
+:::caution
+When installing NPM packages for use in production environments, it is also advisable to pin a specific Pinned Full Release version number. ENSNode patch version increments MAY include breaking changes, so the usage of exact version package specifiers is encouraged.
+
+For example:
+
+✅ `"@ensnode/[package-name]": "1.0.0"`  
+❌ `"@ensnode/[package-name]": "^1.0.0"`
+
+And when installing from the command line:
+
+✅ `pnpm install @ensnode/[package-name]@1.0.0`  
+❌ `pnpm install @ensnode/[package-name]`
+:::
+
+:::caution
+In particular, when deploying ENSNode to production environments, using the `latest` Docker tag without strict pull policies could result in a version mismatch between interdependent ENSNode apps (which will helpfully crash at startup). Due to this, we highly recommend using a specific Pinned Full Release version number tag like `1.0.0` instead of the tag `latest`, which could point to different versions of the ENSNode app depending on the platform's pull policy.
+:::
+
+:::caution
+Each ENSIndexer version update is likely to produce an updated [Ponder Build Id](https://ponder.sh/docs/api-reference/ponder/database). When updating your ENSIndexer version, you should expect to update the `DATABASE_SCHEMA` environment variable to point to a new Postgres Database Schema for a complete reindexing with the new Ponder Build Id. A complete reindexing may take over 24 hours depending on your configuration. ENSNode version updates require special coordination and should not be assumed to be a simple version bump.
+:::
+
 
 ### Snapshot Releases
 :::caution


docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -29,19 +29,16 @@ The `changesets/bot` will automatically comment on your PR to either remind you
...
+2. Click "Run workflow" and select the target branch
+3. Install preview packages with: `npm install @ensnode/package-name@preview-branch-name`
 
 **Important notes:**
-- NPM Package RC versions are ephemeral - they don't create commits or modify the repository
-- Each workflow run increments the RC version (rc.0, rc.1, rc.2, etc.)
-- Docker images are NOT built for RCs
-- No GitHub releases or tags are created for RCs
-- The `main` branch remains unchanged with regular versions
+- Preview versions are ephemeral and don't create commits or modify the repository
```
lightwalker-eth:
```suggestion
- Preview releases are ephemeral and don't create commits or modify the repository
```

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -29,19 +29,16 @@ The `changesets/bot` will automatically comment on your PR to either remind you
 
 Upon the publishing of a new release, your change will be included in the produced packages/images and your contributions will be referenced in the GitHub Release notes.
 
-## Publishing Release Candidates (RC) of packages to NPM
+## Preview Releases
 
-Publishing release candidates (RCs) of packages to NPM supports apps outside of the ENSNode monorepo to make use of prerelease packages for testing new ENSNode versions before making a full ENSNode release. Package RCs are published to NPM with the `rc` dist-tag:
+Commits to `main` automatically publish prerelease versions with the `next` dist-tag to NPM. These allow testing of changes before an official release.
```
lightwalker-eth:
```suggestion
Commits to `main` automatically publish prerelease package versions with the `next` dist-tag to NPM. These enable testing of package changes before an official release.
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      custom_suffix:
+        description: 'Custom suffix (overrides branch name)'
+        type: string
+        required: false
+      skip_npm:
+        description: 'Skip NPM publishing'
```
lightwalker-eth:
```suggestion
        description: 'Skip preview release of packages to NPM'
```

Is that fair?

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+          echo "docker-tag-base=${DOCKER_TAG_BASE}" >> $GITHUB_OUTPUT
+          echo "commit-sha=${COMMIT_SHA}" >> $GITHUB_OUTPUT
+
+          echo "🏷️  NPM dist-tag: ${DIST_TAG}"
+          echo "📸 Snapshot tag: ${SNAPSHOT_TAG}"
+          echo "🐳 Docker tag base: ${DOCKER_TAG_BASE}"
+          echo "📝 Commit SHA: ${COMMIT_SHA}"
+
+  build-and-publish-npm:
+    name: Build & Publish NPM
```
lightwalker-eth:
```suggestion
    name: Build & Publish Packages to NPM
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+        run: |
+          # Use changesets to create snapshot versions
+          # The snapshot tag will be used as part of the version string
+          pnpm changeset version --snapshot ${{ needs.validate-and-prepare.outputs.snapshot-tag }}
+
+          echo "📦 Packages versioned with snapshot tag: ${{ needs.validate-and-prepare.outputs.snapshot-tag }}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish packages with changesets
```
lightwalker-eth:
```suggestion
      - name: Publish packages to NPM with changesets
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+          pnpm changeset publish --no-git-tag --snapshot --tag ${{ needs.validate-and-prepare.outputs.dist-tag }}
+
+          # Extract published version from package.json
+          PUBLISHED_VERSION=$(node -p "require('./packages/ensnode-sdk/package.json').version")
+          echo "published-version=${PUBLISHED_VERSION}" >> $GITHUB_OUTPUT
+
+          echo "✅ Published packages with version: ${PUBLISHED_VERSION}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
lightwalker-eth: Double checking this has a purpose?

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  build-docker-images:
+    name: Build Docker Images
```
lightwalker-eth:
```suggestion
    name: Build & Publish Docker Images to GHCR
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+          pnpm add @ensnode/ens-referrals@${{ needs.validate-and-prepare.outputs.dist-tag }}
+
+          # Or install specific version
+          pnpm add @ensnode/ensnode-sdk@${{ needs.build-and-publish-npm.outputs.published-version }}
+          \`\`\`
+
+          "
+          fi
+
+          # Docker section
```
lightwalker-eth:
```suggestion
          # GHCR section
```

.github/workflows/preview-release.yml
```diff
@@ -0,0 +1,335 @@
...
+
+      - name: Log summary
+        run: |
+          echo "## 📋 Preview Release Summary"
+          echo "- Branch: ${{ github.ref_name }}"
+          echo "- Published version: ${{ needs.build-and-publish-npm.outputs.published-version }}"
+          echo "- NPM dist-tag: ${{ needs.validate-and-prepare.outputs.dist-tag }}"
+          echo "- Docker tag: ${{ needs.validate-and-prepare.outputs.docker-tag-base }}-${{ needs.validate-and-prepare.outputs.commit-sha }}"
+          echo "- NPM publish: ${{ needs.build-and-publish-npm.result }}"
+          echo "- Docker build: ${{ needs.build-docker-images.result }}"
```
lightwalker-eth:
```suggestion
          echo "- GHCR publish: ${{ needs.build-docker-images.result }}"
```

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -29,19 +29,16 @@ The `changesets/bot` will automatically comment on your PR to either remind you
...
 
-1. Ensure your changesets are committed to `main`
-2. Navigate to [Actions > Release NPM RC](https://github.com/namehash/ensnode/actions/workflows/npm-release-rc.yml)
-3. Click "Run workflow" and select the `main` branch
-4. RC packages are published to NPM with the `rc` tag (e.g., `1.0.0-rc.0`)
-5. Install RCs with: `npm install @ensnode/package-name@rc`
+To manually trigger a preview release for a branch that has not been merged to `main`, follow these steps:
+
+1. Navigate to [Actions > Release Preview](https://github.com/namehash/ensnode/actions/workflows/preview-release.yml)
```
lightwalker-eth: Is this "Release Preview" or "Preview Release"?

Since the context is an action, it seems to me that it should always begin with a verb. Ex: "Release Preview" or "Release Snapshot", etc.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
lightwalker-eth: Suggest to document here a link to the specific GitHub Action file that automatically triggers this upon each commit to main.

.github/workflows/snapshot-release.yml
```diff
@@ -0,0 +1,97 @@
+name: Snapshot Release (@next)
+
+on:
```
lightwalker-eth: What about the case that the branch being merged into main is one of our release PRs? In this case would we create both an "official" release AND a "snapshot" release?

It seems that ideally we wouldn't create both an "official" and a "snapshot" release from the same commit?

Appreciate your advice.

.github/workflows/snapshot-release.yml
```diff
@@ -0,0 +1,97 @@
+name: Snapshot Release (@next)
```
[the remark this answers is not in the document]
stevedylandev: At first I agreed that it could be simpler to merge these two workflows, but after some experimenting I think it actually makes it more complicated. Since we're using changesets the existing `release.yml` workflow either:

A. Opens a Release PR via the GitHub bot for a potential new release
B. Publishes a new release if the aforementioned Release PR is merged into `main`

All of this logic is encased inside the changesets GitHub action and bot, so it becomes harder to build on top of it. At lease for now I think it will be best to keep these workflows separate. 

I did remove the `workflow_dispatch` as you are correct we no longer need it. I will be reviewing the entire workflow a bit more closely and ensure a cohesive flow!

.github/actions/setup-node-environment/action.yml
```diff
@@ -0,0 +1,26 @@
+name: 'Setup Node Environment'
```
shrugs: this folder should either use `snake_case` or all action folders should switch to `kebab-case` — up to you

.github/workflows/snapshot_release.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Build & Push
+        uses: ./.github/actions/build_docker_image
+        with:
+          image: ghcr.io/${{ github.repository }}/${{ matrix.app }}
+          dockerfile: apps/${{ matrix.app }}/Dockerfile
+          registry_user: ${{ github.actor }}
+          registry_token: ${{ secrets.GITHUB_TOKEN }}
+          tags: |
+            type=raw,value=next
+            type=raw,value=${{ steps.snapshot-tag.outputs.tag }}
```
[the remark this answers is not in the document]
stevedylandev: Good catch!

.github/actions/setup_node_environment/action.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Setup Node Environment'
```
lightwalker-eth: Maybe this script is more about something like "Install ENSNode"?

.github/workflows/preview_release.yml
```diff
@@ -0,0 +1,337 @@
...
+          # --no-git-tag: Don't create git tags for preview releases
+          # --snapshot: Publish as snapshot/preview release
+          # --tag: Use our custom dist-tag
+          pnpm changeset publish --no-git-tag --snapshot --tag ${{ needs.validate-and-prepare.outputs.dist-tag }}
+
+          # Extract published version from package.json
+          PUBLISHED_VERSION=$(node -p "require('./packages/ensnode-sdk/package.json').version")
+          echo "published-version=${PUBLISHED_VERSION}" >> $GITHUB_OUTPUT
+
+          echo "✅ Published packages to NPM with version: ${PUBLISHED_VERSION}"
```
lightwalker-eth:
```suggestion
          echo "✅ Published release preview packages to NPM with version: ${PUBLISHED_VERSION}"
```

.github/workflows/snapshot_release.yml
```diff
@@ -0,0 +1,87 @@
+name: Snapshot Release (@next)
```
lightwalker-eth: I believe all these workflows should be named such that they begin with a verb. Ex: "Release Snapshot" and the file name updated accordingly.

lightwalker-eth: I think it will be helpful to add some comments up at the top of this file that:
1. Gives a quick summary of what this GitHub action is for.
2. Gives a link to the relevant docs (when forming this link, assume that this PR is already published and that the new releases.mdx file is already available and therefore an anchor link to the relevant section is possible.

Can you please apply this idea across the other release workflows as is relevant? Thanks

.github/workflows/snapshot_release.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Publish snapshot
+        if: steps.snapshot.outputs.hasChanges == 'true'
+        run: pnpm changeset-publish:next
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  build-and-publish-ghcr:
+    name: Build & Publish Snapshot Docker Imagess to GHCR
```
lightwalker-eth:
```suggestion
    name: Build & Publish Snapshot Docker Images to GHCR
```

lightwalker-eth (commented):
@stevedylandev Hey great updates. Reviewed and shared some questions and suggestions. Appreciate your advice 👍

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -35,21 +35,21 @@ There are curently two separate flows for previewing a release: Release Snapshot
 
 ### Release Snapshot
 
-Each commit to `main` will trigger the `snapshot-release.yml` workflow to build and publish packages to NPM and Docker images to the GitHub Container Registry (ghcr). These public releases will be under the tag `@next`, allowing anyone who wants to test a package before the stable release. To use snapshot releases run `pnpm install @ensnode/package-name@next` or `docker run ghcr.io/namehash/ensnode:next`. If the `@latest` tag is used then it will only install or run the latest stable release.
+Each commit to `main` will trigger the `snapshot-release.yml` workflow to build and publish packages to NPM and Docker images to the GitHub Container Registry (ghcr). These public releases will be under the tag `@next`, allowing anyone to use the npm packages or docker images associated with each commit to main. To install snapshot releases run `pnpm install @ensnode/package-name@next` or `docker run ghcr.io/namehash/ensnode:next`. If the `@latest` tag is used then it will only install or run the latest stable release.
```
lightwalker-eth: Please also review and update all references to these files after they are renamed.

lightwalker-eth: The last sentence here about the @latest tag feels like it belongs in a different section?

In my mental model, when we write these docs we're writing not only for the creator of releases, but also the consumer of releases.

My question here comes from a desire to support the consumer of releases. Will it be clear for them how to decide what is the right release to use? If you think this is already clear than all good. Just posing the question.

.github/actions/setup_node_environment/action.yml
```diff
@@ -0,0 +1,16 @@
...
+description: 'Sets up pnpm, Node.js, and optionally installs dependencies'
```
lightwalker-eth: I previously shared a question about why "optionally" here that I see was marked as resolved but I still see the word "optionally" here.

From reading this script, how is it optional that it installs dependencies? It seems to happen each time it runs.

lightwalker-eth (commented):
@stevedylandev Reviewed and shared feedback

.github/workflows/preview_release.yml
```diff
@@ -1,14 +1,14 @@
-name: Preview Release
+name: Release Preview (@preview)
```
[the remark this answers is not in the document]
stevedylandev: Agreed! I made a point to do this before merging. In the last review I actually updated the filenames, but doing so deleted all of the unresolved comments and made it much harder to work through. Because of this I reverted that commit and will be implementing once everything else is good to go.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -35,21 +35,21 @@ There are curently two separate flows for previewing a release: Release Snapshot
...
 2. Click "Run workflow" and select from the following options:
- - Which branch you would like to run the preview release workflow on, ideally the branch for an open PR.
+ - Which branch you would like to run the preview release workflow on, ideally the branch for an open PR. The workflow needs a target branch to build and publish for.
  - Choose whether to build and publish only NPM packages, or to build and publish both NPM packages and Docker images to the GitHub Container Registry (ghcr).
- - (Optional) Provide a custom suffix for the preview release package tag. For example, using `my-feature` would result in a tag name of `@ensnode/package-name@preview-my-feature`.
+ - (Optional) Provide a custom suffix for the preview release package tag. For example, if you had a branch called `feat/add-api-route` and left this custom suffix field blank, the preview release would be `@ensnode/package-name@preview-feat-add-api-route`. If you fill in the custom suffix field with `users-route` then the resulting package name would be `@ensnode/package-name@preview-users-route`.
```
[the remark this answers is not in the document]
stevedylandev: Good catch; it should be tag. Fixed!

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -35,21 +35,21 @@ There are curently two separate flows for previewing a release: Release Snapshot
...
 
-### Release Preveiw
+### Release Preview
 
 To manually trigger a preview release for a branch that has not been merged to `main`, follow these steps:
 
 1. Navigate to [Actions > Release Preview](https://github.com/namehash/ensnode/actions/workflows/preview-release.yml)
 2. Click "Run workflow" and select from the following options:
- - Which branch you would like to run the preview release workflow on, ideally the branch for an open PR.
+ - Which branch you would like to run the preview release workflow on, ideally the branch for an open PR. The workflow needs a target branch to build and publish for.
```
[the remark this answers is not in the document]
stevedylandev: Updated to the following: 

```diff
- - Which branch you would like to run the preview release workflow on, ideally the branch for an open PR. The workflow needs a target branch to build and publish for.
+ - Which branch you would like to run the preview release workflow on. You can choose any branch and it will trigger a preview release for it, however it is recommended to use the branch where there is an active PR so the workflow can add a GitHub comment to the PR. The GitHub comment will contain helpful installation info such as the generated release tag, which would otherwise have to be located manually through the action logs or the publishing registries.
```

> Why is it ideal that the selected branch has an open PR?

Updated reason: You can choose any branch, but it's most beneficial to select one with an open PR so the GitHub action can leave a comment with helpful tag information.

> What happens if I select a branch that doesn't have an open PR.

This should not raise an alarm anymore because of the change in language. "You can choose any branch" as it will still publish a preview release of the target branch, but it might be harder to locate the published tag name unless you visit the npm registry or ghcr.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -25,23 +25,120 @@ The `changesets/bot` will automatically comment on your PR to either remind you
...
+**What to expect:**
+- Experimental and potentially unstable
+- Built from active development branches
+- May not be compatible with other packages
+
+## Version Selection Best Practices
+
+- Check the GitHub releases page for the latest stable version and release notes
+- Pin specific versions in your integrations rather than using tags like `@latest`
+- Use full releases (`@latest`) as much as possible for stability
```
lightwalker-eth: I see this bullet and the one above. They touch on the idea in my new comment up above but they do not sufficiently bring attention to this key issue.

lightwalker-eth (approved):
@stevedylandev Thanks for the updates. One more key remaining issue. Please take the lead to merge when ready 👍

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -25,23 +25,120 @@ The `changesets/bot` will automatically comment on your PR to either remind you
...
+
+## Where to Find Releases
+
+- **NPM Packages**: Available on the [npm registry](https://www.npmjs.com/search?q=%40ensnode) under the `@ensnode` organization
+- **Docker Images**: Available on [GitHub Container Registry](https://github.com/orgs/namehash/packages?repo_name=ensnode)
+- **GitHub Releases**: Full releases are documented with release notes on the [ENSNode GitHub releases page](https://github.com/namehash/ensnode/releases)
+
+## Choosing the Right Release Type
+
+### For Production Use: `@latest` (Full Releases)
```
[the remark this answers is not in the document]
shrugs: yeah we should encourage people to use a specific version not the `latest` tag itself — we're not abiding by real semver so minors can have breaking changes. if it's trivial, it might even be best to remove the `latest` tag entirely, to avoid anyone doing that?

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -90,19 +89,22 @@ When installing ENSNode NPM packages or Docker images, you have several release
...
+- When you need maximum stability and control over updates
 
 **What to expect:**
-- Stable, well-tested releases
-- Complete release notes and changelog entries
+- Stable, well-tested releases with complete release notes
+- The version will never change - you control when to upgrade
+- No unexpected breaking changes or behavior changes
+
+**Important:** Never use the `latest` tag for Docker deployments (e.g., `ghcr.io/namehash/ensnode:latest` or `ghcr.io/namehash/ensnode`). The `latest` tag is a mutable pointer that will automatically pull different versions over time, causing unexpected updates and potential breaking changes in your deployment. Always pin to a specific version number from the [GitHub releases page](https://github.com/namehash/ensnode/releases).
```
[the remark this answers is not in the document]
stevedylandev: To clarify, are you asking to remove any mention of `latest` for Docker images in the docs? Or are you suggesting opening another issue to remove `latest` as a possibility for Docker image publishing?

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -7,41 +7,127 @@ sidebar:
...
+- **Lambda Zip Artifacts**: Available in the Artifact section of a successful workflow run
+
+## Choosing the Right Release Type
+
+### Pinned Full Release Versions
+:::note
+The example version below of `1.0.0` is a placeholder. Visit the [Releases Page](https://github.com/namehash/ensnode/releases) to pick a pinned full release.
+:::
+```bash
+npm install @ensnode/[package-name]@1.0.0
```
stevedylandev: Switched to `[package-name]` and `ensnode/[app-name]` per suggestion from @shrugs

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -7,41 +7,127 @@ sidebar:
...
+# Selecting a Release for Deployment or Installation
+
+When using ENSNode artifacts, you have several release types to choose from.
+
+## Where to Find Releases
+
+- **NPM Packages**: Available on the [npm registry](https://www.npmjs.com/search?q=%40ensnode) under the `@ensnode` organization
+- **Docker Images**: Available on [GitHub Container Registry](https://github.com/orgs/namehash/packages?repo_name=ensnode)
+- **GitHub Releases**: Full releases are documented with release notes on the [ENSNode GitHub releases page](https://github.com/namehash/ensnode/releases)
+- **Lambda Zip Artifacts**: Available in the Artifact section of a successful workflow run
```
stevedylandev: @shrugs @lightwalker-eth open to suggestions on a better way to link/reference Lambda zips. As far as I know they can only be seen in successful workflow runs, e.g. https://github.com/namehash/ensnode/actions/runs/20581435373

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -90,19 +89,22 @@ When installing ENSNode NPM packages or Docker images, you have several release
...
-- When you need maximum stability and reliability
-- When you want thoroughly tested features
+- Production deployments and any critical infrastructure
+- When you need maximum stability and control over updates
 
 **What to expect:**
-- Stable, well-tested releases
-- Complete release notes and changelog entries
+- Stable, well-tested releases with complete release notes
+- The version will never change - you control when to upgrade
```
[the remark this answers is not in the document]
stevedylandev: Updated

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -90,19 +89,22 @@ When installing ENSNode NPM packages or Docker images, you have several release
...
 **When to use:**
-- When you need maximum stability and reliability
-- When you want thoroughly tested features
+- Production deployments and any critical infrastructure
+- When you need maximum stability and control over updates
 
 **What to expect:**
-- Stable, well-tested releases
-- Complete release notes and changelog entries
+- Stable, well-tested releases with complete release notes
```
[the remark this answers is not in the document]
stevedylandev: Updated

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -7,33 +7,33 @@ sidebar:
...
 
 **Important notes:**
-- Changesets are only required for user-facing changes (features, bug fixes)
-- You don't need a changeset for documentation changes or internal refactoring
-- All our packages use "fixed" versioning - they all share the same version number regardless of which package triggered the version bump
+- Changesets are only required for user-facing changes (features, bug fixes).
+- You don't need a changeset for documentation changes or internal refactoring.
+- All our packages use "fixed" versioning - they all share the same version number regardless of which package triggered the version bump.
 
 # Creating a Release
```
lightwalker-eth: Is there a special reason why we should retain the following workflow after the updates being made in this PR? If there's no special benefit to retaining it, I would prefer to remove it.

`.github/workflows/build_ensnode.yml`

@stevedylandev @shrugs Appreciate your advice.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
- - Choose whether to build and publish only NPM packages, or to build and publish both NPM packages and Docker images to the GitHub Container Registry (ghcr).
- - (Optional) Provide a custom suffix for the preview release package tag. For example, if you had a branch called `feat/add-api-route` and left this custom suffix field blank, the preview release would be `@ensnode/package-name@preview-feat-add-api-route`. If you fill in the custom suffix field with `users-route` then the resulting tag name would be `@ensnode/package-name@preview-users-route`.
-3. Install preview packages with: `npm install @ensnode/package-name@preview-branch-name`. 
+ - The branch on which to run the preview release workflow. Any branch other than `main` can be selected. Using a branch with an active PR allows the workflow to add a GitHub comment containing the installation information. If multiple preview releases are triggered for the same brach/PR, the GitHub comment will update with the latest release.
```
lightwalker-eth:
```suggestion
 - The branch on which to run the preview release workflow. Any branch other than `main` can be selected. Using a branch with an active PR allows the workflow to add a GitHub comment containing the installation information. If multiple preview releases are triggered for the same branch/PR, the GitHub comment will update with the latest release.
```

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
lightwalker-eth: Suggested actions:

1. Update the docs to help people find these "Actions Artifacts". To help explain this, suggest to link to the docs here: https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go
2. Update the docs to note how these "Actions Artifacts" are only retained for a max of 90 days, including a link to https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period for more info

I believe we should update our "Full" release workflow so that it publishes the Lambda .zip in a fully durable way that isn't automatically deleted by GitHub after 90 days.

My understanding of the best practice for this is to include the Lambda .zip as a "Release Artifact" such that our "Full" release workflow would be storing the associated Lambda .zip as an additional file in each GitHub Release as shown here:

CleanShot 2025-12-31 at 12 57 33

It also seems attractive to update our release scripts to automatically publish the associated Lambdas as a .zip file to a specific S3 bucket where it automatically names the .zip file to include details such as:
- Is it associated with a preview, a snapshot, or a full release?
- What is the version number / tag associated with the release?

In this way, whenever we want to update the Lambda that's actually running in AWS, we don't need to worry at all about manually building / uploading to S3. Instead we just update the Terraform scripts to reference the desired pre-published lambda .zip file in the target S3 bucket.

@shrugs @jarmint @stevedylandev advice appreciated on what our next steps should be here. Thanks

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
lightwalker-eth: Is there a special reason for Preview Releases to support the case of a branch without a PR?

That seems like an undesirable source of complexity in people's mental models and in operational processes.

Suggest to:
1. Update the Preview Release workflow such that it aborts if the associated branch does not have a PR associated with it.
2. Simplify the docs for Preview Releases such that all complexity about there maybe not being a PR and therefore maybe not having a PR comment, etc.. is removed. Instead we can write about these ideas with black and white certainty: Each Preview Release is associated with a PR, and the PR will have a PR comment.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
 
-### For Production Deployments: Pinned Full Release Versions
+### Pinned Full Release Versions
+:::note
+The example version below of `1.0.0` is a placeholder. Visit the [Releases Page](https://github.com/namehash/ensnode/releases) to pick a pinned full release.
+:::
 ```bash
-npm install @ensnode/package-name@1
-docker run ghcr.io/namehash/ensnode:1
+npm install @ensnode/[package-name]@1.0.0
```
lightwalker-eth: For NPM packages, I believe it's fine to use @latest or to remove any explicit version number is probably better.

In theory, the issue with use of "latest" doesn't impact NPM packages. It impacts Docker images.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -7,33 +7,33 @@ sidebar:
...
 
 The `changesets/bot` will automatically comment on your PR to either remind you to add a changeset or confirm the version changes that will happen when your PR is merged.
 
 **Important notes:**
-- Changesets are only required for user-facing changes (features, bug fixes)
-- You don't need a changeset for documentation changes or internal refactoring
-- All our packages use "fixed" versioning - they all share the same version number regardless of which package triggered the version bump
+- Changesets are only required for user-facing changes (features, bug fixes).
+- You don't need a changeset for documentation changes or internal refactoring.
+- All our packages use "fixed" versioning - they all share the same version number regardless of which package triggered the version bump.
```
lightwalker-eth:
```suggestion
- All our apps and packages (with an exception as noted below) use "fixed" versioning - they all share the same version number regardless of which app or package triggered the version bump.
- An exception to the above "fixed" versioning is the "fallback-ensapi" app. This is a Lambda containing logic specific to NameHash deployments of ENSNode and is versioned independently.
```

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -7,33 +7,33 @@ sidebar:
...
 
 Workflow File: [release.yml](https://github.com/namehash/ensnode/blob/main/.github/workflows/release.yml)
 
-If your PR includes a changeset and is merged to `main` then it will automatically be added to a new automated Release PR by the Changesets bot. As more changesets are added to `main` the Release PR will continue to update. Once the Release PR is merged into `main` then it will trigger a stable release with all of the included changesets. It will also create a release on GitHub with autogenerated release notes. When a full release runs it will build and publish packages to NPM and Docker images to GitHub Container Registry (ghcr).
+If your PR includes a changeset and is merged to `main` then it will automatically be added to a new automated Release PR by the Changesets bot. As more changesets are added to `main` the Release PR will continue to update. Once the Release PR is merged into `main` then it will trigger a stable release with all of the included changesets. It will also create a release on GitHub with autogenerated release notes. When a full release runs it will build and publish all of the monorepo's artifacts (NPM packages, Docker images, and Lambda zips).
```
lightwalker-eth:
```suggestion
If your PR includes a changeset and is merged to `main` then it will automatically be added to a new automated Release PR by the Changesets bot. As more changesets are added to `main` the Release PR will continue to update. Once a Release PR is merged into `main` it triggers a "full" release that will:

- Build and publish all of the monorepo's artifacts (NPM packages, Docker images, and Lambda zips).
- Create a release on GitHub with autogenerated release notes from all the included changesets.
```

lightwalker-eth (commented):
@stevedylandev Reviewed and shared feedback.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
- - Choose whether to build and publish only NPM packages, or to build and publish both NPM packages and Docker images to the GitHub Container Registry (ghcr).
- - (Optional) Provide a custom suffix for the preview release package tag. For example, if you had a branch called `feat/add-api-route` and left this custom suffix field blank, the preview release would be `@ensnode/package-name@preview-feat-add-api-route`. If you fill in the custom suffix field with `users-route` then the resulting tag name would be `@ensnode/package-name@preview-users-route`.
-3. Install preview packages with: `npm install @ensnode/package-name@preview-branch-name`. 
+ - The branch on which to run the preview release workflow. Any branch other than `main` can be selected. Using a branch with an active PR allows the workflow to add a GitHub comment containing the installation information. If multiple preview releases are triggered for the same brach/PR, the GitHub comment will update with the latest release.
+ - Choose which artifacts to build and publish (NPM packages, Docker images, Lambda zips, or any combination).
```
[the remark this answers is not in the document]
stevedylandev: Good point. I've updated `release_preview.yml` to include more options for selecting what to publish, as well as updating the docs to make it simplified. 

https://github.com/namehash/ensnode/pull/1399/commits/601872352832db26bcd54f51325928de7da0bcb6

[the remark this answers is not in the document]
stevedylandev: [See other comment](https://github.com/namehash/ensnode/pull/1399/files#r2655571593)

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
 - **Docker Images**: Available on [GitHub Container Registry](https://github.com/orgs/namehash/packages?repo_name=ensnode)
 - **GitHub Releases**: Full releases are documented with release notes on the [ENSNode GitHub releases page](https://github.com/namehash/ensnode/releases)
+- **Lambda Zip Artifacts**: Available in the Artifact section of a successful workflow run
 
 ## Choosing the Right Release Type
 
-### For Production Deployments: Pinned Full Release Versions
+### Pinned Full Release Versions
+:::note
+The example version below of `1.0.0` is a placeholder. Visit the [Releases Page](https://github.com/namehash/ensnode/releases) to pick a pinned full release.
```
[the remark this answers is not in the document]
stevedylandev: Updated

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
stevedylandev: Agreed; updated workflow and documentation to require a PR for preview releases. https://github.com/namehash/ensnode/pull/1399/commits/16775e2bfbaca8f38113cac9402fe126f836ebaf

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
shrugs: confirming that with the proposed release approach the `build_ensnode.yml` workflow is vestigal. By the way, there's an annoying feature in GitHub, which is that they keep the workflow showing even though it's not in main anymore so that you can access previous history.

Options:
1. rename this to `[Deprecated] Build ENSNode` or something and _then_ remove it, in a followup pr so that the item in this list isn't distracting.
image

2. delete the previous runs of that workflow which will remove it from the list https://github.com/orgs/community/discussions/26256#discussioncomment-13849174

i'd suggest option 1 for simplicity, and the perhaps obscure possibility that deleting the workflow would affect artifact availability for those previous versions

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -60,52 +64,58 @@ Each commit to `main` will automatically trigger the `release_snapshot.yml` work
...
 ### Pinned Full Release Versions
 :::note
-The example version below of `1.0.0` is a placeholder. Visit the [Releases Page](https://github.com/namehash/ensnode/releases) to pick a pinned full release.
+In the Docker example below, `[version]` is a placeholder that must be replaced with a specific version number from the [Releases Page](https://github.com/namehash/ensnode/releases). Pinning to specific versions is critical for Docker images because the `latest` tag is a mutable pointer. Each time you pull or redeploy, you may receive a different version, introducing unexpected breaking changes or bugs into your running deployment without warning.
 :::
+
 ```bash
-npm install @ensnode/[package-name]@1.0.0
-docker run ghcr.io/namehash/ensnode/[app-name]:1.0.0
+npm install @ensnode/[package-name]@latest
```
lightwalker-eth: I think Matt was correct to flag that we should give the specific version info guidance not only for Docker containers, but for all artifacts including NPM packages.

Therefore we should avoid any docs that reference the use of @latest for anything and replace all with a more abstract idea of "[version]"

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -60,52 +64,58 @@ Each commit to `main` will automatically trigger the `release_snapshot.yml` work
...
+- **GitHub Releases**: Full releases are documented with release notes on the [ENSNode GitHub releases page](https://github.com/namehash/ensnode/releases).
+- **Lambda Zip Artifacts**: Available in the [Artifact section](https://github.com/actions/upload-artifact?tab=readme-ov-file#where-does-the-upload-go) of a successful workflow run. These Action Artifacts are [retained for 90 days](https://github.com/actions/upload-artifact?tab=readme-ov-file#retention-period).
 
 ## Choosing the Right Release Type
 
 ### Pinned Full Release Versions
 :::note
-The example version below of `1.0.0` is a placeholder. Visit the [Releases Page](https://github.com/namehash/ensnode/releases) to pick a pinned full release.
+In the Docker example below, `[version]` is a placeholder that must be replaced with a specific version number from the [Releases Page](https://github.com/namehash/ensnode/releases). Pinning to specific versions is critical for Docker images because the `latest` tag is a mutable pointer. Each time you pull or redeploy, you may receive a different version, introducing unexpected breaking changes or bugs into your running deployment without warning.
```
lightwalker-eth:
```suggestion
In the installation examples below, `[version]` is a placeholder that must be replaced with a specific version number from the [Releases Page](https://github.com/namehash/ensnode/releases). Pinning to specific versions is critical for all published assets because the `latest` tag is a mutable pointer. Each time you pull or redeploy, you may receive a different version, introducing unexpected breaking changes or bugs into your running deployment without warning.
```

Goal: See related comment below.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -60,52 +64,58 @@ Each commit to `main` will automatically trigger the `release_snapshot.yml` work
...
+
 ```bash
-npm install @ensnode/[package-name]@1.0.0
-docker run ghcr.io/namehash/ensnode/[app-name]:1.0.0
+npm install @ensnode/[package-name]@latest
+docker run ghcr.io/namehash/ensnode/[app-name]:[version]
 ```
 
-Pinned full releases are recommended for those who want to start using ENSNode artifacts. By using a pinned version you can maintain full control over features or patches that might be included. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
+Pinned full releases are required for those who want to use ENSNode Docker Images. By using a pinned version you can maintain full control over features or patches that might be included. This is specifically important for ENSNode as a new version might require updates to environment variables such as `DATABASE_SCHEMA` or change how ENSNode apps work together. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
```
lightwalker-eth:
```suggestion
Pinned full releases are required for those who want to use any published ENSNode artifacts. By using a pinned version you can maintain full control over features or patches that might be included. This is specifically important for ENSNode as a new version might require updates to environment variables such as `DATABASE_SCHEMA` or change how ENSNode apps work together. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
```

Goal: See related comment above.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -60,52 +64,58 @@ Each commit to `main` will automatically trigger the `release_snapshot.yml` work
...
 
+### Snapshot Releases
 :::caution
-Never use the `latest` tag for Docker deployments (e.g., `ghcr.io/namehash/ensnode:latest` or `ghcr.io/namehash/ensnode`). The `latest` tag is a mutable pointer that will automatically pull different versions over time, causing unexpected updates and potential breaking changes in your deployment. Always pin to a specific version number from the [GitHub releases page](https://github.com/namehash/ensnode/releases).
+The `next` tag is a floating pointer that always references the most recent snapshot release. When using Docker images with the `next` tag, you must run `docker pull` to update your local Docker cache if you want to get the actual latest image. Without pulling, you may continue using an older cached version.
```
lightwalker-eth: Please see my comment below. I'm concerned about how we're currently describing use of the `next` tag. Advice appreciated.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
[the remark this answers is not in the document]
lightwalker-eth: @stevedylandev We should also make a high-visibility callout in our releases docs for how ENSNode is not currently using Semantic Versioning and therefore even patch or minor releases may include breaking changes.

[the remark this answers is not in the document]
stevedylandev: @lightwalker-eth Let me now if this works 67dfba89

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
 
-**Important:** Never use the `latest` tag for Docker deployments (e.g., `ghcr.io/namehash/ensnode:latest` or `ghcr.io/namehash/ensnode`). The `latest` tag is a mutable pointer that will automatically pull different versions over time, causing unexpected updates and potential breaking changes in your deployment. Always pin to a specific version number from the [GitHub releases page](https://github.com/namehash/ensnode/releases).
-
-### For Development/Testing: `@next` (Snapshot releases)
+### Snapshot Releases
 ```bash
-npm install @ensnode/package-name@next  
-docker run ghcr.io/namehash/ensnode:next
+npm install @ensnode/[package-name]@next  
+docker run ghcr.io/namehash/ensnode/[app-name]:next
```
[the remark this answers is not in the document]
stevedylandev: Doesn't our framing around snapshot releases help clarify this point? 

> Snapshot releases should be used by those who need to test features or patches before they are included in full releases. These releases follow the `main` branch and are not referenced in the GitHub Releases page. Instead they are installed by using the `next` tag for published artifacts. 

Isn't the big picture is that someone might install a snapshot release to test the behavior on the `main` branch, not necessary build upon it further?

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
-- Production deployments and any critical infrastructure
-- When you need maximum stability and control over updates
+Pinned full releases are recommended for those who want to start using ENSNode artifacts. By using a pinned version you can maintain full control over features or patches that might be included. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
 
-**What to expect:**
-- Stable, well-tested releases with complete release notes
-- The version will never change - you control when to upgrade
-- No unexpected breaking changes or behavior changes
+:::caution
+Never use the `latest` tag for Docker deployments (e.g., `ghcr.io/namehash/ensnode:latest` or `ghcr.io/namehash/ensnode`). The `latest` tag is a mutable pointer that will automatically pull different versions over time, causing unexpected updates and potential breaking changes in your deployment. Always pin to a specific version number from the [GitHub releases page](https://github.com/namehash/ensnode/releases).
```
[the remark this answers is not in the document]
stevedylandev: In this current PR we are documenting and instructing users not to use the `latest` tag. 

It is possible to completely remove the `latest` tag for future ghcr publishing using the [flavor input](https://github.com/actions-marketplace-validations/docker_metadata-action?tab=readme-ov-file#flavor-input) set to `latest=false`. However this introduces a big problem. Since all of the ENSNode apps already have releases where this option was not set, there is no way to retroactively go back and remove those tags. It means a user who tries to run `docker run ghcr.io/namehash/ensnode/[app-name]:latest` will install the last version that used the `latest` tag.

For example, if we published `ensindexer` with the next patch version `1.3.2` and we have the workflow use `latest=false`, that new version will not have the `latest` tag. However the previous version `1.3.1` will still have the `latest` tag. 

I personally would recommend not taking this path and stick with documentation and instruction that pinned versions should be used instead of the `latest` tag.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
```diff
@@ -46,101 +46,88 @@ If your PR includes a changeset and is merged to `main` then it will automatical
...
-npm install @ensnode/package-name@1
-docker run ghcr.io/namehash/ensnode:1
+npm install @ensnode/[package-name]@1.0.0
+docker run ghcr.io/namehash/ensnode/[app-name]:1.0.0
 ```
 
-**When to use:**
-- Production deployments and any critical infrastructure
-- When you need maximum stability and control over updates
+Pinned full releases are recommended for those who want to start using ENSNode artifacts. By using a pinned version you can maintain full control over features or patches that might be included. Review the release notes on the [Releases Page](https://github.com/namehash/ensnode/releases) to help decide which version to install.
```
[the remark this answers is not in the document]
stevedylandev: Thank you both! Updated: 7e778f8a

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
jarmint: this `:next` tag seems redundant for me and just adds uncertainty. 

the `:latest` tag has it's own use for POCs / Demos / development / test environments / CI builds for pulling cache faster. But as mentioned in already placed document: should never be used for running in production.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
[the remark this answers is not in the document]
jarmint: who are primary target users of those snapshot releases? 

**also RE: manual workflow for PR builds:** i'd suggest to make snapshot releases automatic on every commit on already opened PR instead of manual action.

docs/ensnode.io/src/content/docs/docs/contributing/releases.mdx
lightwalker-eth: @jarmint Let's move this discussion into Slack.

Issue #908, opened 2025-08-11
Built Ponder Client and rename ponder-metadata to ponder-sdk


lightwalker-eth:
This logic now exclusively lives inside ENSIndexer. We shouldn't make an effort for this logic to be usable more broadly outside of ENSIndexer.

tk-o:
@lightwalker-eth I think we can keep the `ponder-metadata` package but turn it private and stop publishing it.

There are couple of reasons for keeping the package:
1. Isolate dependencies for working with, for example, Prometheus metrics.
2. Keep consistent codebase patterns
 - Just like we have `ponder-subgraph` package for managing Subgraph API integration with Ponder. We could also have a separate package for defining Ponder-related ideas in one place.
    - Involves moving https://github.com/namehash/ensnode/tree/v0.35.0/apps/ensindexer/src/api/lib/indexing-status/ponder-metadata into https://github.com/namehash/ensnode/tree/v0.35.0/packages/ponder-metadata

PR #1338, opened 2025-12-02
feat(ensapi): add openapi

stevedylandev:
## Summary

This PR starts implementation of [`hono-openapi`](https://hono.dev/examples/hono-openapi) to complete #1115. Work includes:

- Installing `hono-openapi`
- Updates `validate.ts` to use `validator` from `hono-openapi`
- Adds response zod schemas to ensnode-sdk
- Adds `describeRoute` to resolution endpoints
- Adds `GET /openapi.json` route

The result is an OpenAPI schema that can be fetched from `/openapi.json` from the ENSApi:

```bash
curl http://localhost:4334/openapi.json | jq . 
```

With the schema we can build new documentation that accepts OpenAPI for automatic API references. At least for the API we can write almost all of the documentation inside the routes themselves. Here is an example: 

```ts
app.get(
  "/records/:name",
  describeRoute({
    summary: "Resolve ENS Records",
    description: `Resolves ENS records for a given name, including addresses, text records, and other metadata.

## Query Parameters

- \`addresses\` - Comma-separated list of coin type IDs (e.g., \`60\` for ETH, \`0\` for BTC)
- \`texts\` - Comma-separated list of text record keys (e.g., \`avatar\`, \`com.twitter\`)
- \`name\` - Set to \`true\` to include name resolution
- \`trace\` - Set to \`true\` to include resolution trace information
- \`accelerate\` - Set to \`true\` to attempt real-time acceleration

## Examples

### Resolve address records (ETH and BTC)
\`\`\`
GET /records/example.eth?addresses=60,0
\`\`\`

### Resolve text records (avatar and Twitter)
\`\`\`
GET /records/example.eth?texts=avatar,com.twitter
\`\`\`

### Combined resolution
\`\`\`
GET /records/example.eth?name=true&addresses=60,0&texts=avatar,com.twitter
\`\`\``,
    responses: {
      200: {
        description: "Successfully resolved records",
        content: {
          "application/json": {
            schema: resolver(makeResolveRecordsResponseSchema()),
          },
        },
      },
    },
  }),
```

>[!NOTE]
>I don't like escaping characters in the mark up so I will be looking into another way to do this

This results in a docs page like this: 

CleanShot 2025-12-02 at 10 08 51@2x

Take a look at the current preview built in a separate repo: https://docs-sepia-eight.vercel.app/docs/getApiResolveRecords

## Testing

[...]

Issue #1234, opened 2025-10-31
Further consolidate Ponder-specific logic in ENSIndexer


lightwalker-eth:
# Background

Great progress was made towards this goal in https://github.com/namehash/ensnode/pull/1226

# Goals

Be able to document how all of our Ponder-specific logic is found exclusively in a finite set of places.

# Ideas

Shall we also move the following under `apps/ensindexer/ponder` or otherwise consolidate them together under `apps/ensindexer/src/lib/ponder`?
- [apps/ensindexer/src/ponder/config.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/ponder/config.ts)
- [apps/ensindexer/src/lib/ponder-helpers.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/ponder-helpers.ts)
- [apps/ensindexer/src/lib/ponder-helpers.test.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/ponder-helpers.test.ts)
- [apps/ensindexer/src/lib/plugin-helpers.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/plugin-helpers.ts)
- [apps/ensindexer/src/lib/plugin-helpers.test.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/plugin-helpers.test.ts)
- [apps/ensindexer/src/lib/merge-ponder-configs.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/merge-ponder-configs.ts)
- [apps/ensindexer/src/lib/merge-ponder-configs.test.ts](https://github.com/namehash/ensnode/blob/main/apps/ensindexer/src/lib/merge-ponder-configs.test.ts)
- ... anything else that may be relevant.

# Subtasks

- [ ] Reorganize code
- [ ] Document the finite set of places where all Ponder-specific logic is found.

lightwalker-eth:
A related goal was just discussed here: https://github.com/namehash/ensnode/pull/1406/files/dd19ce9a9c8afbeee220703c683b61195a197e99..57fcd2acfbddbe4f3a5a30fa34b038758266a936#r2628814243

Specifically, it would be nice if we could remove any Ponder dependencies from the `ensnode-schema` package, as this spills Ponder dependencies into all apps building on ENSDb, including ENSApi.

PR #1406, opened 2025-12-16
feat(ensindexer): ENSDb Writer Worker

### LINKED ISSUE #1252: ENSApi: Persist ENSIndexerPublicConfig to the database
# Background

Currently:
- ENSIndexer exposes an API for indexing status and for its config
- ENSApi directly sends API calls to ENSIndexer to get its indexing status and config
- ENSApi uses SWR caching to remember the latest indexing status and config from ENSIndexer, such that ENSApi can survive crashes of ENSIndexer if and only if the ENSApi instance was able to successfully fetch these values from ENSIndexer during the lifecycle of the ENSApi process instance.
- ... Therefore: if ENSIndexer is crashed and ENSApi is restarted / launched, the new ENSApi currently will also fail to start. This would be a bad situation and we want to ensure it can't impact ENSNode deployments.

# Goals

1. Ensure ENSAPI can survive situations where ENSIndexer is down and ENSApi is restarted.
2. Update ENSApi and ENSIndexer so that they stop exchanging the config and indexing status via API calls to ENSIndexer and instead they exchange this state exclusively through ENSDb. Here, ENSIndexer should push this state into ENSDb, while ENSApi should pull this state from ENSDb.
3. Unblock the next phase of advancements on ENSAnalytics. More specifically, as a separate future goal, we want to transition the `ens-referrals` package out of the ENSNode repo and into the `ens-referrals` repo (which would be converted into a monorepo). Then, in the `ens-referrals` repo, we would want to create a new backend service for the leaderboard APIs that would be the first example of a custom app that would build on top of ENSDb and expose a tailored API built specifically for the needs of that app. This would allow us to transition all of the ENSAnalytics out of ENSNode and into this new custom app in the `ens-referrals` repo. Before we take all of these actions though it will be important that we first action the goals described in this issue so that there is a mature mechanism for an app to build custom APIs on top of ENSDb and to integrate with the indexing status and config stored in ENSDb. Related issue: https://github.com/namehash/ensnode/issues/1352
4. Protect ENSNode operators against a scenario where they have all ENSNode services up and running. Then imagine they keep ENSApi up and running while terminating ENSIndexer, wiping the configured `DATABASE_SCHEMA`, and then starting a new ENSIndexer with config changes that are incompatible with the config that has been cached into memory by ENSApi. The logic proposed here aims to ensure that a max of 1 minute might exist where an ENSApi instance is attached to an incompatible `DATABASE_SCHEMA` without knowing it.
5. Create a new `ENSDbClient` class that encapsulates logic for the following in a way that would be nicely reusable across apps building on top of ENSDb, including ENSApi and the future app described above in goal 3.
   1. Manages `SWRCache` that read both the config and indexing status from a `DATABASE_SCHEMA` in ENSDb.
   2. Enforcing that the indexing status and config objects that it caches are compatible. Ex: the chain ids perfectly align. Etc.
   3. Provides a helper function such as `isHealthy` that gives a simple mechanism to identify if both a config and indexing status are held in memory and compatible with each other.
6. Improve the startup sequence of ENSApi such that it doesn't crash itself just because it is waiting on ENSIndexer to successfully initialize itself. Instead, it should patiently wait. Related issue: https://github.com/namehash/ensnode/issues/1303
   1. Update the logic for the `/health` endpoint in ENSApi such that it considers itself "unhealthy" if the `ENSDbClient.isHealthy` is false.

# Non-Goals

1. We should not remove the config or indexing status APIs from ENSIndexer. I believe all of our current usage of these APIs should be removed as a result of the changes described here, but I think it's best to keep these APIs living in ENSIndexer anyway as maybe there will be some special reason for them in the future that we cannot predict now.

# Goals

1. ENSIndexer
   1. Each time it starts it should:
      1. Check if a config has already been saved into the configured `DATABASE_SCHEMA`. If so, read it from the `DATABASE_SCHEMA` and verify that it is compatible with the config object it has built in memory. If it is not compatible, refuse to start and provide a meaningful error message.
      2. If no config has been saved into the configured `DATABASE_SCHEMA` yet, save it there.
   2. At a frequent time interval (ex: maybe once per second?):
      1. Build a new indexing status snapshot in memory. If this fails for some reason, log the error and do nothing for this iteration of the asynchronous background loop.
      2. Is the omnichain indexing status in that snapshot `OmnichainIndexingStatusSnapshotUnstarted`? If so, log the error and do nothing for this iteration of the asynchronous background loop.
      3. Upsert the indexing status snapshot into the `DATABASE_SCHEMA`.
2. ENSApi
   1. Update our logic for `SWRCache` so we update the signature of the `fn` such that it optionally receives a copy of whatever the latest cached value is as input. (Optionally meaning that if no value is already cached it receives undefined, but if a value is already cached it receives that value). This way the `fn` can internally decide special logic it may want to perform as a function of whatever the already cached value is. The purpose of this idea will be explained below.
   2. Each time ENSApi starts it should:
      1. Build a SWRCache for reading the config from the configured `DATABASE_SCHEMA`. This should proactively initialize itself and proactively revalidate itself once per minute.
         1. The `fn` should work to read the config from the configured `DATABASE_SCHEMA` and then validate / deserialize it. Anytime this fetch or validation / deserialization fails, log the error.
         2. The `fn` should make use of the idea described above for the `fn` to optionally receive a copy of whatever the latest cached value as input was. More specifically, each time a new config value is built in memory (not yet returned from `fn`) this function should compare that the config has not changed in a way that would be incompatible. In other words, changing a RPC endpoint is an example of a compatible change, however, changing the indexed ENS namespace is an example of an incompatible change. All transitions from no cached result -> cached result would pass this check, this check is purely for validating the transition from cached result A -> cached result B. If a config change is incompatible, ENSApi should forcefully terminate itself (open to other suggestions).
      2. Build a SWRCache for the latest indexing status snapshot. But now the `fn` associated with this should work to read the indexing status snapshot from the `DATABASE_SCHEMA` rather than reading the latest indexing status snapshot from direct API calls to ENSApi. If no indexing status snapshot is found in the `DATABASE_SCHEMA` then log an error. If it is found, validate it. If validation / deserialization fails, log an error. If it is found and all validation / deserialization succeeds, compare the indexing status against the config as read from the SWRCache that manages the config. If the config is not cached yet, log a message. If the config is cached, validate that the new indexing status is compatible with the config. For example, all the chain ids should match up perfectly. Etc. If the new indexing status is compatible, successfully cached it.
      3. store the indexing status in a proactively revalidating SWRCache. need to persist/cache the ensindexerpublicconfig to database so ensapi can survive restarts without ensindexer availability
      2. Consider itself "unhealthy" until both its SWRCacheIf ENSApi it is not found log an error and terminate. If it is found, validate it. If validation / deserialization fails, log an error and terminate. If it is found and all validation / deserialization succeeds, log the config it will be using to the console and store the cached config in memory indefinitely for the remaining lifecycle of the ENSApi process instance.
   3. Remove the following environment variables from ENSApi: `ENSINDEXER_URL`

tk-o:
Store ENSNode metadata (ENSIndexer Public Config, Indexing Status) into ENSDb.

### ENSNode Schema

New schema was added for ENSNode metadata at `packages/ensnode-schema/src/schemas/ensnode-metadata.schema.ts`.

### ENSNode SDK

A simple HTTP client was built at `packages/ensnode-sdk/src/ensindexer/client.ts`. It's useful for sending HTTP requests over the internal network (akin to in-memory function calls):
- `/health` endpoint — required to respond with HTTP Status OK before any other requests can be made.
- `/api/config` endpoint — returns in-memory `ENSIndexerPublicConfig` object.
- `/api/indexing/status` endpoint — returns in-memory `IndexingStatusResponse` object.

#### ENSDb module
Includes:
- `makeDrizzle` factory function known from ENSApi, which creates a Drizzle client instance.
  - `apps/ensindexer/src/lib/ensdb/drizzle.ts`
- `ENSNodeMetadata` type defining data model associated with `ensnode-metadata.schema.ts`
  - `apps/ensindexer/src/lib/ensdb/ensnode-metadata.ts`
- `EnsDbClient` class for performing ENSDb reads and writes.
  - `apps/ensindexer/src/lib/ensdb/ensdb-client.ts`

#### ENSDb Writer Worker

Initialized at ENSIndexer process startup:
- `apps/ensindexer/ponder/src/ensdb-writer-worker.ts`

Waits for ENSIndexer to become healthy and then:
1. Attempts to upsert `ENSIndexerPublicConfig` into ENSDb.
   - This may fail when ENSDb already had other record for ENSIndexerPublicConfig stored before, and the new record was not compatible with the stored one.
   - This is a single attempt and if it fails, the ENSIndexer application will be terminated with error.
   - Also, the ENSDb Version info is stored into ENSDb at the same time as ENSIndexerPublicConfig record.
2. Attempts to upsert `CrossChainIndexingStatusSnapshot` into ENSDb.
  - This may fail if the indexing status was not available yet, or the omnichain indexing status was "unstarted".
  - This is a recurring attempt; any failed attempt will log an error and there will be another attempt.

Related to #1252


apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,142 @@
...
+  const ensDbMutation = new EnsDbMutation(ensDbClient);
+
+  /**
+   * Handle ENSIndexerPublicConfig Record
+   */
+  const handleEnsIndexerPublicConfigRecord = async () => {
+    // Read stored config and in-memory config.
+    // Note: we wrap read operations in pRetry to ensure all of them are
+    // completed successfully.
+    const [storedConfig, inMemoryConfig] = await pRetry(() =>
```
shrugs: technically it think this should be 

```
await Promise.all([
  pRetry(...),
  pRetry(...)
])
```

this way if one of the reads fails, the other isn't unnecessarily re-fetched

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,142 @@
...
+      Promise.all([ensDbQuery.getEnsIndexerPublicConfig(), ensIndexerClient.config()]),
+    );
+
+    // Validate in-memory config object compatibility with the stored one,
+    // if the stored one is available
+    if (storedConfig) {
+      try {
+        validateENSIndexerPublicConfigCompatibility(storedConfig, inMemoryConfig);
```
shrugs: this is a good check but maybe not even necessary? because ponder already will reject re-indexing if the ponder config differs from this schema's previously used ponder config. obviously ponder config !== ensindexerpublicconfig but it's pretty close — can we think of a situation where that wouldn't be the case?

all that said there's no harm in leaving this check here, so that's fine

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,142 @@
...
+      if (inMemoryIndexingStatus.responseCode !== IndexingStatusResponseCodes.Ok) {
+        throw new Error("Indexing Status must be available.");
+      }
+
+      const { snapshot } = inMemoryIndexingStatus.realtimeProjection;
+      const { omnichainSnapshot } = snapshot;
+
+      // Check if Indexing Status is in expected status.
+      if (omnichainSnapshot.omnichainStatus === OmnichainIndexingStatusIds.Unstarted) {
+        throw new Error("Omnichain Status  must be different that 'Unstarted'.");
```
shrugs: than* unstarted

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      // Upsert ENSIndexerPublicConfig into ENSDb.
+      await ensDbMutation.upsertIndexingStatus(snapshot);
+
+      logger.info("Indexing Status successfully stored in ENSDb.");
```
shrugs: - add a once-per-process log at info level for "Indexing Status successfully stored in ENSDb."
- make this a trace-level call with something like "Indexing Status updated in ENSDb."

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,142 @@
...
+  // 4. Handle ENSIndexer Public Config just once.
+  await handleEnsIndexerPublicConfigRecord();
+
+  // 5. Handle Indexing Status on recurring basis.
+  await handleIndexingStatusRecordRecursively();
+}
+
+// Run ENSDb Writer Worker in a non-blocking way to
+// allow database migrations to proceed in the background.
+setTimeout(ensDbWriterWorker, 0);
```
shrugs: don't need to use settimeout, i woudl just do:

```
ensDbWriterWorker()
  .then() // logger.info ENSdb Writer Worker started
  .cach() // logger.error(error), process.exit
```

apps/ensindexer/src/config/compatibility.ts
```diff
@@ -0,0 +1,64 @@
...
+export type ENSIndexerPublicConfigCompatibilityCheck = Pick<
+  SerializedENSIndexerPublicConfig,
+  "indexedChainIds" | "isSubgraphCompatible" | "namespace" | "plugins"
+>;
+
+/**
+ * Validate if `configB` is compatible with `configA`, such that `configA` is
+ * a subset of `configB`.
+ *
+ * @throws error if 'indexedChainIds' were incompatible.
```
shrugs: nit: simplify to `@throws error if configs are incompatible`

apps/ensindexer/src/lib/ensindexer/client.ts
```diff
@@ -0,0 +1,68 @@
...
+  public async health(): Promise<void> {
+    logger.debug("Fetching ENSIndexer Health status");
+
+    try {
+      await fetch(new URL("/health", this.ensIndexerUrl));
+
+      logger.debug("Fetching ENSIndexer Health status: healthy");
+    } catch {
+      const errorMessage = "Health endpoint for ENSIndexer is not available yet.";
+      logger.error(errorMessage);
```
shrugs: i don't think the client should log errors — the consumer should

apps/ensindexer/src/lib/ensdb/drizzle.ts
```diff
@@ -0,0 +1,41 @@
+// This file was copied 1-to-1 from ENSApi.
```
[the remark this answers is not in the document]
shrugs: likewise for `logger.ts`

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
tk-o: Awesome suggestion, thanks 🚀

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,142 @@
...
+
+        // Throw the error to terminate the ENSIndexer process due to
+        // found config incompatibility
+        throw new Error(errorMessage);
+      }
+    } else {
+      // Upsert ENSIndexerPublicConfig into ENSDb.
+      // Note: we wrap write operation in pRetry to ensure it can complete
+      // successfully, as there will be no other attempt.
+      await pRetry(() => ensDbMutation.upsertEnsIndexerPublicConfig(inMemoryConfig));
```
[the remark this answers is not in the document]
tk-o: I didn't feel 100% ok with this pRetry, perhaps we should the process fail if there's any (very unlikely) issue with the write 👍

apps/ensindexer/src/lib/ensdb/ensdb-mutation.ts
```diff
@@ -0,0 +1,75 @@
...
+import type {
+  SerializedCrossChainIndexingStatusSnapshot,
+  SerializedENSIndexerPublicConfig,
+} from "@ensnode/ensnode-sdk";
+
+import type { EnsDbClient } from "./ensdb-connection";
+import { type EnsNodeMetadata, EnsNodeMetadataKeys } from "./ensnode-metadata";
+
+/**
+ * ENSDb Mutation
```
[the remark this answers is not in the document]
tk-o: Will do 👍

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
tk-o: In fact, there's no need to log the trace for the update event. The lack of error log implies the successful update.

apps/ensindexer/src/lib/ensindexer/client.ts
```diff
@@ -0,0 +1,51 @@
+import type {
+  SerializedENSIndexerPublicConfig,
+  SerializedIndexingStatusResponse,
+} from "@ensnode/ensnode-sdk";
+
+export class EnsIndexerClient {
```
lightwalker-eth: I imagine this should move into the `ensnode-sdk` package?

It's true that we would only use it ourselves within ENSIndexer, but in theory we should put the clients for each of the ENSNode apps into `ensnode-sdk`. At least with my current mental model.

packages/ensnode-schema/src/ponder.schema.ts
```diff
@@ -2,6 +2,7 @@
  * Merge the various sub-schemas into a single ponder (drizzle) schema.
```
lightwalker-eth: In my mental model:

1. There should be absolutely no references to "ponder" at all within this package.
2. All ponder dependencies in this package should be transferred into `ensindexer`. The only thing in our whole monorepo that should have any dependency on Ponder should be the `ensindexer` app. No other app or package should have any Ponder deps.
3. This package should be renamed from `ensnode-schema` to `ensdb-sdk` or something like that. It would not only hold the Drizzle schemas, but it would also be the home for the clients we'll be working to build up that would contain the logic for db read and write operations used across the ENSNode monorepo. Not trying to suggest we completely transfer all db read and write operations into this package within this PR. That's too much for this PR. But sharing the idea here to help share an idea on overall future direction.

What do you think? Appreciate your advice.

pnpm-workspace.yaml
```diff
@@ -8,6 +8,7 @@ catalog:
   "@astrojs/react": ^4.4.1
   "@astrojs/tailwind": ^6.0.2
   "@namehash/namekit-react": 0.12.0
+  "@ponder/client": 0.13.14
   "@ponder/utils": 0.2.14
```
lightwalker-eth: Please see other comments suggesting all ponder deps would exclusively live in `ensindexer`. What do you think?

apps/ensindexer/src/lib/ensindexer/client.ts
```diff
@@ -0,0 +1,51 @@
...
+  constructor(private ensIndexerUrl: URL) {}
+
+  /**
+   * Wait for ENSIndexer to become healthy.
```
lightwalker-eth: We should document here:

1. What does it represent for ENSIndexer to be healthy? What does this guarantee about state?
2. What does it represent for ENSIndexer not to be healthy? What should someone avoid doing if ENSIndexer isn't healthy yet?

apps/ensindexer/src/lib/ensindexer/client.ts
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @throws when the Health endpoint didn't return HTTP Status OK.
+   */
+  public async health(): Promise<void> {
```
lightwalker-eth: Shouldn't there be a distinct representation of each of the following possible results?
1. explicitly received a response, the HTTP response code wasn't 200 OK though.
2. never received any response

apps/ensindexer/src/lib/ensindexer/client.ts
```diff
@@ -0,0 +1,51 @@
...
+      new URL("/api/config", this.ensIndexerUrl),
+    ).then((response) => response.json());
+
+    return ensIndexerPublicConfigSerialized as SerializedENSIndexerPublicConfig;
+  }
+
+  /**
+   * Fetch Indexing Status
+   *
+   * @returns Indexing Status when it's available and with valid.
```
lightwalker-eth: Suggest to refine this sentence.

apps/ensindexer/src/lib/ensindexer/local-client.ts
```diff
@@ -0,0 +1,16 @@
...
+import pRetry from "p-retry";
+
+import { EnsIndexerClient } from "./client";
+
+export const ensIndexerClient = new EnsIndexerClient(config.ensIndexerUrl);
+
+/**
+ * Wait for ENSIndexer to become healthy.
+ *
+ * The global promise that will only resolve after the ENSIndexer has become healthy.
```
lightwalker-eth: I'm not familiar with `pRetry`. If the 5 retries all fail, is it confirmed then it causes ENSIndexer to terminate itself? Appreciate your advice.

apps/ensindexer/src/lib/ensindexer/local-client.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export const waitForEnsIndexerToBecomeHealthy = pRetry(async () => ensIndexerClient.health(), {
+  retries: 5,
```
lightwalker-eth: How do you suggest we handle the ideas discussed in our team call earlier today about ENSIndexer potentially needing to wait up to an hour for ENSRainbow to become both healthy and ready?

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
```diff
@@ -0,0 +1,161 @@
...
+ */
+export class EnsDbClient implements EnsDbClientQuery, EnsDbClientMutation {
+  #db = makeDrizzle({
+    databaseSchema: config.databaseSchemaName,
+    databaseUrl: config.databaseUrl,
+    schema,
+  });
+
+  /**
+   * Upsert ENSIndexer Public Config
```
lightwalker-eth:
```suggestion
   * Get ENSIndexer Public Config
```

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
```diff
@@ -0,0 +1,161 @@
...
+   *
+   * @returns updated record in ENSDb.
```
lightwalker-eth:
```suggestion
   * @returns ENSIndexer public config in ENSDb.
```

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
```diff
@@ -0,0 +1,161 @@
...
+   * @throws when upsert operation failed.
```
lightwalker-eth:
```suggestion
   * @throws when read operation failed.
```

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
```diff
@@ -0,0 +1,161 @@
...
+   */
+  async getEnsIndexerPublicConfig(): Promise<SerializedENSIndexerPublicConfig | undefined> {
```
lightwalker-eth: In my mind, the serialization / deserialization responsibilities should live within this class. Therefore all params passed into all the functions in this file or returned by the functions in this file should be deserialized business-logic-level object types.

This comment applies across the whole file.

shrugs (approved):
LGTM!

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,131 @@
...
+ * Runs the following tasks:
+ * 1) On application startup, attempt to upsert serialized representation of
+ *    {@link ENSIndexerPublicConfig} into ENSDb.
+ * 2) On application startup, and then on recurring basis,
+ *    following the {@link INDEXING_STATUS_RECORD_UPDATE_INTERVAL}, attempt to
+ *    upsert serialized representation of {@link CrossChainIndexingStatusSnapshot}
+ *    into ENSDb.
+ */
+async function ensDbWriterWorker() {
+  console.log("ENSDb Writer Worker: waiting for ENSIndexer to become healthy.");
```
lightwalker-eth: What do you think about potentially having Ponder wait up to an hour to have us pass it a Ponder config while we wait for ENSRainbow to become healthy / ready such that ENSIndexer can complete building its config?

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  // 0. Wait for ENSIndexer to become healthy before running the worker's logic
+  await waitForEnsIndexerToBecomeHealthy;
+
+  console.log("ENSDb Writer Worker: ENSIndexer is healthy, starting tasks.");
+
+  // 1. Create ENSDb Client
+  const ensDbClient = new EnsDbClient();
```
lightwalker-eth: Hmm. I think we need to give more of a detailed focus on the case of a completely fresh `DATABASE_SCHEMA` and how this is initialized by Ponder.

We have the existing code path:
1. ENSIndexer fetches ENSRainbow config from ENSRainbow
2. ENSIndexer can now build its own config
3. ENSIndexer uses its own config to build a Ponder config
4. Ponder uses the Ponder config to initialize `DATABASE_SCHEMA`.

Right? Don't we need to wait until step 4 for the tables in `DATABASE_SCHEMA` to be initialized?

Assuming so, this waiting potentially needs to grow a lot if we add logic to wait up to an hour for step 1 above to be complete, waiting for ENSRainbow to become both healthy and ready.

And during all of this time, what is ENSApi supposed to do? From its perspective, it's been configured to read from `DATABASE_SCHEMA` but ENSIndexer might wait over an hour to create it.

Suggest to take a step back and carefully consider all cases and situations, then to document a plan and share it with the team for review 👍

lightwalker-eth (commented):
@tk-o Thanks for this. Reviewed and shared feedback 👍

packages/ensnode-schema/src/ponder.schema.ts
tk-o: Removing `ponder` dependency from `ensnode-schema` package will possibly require maintaining our own implementation of this directory: https://github.com/ponder-sh/ponder/tree/ponder%400.13.14/packages/core/src/drizzle

I don't think it's practical at the moment, given no plans for ENSNode to migrate away from `ponder`.

pnpm-workspace.yaml
tk-o: If we want to remove these dependencies, we need to implement the functionality provided by these packages on our own:
- `@ponder/client`
  - `setDatabaseSchema`
- `@ponder/utils`
  - `mergeAbis`
  - `replaceBigInts`

pnpm-workspace.yaml
```diff
@@ -18,6 +19,8 @@ catalog:
   date-fns: 4.1.0
   drizzle-orm: "=0.41.0"
   hono: ^4.10.2
+  p-memoize: ^8.0.0
+  p-retry: ^7.1.1
   pg-connection-string: ^2.9.1
   pino: 10.1.0
   ponder: 0.13.16
```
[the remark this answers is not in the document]
tk-o: Right now, we use `ponder` in:
- `packages/ensnode-schema`
  - Removing `ponder` dependency here requires further consideration, as explained [here](https://github.com/namehash/ensnode/pull/1406#discussion_r2630426712) and [there](https://github.com/namehash/ensnode/pull/1406#discussion_r2630342844).
- `packages/ponder-metadata`
  - It should be easy to remove `ponder` as dependency here 👍 Will create a follow-up PR including that change and that also resolves https://github.com/namehash/ensnode/issues/908

apps/ensindexer/src/lib/ensindexer/client.ts
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @returns ENSIndexer Public Config
+   * @throws error when fetching ENSIndexer Public Config failed
+   */
+  public async config(): Promise<SerializedENSIndexerPublicConfig> {
+    const ensIndexerPublicConfigSerialized = await fetch(
+      new URL("/api/config", this.ensIndexerUrl),
+    ).then((response) => response.json());
+
+    return ensIndexerPublicConfigSerialized as SerializedENSIndexerPublicConfig;
```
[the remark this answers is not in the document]
tk-o: >Perhaps the idea in this comment is too big for this PR and should be moved into a separate follow-up issue?
>
>I'm curious why we are returning serialized types from our clients. In my mental model, all of our clients should also take on the responsibility of deserialization and returning business-logic-level objects.

I'll follow up on this request in another PR for which the scope was mentioned [here](https://github.com/namehash/ensnode/pull/1406#discussion_r2630311539). TLDR: we need to split request and response types between `EnsIndexerClient` and `EnsApiClient` (which is called `ENSNodeClient` as of now).

> I'm also curious about the use of https://hono.dev/docs/guides/rpc (hono client, internally within our own client implementations so that hono isn't directly exposed / exported by our packages) and how that relates to the work in progress by @stevedylandev in this PR: https://github.com/namehash/ensnode/pull/1338

@lightwalker-eth could you please expand on specific request for me here? I don't understand the expectation here.

apps/ensindexer/src/lib/ensindexer/local-client.ts
tk-o: This is idea for a separate PR, we need to update the way we use `EnsRainbowApiClient` inside `ensindexer` app, such that, before calling a method on an `ensRainbowApiClient` instance, we await for a health check promise to resolve. It should be similar to how we use `waitForEnsIndexerToBecomeHealthy` promise in the `EnsIndexerClient` use case.

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+  }
+
+  /**
+   * Upsert Indexing Status
+   *
+   * @returns updated record in ENSDb.
+   * @throws when upsert operation failed.
+   */
+  async upsertIndexingStatus(
```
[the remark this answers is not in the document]
tk-o: Answered [here](https://github.com/namehash/ensnode/pull/1406#discussion_r2630593201).

apps/ensindexer/src/lib/ensdb/ensnode-metadata.ts
```diff
@@ -0,0 +1,31 @@
+import type {
+  SerializedCrossChainIndexingStatusSnapshot,
+  SerializedENSIndexerPublicConfig,
+} from "@ensnode/ensnode-sdk";
+
```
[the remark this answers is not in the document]
tk-o: Will do that in a follow up PR.

apps/ensindexer/src/lib/ensdb/drizzle.ts
```diff
@@ -0,0 +1,25 @@
...
+// TODO: deduplicate with apps/ensapi/src/lib/handlers/drizzle.ts when ensnode nodejs internal package is created
```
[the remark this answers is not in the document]
tk-o: Using `setDatabaseSchema` is more like a fix for `drizzle` object instance to use the specified `schemaName` for queries.

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
tk-o: I like the idea for ENSApi to use specific `DATABASE_SCHEMA`. Suggest sticking to that.

ENSApi has to wait for ENSDb to include `EnsNodeMetadata`. Otherwise, there's literally no use case ENSApi could serve. All ENSApi routes require indexing status information, which come from ENSDb.

ENSDb won't have the `EnsNodeMetadata` information until ENSIndexer stores it.

ENSIndexer won't store any `EnsNodeMetadata` information until ENSRainbow is healthy, and ENSIndexer is healthy.

I think we should create a "logical healthcheck" for ENSDb, such that clients connecting to ENSDb could know if the service is ready to be used. For example, we can consider this "logical healthcheck" to ensure that ENSDb includes `EnsNodeMetadata` for both keys: `EnsNodeMetadataKeys.EnsIndexerPublicConfig`, `EnsNodeMetadataKeys.IndexingStatus`.

apps/ensindexer/src/lib/ensindexer/client.ts
tk-o: Implemented `EnsIndexerHealthCheckResults` to handle all possible response result scenarios.

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  /**
+   * Upsert ENSIndexer Public Config
+   *
+   * @returns updated record in ENSDb.
+   * @throws when upsert operation failed.
+   */
+  async upsertEnsIndexerPublicConfig(
+    ensIndexerPublicConfig: SerializedENSIndexerPublicConfig,
+  ): Promise<SerializedENSIndexerPublicConfig> {
```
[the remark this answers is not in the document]
tk-o: I updated all "upsert" methods to have `Promise<void>` return type.

apps/ensindexer/src/lib/ensdb/ensdb-client.ts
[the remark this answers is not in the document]
tk-o: I updated all "upsert" methods to have `Promise<void>` return type.

packages/ensnode-sdk/src/ensindexer/config/compatibility.ts
```diff
@@ -0,0 +1,61 @@
+import type { SerializedENSIndexerPublicConfig } from "@ensnode/ensnode-sdk";
+
```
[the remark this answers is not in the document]
tk-o: Nice idea, will have it moved to be co-located with `EnsIndexerClient` file.

packages/ensnode-schema/src/schemas/ensnode-metadata.schema.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Schema Definitions that hold metadata about the ENSNode instance.
+ */
+
+import { onchainTable } from "ponder";
+
+export const ensNodeMetadata = onchainTable("ensnode_metadata", (t) => ({
```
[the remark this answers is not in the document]
lightwalker-eth: @tk-o Thanks for this detailed investigation and reply 👍 

I've transferred these ideas into https://github.com/namehash/ensnode/issues/1234#issuecomment-3675295303 for future follow up and will mark this conversation thread as resolved for the scope of this PR.

apps/ensindexer/src/lib/ensindexer/local-client.ts
```diff
@@ -2,7 +2,13 @@ import config from "@/config";
 
 import pRetry from "p-retry";
 
-import { EnsIndexerClient } from "./client";
+import { EnsIndexerClient, EnsIndexerHealthCheckResults } from "@ensnode/ensnode-sdk";
+
+/**
+ * How many times retries should be attempted before
+ * {@link waitForEnsIndexerToBecomeHealthy} becomes a rejected promise.
+ */
```
lightwalker-eth:
```suggestion
/**
 * Upon the startup of ENSIndexer, defines the max number failures ENSIndexer is allowed to experience
 * when attempting to get a successful response from its own `/health` endpoint. If these max retries are exceeded
 * {@link waitForEnsIndexerToBecomeHealthy} becomes a rejected promise
 * which ultimately results in ENSIndexer crashing itself.
 */
```

Is that fair?

apps/ensindexer/src/lib/ensindexer/local-client.ts
```diff
@@ -11,6 +17,15 @@ export const ensIndexerClient = new EnsIndexerClient(config.ensIndexerUrl);
...
  */
-export const waitForEnsIndexerToBecomeHealthy = pRetry(async () => ensIndexerClient.health(), {
-  retries: 5,
-});
+export const waitForEnsIndexerToBecomeHealthy = pRetry(
+  async () => {
+    const response = await ensIndexerClient.health();
+
+    if (response !== EnsIndexerHealthCheckResults.Ok) {
+      throw new Error("ENSIndexer is not healthy yet");
```
lightwalker-eth: Would it be helpful to include a log message in here too? Please ignore if not relevant.

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,134 @@
+/**
+ * This file manages syncing ENSNode metadata:
+ * - ENSIndexer Public Config
+ * - Indexing Status
+ * into the ENSDb.
```
lightwalker-eth:
```suggestion
 * This file manages upserting ENSNodeMetadata into ENSDb.
```

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Runs the following tasks:
+ * 1) On application startup, attempt to upsert serialized representation of
+ *    {@link ENSIndexerPublicConfig} into ENSDb.
+ * 2) On application startup, and then on recurring basis,
+ *    following the {@link INDEXING_STATUS_RECORD_UPDATE_INTERVAL}, attempt to
+ *    upsert serialized representation of {@link CrossChainIndexingStatusSnapshot}
+ *    into ENSDb.
+ */
+async function ensDbWriterWorker() {
```
lightwalker-eth:
```suggestion
async function ensDbMetadataUpsertWorker() {
```

Is that fair? Seems much more accurate.

Assuming so, please rename all the related things / comments / files / etc.

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import {
+  CrossChainIndexingStatusSnapshot,
+  type Duration,
+  ENSIndexerPublicConfig,
+  IndexingStatusResponseCodes,
+  OmnichainIndexingStatusIds,
+  validateENSIndexerPublicConfigCompatibility,
+} from "@ensnode/ensnode-sdk";
+
```
lightwalker-eth: Why does this file live under the `/ponder` directory?

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Handle Indexing Status Record Recursively
+   */
+  const handleIndexingStatusRecordRecursively = async () => {
+    try {
+      // Read in-memory Indexing Status.
+      const inMemoryIndexingStatus = await ensIndexerClient.indexingStatus();
+
+      // Check if Indexing Status is available.
+      if (inMemoryIndexingStatus.responseCode !== IndexingStatusResponseCodes.Ok) {
+        throw new Error("Indexing Status must be available.");
```
lightwalker-eth: We need to write better. This should explain why.

apps/ensindexer/ponder/src/ensdb-writer-worker.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  console.log("ENSDb Writer Worker: ENSIndexer is healthy, starting tasks.");
+
+  // 1. Create ENSDb Client
+  const ensDbClient = new EnsDbClient();
+
+  /**
+   * Handle ENSIndexerPublicConfig and ENSDb Version Records
+   */
+  const handleEnsIndexerPublicConfigAndEnsDbVersionRecords = async () => {
```
lightwalker-eth: I note how this function doesn't initialize the metadata field for indexing status.

As a result, ENSDb clients have additional complexity in that they need to prepare for a situation where there may be a long duration of time where they can read 2 of the 3 metadata fields, but not the indexing status field.

What do you think about this? Appreciate your advice. Should we update this function so that it waits until all 3 metadata fields can be initialized at the same time, and then initializes them all together?

apps/ensindexer/src/lib/ensindexer/local-client.ts
```diff
@@ -0,0 +1,31 @@
...
+export const MAX_ENSINDEXER_HEALTHCHECK_ATTEMPTS = 5;
+
+export const ensIndexerClient = new EnsIndexerClient(config.ensIndexerUrl);
+
+/**
+ * Wait for ENSIndexer to become healthy.
+ *
+ * The global promise that will only resolve after the ENSIndexer has become healthy.
+ */
+export const waitForEnsIndexerToBecomeHealthy = pRetry(
```
lightwalker-eth: Is there a special reason to put this function in this file? In my mind it seems it would be better moved into `ensdb-writer-worker.ts`.

lightwalker-eth (commented):
@tk-o Thanks for updates. Reviewed and shared feedback.

PR #1408, opened 2025-12-17
fallback-ensapi

shrugs:
- moves thegraph logic out of ensapi and into sdk internal
- makes datasources and ensnode-sdk treeshakable
- move releases package selection logic to inclusion instead of inclusion
- build/publish integrated with workflow (does NOT include deployment)
- in production `CLOUDFLARE_SECRET` is required for authorizing requests to `/subgraph` (because it's the only route with an expensive operation)
- fetches `THEGRAPH_API_KEY` from secretsmanager
  - expects `THEGRAPH_API_KEY_SECRET_ID` with `THEGRAPH_API_KEY` in the returned JSON obj

https://github.com/namehash/ensnode/actions/runs/20439920414

### DIFF
--- a/apps/fallback-ensapi/README.md
+++ b/apps/fallback-ensapi/README.md
@@ -2,17 +2,19 @@
 
 A lightweight AWS Lambda that provides an infrastructure-level fallback for API requests sent to the ENSApi instances hosted by NameHash.
 
-**fallback-ensapi is written for use within ENSNode's own infrastructure, and may not generalize to other deployments.**
+**fallback-ensapi is written for use within the NameHash team's own infrastructure, and may not generalize to other deployments of ENSNode.**
 
 NameHash operates infrastructure-level monitoring of the health of all of their ENSNode deployments. In the event these monitoring systems determine that a NameHash operated ENSNode deployment has become unhealthy, the monitoring system will automatically redirect all API requests sent to the ENSApi instance associated with an unhealthy ENSNode to this Fallback ENSApi until the load balancer determines that the ENSNode instances are healthy again.
 
-This Fallback ENSApi proxies ENS Subgraph GraphQL API requests originally sent to ENSApi to TheGraph's ENS Subgraph using the provided API key.
+This Fallback ENSApi proxies ENS Subgraph GraphQL API requests originally sent to ENSApi to TheGraph's ENS Subgraph using an API key for TheGraph held in AWS Secrets Manager under the secret id defined by the `THEGRAPH_API_KEY_SECRET_ID` environment variable.
+
+When run in production, it also requires configuration of the `CLOUDFLARE_SECRET` environment variable to avoid exposing our `/subgraph` proxy to the public net.
 
 For data consistency, this fallback is exclusively enabled for NameHash's ENSNode deployments that use a [Subgraph Compatible](https://ensnode.io/docs/reference/subgraph-compatibility/) configuration. More specifically:
 - Mainnet: https://api.mainnet.ensnode.io/
 - Sepolia: https://api.sepolia.ensnode.io/
 
-The following NameHash ENSNode deployments are _not_ [Subgraph Compatible](https://ensnode.io/docs/reference/subgraph-compatibility/) and Fallback ENSApi will return an HTTP 503 (Service Unavailable) in response to any Subgraph API request.
+The following NameHash ENSNode deployments are _not_ fully [Subgraph Compatible](https://ensnode.io/docs/reference/subgraph-compatibility/), as they produce a _superset_ of the data indexed by TheGraph. Fallback ENSApi will return an HTTP 503 (Service Unavailable) in response to any Subgraph API requests for these deployments:
 - Mainnet: https://api.alpha.ensnode.io/
 - Sepolia: https://api.alpha-sepolia.ensnode.io/
 
--- a/apps/fallback-ensapi/package.json
+++ b/apps/fallback-ensapi/package.json
@@ -3,7 +3,7 @@
   "name": "fallback-ensapi",
   "version": "1.3.1",
   "type": "module",
-  "description": "Infrastructure-level ENSNode Fallback API",
+  "description": "Infrastructure-level Fallback for ENSApi",
   "license": "MIT",
   "repository": {
     "type": "git",
--- a/.changeset/vast-lies-design.md
+++ b/.changeset/vast-lies-design.md
@@ -2,4 +2,4 @@
 "fallback-ensapi": minor
 ---
 
-Introduce infrastructure-level fallback-ensapi service, which handles ENSApi requests in the event of catastrophic ENSNode environment failures. Intended for internal ENSNode cloud use only.
+Introduce infrastructure-level fallback-ensapi service, which handles ENSApi requests in the event of ENSNode environment failures. Intended for NameHash hosted ENSNode instances only.


packages/ensnode-sdk/src/shared/config/thegraph.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from "zod/v4";
```
shrugs: moved to shared

packages/ensnode-sdk/src/shared/thegraph.ts
```diff
@@ -0,0 +1,44 @@
+import type { ENSNamespaceId } from "@ensnode/datasources";
```
shrugs: moved to shared

apps/fallback-ensapi/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "private": true,
+  "name": "fallback-ensapi",
```
lightwalker-eth: Prefer that we brainstorm an alternative name for this app. Suggest we align on the name for it in Slack before we lock it in.

One issue I have with `fallback-ensapi` is that (as currently implemented) this is actually a fallback for the ENS Subgraph API, which is just 1 of the APIs in ENSApi.

Additionally, more specifically, this is an infra-level implementation of this fallback. It would be nice to have this distinction be more explicit to help avoid confusion with the app-level implementation of this fallback.

Taking a step back, I also see a broader opportunity for this app to evolve into across time. I'm not trying to suggest we execute on the following ideas right at this moment. Instead, sharing the following ideas as more of a conceptual vision / future narrative for this app's role in the broader ENSNode ecosystem.

Imagine you're another team running your own ENSNode deployment within your own data center. We're already well aware of how there's some meaningful operational complexity and risks to manage here to do it in a mature way. For example: navigating zero-downtime version upgrades, or recovering from issues such as RPCs getting stuck on ENSIndexer, or perhaps some bug in Ponder that causes indexing progress to get stuck where the only way to recover is to restart indexing again from scratch, potentially also wiping Ponder's RPC cache if something got bugged there.

So wouldn't it be nice if we offered a more general solution to this for people hosting their own ENSNode instances?

For example: At an infra-level they could have some load balancer / proxy such as Traefik / etc.. that conditionally fails over to this app if their ENSNode instance is having trouble for some reason. If that happens, they could configure this app to forward requests:
1. ... for the ENS Subgraph API to The Graph (existing functionality of this app)
2. ... for all ENSApi APIs to another instance of ENSApi (new functionality I'm brainstorming here)

This way, as we build up the strength of our own hosting and devops, we can offer this strength and reliability to the community to cover them as a fallback as they go through their own operational processes, pains / issues.

I also see value at a narrative and branding level to give this app a name that has some sex appeal and communicates a positive narrative with positive framing. It should give an emotion of security and reliability in ENSNode as hardcore foundational infrastructure for ENS rather than creating an emotion of failure or insecurity.

What do you think?

apps/fallback-ensapi/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    // provide existing method/body
+    method: c.req.method,
+    body: await c.req.text(),
+    // override headers to just provide Content-Type
+    headers: { "Content-Type": "application/json" },
+  });
+});
+
+// 503 everything else
+app.all("/*", (c) => c.json({ message: "Service Unavailable" }, 503));
```
lightwalker-eth: For the root path, what do you think about applying the following idea for this app? https://github.com/namehash/ensnode/issues/1409

Goals include: In the event of putting out fires during an incident I can imagine it would be useful to see if a request to hostname X is being routed to this app or to ENSApi.

lightwalker-eth (commented):
@shrugs Hey great to see this! Reviewed and shared feedback 👍

apps/ensapi/src/lib/thegraph.ts
```diff
@@ -1,5 +1,5 @@
-import type { ENSNamespaceId } from "@ensnode/datasources";
 import type { TheGraphFallback } from "@ensnode/ensnode-sdk";
+import { makeTheGraphSubgraphUrl } from "@ensnode/ensnode-sdk/internal";
 
 import type { EnsApiConfig } from "@/config/config.schema";
```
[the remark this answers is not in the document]
shrugs: oh, yeah that was a mistake during the moving to shared

apps/fallback-ensapi/package.json
shrugs: 1. this _is_ a fallback for the whole ENSApi, not just the subgraph, it just only has the ability to handle requests for the subgraph. if ENSApi is unhealthy, ALL traffic goes to this service. it _is_ an ENSApi fallback
    - i chose `fallback-ensapi` over `ensapi-fallback` to minimize confusion when reading, but we can swap the order if it's important
3. for this specific usecase you outlined, consumers shouldn't be running this app (fallback-ensapi): they should just load balance to our production instances if their own falls over (and then our production can fall back to this fallback-ensapi if/when our own ensapi is down)
    - IF a consumer wants additional guarantees (i.e. if they want to handle the case that a) their _own_ infra is down and b) _our_ infra is _also_ down and c) they're _just_ using the /subgraph api that has an acceptable fallback, THEN they can run fallback-ensapi themselves, in which case they clearly know what they're doing
3. for the proposed usecase of "fallback to another ensapi instance", that behavior belongs in the load balancer, NOT in this app. this app is the last possible destination for an ENSApi request that can't be handled by other, operating infrastructure. at that point, there shouldn't be any other, operating infrastructure to forward a request to.
4. i don't see this as a user-facing app, it doesn't need an ENS* branded name. it's background infrastructure.

apps/fallback-ensapi/src/lib/parse-host-header.ts
```diff
@@ -0,0 +1,16 @@
+import type { ConfigTemplateId } from "@ensnode/ensnode-sdk/internal";
+
+/**
+ * Parses a Host header value into a ConfigTemplateId.
+ *
+ * @param value the Host header value
+ * @returns a ConfigTemplateId or null if not matched
+ */
+export function parseHostHeader(value: string): ConfigTemplateId | null {
```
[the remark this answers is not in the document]
shrugs: adds complexity and configuration overhead for little benefit, imo

apps/fallback-ensapi/README.md
```diff
@@ -0,0 +1,42 @@
+# Fallback ENSApi
+
+A lightweight AWS Lambda function that handles ENSApi request that cannot be otherwise handled (due to catastrophic system failure):
+- proxies Subgraph GraphQL API requests to The Graph's ENS Subgraph
+- 503s everything else
```
lightwalker-eth:
```suggestion
A lightweight AWS Lambda that provides an infrastructure-level fallback for API requests sent to the ENSApi instances hosted by NameHash.

NameHash operates infrastructure-level monitoring of the health of all of their ENSNode deployments. In the event these monitoring systems determine that a NameHash operated ENSNode deployment has become "unhealthy" for any reason, the monitoring system will automatically redirect all API requests sent to the ENSApi instance associated with an "unhealthy" ENSNode to this "Fallback ENSApi" Lambda instead until the monitoring systems determine that ENSNode instance is "healthy" again.

This "Fallback ENSApi" Lambda proxies ENS Subgraph GraphQL API requests originally sent to ENSApi to The Graph's ENS Subgraph using NameHash's API key for The Graph.

For data consistency, this fallback is exclusively enabled for NameHash's ENSNode deployments that use a "subgraph-style" ENSNode config template. More specifically:
* Mainnet: https://api.mainnet.ensnode.io/
* Sepolia: https://api.sepolia.ensnode.io/

Therefore, if this infrastructure-level fallback is activated for the following NameHash ENSNode deployments that implement an "alpha-style" ENSNode config template, this Lambda will NOT fallback to The Graph and instead will return HTTP 503 (Service Unavailable) errors in response to any API request:
* Mainnet: https://api.alpha.ensnode.io/
* Sepolia: https://api.alpha-sepolia.ensnode.io/

Additionally, note that this Lambda exclusively implements fallbacks for the ENS Subgraph GraphQL API. ENSApi implements a number of additional APIs. Therefore, no matter which NameHash ENSNode deployment is being queried, this Lambda will return HTTP 503 (Service Unavailable) errors in response to any API request that is not for the ENS Subgraph GraphQL API.

## Observability

You can observe if this fallback ENSApi Lambda has been activated for any NameHash deployed ENSNode instance by querying the root path of the ENSApi service. If the page title is "ENSApi" then the fallback is not currently activated, however if the page title is "ENSApi Fallback" then the fallback is activated.
```

apps/fallback-ensapi/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
```
lightwalker-eth:
```suggestion
app.all("/*", (c) => c.json({ message: "Fallback Service Unavailable" }, 503));
```

What do you think? Idea here is to help us identify if a 503 error is coming from this lambda or not.

apps/fallback-ensapi/src/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    <h1>Hello, World!</h1>
+    <p>You've reached the root of the ENSApi Fallback. You might be looking for ENSNode's <a href="https://ensnode.io/docs/">documentation</a>.</p>
+</body>
+</html>
+`),
+);
+
+app.get("/health", async (c) => {
+  if (!THEGRAPH_API_KEY) return c.json({ message: "THEGRAPH_API_KEY not configured" }, 500);
+  return c.json({ message: "ok" });
```
lightwalker-eth:
```suggestion
  return c.json({ message: "Fallback ok" });
```

apps/fallback-ensapi/src/lib/parse-host-header.ts
```diff
@@ -0,0 +1,16 @@
+import type { ConfigTemplateId } from "@ensnode/ensnode-sdk/internal";
+
+/**
+ * Parses a Host header value into a ConfigTemplateId.
```
lightwalker-eth:
```suggestion
 * Parses a Host header value into a ConfigTemplateId.
 *
 * NOTE: This logic is specifically tailored for NameHash hosted ENSNode deployments.
```

apps/fallback-ensapi/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
lightwalker-eth: Thanks for reply 👍 

Some follow ups:

1. Ok cool, then it seems we should create some changeset identifying the creation of this fallback ENSApi Lambda?
2. I'm concerned about each time we introduce exceptions to otherwise universal principles because it complicates narratives and mental models. For example I'm currently sharing a bunch of feedback with @stevedylandev in https://github.com/namehash/ensnode/pull/1399 where we're trying to create nice docs for people new to our team to wrap their head around all the details of how all our different release scripts work. If you think its important for the version of this lambda to be tracked independently then I'm really ok with that but then please take the lead to ensure the necessary exception cases are documented in all the relevant places in PR 1399. It's getting complex to maintain a mental model for all this stuff and each exception adds cost.
3. We're in a bit of a tricky situation where this PR is introducing new complexity to releases in parallel with the work on PR 1399 that is working to provide improved documentation for our releases. Appreciate if you can create a follow-up issue to ensure that after PR 1399 is merged we'll remember to further update the release docs to explain all the relevant details for releases associated with `fallback-ensapi`.

apps/fallback-ensapi/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "type": "module",
+  "description": "ENSNode Fallback API",
```
lightwalker-eth:
```suggestion
  "description": "Infrastructure-level ENSApi fallback",
```

Goal: This provides a fallback to ENSApi but not other services within an ENSNode instance.

lightwalker-eth (approved):
@shrugs Thanks for the updates and replies to my earlier feedback. Completed a fresh round of reviews. Please take the lead to merge this PR when ready 👍

.github/workflows/release.yml
```diff
@@ -71,18 +72,28 @@ jobs:
...
+              or .name == "@ensnode/ponder-subgraph"
+            ))
+
+      - name: Filter Published Packages For Lambdas
+        uses: cloudposse/github-action-jq@main
+        id: publishedLambdas
+        with:
+          compact: true
+          input: ${{ steps.changesets.outputs.publishedPackages }}
+          # filter publishedPackages for lambdas (projects whose artifact is a dist/lambda.zip)
```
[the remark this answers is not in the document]
shrugs: The action as written specifically only uploads dist/lambda.zip. There's only one of these in the monorepo. If in the future we have multiple lambdas and they publish to some other name, we can provide a file argument.

PR #1412, opened 2025-12-18
feat(ensapi): introduces `GET /amirealtime` endpoint

### LINKED ISSUE #834: Create HTTP endpoint to monitor realtime indexing status

tk-o:
Resolves #834

### DIFF
--- a/.changeset/great-wombats-work.md
+++ b/.changeset/great-wombats-work.md
@@ -1,0 +1,5 @@
+---
+"ensapi": minor
+---
+
+Disabled viem's built-in Transport Ranking feature in order to minimize net_listening RPC credit ussage.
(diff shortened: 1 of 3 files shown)


apps/ensapi/src/index.ts
```diff
@@ -51,6 +56,56 @@ app.get("/health", async (c) => {
   return c.json({ ok: true });
 });
 
+// Set default `maxRealtimeDistance` for `GET /amirealtime` endpoint to one minute.
```
shrugs: can we move this into a `src/handlers/amirealtime-api.ts` and mount it like the others? so 

```ts
// in index.ts
app.route('/amirealtime', amirealtimeApi)

// in amirealtime-api.ts
app.get('/', ...)
```

should work

Issue #1407, opened 2025-12-17
Build ENSRainbow config


djstrong:
Make it similar as in ENSIndexer and ENSApi (using schema, Zod).

PR #1425, opened 2025-12-22
Build ENSRainbow config

djstrong:
Related to #1407

# Lite PR

## Summary

- Built new Zod-based configuration system for ENSRainbow, replacing ad-hoc env parsing with structured validation
- Added `/v1/config` endpoint returning public config (version, label set, records count) and deprecated `/v1/version`
- Server now requires an initialized database (precalculated record count) before starting; empty or uninitialized DB yields a clear error and directs you to run ingestion first

---

## Why

- Centralizes configuration management with proper validation and type safety
- Provides comprehensive public config endpoint for clients to discover service capabilities
- Ensures `/v1/config` is always accurate and avoids serving with an empty DB; run ingestion before starting the server
- Improves developer experience with better error messages and validation

---

## Testing

- Added comprehensive test suite for config schema validation (`config.schema.test.ts`) covering success cases, validation errors, invariants, and edge cases
- Updated existing CLI and server command tests to work with new config system
- Verified new `/v1/config` endpoint in server command tests

---

- The old `/v1/version` endpoint is deprecated but still functional for backward compatibility
- Config is built at module load time and will exit process on validation failure (appropriate for CLI apps)
- **Behavior change:** Deployments that previously started the server with an empty DB will now get a clear error and must run ingestion first (documented in the changeset for release notes)

---

### DIFF
--- a/apps/ensrainbow/src/config/config.schema.ts
+++ b/apps/ensrainbow/src/config/config.schema.ts
@@ -2,7 +2,7 @@ import { isAbsolute, resolve } from "node:path";
 
 import { prettifyError, ZodError, z } from "zod/v4";
 
-import { OptionalPortNumberSchema } from "@ensnode/ensnode-sdk/internal";
+import { OptionalPortNumberSchema, PortNumberSchema } from "@ensnode/ensnode-sdk/internal";
 
 import { ENSRAINBOW_DEFAULT_PORT, getDefaultDataDir } from "@/config/defaults";
 import type { ENSRainbowEnvironment } from "@/config/environment";
@@ -74,11 +74,12 @@ export interface ServeCommandCliArgs {
 
 export function buildServeCommandConfig(args: ServeCommandCliArgs): ServeCommandConfig {
   try {
+    const port = PortNumberSchema.parse(args.port);
     const dataDir = AbsolutePathSchemaBase.parse(args["data-dir"]);
-    return { port: args.port, dataDir };
+    return { port, dataDir };
   } catch (error) {
     if (error instanceof ZodError) {
-      throw new Error(`Invalid data-dir: \n${prettifyError(error)}\n`);
+      throw new Error(`Invalid serve command arguments: \n${prettifyError(error)}\n`);
     }
     throw error;
   }
--- a/apps/ensrainbow/src/config/config.schema.test.ts
+++ b/apps/ensrainbow/src/config/config.schema.test.ts
@@ -322,14 +322,43 @@ describe("buildServeCommandConfig", () => {
 
   it("throws on empty data-dir", () => {
     expect(() => buildServeCommandConfig({ port: 4000, "data-dir": "" })).toThrow(
-      /Invalid data-dir/,
+      /Invalid serve command arguments/,
     );
   });
 
   it("throws on whitespace-only data-dir", () => {
     expect(() => buildServeCommandConfig({ port: 4000, "data-dir": "   " })).toThrow(
-      /Invalid data-dir/,
+      /Invalid serve command arguments/,
     );
+  });
+
+  it("throws when port is less than 1", () => {
+    expect(() => buildServeCommandConfig({ port: 0, "data-dir": "/valid/path" })).toThrow(
+      /Invalid serve command arguments/,
+    );
+  });
+
+  it("throws when port is greater than 65535", () => {
+    expect(() => buildServeCommandConfig({ port: 65536, "data-dir": "/valid/path" })).toThrow(
+      /Invalid serve command arguments/,
+    );
+  });
+
+  it("throws when port is negative", () => {
+    expect(() => buildServeCommandConfig({ port: -1, "data-dir": "/valid/path" })).toThrow(
+      /Invalid serve command arguments/,
+    );
+  });
+
+  it("throws when port is a float", () => {
+    expect(() => buildServeCommandConfig({ port: 3000.5, "data-dir": "/valid/path" })).toThrow(
+      /Invalid serve command arguments/,
+    );
+  });
+
+  it("accepts port at boundary values", () => {
+    expect(buildServeCommandConfig({ port: 1, "data-dir": "/valid/path" }).port).toBe(1);
+    expect(buildServeCommandConfig({ port: 65535, "data-dir": "/valid/path" }).port).toBe(65535);
   });
 });
 


apps/ensrainbow/src/config/config.schema.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Invariant enforcement
+   *
+   * We enforce invariants across multiple values parsed with `ENSRainbowConfigSchema`
+   * by calling `.check()` function with relevant invariant-enforcing logic.
+   * Each such function has access to config values that were already parsed.
+   */
+  .check(invariant_dbSchemaVersionMatch);
+
+export type ENSRainbowConfig = z.infer<typeof ENSRainbowConfigSchema>;
```
tk-o: To stay aligned with rest of the codebase, I suggest replacing `ENS*` type name prefix with `Ens*`.
```suggestion
export type EnsRainbowConfig = z.infer<typeof EnsRainbowConfigSchema>;
```

apps/ensrainbow/src/lib/api.ts
```diff
@@ -101,6 +102,29 @@ export async function createApi(db: ENSRainbowDB): Promise<Hono> {
...
+        {
+          status: StatusCode.Error,
+          error: countResult.error,
+          errorCode: countResult.errorCode,
+        },
+        500,
+      );
+    }
+
+    const publicConfig = buildENSRainbowPublicConfig(server.getServerLabelSet(), countResult.count);
```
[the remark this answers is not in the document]
djstrong: It was done this way but `ENSRainbowPublicConfig` was not using any field from `ENSRainbowConfig` and it was causing linter errors/warnings and @tk-o removed passing config.

apps/ensrainbow/src/lib/api.ts
```diff
@@ -101,6 +102,29 @@ export async function createApi(db: ENSRainbowDB): Promise<Hono> {
     return c.json(result, result.errorCode);
   });
 
+  api.get("/v1/config", async (c: HonoContext) => {
+    logger.debug("Config request");
+    const countResult = await server.labelCount();
```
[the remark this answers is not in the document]
djstrong: Yes, it reads one key from db.

apps/ensrainbow/src/config/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  dbSchemaVersion: number;
+
+  /**
+   * Optional label set configuration that specifies which label set to use.
+   *
+   * A label set defines which labels (domain name segments) are available for label healing.
+   * Both `labelSetId` and `labelSetVersion` must be provided together to create a "fully pinned"
+   * label set reference, ensuring deterministic and reproducible label healing.
+   *
+   * If not provided, ENSRainbow will start without any label set loaded, and label healing
```
lightwalker-eth: Not sure what "the management API" is?

I assume ENSRainbow should refuse to start if this is undefined or a valid server label set is not configured?

apps/ensrainbow/src/config/types.ts
```diff
@@ -0,0 +1,84 @@
...
+   * The absolute path to the data directory where ENSRainbow stores its database and other files.
+   *
+   * This directory will contain:
+   * - The LevelDB database file with label set data
+   * - Any temporary files created during operation
+   *
+   * If a relative path is provided in the environment variable, it will be resolved to an
+   * absolute path relative to the current working directory.
+   *
+   * Default: `{cwd}/data`
```
lightwalker-eth: Defaults should not be documented here. Those ideas belong in a different layer.

apps/ensrainbow/src/config/types.ts
```diff
@@ -0,0 +1,84 @@
...
+   * The database schema version expected by the code.
+   *
+   * This version number corresponds to the structure of the LevelDB database that ENSRainbow
+   * uses to store label set data. If the version in the environment doesn't match the version
+   * expected by the code, the application will fail to start.
+   *
+   * This prevents version mismatches between the codebase and the database schema, which could
+   * lead to data corruption or runtime errors.
+   *
+   * Default: {@link DB_SCHEMA_VERSION} (currently 3)
```
lightwalker-eth: Defaults should not be documented here. Those ideas belong in a different layer.

djstrong:
There is a problem with current solution and also layered Configs (EnvConfig -> ArgsConfig). If env value is invalid but arg value is valid then the validation process will fail on building EnvConfig but it should pass IMO.
If we do not validate env values then CLI defaults might be invalid and the validation should be done after merging env values with arg values - is it the correct way?

apps/ensrainbow/src/config/config.schema.ts
```diff
@@ -0,0 +1,85 @@
...
+    error: "Path must be a non-empty string.",
+  })
+  .transform((path: string) => {
+    if (isAbsolute(path)) {
+      return path;
+    }
+    return resolve(process.cwd(), path);
+  });
+
+const DataDirSchema = AbsolutePathSchemaBase;
```
lightwalker-eth: Is there a special reason to define `DataDirSchema`? Why not just reuse `AbsolutePathSchemaBase`?

apps/ensrainbow/src/config/public.ts
```diff
@@ -0,0 +1,18 @@
+import packageJson from "@/../package.json" with { type: "json" };
+
+import type { EnsRainbowServerLabelSet } from "@ensnode/ensnode-sdk";
+import type { EnsRainbow } from "@ensnode/ensrainbow-sdk";
+
+import type { ArgsConfig } from "./types";
+
+export function buildENSRainbowPublicConfig(
```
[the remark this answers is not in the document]
lightwalker-eth: @djstrong Ah I see! Thank you.

To help make what's going on here more self-documenting, what do you think about defining an object such as a `DbConfig` that would hold the `labelSet` and `recordsCount` fields?

This would then make it self-documenting where these values are coming from. They're coming from the db!

And I note how you're not using the `ArgsConfig` at all and its currently listed there only for semantic purposes. Now that I'm better understanding how all the pieces relate here -- maybe we should remove the `ArgsConfig` that's passed to this function and instead have it only take a new `DbConfig` object?

This would make everything more clear. It would then be clear that the `ENSRainbowPublicConfig` is purely a function of the `DbConfig`.

For building a `DbConfig` object, my assumption is that it would be created through a helper function that took both an `EnvConfig` and an `ArgsConfig` as input?

Appreciate your advice.

apps/ensrainbow/src/cli.ts
```diff
@@ -36,6 +28,11 @@ interface IngestProtobufArgs {
   "data-dir": string;
 }
 
+/**
+ * Arguments for the 'serve' command.
+ *
+ * Note: CLI arguments take precedence over environment variables.
+ */
 interface ServeArgs {
```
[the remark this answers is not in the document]
lightwalker-eth: I'm not 100% sure I understand what you're describing here, but assuming so, that sounds good :+1: Please proceed 👍

lightwalker-eth (approved):
@djstrong Sharing replies to questions 👍

apps/ensrainbow/src/lib/server.ts
```diff
@@ -13,11 +13,30 @@ import {
...
+export async function buildDbConfig(server: ENSRainbowServer): Promise<DbConfig> {
+  const countResult = await server.labelCount();
+  if (countResult.status === StatusCode.Error) {
+    throw new Error(
+      `Failed to read record count from database: ${countResult.error} (errorCode: ${countResult.errorCode})`,
+    );
+  }
+
+  return {
+    labelSet: server.getServerLabelSet(),
```
lightwalker-eth: I assume this also could fail, just like `server.labelCount()`?

PR #1427, opened 2025-12-22
WIP refactor(ensindexer): use local ponder client

tk-o:
Local Ponder Client is a point of integration between ENSIndexer and Ponder applications. It uses Ponder SDK to interact with Ponder APIs.

All `@ensnode/ensnode-sdk` dependencies in Ponder SDK were replaced with local implementations.


packages/ponder-sdk/src/shared.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Chain Name
+ *
+ * Used as a type for object keys expressing Ponder ideas, such as
+ * chain status, or chain metrics.
+ */
+export type ChainName = string;
```
lightwalker-eth: A few ideas:

1. I imagine we should move this definition into `chains.ts`? It's deeply related to chains.
2. I imagine we should also define some utility function in `chains.ts` that maps from a `ChainName` as returned by Ponder into a more universally interpretable `ChainId`? At least for the ChainId / ChainName that ENSNode supports (please remember the ideas I shared about the scope of `ponder-sdk`).

packages/ponder-sdk/src/block-refs.ts
```diff
@@ -36,36 +73,40 @@ export interface ChainBlockRefs {
...
+ * - no historical total blocks metric found for a chain,
+ * - could not get BlockRefs for a chain.
  */
 export async function getChainsBlockRefs(
-  chainIds: ChainIdString[],
+  chainNames: ChainName[],
   chainsBlockrange: Record<ChainName, Blockrange>,
-  metrics: PrometheusMetrics,
+  historicalTotalBlocksForChains: Record<ChainName, number>,
   publicClients: Record<ChainName, PublicClient>,
```
lightwalker-eth: This is an example of the spaghetti I'm hoping to see cleaned up. The layers of responsibility are getting all mixed up.

There should be a big alarm bell here: Where are these `publicClients` coming from? Does this idea belong in this layer or another layer? What are your layers? Code should not be placed in random places.

apps/ensindexer/ponder/local-client/local-ponder-client.ts
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Cached Chain Block Refs
+   *
+   * {@link ChainBlockRefs} for each indexed chain.
+   */
+  private chainsBlockRefs = new Map<ChainName, ChainBlockRefs>();
+
+  private readonly ponderClient: PonderClient;
+
+  constructor(ponderApplicationUrl: URL) {
```
lightwalker-eth: Should this only take as input a port as the local ponder client should always be localhost?

PR #1442, opened 2025-12-29
fix: support RegistryOld fallback for Registry node-resolver-relations

shrugs:
# Fix Acceleration Mismatch

---

<details>

1. **Database query logic in `find-resolver.ts:203-229`** - The `or()` clause that adds RegistryOld fallback and the sorting logic that ensures Registry records take precedence over RegistryOld records
2. **Simplified control flow in `find-resolver.ts:242-271`** - Changed from iterating all node-resolver relations to taking just the first (after sorting). Confirm this matches the intended behavior.

</details>

---

## Problem & Motivation

<details>

- **Problem**: Protocol Accelerated Resolution API requests were failing for legacy unmigrated ENS names
- **Root cause**: The ENS Root Registry contract is actually `ENSRegistryWithFallback`, which falls back to `RegistryOld` storage for nodes that haven't been migrated to the new registry. The acceleration code wasn't implementing this fallback logic.
- **Impact**: Any unmigrated name would return NULL_RESULT when queried through the accelerated path, despite having valid resolver data in RegistryOld
- **Why now**: Discovered when testing acceleration with older ENS names that predate the registry migration

</details>

---

## What Changed (Concrete)

<details>

1. Added RegistryOld datasource contract reference to `find-resolver.ts`
2. Modified `nodeResolverRelation` database query to include records from both Registry and RegistryOld when querying the ENS Root Registry
3. Added sorting logic to prefer Registry records over RegistryOld records (implementing fallback semantics)
4. Simplified iteration logic: instead of looping through all node-resolver relations, now takes the first record after sorting
5. Minor cleanup: removed explicit type annotations from Registry event handler (apps/ensindexer/src/plugins/protocol-acceleration/handlers/Registry.ts:102)

</details>

---

## Design & Planning

<details>

**Approach**: Mirror the ENSRegistryWithFallback contract logic in the query layer by fetching from both registries and using sort order to implement precedence.

**Alternatives considered**:
- Assigning registry old records directly to the registry, but hacky.

**Why this approach**:
- Single query with conditional logic keeps performance acceptable
- Sorting implements fallback semantics cleanly without branching
- Minimal changes to existing code paths

</details>

- Planning artifacts: None, straightforward bug fix
- Reviewed / approved by: N/A

---

## Self-Review

<details>

- **Logic simplified**: Original implementation iterated through all node-resolver relations; realized after review that we only need the first record after proper sorting
- **Type safety**: Removed unnecessary function re-wrapping in Registry.ts handler

</details>

- Bugs caught: None beyond the original issue
- Logic simplified: Changed iteration to take first record after sorting
- Naming / terminology improved: N/A
- Dead or unnecessary code removed: Explicit type annotations in Registry.ts

---

<details>

**Impact**:
- API consumers making accelerated resolution requests for legacy/unmigrated ENS names will now get correct results instead of null
- No breaking changes - purely additive fix
- Performance impact negligible (single query with slightly more complex filter)

**Operators**:
- No configuration changes required
- No migration needed - uses existing indexed data
- Can deploy just ENSApi to patch production

**Terminology**:
- Comments at lines 194-199 explain the ENSRegistryWithFallback contract behavior for future maintainers

</details>

- Public APIs affected: None (behavioral fix, same interface)
- Docs updated: Inline comments added explaining fallback logic

---

## Testing Evidence

<details>

**Testing performed**:
- Manual testing with legacy unmigrated ENS names via accelerated resolution API
- Verified fallback order: Registry records preferred when both exist, RegistryOld used when Registry has no record

**Known gaps**:
- No automated test coverage for this specific fallback scenario
- Integration tests would require larger overhead

**If this is wrong**:
- Legacy unmigrated names would return null resolvers (regression to previous bug)
- Or, if sort order is wrong, could return stale RegistryOld data when migrated data exists in Registry

</details>

- Testing performed: Manual validation with legacy ENS name (`d8da6bf26964af9d7eed9e03e53415d37aa96045.addr.reverse`)
- Known gaps: No automated test coverage for registry fallback scenarios
- What reviewers have to reason about manually: Database query result ordering correctness and the sort comparator logic (lines 226-233)

---

## Scope Reductions

<details>

**Deferred**:
- Adding comprehensive test coverage for registry fallback scenarios (would require significant test infrastructure setup)
- Applying similar fallback logic to other protocol acceleration queries (only needed for resolver lookups currently)

</details>

- Follow-ups: Consider adding integration tests for registry fallback once test infrastructure matures
- Why they were deferred: Bug fix is time-sensitive; test infrastructure work is a larger project

---

## Risk Analysis

<details>

**Assumptions**:
- Sort order correctly prioritizes Registry over RegistryOld when nodes match

**Failure modes**:
- If sort logic is incorrect, could return stale RegistryOld resolver when migrated resolver exists

**Blast radius**:
- Limited to Protocol Accelerated Resolution API
- Non-accelerated paths unaffected
- Only affects resolver lookup, not other ENS operations

</details>

- Risk areas: Sort order correctness
- Mitigations or rollback options: Easy revert via deploying previous ENSApi version
- Named owner if this causes problems: @shrugs (PR author)

---

- [x] Relevant changesets are included (or explicitly not required) - ✅ Included in .changeset/fast-geese-poke.md

### DIFF
--- a/.changeset/fast-geese-poke.md
+++ b/.changeset/fast-geese-poke.md
@@ -1,0 +1,5 @@
+---
+"ensapi": minor
+---
+
+Fixed issue regarding Protocol Accelerated Resolution API requests and legacy unmigrated names, which should now resolve correctly when accelerated.


lightwalker-eth (approved):
@shrugs Nice work, thank you. Please take the lead to merge this when ready 🚀

PR #1495, opened 2026-01-08
fix: ens-referrals package

shrugs:
- adds ens-referrals package to fixed
- bumps version to match existing
- let's merge this and see what changeset reports as next version

skipping pr template, very minimal


lightwalker-eth (approved):
@shrugs Looks good. Let's try it 👍

PR #1515, opened 2026-01-13
ENSv2 Event Timestamps

shrugs:
i forgot to include the block timestamp in event metadata for the ensv2 plugin ಠ_ಠ


lightwalker-eth (approved):
@shrugs Good update. Thanks 👍

PR #1516, opened 2026-01-13
feat: new sepolia-v2 namespace for temp ensv2 deployment

shrugs:
makes a new temporary namespace for the test deployments of ensv2 contracts on sepolia.

this pr represents the temporary ad hoc deployment of ensv2 to the sepolia chain as an independent sepolia namespace to avoid interrupting mainstream usage of the existing sepolia namespace

confirmed indexed to latest


lightwalker-eth (commented):
@shrugs Thanks for updates. Seeking your advice, thanks 👍

PR #1531, opened 2026-01-14
attempt to build css before running ensadmin build

notrab:
@Y3drk draft PR to test if Vercel builds properly now with the compiled CSS.


Y3drk (approved):
Thanks for the help @notrab 🫡 . I'll give it a try and maybe try to polish it later on.

PR #1534, opened 2026-01-15
chore: update subgraph docs

lightwalker-eth:
# Background info

- Today I shut down our old GraphNode instances in AWS. They are no longer providing a meaningful benefit and just add cost / ops complexity.

# Goals

- Remove docs related to our old hosted GraphNode instances which no longer exist.
- Take all the ideas that used to live in `subgraph-compatibility.mdx` transfer them into `what-is-the-ens-subgraph.mdx`. Update these ideas a bit based on advancements since the last update to these docs.
- Update links where relevant to avoid creating any broken links (after removing `subgraph-compatibility.mdx`).

# Notes

- No changeset is relevant for this PR.

### DIFF
--- a/docs/ensnode.io/src/content/docs/docs/concepts/what-is-the-ens-subgraph.mdx
+++ b/docs/ensnode.io/src/content/docs/docs/concepts/what-is-the-ens-subgraph.mdx
@@ -75,7 +75,7 @@ The feature set used by `ensjs` and `ens-app-v3` is fully supported: see the [we
 ### Planned Features
 
 - any open [issues](https://github.com/namehash/ensnode/issues) regarding Subgraph-Compatibility
-- [subgraph `_Meta_` object](https://thegraph.com/docs/en/subgraphs/querying/graphql-api/#subgraph-metadata)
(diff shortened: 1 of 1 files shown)


docs/ensnode.io/src/content/docs/docs/concepts/what-is-the-ens-subgraph.mdx
```diff
@@ -29,37 +29,115 @@ Each Graph Node server can run any number of "subgraphs". Each subgraph is essen
...
+
+:::note[Contributions]
+If you'd like to contribute to these features, please [open a Pull Request on GitHub](https://github.com/namehash/ensnode).
+:::
+
+### Unplanned Features
+
+The following features of the subgraph graphql api are explicitly unsupported and are not planned.
+
+- [1-level-nested Entity `_orderBy` param](https://thegraph.com/docs/en/subgraphs/querying/graphql-api/#example-for-nested-entity-sorting)
```
tk-o: Anchor link needs to be updated to point to the exact section on the target page.
```suggestion
- [1-level-nested Entity `_orderBy` param](https://thegraph.com/docs/en/subgraphs/querying/graphql-api/#nested-entity-sorting-example)
```

Issue #1564, opened 2026-01-27
ENSv2 Domain Search/Filters


shrugs:
as specified by https://hackmd.io/@oneofthemanymatts/SkypJFNH-e

Issue #1594, opened 2026-02-01
ENSv2 Refactor Registration/Renewal Superceding Logic


shrugs:
- index all entities with their index
- use a lookup table to go from key -> id and then lookup again to get the latest of each
- update latest registration subquery in find-domains

PR #1595, opened 2026-02-01
ENSv2 Find Domain Ordering Support

shrugs:
closes https://github.com/namehash/ensnode/issues/1564

  1. **`cursorFilter` in `find-domains.ts`** — NULL-handling logic for keyset pagination. NULLs can't participate in tuple comparison in Postgres, so they're handled explicitly. explicit `::text`/`::bigint` casts on cursor values to avoid
  parameter type inference failures in tuple context.
  2. **composite `DomainCursor` design** — cursor encodes `by`/`dir` alongside the sort column value. allows cursor validation (stale cursors with mismatched order args throw) and enables tuple comparison without subqueries. uses superjson for
  bigint encoding; decode is wrapped in try/catch for malformed input.
  3. **`findDomains` CTE structure** — unions v1 and v2 domains, joins head label for partial matching + NAME ordering, joins latest registration for REGISTRATION_* ordering. verify the LEFT JOINs and NULLS LAST behavior.

  ---

  ## Problem & Motivation

  - `Query.domains` and `Account.domains` only supported ordering by `id`, making search/autocomplete UX hard to build
  - need to support ordering by `NAME`, `REGISTRATION_TIMESTAMP`, and `REGISTRATION_EXPIRY`
  - keyset pagination over non-id columns requires composite cursors that include the sort column value (to enable tuple comparison), otherwise pagination breaks under non-unique sort values

  ---

  ## What Changed (Concrete)

  1. new `DomainsOrderBy` enum (`NAME | REGISTRATION_TIMESTAMP | REGISTRATION_EXPIRY`) and `DomainsOrderInput` input type
  2. new `OrderDirection` enum (`ASC | DESC`)
  3. `order` arg added to `Query.domains` and `Account.domains` connections
  4. `DomainCursor` — composite keyset cursor encoding `{ id, by, dir, value }` as base64 superjson. decode catches malformed input with a clear error.
  5. `findDomains` refactored into a CTE-based query builder. NAME ordering uses the head label (the varying part when a concrete path is specified; equals the leaf when there's no concrete path).
  6. `find-domains-by-labelhash-path` extracted to its own file (no behavior change)
  7. `resolveFindDomains` shared resolver replaces duplicated inline pagination logic in `query.ts` and `account.ts`
  8. `cursorFilter` handles NULL and non-NULL cursor values, with explicit `::text`/`::bigint` casts in tuple comparisons
  9. `Event.timestamp` field added (description is "TODO")
  10. `superjson` dep added

  ---

  ## Design & Planning

  - composite cursor required because simple id-based cursors break when ordering by non-unique columns
  - cursor encodes `by`/`dir` to detect stale cursors when callers change order args mid-pagination
  - superjson chosen over custom bigint serializer since it already handles the type safely
  - NULLS LAST is intentional: unregistered domains (NULL registration fields) sort to the end regardless of direction
  - NAME ordering uses head label rather than leaf label — when a concrete path is specified (e.g. `sub1.sub2.exam`), all results share the same leaf, making leaf-based sorting meaningless. head label is the varying part. when there's no
  concrete path, head = leaf = self, so it's equivalent.
  - full canonical name sorting was considered but deferred — requires a second recursive CTE per result and the canonical name isn't materialized. head label + id tiebreaker is sufficient for autocomplete.

  - Planning artifacts: none
  - Reviewed / approved by: n/a

  ---

  ## Self-Review

  fixed cursor pagination bugs with NULL values and direction mismatch after initial draft. caught that leaf label was wrong for NAME ordering (identical across all results when concrete path is specified). added explicit type casts in tuple
  comparison to avoid Postgres parameter type inference failures. wrapped cursor decode in try/catch.

  - Bugs caught: NULL cursor values in tuple comparison, leaf vs head label for NAME ordering, untyped parameters in tuple comparison, unhandled cursor decode errors
  - Logic simplified: extracted `resolveFindDomains` to DRY `query.ts` and `account.ts`
  - Naming / terminology improved: `DomainCursor` interface/namespace dual export pattern
  - Dead or unnecessary code removed: old inline pagination logic in both resolvers, leaf label join (no longer needed)

  ---

  ## Cross-Codebase Alignment

  verified `pothos/plugin-relay` `resolveCursorConnection` contract (`inverted`, `before`, `after`, `limit`). reviewed existing `cursors` util (still used elsewhere, not removed). `find-domains-by-labelhash-path` logic unchanged from prior
  implementation.

  - Search terms used: `resolveCursorConnection`, `findDomains`, `DomainCursor`, `cursors.decode`
  - Reviewed but unchanged: `cursors.ts` (still used by other resolvers)

  ---

  - **callers**: new optional `order` arg on `Query.domains` and `Account.domains`; defaults to `NAME ASC` if omitted. existing queries unaffected.

  - Public APIs affected: `Query.domains(order:)`, `Account.domains(order:)`
  - Docs updated: changeset added (`ensapi` minor)
  - Naming decisions: `DomainsOrderBy` vs `OrderBy` — scoped to avoid collision with future ordering enums on other types

  ---

  ## Testing Evidence

  - Testing performed: manual local integration testing
  - Unit tests for `DomainCursor` encode/decode, and `isEffectiveDesc`
  - Known gaps: no tests for `cursorFilter`, NULL cursor edge cases, cursor decode error handling, or cursor validation errors
  - What reviewers have to reason about manually: the NULL cursor handling branches in `cursorFilter` and the `isEffectiveDesc` XOR logic

  ---

  ## Scope Reductions

  - cursor format validation (decode currently trusts cursor structure after base64/superjson parse) — tracked as TODO in `domain-cursor.ts`
  - replace correlated registration subquery with a lookup table join — tracked as TODO in `find-domains.ts` (#1594)
  - full canonical name sorting (would require second recursive CTE; head label + id tiebreaker is sufficient for now)
  - tests for pagination logic

  - Why they were deferred: keep PR reviewable; lookup table is a separate infrastructure change; full name sorting adds real query cost for marginal UX benefit

  ---

  ## Risk Analysis

  - NULL handling in `cursorFilter` is hand-rolled; if the logic is wrong, pagination silently skips or duplicates rows
  - cursor format change is a breaking change for any client that persists cursors across deploys

  - Risk areas: `cursorFilter` NULL branches, cursor invalidation on deploy
  - Mitigations: cursor validation throws on `by`/`dir` mismatch; explicit casts prevent type inference failures; NULLS LAST is consistent; malformed cursors return clear errors

  ---


apps/ensapi/src/graphql-api/lib/find-domains/find-domains-by-labelhash-path.ts
```diff
@@ -1,146 +1,9 @@
...
-  type ENSv2DomainId,
-  interpretedLabelsToLabelHashPath,
-  type LabelHashPath,
-  type Name,
-  parsePartialInterpretedName,
-} from "@ensnode/ensnode-sdk";
+import type { ENSv1DomainId, ENSv2DomainId, LabelHashPath } from "@ensnode/ensnode-sdk";
 
 import { db } from "@/lib/db";
-import { makeLogger } from "@/lib/logger";
```
shrugs: kept in `find-domains.ts`

apps/ensapi/src/graphql-api/lib/find-domains/find-domains-resolver.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+/**
+ * Shared GraphQL API resolver for domains connection queries, used by Query.domains and
+ * Account.domains.
+ *
+ * @param context - The GraphQL Context, required for Dataloader access
+ * @param args - The GraphQL Args object (via t.connection) + FindDomains-specific args (where, order)
+ */
+export function resolveFindDomains(
```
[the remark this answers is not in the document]
shrugs: it's at this level that we build the specific query (using `find-domains.ts`) and apply the request's pagination constraints

apps/ensapi/src/graphql-api/schema/domain.ts
```diff
@@ -331,3 +332,25 @@ export const AccountDomainsWhereInput = builder.inputType("AccountDomainsWhereIn
...
+
+export const DomainsOrderBy = builder.enumType("DomainsOrderBy", {
+  description: "Fields by which domains can be ordered",
+  values: ["NAME", "REGISTRATION_TIMESTAMP", "REGISTRATION_EXPIRY"] as const,
+});
+
+export type DomainsOrderByValue = typeof DomainsOrderBy.$inferType;
+
+export const DomainsOrderInput = builder.inputType("DomainsOrderInput", {
+  description: "Ordering options for domains query. If no order is provided, the default isASC.",
```
lightwalker-eth:
```suggestion
  description: "Ordering options for domains query. If no order is provided, the default is ASC.",
```

apps/ensapi/src/graphql-api/lib/find-domains/find-domains.ts
```diff
@@ -0,0 +1,310 @@
...
+ *    a. Build recursive CTE to find domains matching the concrete labelHash path
+ *    b. Extract unified structure: {id, ownerId, headLabelHash}
+ * 4. Union v1 and v2 results into domainsBase CTE
+ * 5. Join domainsBase with:
+ *    - headLabel: for partial name matching (LIKE prefix) and NAME ordering
+ *    - latestRegistration: correlated subquery for REGISTRATION_* ordering
+ * 6. Apply filters (owner, partial) in the unified query
+ * 7. Return CTE with columns: id, headLabel, registrationTimestamp, registrationExpiry
+ */
+export function findDomains({ name, owner }: FindDomainsWhereArg) {
```
lightwalker-eth: Super work all around here 💯

PR #1607, opened 2026-02-05
Implement netcat listener for open port scanning on Render

djstrong:
# Lite PR

## Summary

- Added netcat listener during container initialization to keep port open while database downloads/validates
- Added `netcat-openbsd` package to Dockerfile dependencies
- Netcat listener is automatically cleaned up before the actual server starts

---

## Why

During container initialization, ENSRainbow can take a long time to download and validate the database (up to 20 minutes). Without a listener on the port during this phase, Render health checks fail and orchestration systems may mark the container as unhealthy or restart it prematurely. The netcat listener keeps the port open and accepts connections during initialization, then is cleanly terminated when the actual server is ready to start.

---

## Testing

- Tested on Render

---

- It is a quick workaround for Render.

---

### DIFF
--- a/apps/ensrainbow/scripts/entrypoint.sh
+++ b/apps/ensrainbow/scripts/entrypoint.sh
@@ -144,10 +144,11 @@ fi # End of download and extraction block
 
 # 7. Kill netcat before starting the server
 echo "Stopping netcat listener before starting server..."
-cleanup_nc
-# Clear the PID and remove trap since we've manually cleaned up
-NC_PID=""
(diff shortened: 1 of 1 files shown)


apps/ensrainbow/Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM node:24-slim AS runtime
 
 # Install only essential system dependencies for runtime
```
lightwalker-eth: @djstrong We should add documentation here explaining the background info for why netcat is being installed here. We need to document the problem we are solving and how netcat is being used to solve it.

PR #1651, opened 2026-02-18
Claude Code Setup

Goader:
# Claude Code Setup

Sharing my Claude Code setup. It is mostly automatically generated by Claude itself, but with my instructions on what should be covered and quickly verified. Note that some statements may be incorrect; the main purpose is to reduce token usage so that Claude can infer the patterns we use from a simple description rather than manually reading and analyzing the entire codebase.

Besides that, I'm using:

* `claude-code-setup` - general helpers to set up everything
* `claude-md-management` - updating, improving CLAUDE.md
* `code-review` - self-explanatory (haven't tried it yet)
* `context7` - MCP for documentation
* `typescript-lsp` - language server

All of those are official plugins. Haven't explored other plugins yet, and haven't set up any skills for manual testing and whatnot.

Not included, but I'm also using a few hooks, like running a linter after Claude edits the code.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,511 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Repository Overview
+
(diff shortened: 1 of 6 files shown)


shrugs:
closing in favor of #2287

PR #1687, opened 2026-02-26
refactor: use zod-openapi to serve openapi.json route

notrab:
# Lite PR

## Summary

- Replaced `hono-openapi`'s `openAPIRouteHandler` with `@hono/zod-openapi`'s `getOpenAPI31Document()` for serving the `/openapi.json` route, using the existing stub routes and `openapiMeta`.
- Removed unused `hono-openapi` and `@hono/standard-validator` dependencies
- Removed dead `validate.ts` file (was the only consumer of `hono-openapi`, not imported anywhere).

---

## Why

- Completes the migration to `zod-openapi` for the ensapi app. All API routes have already been migrated to use `@hono/zod-openapi`'s `createRoute`, so the spec endpoint should use the same system rather than relying on a separate `hono-openapi` library.

---

## Testing

- Type-checked with `tsc --noEmit` — passes cleanly.
- Verified `hono-openapi` and `@hono/standard-validator` have no remaining imports in the source.

---

- `openapi-meta.ts` was updated to use the dynamic `packageJson.version` instead of the previously hardcoded `"0.0.0"`.
- The localhost development server is appended to the metadata at request time in `index.ts` since it depends on config (which `openapi-meta.ts` intentionally avoids importing to keep stub route generation config-free).

---

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -324,9 +324,6 @@ importers:
       '@hono/otel':
         specifier: ^0.2.2
         version: 0.2.2(hono@4.12.2)
-      '@hono/standard-validator':
-        specifier: ^0.2.2
-        version: 0.2.2(@standard-schema/spec@1.0.0)(hono@4.12.2)
       '@hono/zod-openapi':
         specifier: ^1.2.2
         version: 1.2.2(hono@4.12.2)(zod@4.3.6)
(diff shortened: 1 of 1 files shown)


packages/ensnode-sdk/src/ensapi/api/resolution/zod-schemas.ts
```diff
@@ -38,7 +44,9 @@ export const makeResolvePrimaryNameResponseSchema = () =>
  */
 export const makeResolvePrimaryNamesResponseSchema = () =>
   z.object({
-    names: z.record(z.number(), z.string().nullable()),
+    names: withOpenApi(z.record(z.number(), z.string().nullable()), {
+      propertyNames: { type: "string", pattern: "^[0-9]+$" },
```
lightwalker-eth: Why this pattern?

Issue #1168, opened 2025-10-09
Connection view: update ENSRainbow section values


tk-o:
Depends on:
- #1166 

Currently, the section presents data from `EnsRainbowClientLabelSet`, which is based on ENSIndexer application settings. However, it should use a dynamic `EnsRainbowServerLabelSet` value fetched from ENSRainbow service.

<img width="1266" height="1053" alt="Image" src="https://github.com/user-attachments/assets/bf037ea3-a1c7-48f3-a35d-1a908c9b13a2" />

PR #1705, opened 2026-03-02
Use ENSRainbow public config within ENSIndexer public config

### LINKED ISSUE #1166: Update ENSIndexer version info to fully cover ENSRainbow version info
Creating an object of `ENSIndexerVersionInfo` type should fully cover the value returned from ENSRainbow `/version` endpoint.

In other words, for `ENSIndexerVersionInfo` type fields:
1. The type of `ensRainbow` here should be the type of an `EnsRainbowVersionInfo` and contain nested fields inside of itself.
2. The `ensRainbowSchema` field should be removed at this level. It already lives inside `EnsRainbowVersionInfo`.

_Originally posted by @lightwalker-eth in https://github.com/namehash/ensnode/pull/1157#discussion_r2414044145_

Edit: ENSIndexer should also verify invariants match between the `EnsRainbowVersionInfo` and the `EnsIndexerVersionInfo`. For example: The client and server label sets need to be compatible. When ENSIndexer dynamically fetches `EnsRainbowVersionInfo` from ENSRainbow to build its own `EnsIndexerVersionInfo` if this invariant check fails then ENSIndexer should throw an error such that ultimately the API call to get the config of the ENSIndexer instance should trigger a HTTP 500 error. The cause of this error should be written to error logs to help ENSNode operators figure out they have a misconfiguration between their ENSIndexer and ENSRainbow instances.

Edit: ENSRainbow should stop exposing its schema version through its API. It's schema version is an internal implementation detail that shouldn't be exposed via API. @lightwalker-eth is advancing a PR now that will remove the display of this field from the ENSAdmin UI, but this field still needs to be removed from the ENSRainbow API and also from the ENSIndexer config API.

Image

Edit: We likely shouldn't create any `EnsRainbowVersionInfo` type / interface. Instead we should be creating an `EnsRainbowConfig` (or related public variant). Here we should:
- Apply all the same design patterns for building a config in ENSRainbow that we implemented in ENSIndexer.
   - Therefore, ENSRainbow should build and validate its config from environment variables on startup and then make use of the fully validated config across its code.
   - ENSRainbow should offer its public config via a `/config` API.
   - The ENSRainbow public config object should include within it fields that are NOT set via environment variables but that are relevant for consumers of the `/config` API. For example, it should include a field for the count of rainbow records it is managing.
- We should update the UI in ENSAdmin so that we show this new count of rainbow records field in the "ENSRainbow" panel when displaying the details of the active connection.

tk-o:
# Lite PR

## Summary

- Adds `ensRainbowPublicConfig: EnsRainbowPublicConfig` field to `EnsIndexerPublicConfig`
- Updates all relevant zod schemas, types, and mocks, to follow updated `EnsIndexerPublicConfig` data model
- Adds new invariant to `EnsIndexerPublicConfig` schema which ensures ENSRainbow labelset match between server labelset and client labelset.

---

## Why

- ENSIndexer needs to be able to demonstrate configuration of connected ENSRainbow service, so that ENSApi could pass this information further, to ENSAdmin UI 
- There was [a requirement formed a while ago that aimed for a nested `ensRainbow` field to be added to the ENSIndexer version info object](https://github.com/namehash/ensnode/pull/1157#discussion_r2414044145). I figured, that we could simply reference a full `EnsRainbowPublicConfig` object, so it was natural to use `ensRainbowPublicConfig` to name the new field within the ENSIndexer version info object.

---

## Testing

- Successfully executed lint, typecheck and testing suite.
- Ran ENSIndexer, ENSApi and ENSAdmin locally, together, and observed the ENSAdmin UI:
image

---

- Blocked by PR #1702
  - ENSIndexer public config object (as well as Indexing Status object) has to be read from ENSDb Client.
- Please review this PR commit-by-commit.

---

Resolves #1166, #1168.

### DIFF
--- a/packages/ensnode-sdk/src/ensindexer/config/types.ts
+++ b/packages/ensnode-sdk/src/ensindexer/config/types.ts
@@ -99,7 +99,7 @@ export interface EnsIndexerPublicConfig {
   /**
    * ENSRainbow public config
    *
-   * Represent the public config of connected ENSRainbow instance.
+   * Represents the public config of the connected ENSRainbow instance.
    */
   ensRainbowPublicConfig: EnsRainbowPublicConfig;
 


tk-o:
@greptile review

tk-o:
Converting PR to draft, as PR #1702 needs to be merged first.

packages/ensnode-sdk/src/ensindexer/config/types.ts
```diff
@@ -51,16 +51,9 @@ export interface EnsIndexerVersionInfo {
...
-   **/
-  ensRainbow: string;
-
-  /**
-   * ENSRainbow schema version
-   **/
-  ensRainbowSchema: number;
+   * ENSRainbow public config
+   */
+  ensRainbowPublicConfig: EnsRainbowPublicConfig;
```
lightwalker-eth: This field should be moved out of `EnsIndexerVersionInfo` and into `EnsIndexerPublicConfig`.

Goal: It's not a version number like all the other fields in `EnsIndexerVersionInfo`

Issue #1698, opened 2026-02-27
ENSApi Devnet Integration Test in CI


shrugs:
integrate a devnet integration test into ci
- needs to have postgres/devnet running somehow
- needs to run the suite of apps from the current commit
- likely can be extensible
  - action to run integration tests should be able to accept various configs and run a specific integration testing script in the repo so it can be extended to support things like sepolia integration testing
  - the integration tests would need to become config-dependent/config-scoped. so the current test suite actually represents an integration test against the ens-test-env namespace with these specific plugins etc
  - maybe skip the extensibility for now and just start with a PoC to get the devnet integration tests running in CI

PR #1729, opened 2026-03-05
ENSApi Integration Tests in CI

shrugs:
closes https://github.com/namehash/ensnode/issues/1698

  - **`packages/integration-test-env/src/ci.ts`** — the orchestrator. this is the main deliverable. review the phased startup, cleanup ordering, and abort-flag pattern.
  - **cleanup correctness** — execa handles subprocess cleanup on parent exit, but the reverse-order teardown and container stop logic is custom. verify the SIGINT/SIGTERM handling is sound.
  - **CI workflow** — the new `integration-tests` job in `test_ci.yml`. simple, but confirm the timeout-minutes and runner config are appropriate.

  ---

  ## Problem & Motivation

  - ensapi has integration tests (`*.integration.test.ts`) that test the graphql api against a real stack, but there was no way to run them in CI.
  - running them requires postgres, the ens-test-env devnet, ensrainbow (with a pre-built label database), ensindexer (fully indexed), and ensapi — all coordinated in the right order.
  - without CI coverage, integration test regressions can only be caught locally.

  ---

  ## What Changed (Concrete)

  1. new private workspace package `@ensnode/integration-test-env` at `packages/integration-test-env/` — owns testcontainers, execa, and ensnode SDK dependencies
  2. `ci.ts` orchestrator script that starts postgres + devnet (parallel via testcontainers), downloads ensrainbow database, starts ensrainbow/ensindexer/ensapi from source, polls for indexing completion, runs `pnpm test:integration`, then
  cleans up
  3. new `integration-tests` job in `.github/workflows/test_ci.yml` with 10-minute timeout
  4. `pnpm test:integration:ci` script at monorepo root that delegates to the new package
  5. `pnpm test:integration` script at monorepo root using vitest integration config with `--silent passed-only`
  6. fixed pre-existing broken test: `query.integration.test.ts` v1 eth domain assertion — was `toBeUndefined()` with a TODO, now correctly asserts `toMatchObject()` with `makeENSv1DomainId`
  7. renamed `ENSAPI_GRAPHQL_API_URL` → `ENSNODE_URL` in integration test client config
  8. made `download-prebuilt-database.sh` executable (mode 0644 → 0755)
  9. added `permissions: contents: read` to the workflow file

  ---

  ## Design & Planning

  - no upfront design doc. built iteratively: started with raw `docker run` commands, migrated to testcontainers for postgres (dynamic ports, built-in health checks), then added execa for subprocess management after debugging cleanup issues with
   raw `spawn`.
  - considered running ensrainbow/ensindexer/ensapi as Docker containers too, but running from source means CI tests the actual code in the PR.
  - considered using the ensrainbow Docker entrypoint.sh directly, but it has Docker-specific paths (`/app/...`), a netcat placeholder listener, and a `sleep 2` that make it unsuitable for source-based runs.

  - Planning artifacts: none
  - Reviewed / approved by: n/a

  ---

  ## Self-Review

  - Bugs caught: cleanup was not waiting for children to exit before stopping postgres → DB connection errors. fixed with reverse-order sequential teardown.
  - Logic simplified: replaced manual `spawn` + `waitForExit` + `ChildProcess` tracking with execa (`cleanup: true`, `forceKillAfterDelay`, `reject: false`, `extendEnv`).
  - Naming / terminology improved: `ENSAPI_GRAPHQL_API_URL` → `ENSNODE_URL` for consistency.
  - Dead or unnecessary code removed: removed testcontainers devDeps from ensapi package.json, removed `test:integration:ci` script from ensapi.

  ---

  ## Cross-Codebase Alignment

  - Search terms used: `OmnichainIndexingStatusIds`, `ENSNamespaceIds`, `ensTestEnvChain`, `ENSRAINBOW_DEFAULT_PORT`, `entrypoint.sh`
  - Reviewed but unchanged: `apps/ensrainbow/scripts/entrypoint.sh` (considered reusing, too Docker-specific), `packages/datasources/src/lib/chains.ts` (confirmed devnet hardcodes `localhost:8545`), `packages/ensnode-sdk/src/indexing-status/`
  (confirmed `OmnichainIndexingStatusIds` exports)
  - Deferred alignment: ensrainbow database download/extract logic is duplicated between `entrypoint.sh` and `ci.ts` — should be extracted to a reusable script (follow-up)

  ---

  - no public API changes
  - integration tests now run in CI on every PR — regressions will be caught automatically
  - `ENSAPI_GRAPHQL_API_URL` env var renamed to `ENSNODE_URL` in the integration test client (only used by integration tests, not production)

  - Public APIs affected: none
  - Docs updated: frontmatter in `ci.ts` documents design decisions
  - Naming decisions worth calling out: `ENSNODE_URL` replaces `ENSAPI_GRAPHQL_API_URL`

  ---

  ## Testing Evidence

  - ran the full orchestrator locally — all integration tests pass (104/104)
  - ran in CI via the PR — verified the workflow job executes end-to-end
  - unit tests unaffected (`pnpm test` passes, 84/84 in ensapi)

  - Testing performed: local + CI end-to-end runs
  - Known gaps: no test for the orchestrator itself (it's the test runner). cleanup behavior under various failure modes was tested manually.
  - What reviewers have to reason about manually: whether the cleanup ordering and signal handling are correct in edge cases (crash during startup, crash during indexing, repeated Ctrl-C)

  ---

  ## Scope Reductions

  - **extract ensrainbow `prepare-database.sh`** — the download + extract + validate logic is duplicated between `entrypoint.sh` and `ci.ts`. should be a shared script that both consume. deferred to keep this PR focused on CI plumbing.

  - Follow-ups: reusable `prepare-database.sh` in ensrainbow
  - Why they were deferred: entrypoint.sh refactoring is a separate concern and would expand the blast radius of this PR

  ---

  ## Risk Analysis

  - **devnet image pinned to `main-f476641`** — if that image is removed from ghcr, CI breaks. low risk since it's a published release.
  - **fixed ports for devnet (8545/8000)** — will conflict if another service uses those ports on the CI runner. acceptable because CI runners are ephemeral.
  - **10-minute timeout** — if indexing or image pulls slow down, the job will fail. can be bumped.
  - **testcontainers requires Docker** — CI runner must have Docker available. blacksmith runners do.

  - Risk areas: devnet image availability, port conflicts on CI runners
  - Mitigations or rollback options: revert the workflow job to disable integration tests. the package is private and has no consumers.

  ---

### DIFF
--- a/packages/integration-test-env/LICENSE
+++ b/packages/integration-test-env/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2025 NameHash
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/packages/integration-test-env/README.md
+++ b/packages/integration-test-env/README.md
@@ -1,0 +1,36 @@
+# @ensnode/integration-test-env
+
+Integration test environment orchestration for ENSNode. Spins up the full ENSNode stack against the [ens-test-env](https://github.com/ensdomains/contracts-v2) devnet, runs monorepo-level integration tests, then tears everything down.
+
+## Devnet Image
+
+The current devnet image is pinned to:
+
+```
+ghcr.io/ensdomains/contracts-v2:main-f476641
+```
+
+Update the `DEVNET_IMAGE` constant in the orchestrator source to change the devnet version.
+
+## How It Works
+
+The orchestrator runs a 6-phase pipeline:
+
+1. **Postgres + Devnet** — started in parallel via testcontainers
+2. **ENSRainbow database** — downloads pre-built LevelDB, extracts, starts ENSRainbow from source
+3. **ENSIndexer** — starts from source, waits for health
+4. **Indexing** — polls until omnichain status reaches "Following" or "Completed"
+5. **ENSApi** — starts from source, waits for health
+6. **Integration tests** — runs `pnpm test:integration` at monorepo root
+
+## Usage
+
+```sh
+pnpm start
+```
+
+Works both in CI and locally — just make sure the required ports are available (8545, 8000, 3223, 42069, 4334).
+
+## License
+
+Licensed under the MIT License. See [LICENSE](./LICENSE).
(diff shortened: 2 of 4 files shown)


lightwalker-eth:
@greptile

packages/integration-test-env/src/ci.ts
```diff
@@ -0,0 +1,358 @@
...
+  }
+  throw new Error(`${label} did not become healthy within ${timeoutMs / 1000}s`);
+}
+
+function spawnService(
+  command: string,
+  args: string[],
+  cwd: string,
+  env: Record<string, string>,
+  label: string,
```
lightwalker-eth:
```suggestion
  serviceName: string,
```

Issue #1728, opened 2026-03-05
Remove Core Plugin Config Invariant


shrugs:
- allow the activation of both `subgraph` and `ensv2` plugins at the same time
- manually verify against sepolia
- make sure there are no 'core plugin' references anymore
- restructure the core plugin middleware to just be two separate middlewares, one for each of the apis

PR #1730, opened 2026-03-06
Remove Core Plugin Restriction

shrugs:
closes https://github.com/namehash/ensnode/issues/1728

  ## Summary

  - didn't actually need to change any invariants, this was already possible
  - added warning
  - removed concept of 'core plugin'

  ---

  ## Why

  - the shared middleware unnecessarily coupled the two handlers and enforced a "one core plugin" mental model
  - both plugins should be activatable in parallel for deployments that serve both APIs

  ---

  ## Testing

  - typecheck, lint

  ---

  ## Notes for Reviewer

  - the inline middleware is functionally identical to the relevant branch of the deleted middleware — each handler previously only hit one code path
  - check wording of the warning

  ---

  ## Checklist


lightwalker-eth (approved):
@shrugs This is great!

Issue #1704, opened 2026-03-02
Inconsistent order of Registrar Actions


Goader:
# Incosistent order of Registrar Actions

@Y3drk noted a weird behavior:

> Before I added the ByDecodedReferrer filter to the useRegistarActions call in ENSAwards, I noticed that while no new registrar actions show up, something weird is going on with the animation (see the attached video).
>
> Namely, the last row kept changing. I monitored it for a few iterations of the query and noticed that the API returns a different set of latest registrar-actions (ordered by their timestamp) nearly each time the query refetches. Take a look at the example JSON file (results of a few sequential queries made to the API) I attached. Only the last item keeps changing, and it seems like the rotation of these records is periodic.
>
> I believe it is due to the fact that they share the same exact timestamp, and when ordering, the ties are resolved inconsistently, but I didn't have time to check it in the code.

From the quick investigation, it seems that the reason is indeed the sorting by timestamp, which can be ambiguous in the case of actions with the same timestamp. However, it is not confirmed yet, and therefore needs further validation.

The file [find-registrar-actions.ts](https://github.com/namehash/ensnode/blob/b0de5e9cda0f2919aa201e1ec26a05a6473dc03c/apps/ensapi/src/lib/registrar-actions/find-registrar-actions.ts#L29) has the following function, which seems to be the place responsible for this:

```typescript
/**
 * Build SQL for order clause from provided order param.
 */
function buildOrderByClause(order: RegistrarActionsOrder): SQL {
  switch (order) {
    case RegistrarActionsOrders.LatestRegistrarActions:
      return desc(schema.registrarActions.timestamp);
  }
}
```

PR #1742, opened 2026-03-09
Registrar Actions: Inconsistent Order

Goader:
# Inconsistent Order of Registrar Actions

closes: #1704 

## Summary

- Changed the SQL query to order by Registrar Actions ID, not timestamp

---

## Why

- Registrar Actions order returned from ENSApi is inconsistent, causing undeterministic behavior on the frontend
- Sorting by `id` just like we did with `rev-share-limit` leaderboard building (not that `eventType` is not processed contrary to leaderboard building)
- See: #1704 

---

## Testing

- Automatic, CI testing
- Manual validation (using older version of ENSIndexer and ENSRainbow, but as far as I understand this shouldn't affect the outcome)

---

- Whether `id` is sufficient for the deterministic ordering

---

### DIFF
--- a/packages/ens-referrals/src/v1/award-models/rev-share-limit/leaderboard.ts
+++ b/packages/ens-referrals/src/v1/award-models/rev-share-limit/leaderboard.ts
@@ -21,12 +21,12 @@ import {
   buildReferrerMetricsRevShareLimit,
 } from "./metrics";
 import type { ReferralEvent } from "./referral-event";
-import { sortReferralEvents } from "./sort-referral-events";
 import {
   BASE_REVENUE_CONTRIBUTION_PER_YEAR,
   isReferrerQualifiedRevShareLimit,
   type ReferralProgramRulesRevShareLimit,
 } from "./rules";
+import { sortReferralEvents } from "./sort-referral-events";
 
 /**
  * Represents a leaderboard with the rev-share-limit award model for any number of referrers.


packages/ens-referrals/src/v1/award-models/rev-share-limit/referral-event.ts
```diff
@@ -22,7 +22,7 @@ export interface ReferralEvent {
   /**
    * Registrar action ID.
    *
-   * A deterministic and globally unique identifier for the "logical registrar action"
+   * A deterministic and globally unique fixed-length identifier for the "logical registrar action"
    * associated with the ReferralEvent.
    *
    * A Ponder-encoded checkpoint string that uniquely and deterministically identifies
```
lightwalker-eth: Suggest we not document the implementation details of this `id` string value at this layer of responsibility.

Suggested action items:

1. We should move all the strong documentation about the implementation details of this `id` field into the `ensnode-schema` package. Specifically here: https://github.com/namehash/ensnode/blob/main/packages/ensnode-schema/src/schemas/registrars.schema.ts#L170
2. ... therefore, we can expose the implementation details of this `id` field to those building directly on ENSDb.
3. The documentation about this `id` field within the `ens-referrals` package should treat the `id` field as an opaque identifier without any mention of its implementation details. Within the `ens-referrals` package we should just identify the following invariants about this `id` field:
   1. It is deterministic and globally unique.
   2. Sorting by this value achieves a chronological ordering of each registrar action by the order they were executed onchain.
4. ... all the other implementation details, including any reference to Ponder or details of the "sub-fields" encoded within this string, etc.. should move into `ensnode-schema`.

packages/ensnode-sdk/src/ensapi/api/registrar-actions/request.ts
```diff
@@ -53,7 +53,16 @@ export type RegistrarActionsFilter =
  * Records Orders
  */
 export const RegistrarActionsOrders = {
-  LatestRegistrarActions: "orderBy[timestamp]=desc",
+  /**
+   * Returns registrar actions newest-first.
+   *
+   * Sorts by the `id` field descending. Each `id` is a Ponder checkpoint string
```
lightwalker-eth: This is a great comment, but suggest a few action items:
1. Transfer all the nice ideas in this comment into a new "sort events" utility function in the `ens-referrals` package.
2. Please see the feedback above suggesting to revert the removal of `"orderBy[timestamp]=desc"`. Suggest to add a comment there about how sorting by timestamp not only sorts by timestamp, but further sorts by the chronological execution order of each registrar action.

packages/ens-referrals/src/v1/award-models/rev-share-limit/leaderboard.ts
```diff
@@ -101,11 +100,14 @@ export const buildReferrerLeaderboardRevShareLimit = (
...
-  //    (after zeroing the internal `eventType` digit) yields a total chronological order.
-  const sortedEvents = [...events].sort((a, b) => compareEventIds(a.id, b.id));
+  // 1. Sort events deterministically by id (lexicographic order).
+  //    Ponder checkpoint IDs are constant-length decimal strings, so lexicographic
+  //    order equals chronological order across all fields including eventType.
+  //    NOTE: This relies on the invariant that all ReferralEvents are produced by
+  //    Ponder log (smart-contract event) handlers and therefore share the same
+  //    eventType digit. If mixed event types are ever introduced, ordering by id
+  //    directly would need revisiting.
+  const sortedEvents = [...events].sort((a, b) => (a.id < b.id ? -1 : a.id > b.id ? 1 : 0));
```
lightwalker-eth: Thanks for this 👍 Suggest moving this logic and it's associated comments into a dedicated utility function.

lightwalker-eth (approved):
@Goader Hey looks good. Thanks. Shared a few suggestions. Please feel welcome to merge when ready 👍

PR #1747, opened 2026-03-10
add missing index on domain name column

shrugs:
## Summary

- added a `byName` index on the `name` column of `subgraph_domains` table

## Why

- queries filtering domains by name (e.g. `where: { name: "foo.eth" }`) were doing full table scans — this column had no index despite being a common filter target

## Testing

- not tested directly; this is a declarative schema index addition that ponder applies on next migration
- verified the schema file typechecks

## Checklist


lightwalker-eth (approved):
@shrugs Thanks 👍

Issue #1789, opened 2026-03-18
Extend ENSDb documentation


tk-o:
Documentation present at https://ensnode.io/ensdb/ needs to be updated to present an overview about ENSDb service, and how to use it. There should be examples for interacting with ENSDb with ENSDb Reader and ENSDb Writer classes from ENSDb SDK.

Here's the current outline I created for the ENSDb Docs:
https://docs.google.com/document/d/1EodQIeotwAHB3bKFsb-Tca17tg7b-Ji6tijJ3yuB_Hk/edit?tab=t.0

tk-o:
Resolved with #1942

PR #1771, opened 2026-03-16
WIP ENSDb migrations

tk-o:
# Lite PR

[Tip: Review docs on the ENSNode PR process](https://ensnode.io/docs/contributing/prs)

## Summary

- What changed (1-3 bullets, no essays).

---

## Why

- Why this change exists. Link to related GitHub issues where relevant.

---

## Testing

- How this was tested.
- If you didn't test it, say why.

---

## Notes for Reviewer (Optional)

- Anything non-obvious or worth a heads-up.

---

## Pre-Review Checklist (Blocking)

- [ ] This PR does not introduce significant changes and is low-risk to review quickly.
- [ ] Relevant changesets are included (or are not required)

PR #1754, opened 2026-03-12
ENSDb migrations

tk-o:
# Lite PR

[Tip: Review docs on the ENSNode PR process](https://ensnode.io/docs/contributing/prs)

## Summary

- Configured `drizzle-kit` for managing ENSDb migrations for ENSNode Schema.
  - This is a dev dependency.
- Created ENSDb Writer Worker task to execute migrations mentioned above.
  - Migrations will be auto-executed as the initial task performed by the worker.

---

## Why

- ENSNode Schema is a new schema added to ENSDb and it needs migrations to be executed automatically.

---

## Testing

- I ran local instances of ENSIndexer and ENSApi working together.
  - Migrations created `ensnode` schema in local ENSDb, including nested database object (`ensnode_metadata` table).

---

## Notes for Reviewer (Optional)

- Please review commit by commit.
- This PR will be merged into #1753, and into #1752.

---

## Pre-Review Checklist (Blocking)

- [x] This PR does not introduce significant changes and is low-risk to review quickly.
- [x] Relevant changesets are included (or are not required)

PR #1798, opened 2026-03-19
Integrate ENSDb SDK into ENSIndexer

tk-o:
# Substantial PR

<details>

- ENSIndexer running database migrations for ENSDb.
- ENSDb SDK client implementations.
- I suggest reviewing commit-by-commit.

</details>

---

## Problem & Motivation

<details>

- This PR makes a huge step forward towards the goal of splitting ENSIndexer and ENSApi.
- Also, with this PR, ENSDb becomes a single service to discover data about ENSNode instance. With ENSDb, a user can get a list of all ENSIndexer instances running within the ENSNode instnace, and then, get indexed data from any ENSIndexer instance of their choice. It is feels like creating ENSDb CLI would be possible now.
- Without that PR
  - ENSApi and ENSIndexer must keep using their respective ENSDb Client implementations.
  - ENSNode Metadata must be kept in an "onchain table" as part of ENSIndexer Schema.

Prior discussion
  - https://namehash.slack.com/archives/C086Z6FNBHN/p1771947223699149?thread_ts=1771875835.365639&cid=C086Z6FNBHN
  - https://namehash.slack.com/archives/C086Z6FNBHN/p1773668058899639?thread_ts=1773662316.353819&cid=C086Z6FNBHN

</details>

---

## What Changed (Concrete)

<details>

1. ENSDb SDK now includes two ENSDb client implementations for:
    - quering ENSDb (see `EnsDbReader` and `EnsNodeDbQueries`),
    - mutating ENSDb (see `EnsDbWriter`, `EnsNodeDbMutations`, `EnsNodeDbMigrations`)
2. The `EnsDbClient` implemented previously for ENSIndexer has been replaced with `EnsDbWriter` implementation from ENSDb SDK.
3. ENSIndexer is now responsible for executing database migrations for ENSNode Schema in ENSDb. 
    - See how `migrateEnsNodeDb` function is called in a non-blocking way before `startEnsDbWriterWorker` function is called. This ensures that `ENSDbWriterWorker` won't start until database migrations for ENSNode Schema in ENSDb have been executed successfully.
4. ENSDb SDK does not have dependency on `@ponder/client` package, so we had to copy `setDatabaseSchema` function to be avialable inside ENSDb SDK. You can find is as `EnsDbReader.bindEnsIndexerSchemaWithName`.

</details>

---

## Design & Planning

<details>

There were multiple discussions based on the PRs that attempted to address related issues in the past:
- #1752 
- #1753 
- #1754
- #1771 

Key goals for the architecture were to:
1 . Create two kinds of ENSDb client: one for read-only operations, and the other for read-write operations.
2. Enable creating complex queries where we could join tables across different database schema.
3. Make room in ENSDb for ENSIndexer Schema (multiple database schema instances, one per each ENSIndexer instance within a ENSNode instance), and ENSNode Schema (single database schema instance, aggregating data from multiple ENSIndexer instances).

[The latest architecture design that was approved](https://namehash.slack.com/archives/C086Z6FNBHN/p1773668058899639?thread_ts=1773662316.353819&cid=C086Z6FNBHN) is below:
image

</details>

---

## Self-Review

<details>

I was reviewing the diff multiple times before creating this PR, and the main change I decided to go ahead with was to avoid using `EnsDbWriter` type on the ENSIndexer app side, and instead, focus on using specific inteface types that `EnsDbWriter` implements. This way, I feel the code is more concise and precise, as some conumers of `EnsDbWriter` instance need to run migrations, while others need to just read and write data from/into tables.

I was checking for the list of searched terms, and found `ensIndexerClient` in there. It led to deleting dead code at `apps/ensindexer/src/lib/ensindexer-client/singleton.ts`.

</details>

---

## Cross-Codebase Alignment

<details>
<summary>Related code you checked</summary>

- Search terms used within `./apps/ensindexer`:
  - drizzle
  - ensdb-sdk
  - startEnsDbWriterWorker
  - ensIndexerClient
  
- Reviewed but unchanged:
- Deferred alignment (with rationale):
  - ENSApi codebase was not updated yet, it will be done in a subsequent PR to limit the scope of change in this PR.

</details>

---

<details>

This change affects the ENSNode operators as the ENSNode Schema has to be managed outside of Ponder app. This leads to making sure all ENSNode Schema changes are captured with updated migration files inside `@ensnode/ensdb-sdk`, and executed against the target ENSDb instance.

Call out:
- ENSIndexer Schema definition is defined a single file at `packages/ensdb-sdk/src/ensindexer/index.ts`
- ENSDb can have multiple instances of ENSIndexer Schema, each for a single ENSIndexer instance running within the given ENSNode instance.
- I distingushied terms of "ENSIndexer Schema definition" and "ENSIndexer Schema instance".
  - See code docs for `bindEnsIndexerSchemaWithName` function.

Public APIs affected:
  - ENSDb has a new schema now (ENSNode Schema)

Docs updated:
  - None; related issue logged #1789 

Naming decisions worth calling out:
  - `EnsDbWriter` is a class name, but in fact its an ENSDb client implementing ceriant interfaces. Therefore, the `EnsDbWriter` singleton instance is still referred to as `ensDbClient`.

</details>

---

## Testing Evidence

<details>

Both, `EnsDbReader` and `EnsDbWriter` implementations have been covered with unit tests.

Additionally, I manually tested running ENSIndexer locally in the following scenarios:
- Ran ENSIndexer with `dev` command multiple times with the same config.
  - No issues.
- Ran ENSIndexer with `dev` command multiple times with different plugins, and the same ENSIndexer Schema name
  - There was a developer experience issue where ENSDb Writer Worker fails to upsert ENSIndexer Public Config due to incompatibility between config stored in ENSDb, and config for the current ENSIndexer instance. [I posted a question on the team's channel](https://namehash.slack.com/archives/C086Z6FNBHN/p1773922368429869?thread_ts=1773921566.284039&cid=C086Z6FNBHN) to see how we want to solve that problem.
- Ran ENSIndexer with `start` command multiple times with the same config.
  - No issues.
- Ran ENSIndexer with `start` command multiple times with different plugins, and the same ENSIndexer Schema name
  - As expected, Ponder wouldn't allow ENSIndexer to run. Here's the error message:
  ```
  MigrationError: Schema "t" was previously used by a different Ponder app. Drop the schema first, or use a different schema. Read more: https://ponder.sh/docs/database#database-schema
  ```
- Ran ENSIndexer with `start` command multiple times with different plugins, and different ENSIndexer Schema name
  - No issues.

Testing ENSIndexer with `start` command was done using docker compose setup. This confirms that database migrations are executed correctly when running ENSIndexer in a docker container.

</details>

---

## Scope Reductions

<details>

I haven't integrated ENSDb SDK into ENSApi yet. This will be done right after this PR is merged into `main`.

</details>

---

## Risk Analysis

<details>

The major change in this PR is the introduction of new responsibility for ENSIndexer to handler ENSNode Schema in ENSDb. The migration process requires manual step of generaing updated migartion files in ENSDb SDK and having them comitted to the repository in order to be executed at ENSIndexer runtime.

The risk is that maual steps can be missed. If that happens, ENSDb might not have the database schema that ENSNode apps would expect.

</details>

---

  - *Tip: [Changesets](https://ensnode.io/docs/contributing/releases/#adding-a-changeset-to-your-pr) should be created / written to optimize for the narrative of the next autogenerated [release notes](https://github.com/namehash/ensnode/releases).*
  
Replaces #1752, #1753, #1754, #1771


apps/ensindexer/src/lib/ensdb/migrate-ensnode-db.ts
```diff
@@ -0,0 +1,20 @@
+import { createRequire } from "node:module";
+import { join } from "node:path";
+
+import type { EnsNodeDbMigrations } from "@ensnode/ensdb-sdk";
+
+// Resolve the path to the migrations directory within the ENSDb SDK package
+const migrationsDirPath = join(
+  createRequire(import.meta.url).resolve("@ensnode/ensdb-sdk"),
+  "../../migrations",
```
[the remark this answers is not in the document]
tk-o: This code is fine as is.

packages/ensdb-sdk/src/client/ensdb-reader.ts
```diff
@@ -0,0 +1,170 @@
...
+import { and, eq } from "drizzle-orm/sql";
+
+import {
+  type CrossChainIndexingStatusSnapshot,
+  deserializeCrossChainIndexingStatusSnapshot,
+  deserializeEnsIndexerPublicConfig,
+  type EnsIndexerPublicConfig,
+} from "@ensnode/ensnode-sdk";
+
+import * as ensIndexerSchema from "../ensindexer";
```
lightwalker-eth: I'm worried about calling this idea `ensIndexerSchema`. Doesn't this object include some placeholder ENSIndexer Schema Name which is likely different than whatever `ensIndexerSchemaName` is passed into an `EnsDbReader` instance?

Should we therefore rename this idea from `ensIndexerSchema` to `abstractEnsIndexerSchema` and document in the relevant places how this object just holds a placeholder value for the ENSIndexer Schema Name which needs to be replaced by ...?

tk-o: I will make the `lib/drizzle.ts` file to be the only place to use import the `abstractEnsIndexerSchema` and use that abstract schema to construct a concrete one with `buildEnsIndexerSchema` function.

packages/ensdb-sdk/src/lib/drizzle.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @param ensIndexerSchemaName - The name of the ENSIndexer Schema instance in ENSDb.
+ *
+ * Note: this function is a replacement for `setDatabaseSchema` from `@ponder/client`.
+ */
+function buildEnsIndexerSchema<EnsIndexerSchemaType extends AbstractEnsIndexerSchema>(
+  ensIndexerSchemaName: string,
+): EnsIndexerSchemaType {
+  const ensIndexerSchema = {} as EnsIndexerSchemaType;
+
+  for (const [key, abstractSchemaObject] of Object.entries(abstractEnsIndexerSchema)) {
```
shrugs: why do we clone the abstract schema into the concrete schema here instead of mutating it directly? mutating would avoid the somewhat hacky cloning situation here

apps/ensindexer/src/lib/ensdb/singleton.ts
```diff
@@ -1,16 +1,27 @@
...
  */
-const ensDbSchema = buildEnsDbSchema(ensIndexerSchemaName);
+const ensDbSchema = buildEnsDbSchema(ensIndexerSchema);
 const ensDbDrizzleClient = buildEnsDbDrizzleClient(ensDbConnectionString, ensDbSchema);
 
 /**
  * Singleton instance of ENSDbWriter for the ENSIndexer application.
  */
-export const ensDbWriter = new EnsDbWriter(ensDbDrizzleClient, ensDbSchema, ensIndexerSchemaName);
+export const ensDbWriter = new EnsDbWriter(
```
lightwalker-eth: @tk-o Is there a special reason why we don't move a lot of the logic that's in this file into the `EnsDbWriter` (or reader) constructor so that its constructor just needs:
1. `ensDbConnectionString`
2. `ensIndexerSchemaName`

packages/ensdb-sdk/src/client/ensdb-reader.ts
```diff
@@ -38,12 +38,12 @@ export class EnsDbReader<
...
   /**
-   * ENSDb Schema definition for ENSDb.
+   * "Concrete" ENSIndexer Schema definition for ENSDb.
    *
-   * This is the "concrete" ENSDb Schema in which tables reference
+   * This is the "concrete" ENSIndexer Schema in which tables reference
    * the ENSIndexer Schema name from {@link ensIndexerSchemaName}.
    */
-  protected ensDbSchema: EnsDbSchema<EnsIndexerSchemaType>;
+  protected _ensIndexerSchema: EnsIndexerSchemaType;
```
lightwalker-eth: Should we rename `EnsIndexerSchemaType` to `ConcreteEnsIndexerSchema`?

packages/ensdb-sdk/src/lib/drizzle.ts
```diff
@@ -40,14 +70,14 @@ function buildEnsIndexerSchema<EnsIndexerSchemaType extends AbstractEnsIndexerSc
```
lightwalker-eth: Should we rename this function to `buildConcreteEnsIndexerSchema`?

PR #1800, opened 2026-03-19
Narrative Overhaul: reserve top-level package names

shrugs:
- these stub packages have been published
  - `enssdk`
  - `enskit`
  - `ensskills`
- this package is pending exception because it was too similar to the existing `ens-cli`
  - `enscli`

The PR creates the stubbed packages and can be merged once `enscli` ownership is granted. 

currently owned by `shrugs-namehash`, but access to the packages has been granted to `@ensnode` so our CI should continue to publish these successfully.

In the near future, these can be built upon into actual packages and then published normally.

### DIFF
--- a/packages/enscli/package.json
+++ b/packages/enscli/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "enscli",
+  "version": "0.0.1",
+  "description": "Reserved for the ENSNode project by NameHash Labs. See https://ensnode.io",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/namehash/ensnode.git",
+    "directory": "packages/enscli"
+  },
+  "license": "MIT",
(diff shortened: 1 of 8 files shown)


packages/enscli/package.json
```diff
@@ -0,0 +1,12 @@
+{
```
lightwalker-eth: I believe we'll need to update several config files / CI scripts for these new packages to integrate into our changesets / release processes.

Also, FYI: our release processes are configured to use my personal NPM account token right now (configured as a GitHub Secret). I imagine we might need to do configure something so that my personal NPM account will be able to publish these?

lightwalker-eth (commented):
@shrugs Great to see the NPM reservations. Shared a few suggestions when you're ready.

shrugs:
Superseded by #1846.

PR #1805, opened 2026-03-20
feat(docs/ensnode.io): Add API Reference

shrugs:
## Summary

- Added interactive Scalar API reference component embedded within Starlight documentation
- Overrode Starlight's `TwoColumnContent` to always render the right sidebar column and remove `isolation: isolate` from `.main-pane`
- Disabled dark mode across the docs site
- Removed the separate standalone API reference page

## Why

Embedding the Scalar API reference directly in Starlight provides an integrated documentation experience. 

## How it was tested

- Opened the API reference page and clicked "Test Request" and observed the modal rendering
- Verified other pages with `tableOfContents` enabled still display the TOC and TelegramInvite correctly
- Verified the API reference page (with `tableOfContents: false`) shows TelegramInvite in the right sidebar
- Confirmed content width is consistent between pages with and without TOC

## Notes

- The `TwoColumnContent` override is a near-copy of Starlight's original with three changes
- -  Removed the `toc &&` guard on the `<aside>`
-  - removed `isolation: isolate` from `.main-pane`
- - changed `[data-has-sidebar][data-has-toc]` to `[data-has-sidebar]` in the width selector. These are all internal to the component and can't be achieved by wrapping the default.
- If Starlight updates `TwoColumnContent` upstream, this override may need to be synced.
- Scalar's CSS uses `.scalar-app .z-overlay` (descendant selector) for the modal's z-index, but both classes are on the same element. This is a Scalar bug, but removing `isolation: isolate` makes it moot since the modal's stacking context is no longer trapped.


sevenzing (approved):

notrab (approved):

notrab:
@copilot resolve the merge conflicts in this pull request

notrab (approved):

PR #1808, opened 2026-03-22
Improve openapi generation

sevenzing:
## Summary

* fix current problem that we have to specify two times routes to include them in openapi spec
* moving all vars that initialized at top level of each file into `lazy`. that allows importing them without triggering config initialization 

---

## Why

 * to specify routes only once 

---

## Testing

* Ran `pnpm generate:openapi`
* Ran application and tested that routers are working correctly

---

---

### DIFF
--- a/apps/ensapi/src/config/config.singleton.test.ts
+++ b/apps/ensapi/src/config/config.singleton.test.ts
@@ -22,10 +22,11 @@ describe("ensdb singleton bootstrap", () => {
   });
 
   it("constructs EnsDbReader from real env wiring without errors", async () => {
-    const { EnsDbReader } = await import("@ensnode/ensdb-sdk");
     const { ensDbClient, ensDb, ensIndexerSchema } = await import("@/lib/ensdb/singleton");
 
-    expect(ensDbClient).toBeInstanceOf(EnsDbReader);
+    // ensDbClient is a lazyProxy — construction is deferred until first property access.
+    // Accessing a property triggers EnsDbReader construction; verify it succeeds.
+    expect(ensDbClient.ensIndexerSchemaName).toBe(VALID_SCHEMA_NAME);
     expect(ensDb).toBeDefined();
     expect(ensIndexerSchema).toBeDefined();
   });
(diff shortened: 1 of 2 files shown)


apps/ensapi/src/app.ts
```diff
@@ -0,0 +1,82 @@
...
+app.route("/ensanalytics", ensanalyticsApi);
+
+// use ENSAnalytics API v1 at /v1/ensanalytics
+app.route("/v1/ensanalytics", ensanalyticsApiV1);
+
+// use Am I Realtime API at /amirealtime
+// NOTE: this is legacy endpoint and will be deleted in future. one should use /api/realtime instead
+app.route("/amirealtime", realtimeApi);
+
+// serve pre-generated OpenAPI 3.1 document
```
notrab: The code here actually generates the document dynamically on each request. So we should probably update this comment.

apps/ensapi/src/openapi-document.ts
```diff
@@ -1,50 +1,12 @@
...
-    }
-  }
-
-  return app;
-}
+import app from "./app";
 
 /**
- * Generates an OpenAPI 3.1 document from stub route definitions.
+ * Generates an OpenAPI 3.1 document from the real registered routes.
```
notrab: There was a comment in this file for `createStubRoutesForSpec` that outlined what this file was used for. I think it's important to keep that.

For example, we could add to this comment _why_ it exists. E.g. `this indirection exists so the generation script and the runtime endpoint share the same function`

apps/ensapi/src/cache/indexing-status.cache.ts
```diff
@@ -7,42 +7,62 @@ import {
...
+            );
+            throw error;
+          }),
+      ttl: 5, // 5 seconds
+      proactiveRevalidationInterval: 10, // 10 seconds
+      proactivelyInitialize: true,
+    }),
+);
+
+export const indexingStatusCache = new Proxy({} as IndexingStatusCache, {
```
notrab: What's the purpose of this Proxy instead of using the `lazy` import? Maybe a comment would help, I'm struggling to follow the different use of Proxy.

docs/docs.ensnode.io/ensapi-openapi.json
```diff
@@ -29,35 +29,6 @@
   ],
   "components": { "schemas": {}, "parameters": {} },
   "paths": {
```
lightwalker-eth: I see a pretty big diff in this file. Is the diff in this file only due to reordering but no actual changes?

I'm worried if the ordering of endpoints in this file isn't deterministic for some reason? Ideally any diffs to this file would be meaningful.

Appreciate your advice.

PR #1872, opened 2026-04-04
feat: add ensapi openapi examples

sevenzing:
## Summary
* Add `examples.ts` files for `name-tokens`, `registrar-actions`, and `resolution` — typed example constants used in both routes and tests!
* renamed `name` query param to `reverseName`. only in API, didnt update core logic
* Wire 200/400/500 response schemas with examples
* Add descriptions to all `resolution/registrar-actions` query and path params; fix `beginTimestamp` and `endTimestamp` missing type: `"integer"` in OpenAPI output was missing for some reason, idk why...
* Remove `z.undefined()` from `makeResponsePageContextSchemaWithNoRecords` — `ZodUndefined` has no JSON Schema representation and caused UnknownZodTypeError during OpenAPI generation

---

## Testing

tested it with

```
 pnpm test
 pnpm typecheck
 pnpm generate:openapi -- generated OK with examples
```
---


apps/ensapi/src/app.ts
```diff
@@ -75,7 +75,7 @@ app.get("/health", async (c) => {
 // log hono errors to console
 app.onError((error, ctx) => {
```
lightwalker-eth: For this fallback catch-all error handler, what do you think about adding a prefix to the log messages / `errorResponse that would clearly identify for us that the error went through this fallback catch-all error handler.

Goal: Help us with debugging.

For example, maybe a little prefix such as "Unexpected server error: ..."?

Appreciate your advice.

packages/ensnode-sdk/src/registrars/zod-schemas.ts
```diff
@@ -87,6 +87,35 @@ const makeRegistrarActionPricingSchema = (valueLabel: string = "Registrar Action
       .transform((v) => v as RegistrarActionPricingUnknown),
   ]);
 
+const makeSerializedPriceEthSchema = (valueLabel: string = "Serialized Price ETH") =>
+  z.object({
+    amount: z.string({ error: `${valueLabel} amount must be a string.` }).regex(/^\d+$/, {
```
lightwalker-eth: We should have some other existing Zod schemas you could build upon for non-negative integers so that no regex is needed here.

Goal: Build upon the primitives we've already been creating.

packages/ensnode-sdk/src/registrars/zod-schemas.ts
```diff
@@ -87,6 +87,35 @@ const makeRegistrarActionPricingSchema = (valueLabel: string = "Registrar Action
...
+      error: `${valueLabel} amount must be a non-negative integer string.`,
+    }),
+    currency: z.literal("ETH", {
```
lightwalker-eth: Hmm. I thought we would already have a Zod schema defined for this? Can you please review our existing Zod schemas such as:

1. `makePriceSchema`
2. `makePriceEthSchema`
3. `deserializePriceEth`

Perhaps there's existing code to reuse or refactor?

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
+    .optional()
+    .describe(
+      "Comma-separated list of text record keys to resolve (e.g. 'avatar,description,url').",
+    ),
 });
 
 const selection = z
   .object({
-    name: z.optional(boolstring),
+    nameRecord: z.optional(boolstring),
```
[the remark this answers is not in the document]
lightwalker-eth: Appreciate if you can put a special effort into adding some nice docs about this idea in all the relevant places across our whole monorepo. It's a complex idea and I really want to see us investing strongly in our docs.

This includes docs for this field within the Drizzle schemas that are defined in ENSDb.

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
+  .default(false)
+  .describe("Include detailed resolution trace information in the response.")
+  .openapi({ default: false });
 
-const chainIdsWithoutDefaultChainId = z.optional(
-  stringarray.pipe(z.array(defaultableChainId.pipe(excludingDefaultChainId))),
+const accelerate = z
+  .optional(boolstring)
+  .default(false)
+  .describe("Attempt accelerated CCIP-Read resolution using L1 data.")
```
lightwalker-eth: Our acceleration is not constrained only to L1 data. We can fully accelerate resolutions for some "L2 names" such as jesse.base.eth

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
+  .openapi({
+    default: false,
```
lightwalker-eth: I believe we should update this to accelerate by default. Please feel welcome to create a separate follow-up issue / PR for this.

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
+  });
+const address = makeLowercaseAddressSchema().describe(
+  "EVM wallet address (e.g. '0xd8da6bf26964af9d7eed9e03e53415d37aa96045').",
```
lightwalker-eth: Should we note in this description that it must be all lowercase?

packages/ensnode-sdk/src/ensapi/api/resolution/examples.ts
```diff
@@ -0,0 +1,39 @@
...
+export const resolveRecordsResponseExample = {
+  records: {
+    name: "vitalik.eth",
+    addresses: { "60": "0xd8da6bf26964af9d7eed9e03e53415d37aa96045" },
+    texts: {
+      description: "mi pinxe lo crino tcati",
+    },
+  },
+  accelerationRequested: false,
+  accelerationAttempted: false,
```
lightwalker-eth: This terminology of `accelerationAttempted` is confusing. Can you please create a follow up issue and PR to update it?

The problem is that "attempted" is ambiguous. Ok.. maybe it was "attempted" but was it actually "completed"?

I assume that this field actually represents if acceleration was completed or not?

Assuming so, suggest to rename this field from `accelerationAttempted` to just `accelerated` as that is the more concise way of communicating the idea of "acceleration completed".

docs/docs.ensnode.io/ensapi-openapi.json
```diff
@@ -969,19 +969,66 @@
...
-          { "schema": { "type": "string" }, "required": false, "name": "addresses", "in": "query" },
-          { "schema": { "type": "string" }, "required": false, "name": "texts", "in": "query" },
           {
-            "schema": { "type": "boolean", "default": false },
+            "schema": {
+              "type": "string",
+              "description": "ENS name to resolve (e.g. 'vitalik.eth'). Must be normalized per ENSIP-15."
+            },
+            "required": true,
+            "description": "ENS name to resolve (e.g. 'vitalik.eth'). Must be normalized per ENSIP-15.",
```
lightwalker-eth: I note that the `description` is being duplicated both here and in the `schema` a few lines above.

Maybe this is totally fine and the right way to do things? I'm not sure. If it is, please ignore and mark this as resolved. Just flagging in case there's something for us to optimize here. Appreciate your advice. Thanks.

lightwalker-eth (commented):
@sevenzing Hey good to see the improvements here. Reviewed and shared feedback 👍

apps/ensapi/src/app.ts
sevenzing: We pass this error to `errorResponse` which have whole logic of determining when to return simple `Internal Server Error` and when to return whole error description. I think that first of all we need to store this logic there and second of all current error handling looks fine for me

packages/ensdb-sdk/src/client/zod-schemas/ensdb-config.ts
```diff
@@ -1,31 +1,35 @@
 import { parse as parseConnectionString } from "pg-connection-string";
 import { z } from "zod/v4";
 
-export const EnsDbUrlSchema = z.string().refine(
-  (url) => {
-    try {
-      if (!url.startsWith("postgresql://") && !url.startsWith("postgres://")) {
+export const EnsDbUrlSchema = z
+  .string({
+    error: "ENV variable ENSDB_URL is required.",
```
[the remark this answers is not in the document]
sevenzing: yeah, I also wanted to something like that but then I saw that we only have 2 variables and I was like "dont want to introduce it for just 2 variables"

decided to remove this change from this PR and created sep issue for that. also described my vision of how to do it.

https://github.com/namehash/ensnode/issues/1916

packages/ensnode-sdk/src/registrars/zod-schemas.ts
sevenzing: we need exactly `z.string()` because this is serialized version of `makePriceAmountSchema `. 

`makePriceAmountSchema` has `z.coerce.bigint()` which is good for logic type but not fully supported by openapi. openapi needs to know that this type will be actually string to allow me write examples

packages/ensnode-sdk/src/registrars/zod-schemas.ts
sevenzing: yeah, this is tricky, I also spend a lot of time here. basically replied here: https://github.com/namehash/ensnode/pull/1872#discussion_r3069728541

the main problem is: in JSON REST API `bigint` is presented as `string` because default json integers only support int32. so we need to specify that internal types uses bigint not to allow `123asdasd` string and output type (for JSON transport) use `string` to allow me write examples with typechecking. that's why we need to write a lot of boilerplate for that (I also mentioned it in PR description 😄 )

packages/ensnode-sdk/src/registrars/zod-schemas.ts
```diff
@@ -87,6 +87,35 @@ const makeRegistrarActionPricingSchema = (valueLabel: string = "Registrar Action
...
+      error: `${valueLabel} currency must be set to 'ETH'.`,
+    }),
+  });
+
+/**
+ * Schema for parsing objects into serialized {@link RegistrarActionPricing}.
+ */
+const makeSerializedRegistrarActionPricingSchema = (
```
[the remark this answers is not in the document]
sevenzing: yeah, the same as `makeRegistrarActionPricingSchema` for consistency

packages/ensnode-sdk/src/registrars/zod-schemas.ts
```diff
@@ -194,3 +223,45 @@ export const makeRegistrarActionSchema = (valueLabel: string = "Registrar Action
     makeRegistrarActionRegistrationSchema(`${valueLabel} Registration`),
     makeRegistrarActionRenewalSchema(`${valueLabel} Renewal`),
   ]);
+
+const makeBaseSerializedRegistrarActionSchema = (
```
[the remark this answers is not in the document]
sevenzing: it should be `makeRegistrarActionsResponseSchema` and in this case `generate:openapi` failes:

<details>
  <summary>Commands from my terminal</summary>
  


```bash
> cat apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts| grep "schema: makeRegistrarActionsResponseSchema()" -A 5 -B 5
  responses: {
    200: {
      description: "Successfully retrieved registrar actions",
      content: {
        "application/json": {
          schema: makeRegistrarActionsResponseSchema(),
        },
      },
    },
    400: {
      description: "Invalid query",

~/namehash/ensnode main *9 !2 ?1                 node 24.13.0 21:23:22
> p -F ensapi typecheck

> ensapi@1.9.0 typecheck /Users/levlymarenko/namehash/ensnode/apps/ensapi
> tsgo --noEmit


~/namehash/ensnode main *9 !2 ?1                 node 24.13.0 21:23:40
> p generate:openapi   

> ensnode-monorepo@0.0.1 generate:openapi /Users/levlymarenko/namehash/ensnode
> pnpm -r --if-present generate:openapi

Scope: 24 of 25 workspace projects
docs/docs.ensnode.io generate:openapi$ pnpm --filter ensapi exec tsx…
│ node:internal/modules/run_main:107
│     triggerUncaughtException(
│     ^
│ UnknownZodTypeError {
│   message: 'Unknown zod object type, please specify `type` and oth…
│   data: { currentSchema: { type: 'undefined' }, schemaName: undefi…
│ }
│ Node.js v24.14.0
│ undefined
│ /Users/levlymarenko/namehash/ensnode/apps/ensapi:
│  ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command failed with exit code…
└─ Failed in 3.2s at /Users/levlymarenko/namehash/ensnode/docs/docs.ensnode.io
/Users/levlymarenko/namehash/ensnode/docs/docs.ensnode.io:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @docs/mintlify@1.9.0 generate:openapi: `pnpm --filter ensapi exec tsx --tsconfig tsconfig.json ../../scripts/generate-ensapi-openapi.mts`
Exit status 1
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
 ELIFECYCLE  Command failed with exit code 1.
```
</details>

after that I started my journey to identify the reason of that. I remember that this led me to "I need to create serialized schema, without it I cannot fix it", however I'm trying to investigate the reason right now and i'm a little bit stuck...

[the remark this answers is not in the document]
sevenzing: okay, I remember the problem. 

The core issue is a type mismatch between the `SerializedRegistrarActionsResponseOk` type (what `serializeRegistrarActionsResponse` returns) and the Zod schema inferred type (what the OpenAPI route expects). Specifically, the `pricing.amount` has type `SerializedCurrencyAmount` which is alias to `string`, but the Zod schema infers `never` because of how `z.coerce.bigint()` works — it can't be inferred back to `string` in the output type :( 

BUT the interesting thing is that this code compiles (I commented schema for 500 error):

```js
export const getRegistrarActionsRoute = createRoute({
  method: "get",
  path: "/",
  operationId: "getRegistrarActions",
  tags: ["Explore"],
  summary: "Get Registrar Actions",
  description: "Returns all registrar actions with optional filtering and pagination",
  request: {
    query: registrarActionsQuerySchema,
  },
  responses: {
    200: {
      description: "Successfully retrieved registrar actions",
      content: {
        "application/json": {
          schema: makeRegistrarActionsResponseOkSchema().openapi({
            example: registrarActionsResponseOkExample,
          }),
        },
      },
    },
    400: {
      description: "Invalid query",
      content: { "application/json": { schema: ErrorResponseSchema } },
    },
    500: {
      description: "Internal server error",
      // content: {
      //   "application/json": {
      //     schema: makeRegistrarActionsResponseErrorSchema(),
      //   },
      // },
    },
  },
});
```

but with schema for 500 it throws error!!! I dont know why!!!

```js
export const getRegistrarActionsRoute = createRoute({
  method: "get",
  path: "/",
  operationId: "getRegistrarActions",
  tags: ["Explore"],
  summary: "Get Registrar Actions",
  description: "Returns all registrar actions with optional filtering and pagination",
  request: {
    query: registrarActionsQuerySchema,
  },
  responses: {
    200: {
      description: "Successfully retrieved registrar actions",
      content: {
        "application/json": {
          schema: makeRegistrarActionsResponseOkSchema().openapi({
            example: registrarActionsResponseOkExample,
          }),
        },
      },
    },
    400: {
      description: "Invalid query",
      content: { "application/json": { schema: ErrorResponseSchema } },
    },
    500: {
      description: "Internal server error",
      content: {
        "application/json": {
          schema: makeRegistrarActionsResponseErrorSchema(),
        },
      },
    },
  },
});

```

error is
```bash
...
Type '{ baseCost: { currency: "ETH"; amount: string; }; premium: { currency: "ETH"; amount: string; }; total: { currency: "ETH"; amount: string; }; }' is not assignable to type '{ baseCost: null; premium: null; total: null; }'.
                                Types of property 'baseCost' are incompatible.
                                  Type '{ currency: "ETH"; amount: string; }' is not assignable to type 'null'.

86 app.openapi(getRegistrarActionsRoute, async (c) => {
                                         ~~~~~~~~~~~~~~


Found 1 error in src/handlers/api/explore/registrar-actions-api.ts:86

/Users/levlymarenko/namehash/ensnode/apps/ensapi:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  ensapi@1.9.0 typecheck: `tsgo --noEmit`
Exit status 2
```

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
-const coinType = makeCoinTypeStringSchema();
+const trace = z
+  .optional(boolstring)
+  .default(false)
+  .describe("Include detailed resolution trace information in the response.")
+  .openapi({ default: false });
 
-const chainIdsWithoutDefaultChainId = z.optional(
-  stringarray.pipe(z.array(defaultableChainId.pipe(excludingDefaultChainId))),
+const accelerate = z
```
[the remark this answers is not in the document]
sevenzing: agree, but `tryAccelerate` sounds more logical in this case 👍

apps/ensapi/src/lib/handlers/params.schema.ts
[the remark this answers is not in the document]
sevenzing: yeah, I understand the complexity of this record. I agree that we need to describe this field more. In API docs I would simply have small description and URL to docs where discuss this param in details.

I dont feel good about `advancedOptions` options of something like that. I think making this field optional (as it is right now), renaming to `reverseName` and adding link to docs is enough

docs/ensnode.io/ensapi-openapi.json
```diff
@@ -969,19 +969,66 @@
...
-          { "schema": { "type": "string" }, "required": false, "name": "addresses", "in": "query" },
-          { "schema": { "type": "string" }, "required": false, "name": "texts", "in": "query" },
           {
-            "schema": { "type": "boolean", "default": false },
+            "schema": {
+              "type": "string",
+              "description": "ENS name to resolve (e.g. 'vitalik.eth'). Must be normalized per ENSIP-15."
+            },
+            "required": true,
+            "description": "ENS name to resolve (e.g. 'vitalik.eth'). Must be normalized per ENSIP-15.",
```
sevenzing: yeah, I also dont like it, but this is how `zod-openapi` works, now way to fix it. At least I and LLM didnt find the solution for that. This happens because @hono/zod-openapi follows the OpenAPI 3.x convention where both levels can carry descriptions and it tries to fill both, idk why, hope there is a reason for that :)

apps/ensapi/src/app.ts
```diff
@@ -75,7 +75,7 @@ app.get("/health", async (c) => {
...
   logger.error(error);
-  return errorResponse(ctx, "Internal Server Error");
+  return errorResponse(ctx, error);
```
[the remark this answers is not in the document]
sevenzing: looks like this function is always called when error occurred, that's why we have 500 in https://api.alpha.ensnode.io/api/resolve/records/vitalik.eth

also, I think `errorResponse` should handle case when internal error happens and hide it with `Internal Server Error`: https://github.com/namehash/ensnode/blob/bb1686a34ce1bd36a44598f8de0a24c40a439bc3/apps/ensapi/src/lib/handlers/error-response.ts#L49

apps/ensapi/src/lib/handlers/params.schema.ts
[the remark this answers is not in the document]
sevenzing: let's do it in https://github.com/namehash/ensnode/issues/1918

apps/ensapi/src/lib/handlers/params.schema.ts
sevenzing: https://github.com/namehash/ensnode/issues/1918

packages/ensnode-sdk/src/ensnode/api/resolution/examples.ts
```diff
@@ -0,0 +1,39 @@
...
+export const resolveRecordsResponseExample = {
+  records: {
+    name: "vitalik.eth",
+    addresses: { "60": "0xd8da6bf26964af9d7eed9e03e53415d37aa96045" },
+    texts: {
+      description: "mi pinxe lo crino tcati",
+    },
+  },
+  accelerationRequested: false,
+  accelerationAttempted: false,
```
sevenzing: let's implement it in another PR
https://github.com/namehash/ensnode/issues/1918

apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts
```diff
@@ -96,12 +102,27 @@ export const getRegistrarActionsRoute = createRoute({
   responses: {
     200: {
       description: "Successfully retrieved registrar actions",
+      content: {
+        "application/json": {
+          schema: makeSerializedRegistrarActionsResponseSchema(
+            "Registrar Actions Response",
+          ).openapi({
+            example: registrarActionsResponseOkExample,
```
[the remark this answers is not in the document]
sevenzing: I tried my best to add examples for errors. It always goes with "fixing some random typecheck failure" so I decided to skip `ensanalytics` and `meta` examples

packages/ensnode-sdk/src/ensapi/api/resolution/examples.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Example values for {@link ResolveRecordsResponse}, for use in OpenAPI documentation.
+ */
+export const resolveRecordsResponseExample = {
+  records: {
+    name: "vitalik.eth",
```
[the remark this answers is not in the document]
sevenzing: agree, removed

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
+  });
+const address = makeLowercaseAddressSchema().describe(
+  "EVM wallet address (e.g. '0xd8da6bf26964af9d7eed9e03e53415d37aa96045').",
+);
+const defaultableChainId = makeDefaultableChainIdStringSchema().describe(
+  "Chain ID as a string (e.g. '1' for Ethereum mainnet). Use '0' for the default EVM chain.",
 );
+const coinType = makeCoinTypeStringSchema();
+
+const chainIdsWithoutDefaultChainId = z
```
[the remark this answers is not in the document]
sevenzing: agree on removing `The default EVM chain ID (0) is not allowed.`

regarding `DefaultableChainId` -- not sure if I understood the problem here

apps/ensapi/src/app.ts
sevenzing: okay, I reverted this changes, looks like it need a lot of discussion. let's leave it for next PR

apps/ensapi/src/app.ts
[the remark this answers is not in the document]
sevenzing: okay, I reverted this changes, looks like it need a lot of discussion. let's leave it for next PR

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -37,33 +37,67 @@ const stringarray = z
...
+  .optional(boolstring)
+  .default(false)
+  .describe("Attempt accelerated CCIP-Read resolution using L1 data.")
+  .openapi({
+    default: false,
+  });
+const address = makeLowercaseAddressSchema().describe(
+  "EVM wallet address (e.g. '0xd8da6bf26964af9d7eed9e03e53415d37aa96045').",
+);
+const defaultableChainId = makeDefaultableChainIdStringSchema().describe(
```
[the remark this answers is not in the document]
sevenzing: hm, not sure if anything expect openapi docs needs this description. probably we can do it, but I see no benefits of that

docs/docs.ensnode.io/ensapi-openapi.json
```diff
@@ -1825,23 +1983,427 @@
...
+                                  },
+                                  "pricing": {
+                                    "anyOf": [
+                                      {
+                                        "type": "object",
+                                        "properties": {
+                                          "baseCost": {
+                                            "type": "object",
+                                            "properties": {
+                                              "amount": { "type": "string", "pattern": "^d+$" },
```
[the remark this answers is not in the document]
sevenzing: looks like problem in library, created issue:
https://github.com/asteasolutions/zod-to-openapi/issues/381

apps/ensapi/src/app.ts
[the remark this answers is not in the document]
sevenzing: https://github.com/namehash/ensnode/issues/1921

lightwalker-eth:
@greptile

packages/ensnode-sdk/src/ensnode/api/registrar-actions/serialize.ts
```diff
@@ -20,6 +23,12 @@ export function serializeNamedRegistrarAction({
   };
 }
 
+export function serializeRegistrarActionsResponse(
```
lightwalker-eth, line 26: Can you help me understand the goal for adding these two additional function definitions here?

packages/ensnode-sdk/src/ensapi/api/resolution/examples.ts
```diff
@@ -0,0 +1,38 @@
...
+export const resolveRecordsResponseExample = {
+  records: {
+    addresses: { "60": "0xd8da6bf26964af9d7eed9e03e53415d37aa96045" },
+    texts: {
+      description: "mi pinxe lo crino tcati",
+    },
+  },
+  accelerationRequested: false,
+  accelerationAttempted: false,
+};
```
lightwalker-eth: This feedback applies to all the example responses, not just this one.

It will be great if we can use the type-system to help us catch a need to update these.

This can be achieved by writing, for example: ... `satisfies X` where X is the name of the serialized version of the type for the example response object.

packages/ensnode-sdk/src/ensnode/api/name-tokens/serialize.ts
```diff
@@ -26,6 +28,10 @@ export function serializeRegisteredNameTokens({
   };
 }
 
+export function serializeNameTokensResponse(
+  response: NameTokensResponseOk,
+): SerializedNameTokensResponseOk;
+export function serializeNameTokensResponse<T extends NameTokensResponseError>(response: T): T;
```
lightwalker-eth, line 34: Can you help me understand the goal for adding the 2 extra function defs here?

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -32,33 +32,67 @@ const stringarray = z
...
+    .optional()
+    .describe(
+      "Comma-separated list of text record keys to resolve (e.g. 'avatar,description,url').",
+    ),
 });
 
 const selection = z
   .object({
-    name: z.optional(boolstring),
+    reverseName: z.optional(boolstring),
```
lightwalker-eth: Suggest we not rename this param in this PR. Please revert it. Can you please create a separate follow-up issue for us to improve the naming for this? I believe I've shared a long comment with you somewhere before suggesting how we might want to:

- separate the data model for a selection in the query param / client layer from the data model for a selection in the backend implementation.
- for the query param / client layer, we should make it more friendly for devs new to ENS. Ex: add a new nested object named something like `advancedOptions` and move this `reverseName` field there.
- Add lots of JSDoc in all the relevant places helping devs understand what this `reverseName` lookup actually does.

apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts
```diff
@@ -96,12 +102,31 @@ export const getRegistrarActionsRoute = createRoute({
...
       description: "Invalid query",
+      content: {
+        "application/json": {
+          schema: makeErrorResponseSchema().openapi({ example: errorResponseBadRequestExample }),
+        },
+      },
     },
     500: {
       description: "Internal server error",
+      // TODO: fix the problem with typechecking.
```
lightwalker-eth: Thanks for flagging this opportunity. Can you investigate how we fix it or see if someone on the team could help before we merge this PR?

If we can't find a solution before merging this PR then we need to create a new issue on GitHub and update the comment in the code here to include a link back to that newly created GitHub issue.

I note how there's more than this one place where you identified this issue. Please apply the strategy suggested above to each place this issue exists.

lightwalker-eth (approved):
@sevenzing Nice! Shared some suggestions. Please take the lead to merge when ready 👍

packages/ensnode-sdk/src/ensnode/api/registrar-actions/serialize.ts
sevenzing: for typechecking. without it this function takes enum and returns enum, like

`(Ok | Err) -> (SerializedOk | SerializedErr)`

adding those function signatures allows us to tell compiler more strict typing:

`Ok -> SerializedOk` and `Err -> SerializedErr`

packages/ensnode-sdk/src/ensapi/api/resolution/examples.ts
sevenzing: goot catch! without it wrong example still would be catch during tests, however you suggestion with adding `satisfies X` would save us time during compile time!

packages/ensnode-sdk/src/ensnode/api/registrar-actions/serialize.ts
sevenzing: so, w/o overloads, every caller gets `SerializedRegistrarActionsResponse` (the union / enum) and has to **narrow** it themselves. w/ overloads narrowing applies automatically.

packages/ensnode-sdk/src/ensnode/api/name-tokens/serialize.ts
sevenzing: same as in https://github.com/namehash/ensnode/pull/1872#discussion_r3108774506

apps/ensapi/src/lib/handlers/params.schema.ts
sevenzing: @lightwalker-eth yeah I answered in this discussion: https://github.com/namehash/ensnode/pull/1872#discussion_r3073458419

yeah, agree, let's do it in separate PR. will create an issue for that

apps/ensapi/src/lib/handlers/params.schema.ts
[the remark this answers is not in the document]
sevenzing: https://github.com/namehash/ensnode/issues/1953

apps/ensapi/src/lib/handlers/params.schema.ts
sevenzing: described everything in https://github.com/namehash/ensnode/issues/1953 🙏

apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts
sevenzing: created an issue: https://github.com/namehash/ensnode/issues/1954

sevenzing: fixed it in https://github.com/namehash/ensnode/pull/1957 and merge to this branch

apps/ensapi/src/lib/handlers/params.schema.ts
```diff
@@ -32,22 +32,56 @@ const stringarray = z
...
 
 const rawSelectionParams = z.object({
-  name: z.string().optional(),
-  addresses: z.string().optional(),
-  texts: z.string().optional(),
+  name: z
+    .string()
+    .optional()
+    .describe(
+      "Whether to include the reverse name record in the response, see ENSIP-19 (https://docs.ens.domains/ensip/19/#reverse-resolution)",
```
lightwalker-eth:
```suggestion
      "Whether to include the reverse name record in the response, see ENSIP-3 (https://docs.ens.domains/ensip/3#resolver-interface). Resolving the reverse 'name' resolver record is relevant as an internal implementation detail of the ENS reverse resolution process where the primary ENS name associated with an address is being determined through a multi-step resolution process. For example, querying the (unvalidated!) reverse 'name' resolver record for Vitalik's address (0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045) is achieved by resolving the reverse 'name' resolver record of the reverse name 'd8da6bf26964af9d7eed9e03e53415d37aa96045.addr.reverse' which (at the time of writing) returns 'vitalik.eth'. The ENS reverse resolution process requires that any reverse 'name' resolver record must also pass a forward-resolution validation to be represented as the primary name for an address. More details here: https://docs.ens.domains/web/reverse",
```

I tested the above with this query: https://api.alpha.ensnode.io/api/resolve/records/d8da6bf26964af9d7eed9e03e53415d37aa96045.addr.reverse?name=true&accelerate=true

Related: You can see at https://app.ens.domains/vitalik.eth how `0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045` is the Ethereum address for vitalik.eth which is how I manually constructed the reverse name query referenced above.

docs/ensnode.io/ensapi-openapi.json
```diff
@@ -969,19 +969,38 @@
...
+            "in": "path"
+          },
           { "schema": { "type": "boolean" }, "required": false, "name": "name", "in": "query" },
           { "schema": { "type": "string" }, "required": false, "name": "addresses", "in": "query" },
           { "schema": { "type": "string" }, "required": false, "name": "texts", "in": "query" },
           {
-            "schema": { "type": "boolean", "default": false },
+            "schema": {
+              "type": "boolean",
+              "description": "Include detailed resolution trace information in the response.",
```
lightwalker-eth:
```suggestion
              "description": "Include detailed OpenTelemetry trace information about the resolution in the response.",
```

I know I didn't put this comment in the correct place 😅 Appreciate if you can update the correct place.

Goal: Communicate how the returned trace info is in OpenTelemetry format.

lightwalker-eth (approved):
@sevenzing Great updates here! Shared a few small suggestions please take the lead to merge this as soon as you're ready 👍

PR #1899, opened 2026-04-08
Extend ENSDb migrations

tk-o:
# Lite PR

## Summary

- Added migration that enables the `pg_trgm` extension.
- Added migrations that creates a trigger in ENSDb that is executed each time a schema is dropped from ENSDb. When that happens, we take the name of the schema being dropped, and we delete all records referencing this schema name from ENSNode Metadata table.

---

## Why

- We needed a way to run certain queries against the ENSDb instance just once. The database migrations guarantee that the queries are executed exactly-once.
- Each time ENSIndexer Schema would be dropped from ENSDb, for example, during regular maintenance, the ENSNode Metadata table would still include records referencing the dropped schema. We want to make the ENSDb data model intuitive, and having orphaned data does not support that goal.

---

## Testing

- After creating migration files, I've ran a local ENSIndexer instance and observed that both migrations were executed correctly.
  image
- I ran three local instances of ENSIndexer 
  - `ENSINDEXER_SCHEMA_NAME=abc1 pnpm -F ensindexer start`
  - `ENSINDEXER_SCHEMA_NAME=abc2 pnpm -F ensindexer start`
  - `ENSINDEXER_SCHEMA_NAME=abc3 pnpm -F ensindexer start`
- I saw 9 records in ENSNode Metadata table, 3 records for each ENSIndexer Schema Name.
- I stopped all three local ENSIndexer instances
- I deleted the `abc2` schema
- I observed how only 6 records were remaining in ENSNode Metadata table
  image

---

- Drizzle migrations are guaranteed to run just once for the given ENSDb instance.

---


packages/ensdb-sdk/migrations/0002_manage_orphaned_ensnode_metadata.sql
```diff
@@ -0,0 +1,43 @@
+-- This migration adds an event trigger to handle cleanup of orphaned records
+-- in the `ensnode.metadata` table when an ENSIndexer Schema is dropped.
```
lightwalker-eth: I note that this triggers on any schema that is not the ENSNode schema. We should document this more precisely.

lightwalker-eth (commented):
@tk-o Directionally this looks nice but we should enhance the execution of the idea

Issue #1032, opened 2025-09-09
Support Interpreted Names in Resolution API


shrugs:
support Interpreted Names (normalized except for containing encoded label hashes) in resolution api as input
also requires storage of interpreted names for name and primary name record values
requires lots of testing

lightwalker-eth:
Flagging how we've updated our latest indexed registrations UI component to link to the new name details page:

<img width="2332" height="892" alt="Image" src="https://github.com/user-attachments/assets/f5aaed09-b2b6-4445-a7a9-f78f3ff6ec4a" />

Which then encounters an error when the linked name includes encoded labelhashes:

<img width="2628" height="1548" alt="Image" src="https://github.com/user-attachments/assets/310de262-81e7-4f8e-b801-18a28bbdab7f" />

shrugs:
based on my current read of the ens contracts, i'm not sure that we should be using encoded-labelhash-aware logic for resolution, so the 'solution' is to just resolve that name as a `LiteralName` and call it a day

PR #1908, opened 2026-04-09
Refactor/enssdk refactor continued

### LINKED ISSUE #1876: `enssdk`: Further migrate types and libs to enssdk
reference: https://github.com/namehash/ensnode/pull/1867#discussion_r3039204064

- [x] `NormalizedAddress` branded type refactor, `asLowerCaseAddress` becomes `asNormalizedAddress` etc
- [x] definitely need something like `asInterpretedName` to go from Name to InterpretedName with a helpful api for consumers like the example app parsing url params
  - current it just validates, but we should allow converstion
- [x] refactor the `packages/enssdk/src/lib/names.ts` utils to accept InterpretedName instead of Name or `NormalizedName`
  - likely allows us to get rid of `NormalizedName` entirely — make sure that the protocol acceleration lib can use `InterpretedName` instead of `NormalizedName`
  - update the test case in `apps/ensapi/src/lib/handlers/params.schema.test.ts` to test interpreted names as well as normalized names
- [x] clean up `apps/ensrainbow/src/commands/convert-csv-command.test.ts` to use the `await expect().resolves` pattern
- [x] manually test resolution with an InterpretedName, making sure to hit `packetToBytes` at
  - `apps/ensapi/src/lib/resolution/resolve-calls-and-results.ts`
  - `apps/ensapi/src/lib/protocol-acceleration/find-resolver.ts` (`findResolverWithUniversalResolver`)
  - `apps/ensapi/src/lib/resolution/resolve-with-universal-resolver.ts`

shrugs:
closes https://github.com/namehash/ensnode/issues/1876

1. resolution libs now operate over (normalized) `InterpretedName`s — deferring further constraints to https://github.com/namehash/ensnode/issues/1920
2. `NormalizedAddress` is an unbranded alias for `0x{string}`; all indexer/ens-referrals callsites cast rather than validate — spot-check that boundary validation still happens where it matters. the important entrypoints for the ens-referrals, ensanalytics, registrar-actions, and associated modules have all been checked and are enforced by zod schemas.
4. new `enskit/react` submodule export + `EnsureInterpretedName` relocation — verify the package surface is what you expect

## What Changed (Concrete)

1. removed `NormalizedName`; resolution libs + callsites now use `InterpretedName`
2. introduced `NormalizedAddress` (unbranded alias for `0x{string}`) replacing `asLowerCaseAddress` + `makeLowerCaseAddressSchema`; updated all indexer + ens-referrals callsites
3. added `enskit/react` submodule export; moved `EnsureInterpretedName` there
5. expanded `literalNameToInterpretedName` logic; added `nameToInterpretedName` helper and tests
6. omnigraph: local resolvers, bigint deserialization, `bigint[]` support, semantic scalars (`Duration` moved to enssdk)
7. encoded-referrer: DRY'd encoded labelhash parsing, fixed size check, `decodeEncodedReferrer` now returns `NormalizedAddress`
8. root name handling in `parse-reverse-name`; `ENS_ROOT_NAME` / `ENS_ROOT_NODE` cleanup
9. dropped `@urql/introspection` codegen and the separate `generated/introspection.ts`, can use the gql.tada generated file for urql introspection
10. misc: typos/docstrings

## Design & Planning

- follow-up to the enssdk refactor tracked in #1876; planning was lightweight and incremental — decisions landed in commits and copilot notes rather than a design doc
- Planning artifacts: issue #1876

## Self-Review

- Bugs caught: referrer size check, root name in `parse-reverse-name`, partial-data guard in cache-exchange, encoded labelhash validation in interpretation
- Logic simplified: `ensure-interpreted-name` streamlined, `validateAddress` refactored, `toNormalizedAddress` input type relaxed
- Naming / terminology improved: `NormalizedName` → `InterpretedName` project-wide; `Address` → `NormalizedAddress` where the invariant holds
- Dead / unnecessary code removed: dropped ambiguous `readFragment` re-export in `enskit/react/omnigraph`; removed `NormalizedName` type + schema

- Public APIs affected: enssdk type exports (`NormalizedName` removed), enskit gains `/react` submodule, omnigraph scalars reshuffled
- Docs updated: schema.graphql regenerated, enssdk `_lib/README.md`, AGENTS.md note on running test files
- Decisions worth calling out: `NormalizedAddress` is intentionally unbranded — it's a documentation alias, not a runtime invariant. We should probably brand it in the near future?

## Testing Evidence

- Testing performed: added tests for `interpreted-names-and-labels`, `encoded-referrer`, `packetToBytes`, `resolve-with-universal-resolver` integration, omnigraph builder
- Known gaps: no e2e coverage for the full indexer re-run — type system carries most of the refactor safety
- What reviewers have to reason about manually: the `NormalizedAddress` cast sites (unbranded → trust the caller)

## Scope Reductions

- Follow-ups: https://github.com/namehash/ensnode/issues/1920
- Why deferred: needs discussion

## Risk Analysis

- Risk areas: any place an `InterpretedName` is passed where a literal name was previously expected; `NormalizedAddress` casts on data from untrusted sources
- Mitigations / rollback: revert is clean — changes are type-level + mechanical


apps/ensapi/src/lib/resolution/forward-resolution.ts
```diff
@@ -57,10 +55,6 @@ import {
 const logger = makeLogger("forward-resolution");
 const tracer = trace.getTracer("forward-resolution");
 
-// NOTE: normalize generic name to force the normalization lib to lazy-load itself (otherwise the
```
shrugs, line 60: just didn't think it was worth

apps/ensapi/src/lib/protocol-acceleration/find-resolver.ts
```diff
@@ -147,7 +145,8 @@ async function findResolverWithUniversalResolver(
       }
 
       // UniversalResolver returns the offset in bytes within the DNS Encoded Name where the activeName begins
-      const activeName: Name = bytesToPacket(dnsEncodedNameBytes.slice(offset));
+      // Invariant: the decoded name must be an InterpretedName
+      const activeName = asInterpretedName(bytesToPacket(dnsEncodedNameBytes.slice(offset)));
```
lightwalker-eth, line 149: Shouldn't we say that `bytesToPacket` returns a `LiteralName` and that we therefore need to transform from the `LiteralName` to an `InterpretedName`?

apps/ensapi/src/lib/resolution/forward-resolution.ts
```diff
@@ -95,9 +90,12 @@ export async function resolveForward<SELECTION extends ResolverRecordsSelection>
   selection: ForwardResolutionArgs<SELECTION>["selection"],
   options: Omit<Parameters<typeof _resolveForward>[2], "registry">,
 ): Promise<ForwardResolutionResult<SELECTION>> {
+  // Invariant: Name must be an InterpretedName
+  const interpretedName = asInterpretedName(name);
```
lightwalker-eth, line 95: We have a related issue open: https://github.com/namehash/ensnode/issues/1032

What do you suggest we do in relation to issue 1032 within the scope of this PR (PR 1908)? Should we work to confirm that issue 1032 can be successfully closed, or maybe keep issue 1032 open and make any new relevant comments on it that might help us with working on it later?

lightwalker-eth, line 95: I believe the `name` param (currently typed as `ForwardResolutionArgs<SELECTION>["name"]`) should be updated so that it is already a validated `InterpretedName` and that the validation that it is a valid interpreted name might be handled by Zod when we're parsing the query params?

We could then also remove the need for creating `interpretedName` here as a distinct field from the `name` param.

apps/ensapi/src/omnigraph-api/lib/get-canonical-path.ts
```diff
@@ -63,7 +63,7 @@ export async function getV1CanonicalPath(domainId: ENSv1DomainId): Promise<Canon
 
   // v1Domains are canonical if the TLD's parent is ROOT_NODE (ROOT_NODE itself does not exist in the index)
```
lightwalker-eth: Suggest to look for any occurrences of `ROOT_NODE` and change them to `ENS_ROOT_NODE`.

apps/ensapi/src/omnigraph-api/schema/permissions.integration.test.ts
```diff
@@ -35,14 +37,14 @@ describe("Permissions", () => {
   type PermissionsResult = {
     permissions: {
       id: string;
-      contract: { chainId: number; address: Address };
+      contract: { chainId: number; address: NormalizedAddress };
```
lightwalker-eth:
```suggestion
      contract: { chainId: ChainId; address: NormalizedAddress };
```

wdyt?

apps/ensindexer/src/plugins/ensv2/handlers/ensv1/BaseRegistrar.ts
```diff
@@ -52,8 +52,8 @@ export default function () {
     }: {
       context: IndexingEngineContext;
       event: EventWithArgs<{
-        from: Address;
-        to: Address;
+        from: NormalizedAddress;
+        to: NormalizedAddress;
         tokenId: bigint;
```
lightwalker-eth:
```suggestion
        tokenId: TokenId;
```

wdyt?

apps/ensindexer/src/plugins/ensv2/handlers/ensv1/BaseRegistrar.ts
```diff
@@ -98,7 +98,7 @@ export default function () {
     context: IndexingEngineContext;
     event: EventWithArgs<{
       id: bigint;
-      owner: Address;
+      owner: NormalizedAddress;
       expires: bigint;
```
lightwalker-eth: Maybe we introduce a type alias for `BigUnixTimestamp` that's a `bigint` rather than how `UnixTimestamp` is a `number`.

Just an idea. Up to you 👍

apps/ensindexer/src/plugins/ensv2/handlers/ensv2/ENSv2Registry.ts
```diff
@@ -332,7 +332,7 @@ export default function () {
     event,
   }: {
     context: IndexingEngineContext;
-    event: EventWithArgs<{ id: bigint; to: Address }>;
+    event: EventWithArgs<{ id: bigint; to: NormalizedAddress }>;
```
lightwalker-eth: I see a number of fields in our token logic named `id` with type `bigint`. Should we apply the `TokenId` type alias on these to give them more semantic meaning?

apps/ensindexer/src/plugins/ensv2/handlers/ensv2/ENSv2Registry.ts
```diff
@@ -46,9 +46,9 @@ export default function () {
       labelHash: LabelHash;
       label: string;
```
lightwalker-eth, line 49: I guess this would be a `LiteralLabel`?

apps/ensindexer/src/plugins/ensv2/handlers/ensv2/ETHRegistrar.ts
```diff
@@ -54,12 +54,12 @@ export default function () {
       event: EventWithArgs<{
         tokenId: TokenId;
         label: string;
```
lightwalker-eth, line 59: I assume this is a `LiteralLabel`?

apps/ensindexer/src/plugins/ensv2/handlers/ensv2/ETHRegistrar.ts
```diff
@@ -54,12 +54,12 @@ export default function () {
...
-        subregistry: Address;
-        resolver: Address;
+        owner: NormalizedAddress;
+        subregistry: NormalizedAddress;
+        resolver: NormalizedAddress;
         duration: bigint;
         referrer: EncodedReferrer;
-        paymentToken: Address;
+        paymentToken: NormalizedAddress;
         base: bigint;
```
lightwalker-eth: Not sure if these values are exposed through our Omnigraph API?

It might be nice to define a type alias for `Wei`

apps/ensindexer/src/plugins/subgraph/shared-handlers/NameWrapper.ts
```diff
@@ -356,7 +357,7 @@ export const makeNameWrapperHandlers = () => {
       event,
     }: {
       context: IndexingEngineContext;
-      event: EventWithArgs<{ ids: readonly bigint[]; to: Address }>;
+      event: EventWithArgs<{ ids: readonly bigint[]; to: NormalizedAddress }>;
```
lightwalker-eth: I guess these ids are `TokenId[]`?

packages/ens-referrals/src/v1/award-models/rev-share-limit/rules.ts
```diff
@@ -162,7 +163,7 @@ export function isReferrerQualifiedRevShareLimit(
   totalBaseRevenueContribution: PriceUsdc,
   rules: ReferralProgramRulesRevShareLimit,
 ): boolean {
-  const normalizedReferrer = normalizeAddress(referrer);
+  const normalizedReferrer = toNormalizedAddress(referrer);
   const isAdminDisqualified = rules.disqualifications.some(
     (d) => d.referrer === normalizedReferrer,
   );
```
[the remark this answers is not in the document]
lightwalker-eth: This is a good suggestion from CodeRabbit. I created a follow-up issue for Mykola: https://github.com/namehash/ensnode/issues/1910

packages/ensnode-sdk/src/omnigraph-api/example-queries.ts
```diff
@@ -28,12 +30,12 @@ const ENS_TEST_ENV_V2_ETH_REGISTRAR = maybeGetDatasourceContract(
...
+const DEVNET_DEPLOYER = toNormalizedAddress("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266");
+const DEVNET_OWNER = toNormalizedAddress("0x70997970C51812dc3A010C7d01b50e0d17dc79C8");
 // biome-ignore lint/correctness/noUnusedVariables: keeping it around for the future
-const DEVNET_USER = "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC";
+const DEVNET_USER = toNormalizedAddress("0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC");
 
-const VITALIK_ADDRESS = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045";
+const VITALIK_ADDRESS = toNormalizedAddress("0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045");
 
 const DEVNET_NAME_WITH_OWNED_RESOLVER = "example.eth";
```
lightwalker-eth: I assume we want to make this an interpreted name?

packages/ensnode-sdk/src/registrars/zod-schemas.ts
```diff
@@ -127,7 +127,7 @@ const makeRegistrarActionReferralSchema = (valueLabel: string = "Registrar Actio
           { bytesCount: ENCODED_REFERRER_BYTE_LENGTH },
           `${valueLabel} Encoded Referrer`,
         ),
-        decodedReferrer: makeLowercaseAddressSchema(`${valueLabel} Decoded Referrer`),
+        decodedReferrer: makeNormalizedAddressSchema(`${valueLabel} Decoded Referrer`),
       })
       .check(invariant_registrarActionDecodedReferrerBasedOnRawReferrer),
```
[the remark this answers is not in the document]
lightwalker-eth: This looks like a good catch by CodeRabbit about line 99 in this file.

packages/enssdk/src/lib/types/evm.ts
```diff
@@ -1,9 +1,26 @@
 import type { Address as ViemAddress, Hex as ViemHex } from "viem";
 
-// re-export viem Address/Hex for consistency
+/**
+ * Represents an Hex string, in the format `0x{string}`.
```
lightwalker-eth:
```suggestion
 * Represents a Hex string, in the format `0x{string}`.
```

packages/enssdk/src/lib/parse-reverse-name.ts
```diff
@@ -21,11 +21,9 @@ const REVERSE_NAME_REGEX = /^([0-9a-fA-F]+)\.([0-9a-f]{1,64}|addr|default)\.reve
 const parseAddressLabel = (addressLabel: Label): Address => {
```
lightwalker-eth: I believe this should be a `LiteralLabel` but not 100% sure how this value is getting passed around.

packages/enssdk/src/lib/normalization.ts
```diff
@@ -18,8 +18,8 @@ export const normalizeName = (name: Name): InterpretedName =>
  * @see https://docs.ens.domains/ensip/15
  */
 export const normalizeLabel = (label: Label): InterpretedLabel => {
-  // empty string cannot be a label
-  if (label === "") throw new Error("Empty string is not a valid Label.");
+  // empty string cannot be a normalized label
+  if (label === "") throw new Error("Empty string cannot be normalized.");
```
lightwalker-eth:
```suggestion
  if (label === "") throw new Error("Empty label cannot be normalized.");
```

In the context of a name (not a label) the empty string is normalized.

packages/ensnode-sdk/src/shared/interpretation/interpret-record-values.ts
```diff
@@ -9,20 +9,20 @@ import { hasNullByte } from "../null-bytes";
...
  * a) null, representing a non-existant or deletion of the record, or
- * b) a normalized, non-empty-string Name.
+ * b) an {@link InterpretedName}.
  *
  * @param value - The name record value string to interpret.
  * @returns The interpreted name string, or null if deleted.
  */
-export function interpretNameRecordValue(value: string): NormalizedName | null {
-  // empty string is technically a normalized name, representing the ens root node, but in the
+export function interpretNameRecordValue(value: string): InterpretedName | null {
```
lightwalker-eth: My understanding is that the `value` param should be a `LiteralName`. Is that right?

packages/ensnode-sdk/src/shared/interpretation/interpret-record-values.ts
```diff
@@ -9,20 +9,20 @@ import { hasNullByte } from "../null-bytes";
...
+  // empty string is technically an InterpretedName, representing the ens root node, but in the
   // context of a name record value, empty string is emitted when the user un-sets the record (this
   // is because the abi of this event is only capable of expressing string values, so empty string
   // canonically represents the non-existence or deletion of the record value)
   if (value === "") return null;
 
   // if not normalized, is not valid `name` record value
-  if (!isNormalizedName(value)) return null;
+  if (!isInterpretedName(value)) return null;
```
lightwalker-eth, line 25: I think it would be correct that we exclusively allow `LiteralNames` that are already a "normalized name" here?

For example, let's imagine a case that the LiteralName contained an encoded labelhash. As I understand we would want to reject that in this context?

packages/ensdb-sdk/src/ensindexer-abstract/protocol-acceleration.schema.ts
```diff
@@ -28,13 +36,9 @@ export const reverseNameRecord = onchainTable(
     /**
      * Represents the ENSIP-19 Reverse Name Record for a given (address, coinType).
      *
-     * The value of this field is guaranteed to be a non-empty-string normalized ENS name (see
-     * `interpretNameRecordValue` for additional context and specific guarantees). Unnormalized
-     * names and empty string values are interpreted as a deletion of the associated Reverse Name
-     * Record entity (represented in the schema as the _absence_ of a relevant Reverse Name Record
-     * entity).
+     * The value of this field is guaranteed to be a non-empty {@link InterpretedName}.
```
lightwalker-eth, line 39: Please see another comment I shared on where the interpretation of these values happens in our indexing logic.

It's my understanding that we would want any non-null values held in this field to be guaranteed to be a normalized name.

It's of course all good to type this as an `InterpretedName` as all normalized names are interpreted names too. But perhaps the JSDoc here might be updated based on these ideas.

packages/ensdb-sdk/src/ensindexer-abstract/protocol-acceleration.schema.ts
```diff
@@ -124,14 +128,9 @@ export const resolverRecords = onchainTable(
...
-     * The emitted record values are interpreted according to `interpretNameRecordValue` — unnormalized
-     * names and empty string values are interpreted as a deletion of the associated record (represented
-     * here as `null`).
-     *
-     * If set, the value of this field is guaranteed to be a non-empty-string normalized ENS name
-     * (see `interpretNameRecordValue` for additional context and specific guarantees).
+     * If present, the value of this field is guaranteed to be a non-empty {@link InterpretedName}.
      */
-    name: t.text(),
+    name: t.text().$type<InterpretedName>(),
```
lightwalker-eth, line 133: Please see my related feedback on the other `name` field in this file 👍

apps/ensapi/src/handlers/api/explore/name-tokens-api.ts
```diff
@@ -83,7 +83,9 @@ app.openapi(getNameTokensRoute, async (c) => {
       );
     }
 
-    const parentNode = namehashInterpretedName(getParentNameFQDN(name));
+    // biome-ignore lint/style/noNonNullAssertion: getParentInterpretedName only throws if name === ENS_ROOT_NAME but we handle that case above
```
lightwalker-eth: Perhaps remove this ignore and the `!` and instead do the check here for `parentName` being null and if so move the code here that generates the 404 error for name param not being the ENS Root.

apps/ensapi/src/lib/resolution/packetToBytes.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  it("correctly DNS-encodes an InterpretedName containing an encoded labelhash label", () => {
+    const name = asInterpretedName(
+      interpretedLabelsToInterpretedName([ENCODED_LABEL_HASH, asInterpretedLabel("eth")]),
+    );
+    const expectedLabelHex = stringToHex(ENCODED_LABEL_HASH).slice(2);
+
+    // The encoded labelhash label (e.g. "[9c22ff...b658]") is 66 characters and should
+    // be DNS-encoded as a single label: length prefix 0x42 (66) followed by its UTF-8 bytes,
+    // followed by the 'eth' label with length (03 65 74 68) and null terminator (00)
+    const result = bytesToHex(packetToBytes(name));
```
lightwalker-eth, line 26: I believe `packetToBytes` should exclusively operate on `LiteralName` values and never on an `InterpretedName`.

Suggest we build our own wrapper implementation of `packetToBytes` that's optimized for our type system.

packages/enskit/src/react/omnigraph/components/ensure-interpreted-name.tsx
```diff
@@ -0,0 +1,34 @@
...
+type MalformedNameRenderer = (name: string) => ReactNode;
+type InterpretedNameRenderer = (name: InterpretedName) => ReactNode;
+
+/**
+ * Renders a Name by ensuring it is an InterpretedName. If the name is already
+ * interpreted, renders via `children`. If not, attempts to interpret it and
+ * renders via `interpreted` (enabling e.g. a redirect). If interpretation
+ * fails (malformed or unnormalizable), renders via `malformed`.
+ */
+export function EnsureInterpretedName({
```
lightwalker-eth: Nice!

Only feedback is to ask if this should live under `/omnigraph/...` or not? As this looks like a utility that could be useful in ENSKit outside the scope of the Omnigraph.

packages/enssdk/src/lib/address.ts
```diff
@@ -1,11 +1,35 @@
-import type { Address } from "./types";
+import { isAddress } from "viem";
+
+import type { Address, NormalizedAddress } from "./types";
+
+/**
+ * Determines whether an {@link Address} is a {@link NormalizedAddress}.
+ */
+export function isNormalizedAddress(address: Address): address is NormalizedAddress {
+  const isNormalized = address === address.toLowerCase();
```
lightwalker-eth:
```suggestion
  const isLowercase = address === address.toLowerCase();
```

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -13,6 +18,27 @@ import type {
...
+  return interpretedLabelsToInterpretedName(
+    name
+      .split(".")
+      .map((label) => {
+        if (isEncodedLabelHash(label)) return label;
+        if (isNormalizedLabel(label)) return label;
+        return encodeLabelHash(labelhashLiteralLabel(asLiteralLabel(label)));
+      })
+      .map(asInterpretedLabel),
+  );
```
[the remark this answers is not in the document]
lightwalker-eth: @shrugs This looks like a nice catch from Copilot. If the user input name is "" then the returned `InterpretedName` should also be "".

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -13,6 +18,28 @@ import type {
...
+ *   but is distinct in that we're interpreting a Name (i.e. from user input) rather than a previously
+ *   interpreted InterpretedName.
+ */
+export function nameToInterpretedName(name: Name): InterpretedName {
+  return interpretedLabelsToInterpretedName(
+    name
+      .split(".")
+      .map((label) => {
+        if (isEncodedLabelHash(label)) return label;
+        if (isNormalizedLabel(label)) return label;
```
lightwalker-eth: Am I correct to understand that the `nameToInterpretedName` function is intended for the goal of parsing names from user input?

Assuming so:

1. Perhaps change the input param so `name` is renamed to `maybeName` and its type changes from `Name` to `string`.
2. Perhaps trim the input? There might be some optional options passed into this function where the logic to trim the input defaults to true.
3. If a label is not normalized but is normalizable we shouldn't convert it into an encoded labelhash. Instead we should normalize it.
4. In the case of a label not being normalizable, in most situations I don't believe the ideal behaviour would be to convert it into an encoded labelhash. Instead, I believe in most situations the ideal behaviour would be to throw an error. I can imagine some optional options being passed into this function where the option to reject unnormalizable labels would default to true.
5. We probably would want an option to reject the empty name. By default the empty name would be rejected.
6. Maybe we rename this function from `nameToInterpretedName` to `userInputToInterpretedName`?

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -182,42 +216,34 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
   return { concrete, partial };
 }
 
+/**
+ * Validates and casts a string to a {@link LiteralLabel}.
+ * A LiteralLabel is a label as it literally appears onchain.
```
lightwalker-eth:
```suggestion
 * Casts a string to a {@link LiteralLabel}.
 * A LiteralLabel is a label that should be interpreted as a string literal. It may or may not be normalized or normalizable. It may also be formatted as an encoded labelhash, but that does not mean it should be interpreted as an encoded labelhash.
```

Goals include expanding the definition here for cases that aren't strictly onchain. Ex: A literal label can come from any source of input, such as a database, somewhere in DNS, etc.

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -182,42 +216,34 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
...
+ */
+export function asLiteralLabel(label: Label): LiteralLabel {
```
lightwalker-eth, line 315: Maybe we completely remove the `Label` and `Name` type aliases? I worry they will be a source of problems and that only our more precise type names should be used.

The only case I can think of right now that's not covered by our more formal type names would be what happens when we beautify an `InterpretedName`. This doesn't return an `InterpretedName`. Instead it returns what we might give a new type alias for: a `BeautifiedName`, which is the same as an `InterpretedName` except it loosens the constraints such that a label might be normalizable but not necessarily normalized.

In other words:
- An `InterpretedName` is all *normalized* labels or encoded labelhashes.
- A `BeautifiedName` is all *normalizable* labels or encoded labelhashes.

When we display names in a UI we should render them as `BeautifiedName` values.

Please also note that I believe raffy's beautify name function won't properly handle labels that are encoded labelhashes and would throw on those. So we would need to implement our own wrapper around raffy's beautify name function to correctly convert any `InterpretedName` into a `BeautifiedName`.

Appreciate your advice 👍

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -182,42 +216,34 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
...
+export function asInterpretedLabel(label: Label): InterpretedLabel {
+  if (isInterpetedLabel(label)) return label;
 
-  return label as InterpretedLabel;
+  throw new Error(`Not a valid InterpretedLabel: '${label}'`);
 }
 
 /**
  * Validates and casts a string to an {@link InterpretedName}.
  * An InterpretedName is composed entirely of InterpretedLabels joined by dots.
```
lightwalker-eth: Suggest we also make it explicit how an empty name is also an InterpretedName.

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -182,42 +216,34 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
...
+ * Validates and casts a string to a {@link LiteralLabel}.
+ * A LiteralLabel is a label as it literally appears onchain.
+ */
+export function asLiteralLabel(label: Label): LiteralLabel {
+  return label as LiteralLabel;
+}
+
 /**
  * Validates and casts a string to an {@link InterpretedLabel}.
  * An InterpretedLabel is either a normalized label or an EncodedLabelHash.
```
lightwalker-eth, line 321: One worry I have is about how we handle the parsing of EncodedLabelHash values. Do we need to modify anything here in relation to case sensitivity?

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -182,42 +216,34 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
...
-/**
- * Validates and casts a string to a {@link LiteralLabel}.
- * A LiteralLabel is a label as it literally appears onchain.
- *
- * @throws if the input is empty
+ * @throws if the input cannot be interpreted into an InterpretedName
  */
-export function asLiteralLabel(label: string): LiteralLabel {
-  if (label === "") throw new Error("LiteralLabel must not be empty");
+export function asInterpretedName(name: Name): InterpretedName {
```
lightwalker-eth, line 338: This feedback is in relation to your work on `nameToInterpretedName` which is really awesome.

A worry I have is that devs using this package will be lazy with how they use `asInterpretedName`. Specifically worried that they will use it when parsing names from user input instead of using `nameToInterpretedName`.

Perhaps we should rename `asInterpretedName` to something that makes it more explicit it is just for "system inputs" and not user inputs? Appreciate your advice.

lightwalker-eth (approved):
@shrugs Awesome updates here 😄 Reviewed and shared some suggestions 👍 Appreciate your advice!

packages/enssdk/src/omnigraph/introspection.ts
```diff
@@ -1 +1,2 @@
-export { introspection } from "./generated/introspection";
+export { introspection } from "./generated/graphql-env";
+// export { introspection } from "./generated/introspection";
```
lightwalker-eth: Adding comment as I imagine you were planning to look at this again.

packages/enskit/src/react/omnigraph/_lib/local-bigint-resolvers.ts
```diff
@@ -0,0 +1,35 @@
+import type { Resolver, ScalarObject } from "@urql/exchange-graphcache";
+
+import { type IntrospectionSchema, unwrapType } from "./introspection-helpers";
+
+// graphcache's ResolverResult type doesn't include bigint, but the value is stored
+// in the normalized cache and returned to the consumer as-is, so bigint works at runtime
+// the load-bearing piece of
```
lightwalker-eth: Love all the comments. I think this one is a work in progress

packages/enskit/src/react/omnigraph/_lib/cache-exchange.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const omnigraphCacheExchange = cacheExchange({
+  schema: introspection,
+  keys: {
+    // by default, all Entities are assumed to match the Relay spec, and graphcache treats
+    // them as keyable by `id`. if it encounters an Entity with no `id` field and no other
+    // special handling here in the cacheExchange.keys definitions, it will issue a warning.
+
+    // AccountIds are keyable by stringifying them
+    AccountId: (data) => stringifyAccountId(data as unknown as AccountId),
```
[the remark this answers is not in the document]
lightwalker-eth: This feedback from CodeRabbit might be valid. Appreciate your advice 🤔

packages/enskit/src/react/omnigraph/_lib/README.md
```diff
@@ -0,0 +1 @@
+enskit/react/omnigraph libs or helpers not meant for export to consumers
```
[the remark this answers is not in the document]
lightwalker-eth: Looks like a good minor comment from Coderabbit

apps/ensapi/src/lib/resolution/resolve-with-universal-resolver.integration.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      executeResolveCallsWithUniversalResolver({
+        name: NAME,
+        calls: CALLS,
+        publicClient,
+      }),
+    ).resolves.toMatchObject([{ result: EXPECTED_DESCRIPTION }]);
+  });
+
+  /**
+   * NOTE(shrugs): This was contrary to my expectations, but the NameCoder (in both ENSv1 and ENSv2)
```
lightwalker-eth, line 52: Nice work flagging this issue. It goes pretty deep. How about we have a dedicated sync up about it on Monday?

lightwalker-eth (approved):
@shrugs Great updates! Shared a few very minor suggestions. Looks really good!

packages/ensnode-sdk/src/shared/interpretation/interpret-record-values.ts
shrugs: deferred to https://github.com/namehash/ensnode/issues/1920

packages/ensdb-sdk/src/ensindexer-abstract/protocol-acceleration.schema.ts
shrugs: deferred to https://github.com/namehash/ensnode/issues/1920

apps/ensapi/src/lib/resolution/packetToBytes.test.ts
shrugs: agree https://github.com/namehash/ensnode/issues/1920

packages/enssdk/src/lib/interpreted-names-and-labels.ts
shrugs: hmm, i see the concern, but i'm loath to change the existing `as*` and `to*` pattern for these branded types. hopefully with enough example cases and public-facing usage of the `literalNameToInterpretedName` function this won't be an issue?

packages/enssdk/src/lib/interpreted-names-and-labels.ts
shrugs: good catch, handled

shrugs:
@greptile re-review

shrugs:
@greptile re-review

lightwalker-eth:
@greptile

apps/ensapi/src/lib/resolution/forward-resolution.ts
```diff
@@ -139,6 +140,14 @@ async function _resolveForward<SELECTION extends ResolverRecordsSelection>(
           // Validate Input
           //////////////////////////////////////////////////
 
+          // TODO: technically InterpretedNames are not resolvable, since ENS contracts are not
```
lightwalker-eth, line 143: Comment here is great 👍 

Suggest to add a link to the related issue here too: https://github.com/namehash/ensnode/issues/1920

packages/ens-referrals/src/referrer-metrics.ts
```diff
@@ -28,7 +27,7 @@ export interface ReferrerMetrics {
    *
    * @invariant Guaranteed to be a valid EVM address in lowercase format
```
lightwalker-eth: Thanks for the update here 👍 We can also simplify the JSDocs now too.

packages/ens-referrals/src/v1/award-models/rev-share-limit/referral-event.ts
```diff
@@ -12,7 +12,7 @@ export interface ReferralEvent {
   /**
    * The fully lowercase Ethereum address of the referrer.
```
lightwalker-eth: We can also simplify JSDocs here now too 👍

packages/ens-referrals/src/v1/referrer-metrics.ts
```diff
@@ -19,7 +18,7 @@ export interface ReferrerMetrics {
    *
    * @invariant Guaranteed to be a valid EVM address in lowercase format
```
lightwalker-eth: Found a JSDoc refinement opportunity

packages/enssdk/src/lib/parse-reverse-name.ts
```diff
@@ -41,7 +38,10 @@ const parseCoinTypeLabel = (coinTypeLabel: Label): CoinType => {
 /**
  * Parse the address and coinType out of an ENSIP-19 reverse name.
  */
-export function parseReverseName(name: Name): { address: Address; coinType: CoinType } | null {
+export function parseReverseName(name: Name): {
```
lightwalker-eth: Should this be a `LiteralName`?

examples/enskit-react-example/src/DomainView.tsx
```diff
@@ -92,13 +87,30 @@ export function DomainView() {
...
+      }}
+      //
+      // this isn't an InterpretedName, but it was coerced to an InterpretedName: redirect the user to the canonical url
+      coerced={(name) => <Navigate to={`/domain/${name}`} replace />}
+      //
+      // this name can't conform to InterpretedName nor can it be coerced: it is malformed: show an error
       malformed={(name) => (
         <div>
-          <h2>{name} is malformed</h2>
+          <h2>{name} could not be understood</h2>
```
lightwalker-eth:
```suggestion
          <h2>Invalid name: {name}</h2>
```

I'm also not sure if there can be a potential security risk of putting an unnormalized name into HTML like this. Maybe it's possible to do something crazy like building a query param that once urldecoded is some HTML that embeds an import of some javascript file or something crazy? Not confident in my understanding of how all the related operations are exactly implemented.

Appreciate your advice. Also open to something like:

```suggestion
          <h2>Invalid name</h2>
```

I suppose we could validate what's ok with a little manual test? I assume no need for an automated test (complicated?). Just curious about this from first-principles.

examples/enskit-react-example/src/DomainView.tsx
```diff
@@ -55,7 +50,7 @@ function RenderDomain({ name }: { name: InterpretedName }) {
 
   if (fetching) return <p>Loading...</p>;
   if (error) return <p>Error: {error.message}</p>;
-  if (!data?.domain) return <p>A Domain with InterpretedName '{name}' was not found.</p>;
+  if (!data?.domain) return <p>A Domain with name '{name}' was not found.</p>;
```
lightwalker-eth:
```suggestion
  if (!data?.domain) return <p>No domain was found with name '{name}'.</p>;
```

packages/ensnode-sdk/src/registrars/encoded-referrer.test.ts
```diff
@@ -24,7 +24,7 @@ describe("encoded referrer", () => {
       const result = decodeEncodedReferrer(input);
 
       // assert
-      expect(result).toEqual(vitalikEthAddressChecksummed);
+      expect(result).toEqual(vitalikEthAddressLowercase);
```
lightwalker-eth, line 30: We can also update the type of `vitalikEthAddressLowercase`.

packages/ensnode-sdk/src/registrars/encoded-referrer.ts
```diff
@@ -49,26 +49,26 @@ export const ZERO_ENCODED_REFERRER: EncodedReferrer = pad("0x", {
 });
 
 /**
- * Build an {@link EncodedReferrer} value for the given {@link Address}
+ * Build an {@link EncodedReferrer} value for the given {@link NormalizedAddress}
  * according to the referrer encoding with left-zero-padding.
  */
-export function buildEncodedReferrer(address: Address): EncodedReferrer {
-  const lowercaseAddress = address.toLowerCase() as Address;
+export function buildEncodedReferrer(address: NormalizedAddress): EncodedReferrer {
```
lightwalker-eth, line 55: I was surprised to see the logic in this file inside the `ensnode-sdk` package!

I thought we had already moved all the ideas in this file into the `ens-referrals` package, but I see now that I am wrong.

In my mental model:
1. We want functions like `buildEncodedReferrer` and `decodeEncodedReferrer` to be utilities for the community to use that are building on the ENS Referral Program.
2. `buildEncodedReferrer` should be optimized for helping 3rd parties use it successfully with minimal friction. Therefore:
   1. The `address` param should be an `Address` rather than a `NormalizedAddress`.
   2. Instead of throwing an error if the provided `address` param is not a `NormalizedAddress`, we should internally attempt to convert the `address` param value to a `NormalizedAddress` and only throw an error if that fails.

We should also update the readme file of the `ens-referrals` package in the ["Other Utilities" section.](https://github.com/namehash/ensnode/tree/main/packages/ens-referrals#other-utilities) so that it includes a brief description of the `buildEncodedReferrer` function. This function is relevant for the context of apps integrating with the ENS Referral Program at a smart-contract layer and who are looking for a little utility function for building the params they pass onchain when making a registration or renewal.

As the mirror of `buildEncodedReferrer`, `decodeEncodedReferrer` should also be included in the `ens-referrals` package, but I don't believe it is worthy of inclusion in the README.md file since it is relevant for a much more narrow audience (ex: indexer implementations).

I also note that `ensindexer` currently doesn't include a dependency on the `@namehash/ens-referrals` package, but I think it should be fine to make a change here so that it does. We can note how the `ensapi` app already includes a dependency on `@namehash/ens-referrals` -- it's fine for `ensindexer` to add this dependency too.

I believe that the `ensdb-sdk` package should also add the `@namehash/ens-referrals` package as a dependency.

Files such as `packages/ensnode-sdk/src/registrars/encoded-referrer.test.ts` should also move into `ens-referrals`.

----

Edit: One other option could be to move these utility functions into `enssdk` instead of `ens-referrals`, but then to talk about them in the readme of the `ens-referrals` package. I'd be open to that too.

---

Please also note this related follow-up issue I created: https://github.com/namehash/ensnode/issues/1922

packages/enssdk/src/lib/labelhash.ts
```diff
@@ -51,12 +51,45 @@ export function isLabelHash(maybeLabelHash: string): maybeLabelHash is LabelHash
 export const encodeLabelHash = (labelHash: LabelHash): EncodedLabelHash =>
   `[${labelHash.slice(2)}]`;
 
+/**
+ * Parses an Encoded LabelHash (`[hash_without_0x_prefix]`) as a {@link LabelHash},
```
lightwalker-eth:
```suggestion
 * Parses an Encoded LabelHash (`[labelhash_without_0x_prefix]`) as a {@link LabelHash},
```

packages/enssdk/src/lib/labelhash.ts
```diff
@@ -51,12 +51,45 @@ export function isLabelHash(maybeLabelHash: string): maybeLabelHash is LabelHash
 export const encodeLabelHash = (labelHash: LabelHash): EncodedLabelHash =>
```
lightwalker-eth, line 51: I think it will be good for us to add formalism to our `LabelHash` and `EncodedLabelHash` types, similar to the formalism you've added to our `NormalizedAddress` type.

I'm concerned about ambiguity on the capitalization of hex digits inside encoded labelhash values or when comparing labelhash values for equality.

I suppose the same idea can be applied to our `Node` type which should also guarantee a fully lowercase hex value of a specific length.

Happy to move these ideas into a separate follow-up issue 👍

apps/ensapi/src/omnigraph-api/schema/permissions.integration.test.ts
```diff
@@ -40,30 +40,28 @@ const EAC_ROLES_CHANGED_SELECTOR = toEventSelector(
   )!,
 );
 
+// the type of the PermissionUser returned by the omnigraph, to DRY up the types below
+type PermissionUser = {
+  id: PermissionsUserId;
+  resource: string;
```
lightwalker-eth, line 46: There's several places in this file where we define a field named `resource`. I assume they aren't a `PermissionsResourceId`? Is there a semantic type we could use for this? If it's complicated to add now then no worries happy to create a separate follow-up issue for it 👍 Appreciate your advice.

apps/ensapi/src/omnigraph-api/schema/permissions.integration.test.ts
```diff
@@ -40,30 +40,28 @@ const EAC_ROLES_CHANGED_SELECTOR = toEventSelector(
...
+  user: { address: NormalizedAddress };
+  roles: string;
```
lightwalker-eth, line 48: Is there a semantic type we could use for this? If it's complicated to add now then no worries happy to create a separate follow-up issue for it 👍

examples/enskit-react-example/src/DomainView.tsx
```diff
@@ -92,13 +87,30 @@ export function DomainView() {
   // here we ensure that the provided /domain/:name parameter is an InterpretedName
   return (
     <EnsureInterpretedName
-      name={params.name}
-      // this isn't an InterpretedName, but it can conform to InterpretedName: redirect the user
-      interpreted={(name) => <Navigate to={`/domain/${name}`} replace />}
-      // this name can't conform to InterpretedName: it is malformed or contains unnormalizable Labels
+      name={asLiteralName(params.name)}
```
lightwalker-eth, line 90: I've gone back and forth over time wondering how we should classify names that are passed in URL params. It's a tricky question.

I don't believe we should treat them `asLiteralName` though, because we can consider what that would do for the case that the name passed in the query param was an `InterpretedName`. That would cause all the labels with encoded labelhashes to be interpreted as unnormalized labels (not encoded labelhashes) and therefore to be transformed into a new (and different!) encoded labelhash.

I believe an elegant solution could be achieved if we introduced new type aliases for: `UserInputName` and `UserInputLabel`. The `UserInputName` type alias would say it is a `string` consisting of 0 or more `UserInputLabel` (0 labels means it can be the empty string). The `UserInputLabel` type alias would say it is a string that may or may not be a normalized label (with a special note that it also can be an empty string, consider the case of a user input name such as "abc...xyz" -- however if any labels are formatted as encoded labelhashes they should be interpreted as encoded labelhashes and not `LiteralLabel`.

Then we could define a utility function for converting from a `UserInputName` -> `InterpretedName`. I note how this PR is introducing a new `literalNameToInterpretedName` function that includes some cool options. My first instinct is that we should change the `literalNameToInterpretedName` so that it works to transform `UserInputName` rather than `LiteralName` into `InterpretedName`.

I think it's nice to still have a `literalNameToInterpretedName` utility function, but in my mind it shouldn't have any options to it. It should be a strongly-opinionated operation where each label that is not a normalized label is transformed into an encoded labelhash. In my mind, the situation that benefits from options is a utility function for converting from a `UserInputName` -> `InterpretedName`. That's because depending on the context, different strategies for interpreting user input may be desired.

If we actioned the above ideas, we could then treat the case of parsing name inputs from query params or user input text fields using the new `UserInputName` -> `InterpretedName` utility function.

For a new `UserInputName` -> `InterpretedName` utility function:

1. Empty name -> optionally reject with an error. By default, option to reject with an error is true.
2. Empty label (ex: a name such as "abc..xyz") -> always reject with error, no option to allow it.
3. Normalized label -> no transformations needed, retain as-is.
4. Unnormalized but normalizable label -> optionally reject with an error. By default, option to reject is false. When option to reject is false, these labels should be transformed into their normalized form.
5. Unnormalized label that's formatted as an encoded labelhash -> I see 3 potential options here: reject with error, interpret as `InterpretedLabel`, or interpret as `LiteralLabel` that would then be transformed into an encoded labelhash. By default, this should interpret as `InterpretedLabel`.
6. Unnormalized label that's NOT formatted as an encoded labelhash -> I see 2 potential options here: reject with error or interpret as `LiteralLabel` that would be transformed into an encoded labelhash. By default, this should reject with error.

Let's assume we did all of the above. Assuming so, let's try to attack how a few cases would be handled:

### Case 1. A label is formatted as an encoded labelhash, but it was intended to be interpreted as a `LiteralLabel` and not an `InterpretedLabel`.

Fair enough. This is subjective and each situation can pick their own strategy for handling it based on the 3 potential option values I suggested in point 5 above.

### Case 2. A label is not normalized and is not formatted as an encoded labelhash.

First, let's note how such a label would be a violation of the `InterpretedName` type definition, but it would fully comply with the `UserInputName` type definition. All good there!

Up above, in points 2, 4, and 6 I believe all the possible cases of an unnormalized label that's not formatted as an encoded labelhash are well covered with a good plan.

The utility function and types proposed here could then be a mature answer for handling URL query params and user input text fields.

Appreciate your advice 👍

packages/enskit/src/react/components/ensure-interpreted-name.tsx
```diff
@@ -0,0 +1,59 @@
...
+ *   unnormalizable-Label handling. When `options.coerceUnnormalizedLabelsToNormalizedLabels` is `true` (default), an
+ *   unnormalized Label is passed through ENSIP-15 normalization (e.g. `"Vitalik"` → `"vitalik"`); when `false`, any
+ *   unnormalized Label causes `malformed` to be invoked — no normalization is attempted and
+ *   `coerceUnnormalizableLabelsToEncodedLabelHashes` is not consulted. When
+ *   `options.coerceUnnormalizableLabelsToEncodedLabelHashes` is `true`, a Label that cannot be normalized is replaced
+ *   with the EncodedLabelHash of its literal bytes and `coerced` is invoked; when `false` (default), encountering
+ *   such a Label causes `malformed` to be invoked. Only consulted when
+ *   `coerceUnnormalizedLabelsToNormalizedLabels` is `true`.
+ */
+export function EnsureInterpretedName({
```
lightwalker-eth, line 32: This is really cool 👍

Please see my other really big feedback comment suggesting a utility function that transforms from a `UserInputName` -> `InterpretedName` with a set of options.

Suggest that we update the logic here to align with that, including:
1. This should take as input a `UserInputName` rather than a `LiteralName`.
2. The `MalformedNameRenderer` could take as input the raw `UserInputName` that was malformed.

A new suggestion would be to update the `coerced` case so that it is a new `CoercedNameRenderer` that receives both the raw `UserInputName` and the coerced `InterpretedName` as params. Maybe it's useful in some situation to pass both?

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -19,24 +19,115 @@ import type {
...
- * @dev encodes unnormalized labels in `name` to make it InterpretedName.
- * @dev this is similar to the concept of Reinterpretation (packages/enssdk/src/lib/reinterpretation.ts)
- *   but is distinct in that we're interpreting a Name (i.e. from user input) rather than a previously
- *   interpreted InterpretedName.
+ * @throws if `name` cannot be coerced into an {@link InterpretedName} under the provided `options`.
+ *
+ * @dev casts to {@link InterpretedLabel} to avoid an additional unnecessary `asInterpretedLabel` pass.
  */
-export function nameToInterpretedName(name: Name): InterpretedName {
+export function literalNameToInterpretedName(
```
lightwalker-eth, line 66: This is very cool 👍 I shared some feedback on this function in a few other comments.

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -217,8 +295,22 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
 }
 
 /**
- * Validates and casts a string to a {@link LiteralLabel}.
- * A LiteralLabel is a label as it literally appears onchain.
+ * Casts a string to a {@link LiteralName}.
+ *
+ * A LiteralName is a name that should be interpreted as a string literal. It may or may not be
+ * normalized or normalizable. It may also include labels formatted as an encoded labelhash, but that
+ * does not mean it should be interpreted as perhaps containing encoded labelhashes.
```
lightwalker-eth:
```suggestion
 * normalized or normalizable. It may also include labels formatted as an encoded labelhash, but
 * such labels must be interpreted literally and not as an encoded labelhash.
```

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -217,8 +295,22 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
...
+export function asLiteralName(name: Name): LiteralName {
+  return name as LiteralName;
+}
+
+/**
+ * Casts a string to a {@link LiteralLabel}.
+ *
+ * A LiteralLabel is a label that should be interpreted as a string literal. It may or may not be
+ * normalized or normalizable. It may also be formatted as an encoded labelhash, but that does not
+ * mean it should be interpreted as an encoded labelhash.
```
lightwalker-eth:
```suggestion
 * normalized or normalizable. It may also be formatted as an encoded labelhash, but 
 * such labels must be interpreted literally and not as an encoded labelhash.
```

packages/enssdk/src/lib/interpreted-names-and-labels.ts
lightwalker-eth: @shrugs I continue to think it would be nice if we completely removed all use of the generic `Name` or `Label` types across our whole monorepo.

Please see my other comments suggesting the introduce new types for `UserInputName` and `UserInputLabel`.

We can then also add types for `BeautifiedName` and `BeautifiedLabel` based on the definition I shared in the parent comment I'm now replying to.

For our utility functions that work on unknown name / label values, those could operate on raw `string` values as we haven't validated yet if they are an `InterpretedName` (etc..) or not.

packages/enssdk/src/lib/interpreted-names-and-labels.test.ts
```diff
@@ -202,4 +205,159 @@ describe("interpretation", () => {
       },
     );
   });
+
+  describe("literalNameToInterpretedName", () => {
```
lightwalker-eth, line 209: Nice work on these tests 👍

lightwalker-eth (commented):
@shrugs Hey super work on the updates here! I think we're close. Shared a few comments with suggestions 👍

examples/enskit-react-example/src/DomainView.tsx
shrugs: likely fine since it'll be escaped by react into a raw string

packages/enssdk/src/lib/labelhash.ts
shrugs: https://github.com/namehash/ensnode/issues/1920

apps/ensapi/src/omnigraph-api/schema/permissions.integration.test.ts
shrugs: in this case it's the stringified bigint of the EAC contract's `resource`, so it likely needs a new semantic type down the road. these types here are just for the test file, low priority

apps/ensapi/src/omnigraph-api/schema/permissions.integration.test.ts
shrugs: similar to `resource`, should introduce them

packages/enskit/src/react/components/ensure-interpreted-name.tsx
shrugs: agree, deferred to https://github.com/namehash/ensnode/issues/1926

examples/enskit-react-example/src/DomainView.tsx
shrugs: agree, deferred to https://github.com/namehash/ensnode/issues/1926

packages/ensnode-sdk/src/registrars/encoded-referrer.ts
shrugs: agree, deferred to https://github.com/namehash/ensnode/issues/1927, working on a plan with claude now

packages/enssdk/src/lib/interpreted-names-and-labels.ts
```diff
@@ -298,8 +298,8 @@ export function parsePartialInterpretedName(partialInterpretedName: Name): {
  * Casts a string to a {@link LiteralName}.
  *
  * A LiteralName is a name that should be interpreted as a string literal. It may or may not be
- * normalized or normalizable. It may also include labels formatted as an encoded labelhash, but that
- * does not mean it should be interpreted as perhaps containing encoded labelhashes.
+ * normalized or normalizable. It may also include labels formatted as an EncodedLabelHash, but that
```
lightwalker-eth:
```suggestion
 * normalized or normalizable. It may also include labels formatted as an EncodedLabelHash, but
```

packages/enssdk/src/lib/parse-reverse-name.test.ts
```diff
@@ -132,20 +134,21 @@ const negativeCases: string[] = [
...
+      // NOTE: directly cast to avoid running asInterpretedName validation
+      expect(parseReverseName(input as InterpretedName)).toBeNull();
     });
   });
 
   it("parses constructed names", () => {
     CHAIN_IDS.forEach((chainId) => {
       const coinType = evmChainIdToCoinType(chainId);
-      expect(parseReverseName(reverseName(EXAMPLE_ADDRESS, coinType))).toEqual({
+      expect(parseReverseName(asInterpretedName(reverseName(EXAMPLE_ADDRESS, coinType)))).toEqual({
```
lightwalker-eth: I assume no need to call `asInterpretedName` name here if `reverseName` now returns an `InterpetedName`.

packages/enssdk/src/lib/reverse-name.ts
```diff
@@ -34,7 +34,7 @@ export const coinTypeReverseLabel = (coinType: CoinType): Label => coinType.toSt
  * reverseName("0x1234", BigInt(0x5678)) // "1234.5678.reverse"
  * ```
  */
-export function reverseName(address: NormalizedAddress, coinType: CoinType): Name {
+export function reverseName(address: NormalizedAddress, coinType: CoinType): InterpretedName {
   const label = addrReverseLabel(address);
```
lightwalker-eth, line 46: Suggest to review the typing that produces `label` and `middle` (ex: `coinTypeReverseLabel`) to ensure all the related typing produces `InterpretedLabel` for each of these labels.

If nothing to action here then all good, but maybe something so suggesting this.

packages/enssdk/src/lib/reverse-name.ts
```diff
@@ -48,5 +48,5 @@ export function reverseName(address: NormalizedAddress, coinType: CoinType): Nam
     }
   })();
 
-  return `${label}.${middle}.reverse`;
+  return `${label}.${middle}.reverse` as InterpretedName;
```
lightwalker-eth: If each of these labels are `InterpretedLabel` then we could use the utility function you built that converts an array of `InterpetedLabel` into an `InterpretedName`.

lightwalker-eth (approved):
@shrugs Looks good nice work! Happy to merge when you're ready ✅

PR #1915, opened 2026-04-11
fix(security): update deps to pass audit

tk-o:
Update npm dependencies to address:
- https://github.com/namehash/ensnode/security/dependabot/151
and allow the NPM dependency audit to pass successfully.


lightwalker-eth (approved):
@tk-o Thanks!

PR #1928, opened 2026-04-14
fix(ensindexer): hot-reload safe writer worker

### LINKED ISSUE #1432: indexing status api is kinda broken in development or serve mode and it's really annoying to work around

tk-o:
# fix(ensindexer): hot-reload safe writer worker

## Reviewer Focus

- `apps/ensindexer/src/lib/local-ponder-context.ts` — the `getApiShutdown()` / `getShutdown()` function pattern. The staleness contract (always read fresh, never cache) is enforced ergonomically by making call sites function calls rather than property accesses.
- `apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts` — the `.run().catch()` clean-stop discrimination: `(abortSignal.aborted || ensDbWriterWorker !== worker) && isAbortError(error)`. Both abort-source paths must be validated to avoid masking real errors AND to avoid misclassifying intentional supersession as a fatal error.
- `apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts` — the `stopRequested` guard. `stop()` flips it; `run()` checks via `checkCancellation()` between every async setup step so a stop during startup actually cancels.

## Problem

ponder dev-mode hot reload of indexing handler files crashed ensindexer with `Error: EnsDbWriterWorker has already been initialized` and `process.exit(1)`.

Root cause: ponder kills and replaces `common.shutdown` and `common.apiShutdown` on every indexing reload (`ponder/src/bin/commands/dev.ts:95-101`), then re-execs the api entry file. but vite-node only invalidates the changed file's dep tree — api-side singleton modules stay cached. so module-level state survives across reloads (singleton check throws), and any reference cached during the first boot goes stale.

resolves #1432.

## What Changed

1. **`local-ponder-context.ts`** — stable fields stay as an eager `const`. reload-scoped fields are exposed as functions `getApiShutdown()` / `getShutdown()` that re-read `globalThis.PONDER_COMMON` on every call. the function-call shape makes the fresh-read contract visible at every call site, so accidental caching reads as obviously wrong (`const sig = getApiShutdown().signal` clearly caches a function result; the equivalent property access wouldn't).
2. **`singleton.ts`** — `startEnsDbWriterWorker()` is reset-aware. drops the throw-on-re-init, awaits any prior worker's `stop()`, registers cleanup via `getApiShutdown().add(...)` so ponder properly awaits worker shutdown during its kill sequence. extracted `gracefulShutdown(worker, reason)` helper. `.run().catch()` distinguishes intentional stops (signal aborted OR worker superseded in singleton, AND error is `AbortError`) from real failures, which fail-fast via `process.exit(1)`.
3. **`ensdb-writer-worker.ts`** — `stop()` is async. `inFlightSnapshot` tracks the latest upsert with skip-overlap so `stop()` deterministically awaits one promise. new `stopRequested` flag + `checkCancellation()` helper let `stop()` cancel an in-progress `run()` startup before it arms the recurring interval. `run()` accepts an optional external `AbortSignal` for the same purpose.
4. **`packages/ponder-sdk`** — `PonderClient` accepts an optional `AbortSignalGetter` (typed alias). `LocalPonderClient` forwards it. invoked at fetch time inside `health/metrics/status` so requests use the current signal instead of a captured-at-construct reference. `PonderAppShutdownManager` interface + `isPonderAppShutdownManager` guard exported alongside `PonderAppContext` since they mirror a Ponder runtime type.

## Self-Review

- bugs caught during review iteration: typescript narrowing of `globalThis.PONDER_COMMON` lost when read moved into a function (added explicit check); `isAbortError` originally only checked `instanceof Error` and missed `DOMException` (the actual fetch-abort rejection type); the catch-path discriminator went through `||` → `&&` → `(signal.aborted || superseded) && isAbortError` as the supersession case from the defensive cleanup path was identified.
- logic simplified: replaced an initial reactive `Proxy` over `globalThis.PONDER_COMMON` with explicit getter functions. the proxy was technically correct but hid the staleness contract behind property-access syntax. -19 lines.
- naming: introduced "reload-scoped" vs "stable" field distinction to make the contract teachable.
- dead code removed: the `typeof ensDbWriterWorker !== "undefined"` throw is gone — re-init is the expected path.

## Cross-Codebase Alignment

- audited every api-side singleton (`ensDbClient`, `ensRainbowClient`, `publicConfigBuilder`, `indexingStatusBuilder`, `localPonderClient`) for stale-reference hazards. only `localPonderClient` captured a runtime mutable (the abort signal), addressed via the getter-pattern threading.
- `publicConfigBuilder` and `indexingStatusBuilder` cache `immutablePublicConfig` / `_immutableIndexingConfig` after first build. could mask config changes if you edit `config.ts` and reload — pre-existing pre-fix behavior, not the crash, deferred.

- `PonderClient` constructor: optional second arg `getAbortSignal?: AbortSignalGetter`. backwards compatible.
- `LocalPonderClient` constructor: optional 5th arg, same. backwards compatible.
- `EnsDbWriterWorker.stop()` is now `async` and returns `Promise<void>`. existing call sites (production + tests) updated to await.
- new exports from `@ensnode/ponder-sdk`: `AbortSignalGetter`, `PonderAppShutdownManager`, `isPonderAppShutdownManager`.

## Testing

- `pnpm -F ensindexer typecheck` ✓
- `pnpm -F @ensnode/ponder-sdk typecheck` ✓
- `pnpm test --project ensindexer --project @ensnode/ponder-sdk` → 268 passed (4 new `PonderClient` tests for the getAbortSignal getter pattern).
- `pnpm lint` ✓
- manual: `pnpm -F ensindexer dev` against ens-test-env, edited an indexing handler multiple times. confirmed: hot reload completes, no "already been initialized", no `ELIFECYCLE`, indexing resumes, snapshots continue upserting.

no automated hot-reload test infrastructure exists; manual smoke test is the only path.

## Risk Analysis

- the catch path's intentional-stop discrimination has been the most-reviewed area. final shape: `(signal.aborted || ensDbWriterWorker !== worker) && isAbortError(error)` — both an abort-source signal AND the AbortError name are required, ruling out both "real error during shutdown" and "AbortError from a cross-contamination race that didn't actually stop us."
- `stopRequested` + the external `AbortSignal` provide two independent cancellation channels. `checkCancellation()` honors both.
- `process.exit(1)` on real worker errors is explicit and fail-fast. can't rethrow from the fire-and-forget catch (would become an unhandled rejection rather than reaching api/index.ts).
- the function-call getter pattern can still be misused (capture the result), but the misuse reads as obviously wrong, which is the whole improvement over the proxy.
- mitigations / rollback: revert the commit; the previous (broken) behavior is well understood.
- named owner: @shrugs

## Pre-Review Checklist

### DIFF
--- a/apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
+++ b/apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
@@ -77,14 +77,17 @@ export async function startEnsDbWriterWorker(): Promise<void> {
     .run(abortSignal)
     // Handle any uncaught errors from the worker
     .catch(async (error) => {
-      // Treat as a clean stop only when BOTH the captured shutdown signal
-      // is aborted AND the error is an AbortError. Either condition alone
-      // could mask a real failure: a non-AbortError thrown after Ponder
-      // killed the signal is still a bug worth surfacing, and an
-      // AbortError without our signal aborted means it came from
-      // somewhere else (e.g. a reactive-getter race) and shouldn't be
-      // silently swallowed.
-      if (abortSignal.aborted && isAbortError(error)) {
+      // Treat as a clean stop when the error is an AbortError AND the
+      // worker was intentionally stopped — either because Ponder aborted
+      // the captured shutdown signal, or because this worker has been
+      // superseded in the singleton (the defensive stale-instance
+      // cleanup path inside `startEnsDbWriterWorker` calls
+      // `worker.stop()` on the old worker, which causes its `run()` to
+      // throw AbortError without `abortSignal.aborted` necessarily
+      // being true). Requiring `isAbortError(error)` keeps unrelated
+      // failures from being silently swallowed.
+      const intentionallyStopped = abortSignal.aborted || ensDbWriterWorker !== worker;
+      if (intentionallyStopped && isAbortError(error)) {
         await gracefulShutdown(worker, "API shutdown (run aborted)");
         return;
       }
--- a/apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
+++ b/apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
@@ -1,40 +1,86 @@
 import { ensDbClient } from "@/lib/ensdb/singleton";
 import { indexingStatusBuilder } from "@/lib/indexing-status-builder/singleton";
 import { localPonderClient } from "@/lib/local-ponder-client";
+import { localPonderContext } from "@/lib/local-ponder-context";
 import { logger } from "@/lib/logger";
 import { publicConfigBuilder } from "@/lib/public-config-builder/singleton";
 
 import { EnsDbWriterWorker } from "./ensdb-writer-worker";
 
-let ensDbWriterWorker: EnsDbWriterWorker;
+let ensDbWriterWorker: EnsDbWriterWorker | undefined;
+
+function isAbortError(error: unknown): boolean {
+  return error instanceof Error && error.name === "AbortError";
+}
 
 /**
- * Starts the EnsDbWriterWorker in a new asynchronous context.
+ * Start (or restart) the EnsDbWriterWorker.
  *
- * The worker will run indefinitely until it is stopped via {@link EnsDbWriterWorker.stop},
- * for example in response to a process termination signal or an internal error, at
- * which point it will attempt to gracefully shut down.
+ * Called from `apps/ensindexer/ponder/src/api/index.ts` on every Ponder
+ * API exec. Ponder re-executes the API entry file on hot reload, but this
+ * module is cached by vite-node, so module-level state survives across
+ * reloads. This function therefore must:
  *
- * @throws Error if the worker is already running when this function is called.
+ *   1. Be idempotent — treat a re-call as "the previous instance is dead,
+ *      replace it" rather than throwing.
+ *   2. Re-bind reload-scoped resources (e.g. `apiShutdown`) fresh from
+ *      `localPonderContext` on every call. Never hoist them to module
+ *      scope. See `local-ponder-context.ts` for the staleness contract.
  */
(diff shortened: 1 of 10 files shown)


apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
```diff
@@ -29,12 +30,21 @@ export function startEnsDbWriterWorker() {
     localPonderClient,
   );
 
+  logger.info({ msg: "StartingEnsDbWriterWorker..." });
+
+  localPonderContext.abortSignal.addEventListener("abort", () => {
+    logger.info({ msg: "Shutting down EnsDbWriterWorker..." });
+    ensDbWriterWorker?.stop();
```
shrugs: nit: extract to a `gracefulShutdown` helper?

shrugs:
@greptile

shrugs:
@greptileai please re-review — addressed the AND/OR catch path tightening and the unhandled-rejection issue you flagged

shrugs:
@greptileai please re-review — addressed coderabbit's stop()-during-run race (added internal stopRequested guard) and the fail-fast concern (process.exit(1) on real worker errors, since the catch is on a fire-and-forget promise)

shrugs:
@greptileai please re-review — addressed copilot's catch-path classification: defensive stale-instance cleanup now correctly handles AbortError from the new stopRequested guard via an `ensDbWriterWorker !== worker` supersession check

shrugs:
@greptileai please re-review — refactored localPonderContext from a Proxy to explicit `getApiShutdown()` / `getShutdown()` getter functions, dropping ~19 lines and surfacing the staleness contract at every call site

lightwalker-eth (commented):
Appreciate the value of the work being advanced here. Will defer to @shrugs and @tk-o to decide when this PR would be ready for merge.

PR #1930, opened 2026-04-14
fix(security): override follow-redirects to ^1.16.0

shrugs:
## Summary
- adds a `pnpm.overrides` entry pinning `follow-redirects` to `^1.16.0` to resolve GHSA-r4q5-vmmm-2653 (moderate)
- transitive via `docs/ensnode.io > astro-icon > @iconify/tools > axios > follow-redirects`, so an override is used instead of bumping a direct dep
- `pnpm audit --audit-level=moderate` is clean after the change (remaining `high` is the existing CVE-2026-39356 already ignored in `auditConfig`)

## Test plan
- [ ] CI passes
- [ ] `pnpm audit --audit-level=moderate` reports no vulnerabilities


lightwalker-eth (approved):
👍

PR #1932, opened 2026-04-15
fix: minor fixes from copilot after big pr

shrugs:
via these comments https://github.com/namehash/ensnode/pull/1908#pullrequestreview-4108665410


shrugs:
merged as hotfix

PR #1936, opened 2026-04-15
ci: bump pnpm to 10.33.0 and swap audit for osv-scanner

shrugs:
## Summary

- bump pnpm to 10.33.0 across `packageManager` fields and `.tool-versions`
- remove the broken `pnpm audit` CI job
- add `.github/workflows/osv-scanner.yml` using `google/osv-scanner-action@v2.3.5` (delta scan on PRs, full scan on push to main)
- add `pnpm audit:osv` script (`osv-scanner scan source --lockfile pnpm-lock.yaml`) for local parity
- port the drizzle-orm CVE-2026-39356 ignore from `pnpm.auditConfig` to `osv-scanner.toml`

## Why

`pnpm audit` is currently broken on every published pnpm version (10.x and 11.0.0-rc.0) because npm retired the legacy audit endpoints — see [pnpm/pnpm#11265](https://github.com/pnpm/pnpm/issues/11265). The maintainer has committed to fixing it in v11 but not v10, and v11 is still an RC. CI is flaky for anyone hitting the brownout window.

osv-scanner reads `pnpm-lock.yaml` directly and queries the OSV database, so it has no dependency on the retired npm endpoint and no dependency on pnpm shipping a fix.

## Testing

local `pnpm audit:osv` against the worktree:

```
Scanned .../pnpm-lock.yaml file and found 1888 packages
Loaded filter from: .../osv-scanner.toml
GHSA-gpj5-g38j-94v9 and 1 alias have been filtered out because: drizzle-orm SQL injection via dynamic identifiers (CVE-2026-39356). Not exploitable in ENSNode: ...
Filtered 1 vulnerability from output
No issues found
```

the drizzle ignore is honored, no unfiltered findings, runs in ~30ms.

CI side runs via the reusable workflow end-to-end (not executable locally); the workflow auto-loads `osv-scanner.toml` at the scan root.

- **script name**: `pnpm audit` is a pnpm builtin and shadows user scripts with the same name, so the script is `audit:osv` — invoke with `pnpm audit:osv`
- **lockfile flag**: the script uses `--lockfile pnpm-lock.yaml` rather than `scan source .` because the repo's root `.gitignore` contains `.claude`, which makes osv-scanner skip the entire directory when run inside a `.claude/worktrees/...` worktree. scanning the lockfile directly sidesteps the walk entirely and is faster anyway (~30ms vs ~11s).
- **CI workflow layout**: PR variant (`osv-scanner-reusable-pr.yml`) only reports *new* vulns vs base, so it won't block merges over pre-existing lowsev findings. push variant (`osv-scanner-reusable.yml`) does a full scan on main and uploads SARIF to the Security tab — needs `security-events: write`, granted at the workflow level.
- **ignore config format**: osv-scanner uses OSV-native IDs (GHSA), and auto-applies to all aliases — `GHSA-gpj5-g38j-94v9` covers `CVE-2026-39356`.
- the `pnpm.overrides` for follow-redirects etc remain in place unchanged.

## Checklist


lightwalker-eth (approved):
@shrugs Nicely done 👍

PR #1938, opened 2026-04-15
Using `NormalizedAddress` in referral program logic

### LINKED ISSUE #1910: Refine use of `NormalizedAddress` in referral program logic
Image

Note how in PR 1908 a new type alias for `NormalizedAddress` and utility function `toNormalizedAddress` were defined in `enssdk`. We should apply the ideas from CodeRabbit in the screenshot above to optimize our referral program logic.

Specific action items here are:

1. Change the use of `Address` to `NormalizedAddress` in referral program logic.
2. Wherever an address value is captured from an input source, call `toNormalizedAddress` on it. Therefore it is truly guaranteed and enforced that it is a normalized address.
3. Wherever we are comparing two `NormalizedAddress` values for equality, we don't need to do the whole `toLowercase()` or `normalizeAddress` kind of thing anymore. If we're using functions such as `isAddress` or `isAddressEqual` these aren't need anymore.

Goader:
# Using `NormalizedAddress` in referral program logic

closes: #1910 

## Summary

- As suspected, most of the work was done by @shrugs in #1908 
- A few comments were updated, and the `validateAddress` was renamed to `validateNormalizedAddress`

---

## Why

- see: #1910 

---

## Testing

---

## Notes for Reviewer

- Not sure about the example in the README.md (changing import from `viem` to `enssdk`)

---


packages/ens-referrals/src/address.ts
```diff
@@ -1,7 +1,7 @@
 import type { NormalizedAddress } from "enssdk";
 import { isNormalizedAddress } from "enssdk";
 
-export const validateAddress = (address: NormalizedAddress): void => {
+export const validateNormalizedAddress = (address: NormalizedAddress): void => {
```
tk-o, line 4: Feel free to have a look at `Unvalidated<T>` type from ENSNode SDK. It's a helper type that allows to distinguish unvalidated value from the validated one (even if they are technically the same runtime type).

```suggestion
export const validateNormalizedAddress = (address: Unvalidated<NormalizedAddress>): void => {
```

tk-o (approved):
LGTM

packages/ens-referrals/src/address.ts
Goader: Thank you for the tip! In this case validation functions work more like runtime assertions and guards rather than a way to handle unvalidated objects. From what I see in the `Unvalidated<T>` documentation, assertions and guards are not exactly the purpose of it, no?

PR #1957, opened 2026-04-20
Fix type inference

### LINKED ISSUE #1954: Fix type mismatch in Registrar Actions API route 500 response schema
When adding the `500` error response schema to `getRegistrarActionsRoute` and `getRegistrarActionsByParentNodeRoute` in `registrar-actions-api.routes.ts`, Hono's route type checker rejects the schema with:

```
Type '{ currency: "ETH"; amount: string; }' is not assignable to type 'null'.
```

My idea is that the root cause is a mismatch between the pricing field in the serialized response `(amount: string)` and the type Hono infers from the zod schema (which collapses to `amount: never`). The `500` response content block is currently commented out as a workaround: https://github.com/namehash/ensnode/blob/a8b55d3ceaedae178237658a0026d2e07c27e640/apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts#L123-L130

Discovered in #1872.

tk-o:
For OpenAPI Spec, we use the Zod schemas that describe "serialized" data model that is in fact returned in HTTP responses

# Lite PR

## Summary

- Applied the fix to use `makeSerializedRegistrarActionsResponseOkSchema` to describe the "serialized" data model returned from the Registrar Actions API handlers.

---

## Why

The type error exists in registrar-actions-api.ts (the handler), not in the routes file. The handler returns SerializedRegistrarActionsResponseOk (with string amounts), but the route schema expects domain types (bigint amounts).

When the 500 response was commented out, the route definition was incomplete, which likely caused TypeScript to skip full type validation between the handler return type and the route's expected response type. The `createRoute` function from `@hono/zod-openapi` uses the response schemas to infer what handlers should return - an incomplete route definition may not trigger this inference properly.

The real issue is the mismatch between:
- Registrar Actions API Handler returns: `SerializedRegistrarActionsResponseOk` (string amounts)
- Registrar Actions API Route Schema expects: Domain types from `makeRegistrarActionsResponseOkSchema` (bigint amounts)

---

## Testing

- Ran testing suite, typechecks.

---

- Resolves #1954 

---

### DIFF
--- a/apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts
+++ b/apps/ensapi/src/handlers/api/explore/registrar-actions-api.routes.ts
@@ -11,7 +11,8 @@ import {
   makeLowercaseAddressSchema,
   makeNodeSchema,
   makePositiveIntegerSchema,
-  makeRegistrarActionsResponseOkSchema,
+  makeRegistrarActionsResponseErrorSchema,
+  makeSerializedRegistrarActionsResponseOkSchema,
   makeUnixTimestampSchema,
   registrarActionsResponseOkExample,
 } from "@ensnode/ensnode-sdk/internal";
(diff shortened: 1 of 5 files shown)


sevenzing (commented):
ltgm

PR #1975, opened 2026-04-21
[WIP] Split ENSApi config data model

### LINKED ISSUE #1806: Split config objects in ENSApi
ENSApi should have more than one config object. There's an idea where ENSApi uses two config objects:
1. The env-based config built from env vars, available as a static import. 
2. The ENSDb-based config (including ENSIndexer Public Config), derived from async ENSDb read, that is shared via middleware.

It's best to start with a design for the requested changes, as there might be important details that have not yet been discovered.

More details in thse threads:
- https://namehash.slack.com/archives/C086Z6FNBHN/p1773092154129949
- https://namehash.slack.com/archives/C086Z6FNBHN/p1774338589997209

tk-o:
# Lite PR

## Summary

- ENSApi Config is based purely on env vars
- All dynamic config is based on the `EnsNodeStackInfo` object that is loaded from cache for every request. If the `EnsNodeStackInfo` object is not available, no relevant HTTP routes (the ones that rely on the `EnsNodeStackInfo` object) return 503.

---

## Why

- We wanted to simplify data model for handling the ENSApi Config.

---

## Testing

- How this was tested.

---

- Review commit-by-commit
- Resolves #1806 

---

- [ ] Relevant changesets are included (or are not required)

### DIFF
--- a/apps/ensapi/src/lib/protocol-acceleration/find-resolver.ts
+++ b/apps/ensapi/src/lib/protocol-acceleration/find-resolver.ts
@@ -1,5 +1,3 @@
-import config from "@/config";
-
 import { bytesToPacket } from "@ensdomains/ensjs/utils";
 import { SpanStatusCode, trace } from "@opentelemetry/api";
 import {
@@ -14,12 +12,11 @@ import {
 import { isAddressEqual, type PublicClient, toHex, zeroAddress } from "viem";
 import { packetToBytes } from "viem/ens";
 
-import { DatasourceNames, getDatasource } from "@ensnode/datasources";
+import { DatasourceNames, type ENSNamespaceId, getDatasource } from "@ensnode/datasources";
 import { accountIdEqual, getDatasourceContract, isENSv1Registry } from "@ensnode/ensnode-sdk";
 
 import { ensDb } from "@/lib/ensdb/singleton";
 import { withActiveSpanAsync, withSpanAsync } from "@/lib/instrumentation/auto-span";
-import { lazyProxy } from "@/lib/lazy";
 
 type FindResolverResult =
   | {
@@ -37,12 +34,6 @@ const NULL_RESULT: FindResolverResult = {
 
 const tracer = trace.getTracer("find-resolver");
 
-// lazyProxy defers construction until first use so that this module can be
-// imported without env vars being present (e.g. during OpenAPI generation).
-const ensv1RegistryOld = lazyProxy(() =>
-  getDatasourceContract(config.namespace, DatasourceNames.ENSRoot, "ENSv1RegistryOld"),
-);
-
 /**
  * Identifies `name`'s active resolver in `registry`.
  *
@@ -53,12 +44,14 @@ const ensv1RegistryOld = lazyProxy(() =>
  * - TODO: any ENSv2 Registry
  */
 export async function findResolver({
+  namespace,
   registry,
   name,
   accelerate,
   canAccelerate,
   publicClient,
 }: {
+  namespace: ENSNamespaceId;
   registry: AccountId;
   name: InterpretedName;
   accelerate: boolean;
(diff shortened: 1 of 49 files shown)


tk-o:
@greptile review

packages/ensnode-sdk/src/ensnode/client.test.ts
```diff
@@ -69,35 +69,6 @@ const EXAMPLE_ENSAPI_CONFIG_RESPONSE = {
...
-    versionInfo: {
-      ponder: "0.11.43",
-      ensDb: "0.32.0",
-      ensIndexer: "0.32.0",
-      ensNormalize: "1.11.1",
-    },
-  },
 } satisfies SerializedEnsApiPublicConfig;
 
 const EXAMPLE_ENSDB_PUBLIC_RESPONSE = {
```
lightwalker-eth, line 74:
```suggestion
const EXAMPLE_ENSDB_PUBLIC_CONFIG = {
```

packages/ensnode-sdk/src/ensnode/client.test.ts
```diff
@@ -69,35 +69,6 @@ const EXAMPLE_ENSAPI_CONFIG_RESPONSE = {
     canFallback: false,
```
lightwalker-eth, line 69: Several lines above we should rename `EXAMPLE_ENSAPI_CONFIG_RESPONSE` to `EXAMPLE_ENSAPI_PUBLIC_CONFIG`

packages/ensnode-sdk/src/ensnode/client.test.ts
```diff
@@ -69,35 +69,6 @@ const EXAMPLE_ENSAPI_CONFIG_RESPONSE = {
...
-      PluginName.Registrars,
-    ],
-    versionInfo: {
-      ponder: "0.11.43",
-      ensDb: "0.32.0",
-      ensIndexer: "0.32.0",
-      ensNormalize: "1.11.1",
-    },
-  },
 } satisfies SerializedEnsApiPublicConfig;
```
lightwalker-eth, line 73: I'm interested to also see the following updates. Could we include them in this PR?

1. Introduce a new type `EnsIndexerStackInfo` that has two fields: `ensIndexer` and `ensRainbow`, each the relevant public config type.
2. Use `EnsIndexerStackInfo` to split `ensRainbowPublicConfig` from `EnsIndexerPublicConfig`.
3. Remove the count from `ensRainbowPublicConfig` so that ENSRainbow can immediately return its config without waiting.
4. Guarantee that both `EnsIndexerStackInfo` and `EnsNodeStackInfo` have all their fields non-optional. In other words, there's never a case where `ensRainbow` is undefined.
5. Update the data model for `ensRainbowPublicConfig` so that it aligns with the `versionInfo` standards from the other config types. It shouldn't have `version` as a root-level field. Instead it should have a `versionInfo` object with a field inside of it for `ensRainbow` that might be a value such as "0.31.0".

apps/ensapi/src/lib/protocol-acceleration/find-resolver.ts
```diff
@@ -37,12 +34,6 @@ const NULL_RESULT: FindResolverResult = {
 
 const tracer = trace.getTracer("find-resolver");
 
-// lazyProxy defers construction until first use so that this module can be
-// imported without env vars being present (e.g. during OpenAPI generation).
-const ensv1RegistryOld = lazyProxy(() =>
```
lightwalker-eth, line 42: Can you send me a message in Slack when you have an opportunity to reply to this? I'm curious to understand if the work in this PR enables us to completely remove the use of this `lazyProxy` stuff? As I remember it was a source of some complexity for us and we had the goal of fully removing it.

apps/ensapi/src/lib/public-client.ts
```diff
@@ -3,30 +3,36 @@ import config from "@/config";
...
 const _cache = new Map<ChainId, PublicClient>();
 
 /**
  * Gets a viem#PublicClient for the specified `chainId` using the ENSApiConfig's RPCConfig. Caches
  * the instance itself to minimize unnecessary allocations.
  */
-export function getPublicClient(chainId: ChainId): PublicClient {
-  // Invariant: ENSApi must have an rpcConfig for the requested `chainId`
-  const rpcConfig = config.rpcConfigs.get(chainId);
+export function buildPublicClientForRootChain(namespace: ENSNamespaceId): PublicClient {
```
lightwalker-eth, line 15: @tk-o Hey, please ask Matt to review this when you're ready. Please use special care to ensure this is 100% correct.

apps/ensapi/src/lib/resolution/forward-resolution.ts
```diff
@@ -88,18 +87,20 @@ const tracer = trace.getTracer("forward-resolution");
  * }
  */
 export async function resolveForward<SELECTION extends ResolverRecordsSelection>(
+  namespace: ENSNamespaceId,
+  plugins: string[],
```
lightwalker-eth, line 91: Do you have any other suggestions for how we might pass these values around?

I dislike how we are decomposing our config objects / stack objects into independent fields where invariants are being erased because now they are just random fields that are theoretically independent of each other.

Why can't we still have some singleton object such as `stack` (rather than `config`) that would hold the stack. You could implement some getter on it that would throw an error if for some reason the data for the stack hadn't been initialized already.

apps/ensapi/src/middleware/stack-info.middleware.ts
```diff
@@ -52,3 +54,20 @@ export const stackInfoMiddleware = producing(
...
+/**
+ * Ensures that the ENSNode Stack Info is available in the Hono context.
+ * If the Stack Info is not available, an error is thrown.
+ *
+ * This is a type guard that narrows the type of `c.var.stackInfo` to {@link EnsNodeStackInfo} in all downstream middlewares and handlers.
+ * We leverage the fact that {@link stackInfoMiddleware} is the first middleware for all ENSApi routes.
+ *
+ * @param c The Hono context
+ */
+export function ensureEnsNodeStackInfoAvailable<T extends Context>(
```
lightwalker-eth, line 67: I dislike this approach. If the stack info isn't loaded yet it shouldn't result in a HTTP 500 error. It should result in a HTTP 503 error.

lightwalker-eth (commented):
@tk-o Overall I like where this is going. Shared some preliminary feedback 👍

PR #1986, opened 2026-04-23
Prevent overlapping referral program editions

### LINKED ISSUE #1959: Prevent overlapping referral program editions
# Goal

Simplify and constrain the edge cases in our data model by making it impossible for a single referral to be associated with multiple referral program editions.

# Spec

1. Each referral should be associated with 0 or 1 referral program editions and never 2+.
2. To enforce this, suggest that we add more invariant checks to how a referral program config set is defined. If we enforce that for a specific registryId no two referral program editions can overlap in time then we achieve this constraint. When implementing this invariant, please take special note of how the start / end time for an edition is inclusive.

Goader:
# Prevent overlapping referral program editions

closes: #1959 

## Summary

- Added base referral program edition config type to be a common root between config and summary types
- Added verification for the overlap

---

## Why

- See: #1959 

---

## Testing

- Added new tests for the added functionality

---


packages/ens-referrals/src/edition.ts
```diff
@@ -78,6 +89,68 @@ export function validateReferralProgramEditionConfigSet(
...
+ *
+ * @param editions - Array of editions to check
+ * @returns A `[a, b]` tuple of the first offending pair, or `null` if none
+ */
+export function findOverlappingEditionPair<T extends BaseReferralProgramEditionConfig>(
+  editions: readonly T[],
+): readonly [T, T] | null {
+  const byRegistry = new Map<string, T[]>();
+  for (const edition of editions) {
+    const key = `${edition.rules.subregistryId.chainId}:${edition.rules.subregistryId.address}`;
```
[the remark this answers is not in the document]
Goader: `edition.rules.subregistryId.address` has `NormalizedAddress` type, so it is guaranteed to be normalized

packages/ens-referrals/src/edition.ts
```diff
@@ -102,6 +175,8 @@ export function buildReferralProgramEditionConfigSet(
     throw new Error(`Duplicate edition config slugs detected: ${duplicates.join(", ")}`);
   }
 
+  validateNonOverlappingEditionTimes(configs);
```
lightwalker-eth: Is there a special reason not to merge this idea into the existing `validateReferralProgramEditionConfigSet` function?

lightwalker-eth (approved):
@Goader Nice work! 1 small question only. Please take the lead to merge when ready 👍

packages/ens-referrals/src/edition.ts
Goader: No reason, just missed it. Thank you!

PR #1994, opened 2026-04-24
tests: add seed script to improve resolution tests

sevenzing:
## Summary

Add seeding mechanism for devnet to improve our testing

---

## Why

We cannot test ensapi right now

---


packages/datasources/src/lib/chains.ts
```diff
@@ -4,7 +4,7 @@ import { type Chain, localhost } from "viem/chains";
  * The Default Chain Id for Devnet
  * @see https://github.com/ensdomains/contracts-v2/blob/762de44d60b2588b2e92a6d29df941c4de821ae6/contracts/script/setup.ts#L40
  */
-const DEVNET_DEFAULT_CHAIN_ID = 0xeeeeed;
+const DEVNET_DEFAULT_CHAIN_ID = 1;
```
sevenzing: seed script doesnt work without it. view throws error:

```
> p test:integration:ci           

> ensnode-monorepo@0.0.1 test:integration:ci /Users/levlymarenko/namehash/ensnode
> pnpm -F @ensnode/integration-test-env start


> @ensnode/integration-test-env@1.10.1 start /Users/levlymarenko/namehash/ensnode/packages/integration-test-env
> CI=1 tsx src/orchestrator.ts

[ci] Starting integration test environment...
[ci] Software versions:
[ci]   Node.js:  v24.14.0
[ci]   pnpm:     10.33.0
[ci]   Docker:   Docker version 28.5.1, build e180ab8
[ci] Starting Postgres and devnet...
[ci] Postgres is ready (port 5432)
[ci] Devnet is ready (port 8545)
[ci] Seeding devnet...
[ci] ERROR: ContractFunctionExecutionError: Transaction creation failed.

URL: http://localhost:8545
Request body: {"method":"eth_sendRawTransaction","params":["0x02f8d383eeeeed8181843b9aca00843b9aca0882b6db9459b670e9fa9d0a427751af201d676719a970857b80b864c47f002700000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000008746573742e657468000000000000000000000000000000000000000000000000c001a0406e4a76ab2859144894f8018c81729397c93bb36575f7d1980e9fd187d5692fa0214a9bc10847d4e049f7a6479fcb466379f09ceafeb988c22a3f0a29fbcd4d53"]}
 
Request Arguments:
  from:  0x70997970C51812dc3A010C7d01b50e0d17dc79C8
  to:    0x59b670e9fA9D0A427751Af201D676719a970857b
  data:  0xc47f002700000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000008746573742e657468000000000000000000000000000000000000000000000000
 
Contract Call:
  address:   0x59b670e9fA9D0A427751Af201D676719a970857b
  function:  setName(string name)
  args:             (test.eth)
  sender:    0x70997970C51812dc3A010C7d01b50e0d17dc79C8

Docs: https://viem.sh/docs/contract/writeContract
Details: invalid chain id for signer
```

sevenzing: I can leave this line as it was and pass chainId=1 to `createWalletClient` in `seed/index.ts`, but dont like that idea

packages/ensnode-sdk/src/shared/devnet/addresses.ts
```diff
@@ -0,0 +1,97 @@
+/**
```
sevenzing: decided to keep all addresses from `pnpm devnet` logs

packages/integration-test-env/src/seed/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  type Chain,
+  createWalletClient,
+  http,
+  type Transport,
+  type WalletClient,
+} from "viem";
+import { mnemonicToAccount } from "viem/accounts";
+
+import { ensTestEnvChain } from "@ensnode/datasources";
+import { DEVNET_MNEMONIC } from "@ensnode/ensnode-sdk/internal";
```
shrugs: imo move this to the integration test env, no need for it to be shared, right?

packages/integration-test-env/src/seed/abi.ts
```diff
@@ -0,0 +1,23 @@
+import { parseAbi } from "viem";
+
+export const ethReverseRegistrarAbi = parseAbi([
+  // https://github.com/ensdomains/contracts-v2/blob/42f2016e7ba87eb3854afe51ad55186a16b32a74/contracts/src/reverse-registrar/L2ReverseRegistrar.sol#L117-L119
+  "function setName(string name) returns (bytes32)",
+]);
+
+// https://github.com/ensdomains/contracts-v2/blob/42f2016e7ba87eb3854afe51ad55186a16b32a74/contracts/test/utils/resolutions.ts#L28
+export const publicResolverAbi = parseAbi([
```
shrugs: all of this should be available through `@ensnode/datasources` — `ResolverABI`, `UniversalResolverABI` — maybe not the reverse registrar abi, but perhaps just export it if not — let's keep the abi defs as centralized as possible.

packages/ensnode-sdk/src/shared/devnet/addresses.ts
shrugs: hmm. 

1) move this file/export into `@ensnode/datasources` and export as `DEVNET_ADDRESSES`, perhaps gated behind a new submodule export `@ensnode/datasources/devnet`
2) re-use these definitions within the ens-test-env datasource so it reads like

```ts
SomeContract: {
  address: DEVNET_ADDRESSES.SomeContract
}
```

3) ensure all of these addresses are `NormalizedAddress` by manually lowercasing them (select all of them then `> Transform to Lowercase`)

4) update the comment to say that these addresses are correct for a specific devnet commit (and point to the docker-compose, don't say "commit abcdefh" because that will get out of date).

packages/ensnode-sdk/src/shared/devnet/accounts.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { toNormalizedAddress } from "enssdk";
+
+/**
+ * Standard Hardhat/Anvil mnemonic used by the ens-test-env devnet.
+ */
+export const DEVNET_MNEMONIC = "test test test test test test test test test test test junk";
+
+export const DEVNET_ACCOUNTS = {
+  /** Account index 0 — has REGISTRAR role on ETHRegistry */
```
shrugs: we have access to viem here, so:

1) see if there's an export of the default test mnemonic from viem so we don't re-define
2) derive the devnet user addresses rather than hardcoding

packages/ensnode-sdk/src/shared/devnet/data.ts
```diff
@@ -0,0 +1,11 @@
+import type { Hex } from "viem";
+
+export const DEVNET_BYTES: Record<string, Hex> = {
```
shrugs: write this as 

`const DEVNET_BYTES = {} as const satisfies Record<string, Hex>` to avoid all of those `as` casts

shrugs: and we def need a comment and a `@see` link to understand where these values come from and how they're used

packages/ensnode-sdk/src/shared/devnet/addresses.ts
shrugs: let's move all of these files into datasources actually, which already has a viem dep, and definitely do the submodule export to keep it clean

shrugs: i want to

```ts
import {accounts, contracts, datas} from `@ensnode/datasources/devnet`
```

by the way, the devnet vs ens-test-env naming is difficult; everything was ens-test-env and that's mostly what we use in ensnode, but the ens team switched to 'devnet' to identify the actual chain and docker image, but the "ens-test-env" project lives on as a harness to run the devnet container... so use your best judgement here, and `devnet` is likely the correct _internal_ naming to use

packages/datasources/src/lib/chains.ts
shrugs: this is correct, the devnet's chain id is now `1`, if the commit has been updated.

just inline the config now so it's

```ts 
/**
 * @see blah
 */
export const ensTestEnvChain = {
  ...localhost,
  chainId: 1,
  // ... rest
}
```

packages/integration-test-env/src/seed/index.ts
```diff
@@ -0,0 +1,46 @@
...
+export type DevnetWalletClient = WalletClient<Transport, Chain, Account>;
+
+export type DevnetWalletClients = {
+  deployer: DevnetWalletClient; // index 0 — has REGISTRAR role on ETHRegistry
+  owner: DevnetWalletClient; // index 1 — DEVNET_OWNER, owns test.eth
+  user: DevnetWalletClient; // index 2 — DEVNET_USER
+  user2: DevnetWalletClient; // index 3 — DEVNET_USER2
+};
+
+function createDevnetWalletClients(rpcUrl: string): DevnetWalletClients {
```
shrugs, line 36: just use inference if you can:

```ts
function createDevnetWalletClients(rpcUrl: string) {
  // ....
}
```

and if you need the type to be reusable, just do

```ts
export type DevnetWalletClients = ReturnType<typeof createDevnetWalletClients>
```

packages/integration-test-env/src/seed/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  const transport = http(rpcUrl);
+  const mkClient = (addressIndex: number) =>
```
shrugs: nit: `clientForAddressIndex`

packages/integration-test-env/src/seed/resolver-records.ts
```diff
@@ -0,0 +1,142 @@
...
+import { DEVNET_ACCOUNTS, DEVNET_BYTES, DEVNET_CONTRACTS } from "@ensnode/ensnode-sdk/internal";
+
+import { publicResolverAbi, universalResolverV2Abi } from "./abi";
+import type { DevnetWalletClient, DevnetWalletClients } from "./index";
+
+const RESOLVER = DEVNET_CONTRACTS.permissionedResolver;
+
+export async function seedResolverRecords(clients: DevnetWalletClients): Promise<void> {
+  const node = namehash("test.eth");
+  await assertTestEthResolver(clients.owner.transport.url);
```
shrugs: yeah what's the purpose of doing this assert?

packages/integration-test-env/src/seed/resolver-records.ts
```diff
@@ -0,0 +1,142 @@
...
+    abi: universalResolverV2Abi,
+    functionName: "findResolver",
+    args: [toHex(packetToBytes("test.eth"))],
+  });
+  if (activeResolver.toLowerCase() !== RESOLVER.toLowerCase()) {
+    throw new Error(`test.eth resolver mismatch: active=${activeResolver}, expected=${RESOLVER}`);
+  }
+}
+
+async function setTextRecord(
```
shrugs, line 64: allow these helpers to act on any `resolver` address

packages/integration-test-env/src/seed/resolver-records.ts
```diff
@@ -0,0 +1,142 @@
+import { type Address, createPublicClient, type Hex, http, namehash, toHex } from "viem";
+import { packetToBytes } from "viem/ens";
+
+import { ensTestEnvChain } from "@ensnode/datasources";
+import { DEVNET_ACCOUNTS, DEVNET_BYTES, DEVNET_CONTRACTS } from "@ensnode/ensnode-sdk/internal";
+
+import { publicResolverAbi, universalResolverV2Abi } from "./abi";
+import type { DevnetWalletClient, DevnetWalletClients } from "./index";
+
+const RESOLVER = DEVNET_CONTRACTS.permissionedResolver;
```
shrugs: the permissionedResolver address exported by the devnet may be a factory not a specific impl; each address gets their own PermRes in ensv2, so it'd be best to make sure we're acting on an instance of a PermRes that's tied to a specific address

packages/integration-test-env/src/orchestrator.ts
```diff
@@ -248,7 +250,15 @@ async function main() {
   const postgresPort = postgresContainer.getMappedPort(5432);
   const ENSDB_URL = `postgresql://postgres:password@localhost:${postgresPort}/postgres`;
   log(`Postgres is ready (port ${postgresPort})`);
-  log("Devnet is ready");
+  const devnetContainer = composeEnvironment.getContainer("devnet");
+  const devnetPort = devnetContainer.getMappedPort(8545);
+  const devnetRpcUrl = `http://localhost:${devnetPort}`;
+  log(`Devnet is ready (port ${devnetPort})`);
+
+  // Phase 1.5: Seed devnet with test data (before indexing starts)
```
shrugs: nit: update all the step numbers to whole numbers (sorry) so this becomes 2

shrugs (commented):
LOVE the direction, way better than forking; i'd forgotten that we just know the test mnemonic and can hook in like that. definitely the correct direction, thanks for prototyping this

shrugs:
i didn't see the devnet commit version be updated in this pr, make sure that's done in the docker-compose so that the orchestrator is definitely using the correct version and the chain id update is locked in. all devnet-specific stuff must be pinned to a specific commit or we're going to have a really annoying time when things change behind the scenes (as they do a LOT)

packages/integration-test-env/src/seed/resolver-records.ts
sevenzing: it was hard for me to find resolver of this name. I was only able to do it using that function, so like actually call `universalResolverV2.findResolver`. Decided to keep it just in case. 

Will add a comment here describing the problem

packages/integration-test-env/src/orchestrator.ts
sevenzing: ahahha sure

packages/integration-test-env/src/seed/index.ts
sevenzing: dont like it, `addressIndex` is used by viem, I think it's fine

packages/integration-test-env/src/seed/index.ts
sevenzing: failed to use type inference there. typescript said "failed to inference type due to maximum recursion bla bla bla, write explicit"

maybe me skill issue, you can make a commit with fix you want

packages/ensnode-sdk/src/shared/devnet/data.ts
sevenzing: those constants made by me, it's not from ens-test-env, but it will be part of `seedDevnet` process. 
I will add comment about that

packages/ensnode-sdk/src/shared/devnet/accounts.ts
sevenzing: 1. https://github.com/search?q=repo%3Awevm%2Fviem+test+test+test+test+test+test+test+test+test+test+test+junk&type=code

not really, they define it but not in public way

packages/integration-test-env/src/seed/resolver-records.ts
sevenzing: this permissionedResolver is a resolver for all names in devnet 👀

apps/ensapi/src/handlers/api/resolution/resolve-records.integration.test.ts
```diff
@@ -124,18 +124,109 @@ describe("GET /api/resolve/records/:name", () => {
...
         accelerationRequested: false,
         accelerationAttempted: false,
       },
     },
+    // -- Text records (seeded in devnet) --
+    {
+      description: "resolves avatar text record for test.eth",
+      name: "test.eth",
+      query: "texts=avatar",
+      expectedStatus: 200,
```
shrugs: apologies for this nit, but could we shift the structure of this data to `expected: { status: number, body: unknown }` ? sorry to ask you to spend tokens on this, but it reads a lot better to me. feel free to ignore

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,134 @@
+import type { NormalizedAddress } from "enssdk";
+import { toNormalizedAddress } from "enssdk";
+import type { Hex } from "viem";
+import { mnemonicToAccount } from "viem/accounts";
+
+/**
+ * Deterministic contract addresses for the ENS contracts-v2 devnet used by ens-test-env.
+ * Source of truth is the devnet deployment used by this repository's test harness and compose setup:
+ * @see https://github.com/ensdomains/contracts-v2
+ * @see https://github.com/ensdomains/ens-test-env
```
shrugs: `@see docker/services/devnet.yml`

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,134 @@
...
+ */
+export const contracts = {
+  // -- DNS --
+  dnssecGatewayProvider: "0x5fbdb2315678afecb367f032d93f642f64180aa3",
```
shrugs: let's rename these to map 1:1 with the output of the devnet, including the CamelCaseContractName, this way an ai-agent can accept the output of the devnet and trivially format it into this datastructure and we have no place for misalignment

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,134 @@
...
+  staticBulkRenewal: "0x7a9ec1d04904907de0ed7b6839ccdd59c3716ac9",
+  dummyOracle: "0x2b0d36facd61b71cc05ab8f3d2355ec3631c0dd5",
+  exponentialPremiumPriceOracle: "0xfbc22278a96299d91d41c453234d97b4f5eb9b2d",
+
+  // -- Mock Tokens --
+  mockUsdc: "0xfd471836031dc5108809d173a067e8486b9047a3",
+  mockDai: "0xcbeaf3bde82155f56486fb5a1072cb8baaf547cc",
+} as const satisfies Record<string, NormalizedAddress>;
+
+const mnemonic = "test test test test test test test test test test test junk";
```
shrugs, line 99: add a comment that this needs to match the mnemonic in devnet and add an @see to that location in the code

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,134 @@
...
+
+function createAccount(addressIndex: number, resolver: string) {
```
shrugs: just accept `index: number, resolver: NormalizedAddress` and no need to toNormalizedAddress(resolver) later, we should trust the type (it will becomes a branded type in the near future)

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Named accounts from the ens-test-env devnet.
+ * They are NOT real Ethereum Mainnet or testnet addresses.
+ * You can use `pnpm devnet` to see actual data in devnet
+ *
+ * @see https://github.com/ensdomains/ens-test-env
+ */
+export const accounts = {
+  deployer: createAccount(0, "0x1F2Ce8886692b90F5754a7d428a2336800a5911B"),
```
shrugs: If following my normalized address advice from above, make sure to lowercase these addresses.

packages/datasources/src/devnet/namespace.ts
```diff
@@ -1,24 +1,25 @@
 import { zeroAddress } from "viem";
```
shrugs: should name this file `datasources/src/devnet.ts` to match the other namespace files BUT if we're going to rename it to devnet, we should just do that refactor more broadly and entirely remove the outdated 'ens-test-env' terminology. for this PR, though, to keep scope, keep this file and the namespace called `ens-test-env.ts`

packages/datasources/package.json
```diff
@@ -18,14 +18,21 @@
...
   "publishConfig": {
     "access": "public",
     "exports": {
-      "default": "./dist/index.js",
-      "types": "./dist/index.d.ts"
+      ".": {
+        "default": "./dist/index.js",
+        "types": "./dist/index.d.ts"
+      },
+      "./devnet": {
```
shrugs, line 32: confirming that the ens-test-env namespace definition should live in the `.` module and the devnet constants and such should be in the devnet submodule

packages/integration-test-env/src/orchestrator.ts
```diff
@@ -264,8 +266,9 @@ async function main() {
   log(`ENSDb is ready (port ${ENSDB_PORT})`);
 
   // Devnet Chain Id check
+  const devnetRpcUrl = ensTestEnvChain.rpcUrls.default.http[0];
```
shrugs: place this as a constant, perhaps just `RPC_URL` in the `// shared constants` block

shrugs (commented):
Really great direction

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Named accounts from the ens-test-env devnet.
+ * They are NOT real Ethereum Mainnet or testnet addresses.
+ * You can use `pnpm devnet` to see actual data in devnet
+ *
+ * @see https://github.com/ensdomains/ens-test-env
+ */
+export const accounts = {
+  deployer: createAccount(0, "0x1f2ce8886692b90f5754a7d428a2336800a5911b" as NormalizedAddress),
```
lightwalker-eth: Instead of `... as NormalizedAddress` could you try calling the `asNormalizedAddress` utility function?

packages/datasources/src/devnet/constants.ts
```diff
@@ -0,0 +1,139 @@
...
+  owner: createAccount(1, "0x5ea90acf6555276660760fe629d72932c91f4b8e" as NormalizedAddress),
+  user: createAccount(2, "0xb63ae54076c1c281ec9395b290add470e69140c6" as NormalizedAddress),
+  user2: createAccount(3, "0x5380066832977eb36353fd2b01fb92e751636b84" as NormalizedAddress),
+} as const;
+
+/**
+ * Fixtures for seeding the devnet with test data.
+ */
+export const addresses = {
+  one: toNormalizedAddress(`0x${"1".repeat(40)}`),
```
lightwalker-eth: This can be `asNormalizedAddress` because it should already be normalized and we are asserting that it already is.

lightwalker-eth (approved):
@sevenzing Nice work 😄 Please feel welcome to merge when ready 👍

PR #1996, opened 2026-04-24
Fix ENSNode Schema migrations execution

tk-o:
Use advisory locks mechanism provided by ENSDb to ensure that any pending ENSNode Schema migrations are executed exactly once, even if multiple ENSIndexer instances attempt running migrations conurrently.

# Lite PR

## Summary

- The `migrateEnsNodeSchema` method in `EnsDbWriter` class is now race-condition-safe.
  - The standard fix is a Postgres advisory lock — cheap, session-scoped, and automatically released on disconnect

---

## Why

I noticed that the ENSDb instance in one of hosted ENSNode environments has five records in `ensnode.__drizzle_migrations` table with the same `hash` column value. It means that each of the five ENSIndexer instances in this ENSNode environment has executed the same pending migration file on the same ENSNode Schema.
image

To prevent any issues with running ENSNode Schema migrations by multiple ENSIndexer instances, we need to apply a locking mechanism so that only one ENSIndexer instance can attempt executing the pending migrations for the ENSNode Schema at a time.

---

## Testing

- I ran multiple ENSIndexer instances locally at the same time and noticed only one record was added into the `ensnode.__drizzle_migrations` table for one pending migration file. The lock works 👍 

---

---

### DIFF
--- a/.changeset/tough-clubs-eat.md
+++ b/.changeset/tough-clubs-eat.md
@@ -1,5 +1,5 @@
 ---
-"@ensnode/ensdb-sdk": minor
+"@ensnode/ensdb-sdk": patch
 ---
 
 Made `EnsDbWriter.migrateEnsNodeSchema` race-condition safe.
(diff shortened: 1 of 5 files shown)


tk-o:
@greptile review

lightwalker-eth (approved):
@tk-o Nice improvement 👍 Looks good!

PR #1997, opened 2026-04-24
Introduce indexing metadata context data model

### LINKED ISSUE #1417: Gracefully support up to 1-hour delays for ENSRainbow to become ready
Related comments:
1. https://github.com/namehash/ensnode/pull/1406/files#r2629873476
2. https://github.com/namehash/ensnode/pull/1406/files#r2629902065
3. https://github.com/namehash/ensnode/pull/1406/files#r2628891978

### LINKED ISSUE #1817: Allow ENSIndexer to run ENSDb migrations and populate ENSNode Metadata without waiting for ENSRainbow
Currently, an ENSIndexer instance has a dependency on the related ENSRainbow instace, such that until the ENSRainbow instance is ready to serve HTTP traffic, the ENSIndexer instance:
1. Cannot start indexing onchain data.
2. Cannot determine the `EnsIndexerPublicConfig` object, which prevents ENSDb Writer Worker from writing `EnsIndexerPublicConfig` record to ENSDb, and as a result, crashes the ENSIndexer instance.

In a cold-start scenario, it might take more than 30 minutes for the ENSRainbow instance to reach the readiness and open its HTTP server.

While the 1st issue remains unadressed, we can and should address the 2nd isssue. 

We could make the `ensRainbowPublicConfig` field on `EnsIndexerPublicConfig` data model optional. By doing so, the ENSIndexer instance will be able to determine the `EnsIndexerPublicConfig` object even during the ENSRainbow instance cold-start period. Having `EnsIndexerPublicConfig` object, ENSDb Writer Worker will be able to populate ENSDb with ENSNode Meetadata (including ENSIndexer Public Config, and Indexing Status where all chains in queued status).

Key points to reflect on while designing the solution for this issue:
1. The deeper goal here is we want to make it so that the startup dependencies across our services are softened to improve the experience of operators launching services.
2. Fetching the `EnsRainbowPublicConfig` object will require multiple retries. Once the config object was fetched successfully, it should remain cached indefinitely.
3. When `EnsRainbowPublicConfig` object becomes available in the cache, the ENSDb Writer Worker should update the ENSIndexer Public Config record in ENSNode Metadata table to and update its `ensRainbowPublicConfig` column from `undefined` to the cached `EnsRainbowPublicConfig` object.
4. Changing the type of `EnsIndexerPublicConfig['ensRainbowPublicConfig']` field from `EnsIndexerPublicConfig` to `EnsIndexerPublicConfig | underfined` will create side effects on `EnsIndexerPublicConfig` data model consumers, for example, ENSApi, ENSAdmin. Please make sure to take that into consideration while designing the updates for this issue.

tk-o:
# Introduce `IndexingMetadataContext` data model

> Resolves [#1884](https://github.com/namehash/ensnode/issues/1884)

---

<details>

This PR spans two apps: **ENSIndexer** (producer) and **ENSApi** (consumer), connected through ENSDb.

ENSIndexer used to write **three separate** ENSNode Metadata records into ENSDb (`ensdb_version`, `ensindexer_public_config`, `ensindexer_indexing_status`). It now builds a **single** `IndexingMetadataContext` record — holding both `CrossChainIndexingStatusSnapshot` and `EnsIndexerStackInfo` — and writes it into ENSNode Metadata table in ENSDb under one key (`"indexing_metadata_context"`). ENSApi, in turn, reads that same single record to obtain `EnsIndexerStackInfo`, and uses it for loading data into `IndexingStatusCache` and `EnsNodeStackInfoCache`.

The bulk of the change is in ENSIndexer. The changes in ENSApi are limited to updating its caches to consume the new `IndexingMetadataContext` record.

**Review these three areas in order:**

1. **Producer: `IndexingMetadataContextBuilder.getIndexingMetadataContext()`** (`apps/ensindexer/src/lib/indexing-metadata-context-builder/indexing-metadata-context-builder.ts:82`) — the algorithm that decides what `IndexingMetadataContextInitialized` gets written into ENSDb. It fetches three data sources in parallel (stored `IndexingMetadataContext`, in-memory `OmnichainIndexingStatusSnapshot`, in-memory `EnsIndexerStackInfo`), branches on `statusCode` to distinguish fresh-start from restart, and enforces `EnsIndexerPublicConfig` compatibility on restart (skipped in dev mode). Understand this function first — every downstream consumer of `IndexingMetadataContext` depends on what it produces.

2. **Producer: `initIndexingOnchainEvents()`** (`apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts:50`) — the sequence that executes once per ENSIndexer startup, guaranteeing the `IndexingMetadataContext` record is populated before any onchain event handlers fire. The steps: ENSDb health → schema migration → ENSRainbow health → upsert `IndexingMetadataContextInitialized` → ENSDb ready → ENSRainbow ready → start recurring writer. Confirm the ordering and that each step's failure mode (process exit) is appropriate.

3. **Consumer: ENSApi reads the record** (`apps/ensapi/src/cache/stack-info.cache.ts`, `apps/ensapi/src/cache/indexing-status.cache.ts`) — the two ENSApi caches that read `IndexingMetadataContext` from ENSDb and use it to power the ENSApi configuration and API endpoints. `stack-info.cache.ts` extracts `EnsIndexerStackInfo` and composes `EnsNodeStackInfo` for the Indexing Status API; `indexing-status.cache.ts` extracts `CrossChainIndexingStatusSnapshot` for status queries. Confirm the read paths match what the ENSIndexer producer writes.

</details>

---

## Problem & Motivation

<details>

- Three independent ENSNode Metadata records — `ensdb_version`, `ensindexer_public_config`, `ensindexer_indexing_status` — were written to and read from ENSDb separately via `EnsDbReader`/`EnsDbWriter`, despite forming a logical unit: the state of an ENSIndexer instance. Maintaining three separate read/write paths added friction and opened the door to inconsistent reads where one record existed but another didn't.
- `EnsDbWriterWorker` was responsible for fetching and writing all three, each with its own retry logic. Config compatibility validation, DB version writes, and indexing status updates were interleaved in one class with no clear separation of concerns.
- ENSIndexer HTTP endpoints (`/config`, `/indexing-status`) read `EnsIndexerPublicConfig` and `CrossChainIndexingStatusSnapshot` from ENSDb — data that had been round-tripped through `EnsDbWriterWorker`. The in-memory `PublicConfigBuilder` and `IndexingStatusBuilder` are the ultimate source of truth within the ENSIndexer app; ENSDb was functioning as a stale cache of their output.
- ENSApi consumed `EnsNodeStackInfo` (a composite requiring `EnsDbPublicConfig`, `EnsIndexerPublicConfig`, and `EnsRainbowPublicConfig` from multiple ENSDb records). There was no single record representing the full ENSIndexer stack, forcing ENSApi to reassemble it.
- `initializeIndexingSetup` was unreliable as a startup hook: Ponder only invokes setup handlers if (a) Ponder Checkpoints haven't been initialized yet AND (b) at least one setup handler is registered. The `ensv2` plugin registers no setup handlers, so `initializeIndexingSetup` could be skipped entirely — yet database migrations must run every startup regardless.

</details>

---

## What Changed (Concrete)

<details>

1. **New `IndexingMetadataContext` data model** (`packages/ensnode-sdk/src/ensnode/metadata/`) — a discriminated union with `statusCode: "uninitialized" | "initialized"`. The `IndexingMetadataContextInitialized` variant holds `indexingStatus: CrossChainIndexingStatusSnapshot` and `stackInfo: EnsIndexerStackInfo`. The `IndexingMetadataContextUninitialized` variant carries no data. Includes full serialization (`SerializedIndexingMetadataContext*`), deserialization, validation (Zod schemas), and builder functions (`buildIndexingMetadataContextUninitialized()`, `buildIndexingMetadataContextInitialized()`).

2. **`EnsNodeMetadataKeys` reduced to a single key** (`packages/ensdb-sdk/src/client/ensnode-metadata.ts`) — the previous standalone keys `EnsDbVersion`, `EnsIndexerPublicConfig`, and `EnsIndexerIndexingStatus` are removed. Only `IndexingMetadataContext: "indexing_metadata_context"` remains. The `EnsNodeMetadata` union type now contains only `EnsNodeMetadataIndexingMetadataContext`. The `SerializedEnsNodeMetadata` union mirrors this.

3. **`getIndexingMetadataContext()` added to `EnsDbReader`** (`packages/ensdb-sdk/src/client/ensdb-reader.ts:176`) — reads the `IndexingMetadataContextInitialized` serialized record from ENSDb by its `key` (`"indexing_metadata_context"`). Returns `buildIndexingMetadataContextUninitialized()` when no record exists, making the "not yet populated" case explicit rather than returning `undefined`.

4. **New `EnsDbReader` health/readiness methods** (`packages/ensdb-sdk/src/client/ensdb-reader.ts:134-158`):
   - `isHealthy()` — executes `SELECT 1` to verify connectivity.
   - `isReady()` — checks `isHealthy()` AND that the stored `IndexingMetadataContext.statusCode === "initialized"`, meaning the ENSNode Schema has been fully initialized for this ENSIndexer instance.

5. **`upsertIndexingMetadataContext()` added to `EnsDbWriter`** (`packages/ensdb-sdk/src/client/ensdb-writer.ts:64-71`) — serializes an `IndexingMetadataContextInitialized` and upserts it into the `ensnode_metadata` table under the `"indexing_metadata_context"` key. Replaces three removed methods: `upsertEnsDbVersion()`, `upsertEnsIndexerPublicConfig()`, and `upsertIndexingStatusSnapshot()`.

6. **`EnsDbWriterWorker` radically simplified** (`apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts`):
   - Removed `PublicConfigBuilder` and `LocalPonderClient` dependencies. Now takes `IndexingMetadataContextBuilder` instead of `IndexingStatusBuilder`.
   - Single recurring task: call `indexingMetadataContextBuilder.getIndexingMetadataContext()` (which internally handles all data fetching, invariant checks, and `IndexingMetadataContextInitialized` construction) and upsert the result via `ensDbClient.upsertIndexingMetadataContext()`.
   - Removed methods: `getValidatedEnsIndexerPublicConfig()`, `getValidatedIndexingStatusSnapshot()`.
    - Error handling: errors in the update loop are logged and the worker continues running. Previously, errors during config fetching or compatibility validation would rethrow and terminate the worker; now those responsibilities are handled by `IndexingMetadataContextBuilder` during startup instead of recurring in the worker loop.

7. **`initIndexingOnchainEvents()` extracted into its own module, replacing inline init functions in `ponder.ts`** — the `initializeIndexingSetup()` and `initializeIndexingActivation()` functions that were previously defined inline in `apps/ensindexer/src/lib/indexing-engines/ponder.ts` are removed. The new `initIndexingOnchainEvents()` function in `apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts` consolidates all startup logic into one place. The `ponder.ts` file was also restructured:
   - `EventTypeIds.Onchain` renamed to `EventTypeIds.OnchainEvent`.
   - Setup event handlers now return immediately (no init logic), since setup handlers may be skipped if (a) Ponder Checkpoints already exist (restart) or (b) no setup handler is registered (e.g., the `ensv2` plugin has none).
   - Onchain event handler preconditions now use `await import("./init-indexing-onchain-events").then(m => m.initIndexingOnchainEvents())` — dynamic import required because Ponder's build system prohibits static imports of modules that transitively depend on `ponder:api` from outside the `ponder/src/api` directory.
   - Module-level JSDoc significantly expanded to document Ponder's startup sequence.

8. **`initIndexingOnchainEvents()` orchestrates the full startup sequence** (`apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts:50-101`):
   - `ensDbClient.isHealthy()` — verify ENSDb connectivity before any operations.
   - `migrateEnsNodeSchema()` — apply pending ENSNode Schema migrations.
   - `waitForEnsRainbowToBeHealthy()` — fast health check (few retries) to confirm ENSRainbow HTTP server responds, so `EnsRainbowApiClient.config()` can be called.
   - `upsertIndexingMetadataContextRecord()` — build and upsert the `IndexingMetadataContextInitialized` into ENSDb, making `EnsIndexerStackInfo` available for ENSApi to read.
   - `ensDbClient.isReady()` — defensive check that the `IndexingMetadataContext` is now `Initialized` in ENSDb.
   - `waitForEnsRainbowToBeReady()` — long-retry readiness check (up to 1 hour) to confirm ENSRainbow can serve label healing requests.
   - `startEnsDbWriterWorker()` — begin periodic refresh of the `IndexingMetadataContext.indexingStatus` in ENSDb.
   - Any failure logs and exits with `process.exitCode = 1`.

9. **New `IndexingMetadataContextBuilder` class** (`apps/ensindexer/src/lib/indexing-metadata-context-builder/`) — fetches three data sources in parallel:
    - Stored `IndexingMetadataContext` from ENSDb (via `EnsDbReader.getIndexingMetadataContext()`)
    - In-memory `OmnichainIndexingStatusSnapshot` (via `IndexingStatusBuilder`)
    - In-memory `EnsIndexerStackInfo` (via `StackInfoBuilder`)
    Builds a new `CrossChainIndexingStatusSnapshot` with current `snapshotTime`, then constructs an `IndexingMetadataContextInitialized`. Enforces invariants:
    - Fresh start (`statusCode === "uninitialized"`): `OmnichainIndexingStatusSnapshot.omnichainStatus` must be `"unstarted"`.
    - Restart (`statusCode === "initialized"`): validates `EnsIndexerPublicConfig` compatibility between the stored `IndexingMetadataContextInitialized.stackInfo.ensIndexer` and the current in-memory `EnsIndexerPublicConfig`, unless Ponder is in dev mode (allows overriding stored config for development).

10. **New `StackInfoBuilder` class** (`apps/ensindexer/src/lib/stack-info-builder/`) — builds `EnsIndexerStackInfo` from `EnsDbPublicConfig` (via `EnsDbReader.buildEnsDbPublicConfig()`), `EnsIndexerPublicConfig` (via `PublicConfigBuilder.getPublicConfig()`), and `EnsRainbowPublicConfig` (via `EnsRainbowApiClient.config()`). Caches the result for the ENSIndexer instance lifetime.

11. **ENSIndexer HTTP endpoints now use in-memory builders** (`apps/ensindexer/ponder/src/api/handlers/ensnode-api.ts`):
    - `/config` reads from `publicConfigBuilder.getPublicConfig()` directly (previously read `EnsIndexerPublicConfig` from ENSDb).
    - `/indexing-status` reads from `indexingStatusBuilder.getOmnichainIndexingStatusSnapshot()` and builds `CrossChainIndexingStatusSnapshot` in real time via `buildCrossChainIndexingStatusSnapshotOmnichain()` (previously read a potentially stale snapshot from ENSDb).
    - Removed `ensDbClient` import from this handler entirely.

12. **Migration and worker startup moved out of Ponder API entry point** (`apps/ensindexer/ponder/src/api/index.ts`) — the calls to `migrateEnsNodeSchema()` and `startEnsDbWriterWorker()` are removed from the API module. They now execute inside `initIndexingOnchainEvents`, which is called from `eventHandlerPreconditions()` in `ponder.ts`.

13. **New `waitForEnsRainbowToBeHealthy()`** (`apps/ensindexer/src/lib/ensrainbow/singleton.ts`) — distinct from the existing `waitForEnsRainbowToBeReady()`. The ENSRainbow health endpoint (`/health`) returns 200 as soon as the HTTP server is created, enabling `EnsRainbowPublicConfig` to be fetched from `/v1/config`. The ENSRainbow readiness endpoint (`/ready`) returns 200 only after internal initialization completes and label healing (`/v1/heal`) is available. ENSIndexer needs `EnsRainbowPublicConfig` as early as possible (so `EnsIndexerStackInfo` can be upserted into ENSDb for ENSApi to read), then needs readiness before processing any onchain events.

14. **ENSApi caches updated** (`apps/ensapi/src/cache/`):
    - `indexing-status.cache.ts` — now reads `IndexingMetadataContext` from ENSDb (via `EnsDbReader.getIndexingMetadataContext()`). Throws if `statusCode` is `"uninitialized"`, causing the SWRCache to fall back to the previously cached value.
    - `stack-info.cache.ts` — now reads `IndexingMetadataContext` from ENSDb, extracts `EnsIndexerStackInfo` from `IndexingMetadataContextInitialized.stackInfo`, and builds `EnsNodeStackInfo` by composing it with `EnsApiPublicConfig`.
    - Both caches reference `EnsNodeMetadataKeys.IndexingMetadataContext` for their ENSDb lookup key.

15. **ENSApi config schema updated** (`apps/ensapi/src/config/config.schema.ts`) — `buildConfigFromEnvironment()` now fetches `EnsIndexerPublicConfig` from `EnsNodeStackInfo.ensIndexer` (built by `stackInfoCache` from the `IndexingMetadataContext` in ENSDb), replacing the standalone `ensindexer_public_config` record lookup.

16. **New `config.schema.mock.ts`** (`apps/ensapi/src/config/config.schema.mock.ts`) — provides mock `EnsApiEnvironment`, `SerializedEnsIndexerStackInfo`, `SerializedCrossChainIndexingStatusSnapshot`, and a deserialized `IndexingMetadataContextInitialized` for use by `config.schema.test.ts`.

17. **`EnsIndexerStackInfo` introduced as base type** in `ensnode-sdk`; `EnsNodeStackInfo` refactored to extend it. `EnsIndexerStackInfo` is the ENSIndexer-scoped type stored in `IndexingMetadataContextInitialized` (ensDb + ensIndexer + ensRainbow configs). ENSApi composes `EnsNodeStackInfo = EnsIndexerStackInfo + EnsApiPublicConfig` for Indexing Status API responses. This distinction is reflected in type exports.

18. **Integration test orchestrator updated** (`packages/integration-test-env/src/orchestrator.ts`) — the `pollIndexingStatus` function now calls `EnsDbReader.getIndexingMetadataContext()` and checks `statusCode === "initialized"` instead of calling the removed `getIndexingStatusSnapshot()`.

19. **Tests** added/updated:
    - `IndexingMetadataContextBuilder` tests (7 tests) — fresh start, restart with compatibility validation (pass/fail), dev mode bypasses validation, dev mode tolerates incompatible config, parallel fetch ordering, invariant enforcement.
    - `StackInfoBuilder` tests (6 tests) — successful build, caching, error propagation from all three sources.
    - `EnsDbWriterWorker` tests (8 tests) — updated for new `IndexingMetadataContextBuilder` dependency, error recovery, interval lifecycle.
    - `EnsDbReader` and `EnsDbWriter` tests updated for new `IndexingMetadataContext` methods.
    - `config.schema.test.ts` updated to use mock data from `config.schema.mock.ts`.

</details>

---

## Design & Planning

<details>

- **Why a single `IndexingMetadataContext` record instead of three separate metadata records?** The three records (`ensdb_version`, `ensindexer_public_config`, `ensindexer_indexing_status`) formed one logical unit — the state of an ENSIndexer instance. Storing them as a single `IndexingMetadataContext` record eliminates ordering problems (e.g., writing `ensdb_version` before `ensindexer_public_config` before `ensindexer_indexing_status`) and reduces the surface area for reads that see an incomplete state. It also means ENSApi (which needs `EnsNodeStackInfo` = `EnsIndexerStackInfo` + `EnsApiPublicConfig`) reads one record rather than three.

- **Why a discriminated union with `"uninitialized"` / `"initialized"` status codes?** The "no record has been written yet" case is normal (first deployment, fresh ENSDb) and must be handled explicitly. A union type makes this visible at the type level. Previously, each individual record read returned `undefined` when nothing had been written yet — the `statusCode` discriminator replaces multiple `undefined` checks with a single comparison.

- **Why `IndexingMetadataContextBuilder` and `StackInfoBuilder` as separate classes?** `StackInfoBuilder` (building `EnsIndexerStackInfo`) is reusable beyond the `IndexingMetadataContextBuilder` — it could be used directly by ENSApi if needed. `IndexingMetadataContextBuilder` owns the invariant checks (fresh-start indexing status requirement, config compatibility on restart) that are specific to ENSIndexer startup and should not be mixed into a general-purpose stack info builder.

- **Why extract init logic into a dedicated `initIndexingOnchainEvents()` function?** Previously, Ponder startup init was split across two inline functions in `ponder.ts`: `initializeIndexingSetup()` (called before setup handlers) and `initializeIndexingActivation()` (called before onchain events). Ponder setup handlers are not guaranteed to execute — they are skipped if (a) Ponder Checkpoints exist from a prior run (restart), or (b) no setup handler is registered (e.g., the `ensv2` plugin registers none). Consolidating all init into `initIndexingOnchainEvents()`, which runs as a precondition to every onchain event handler dispatch, guarantees execution on every startup. The logic was also extracted into its own file (`init-indexing-onchain-events.ts`) to keep `ponder.ts` focused on indexing engine abstraction.

- **Why two separate ENSRainbow readiness checks?** ENSRainbow's `/health` endpoint returns 200 as soon as the HTTP server starts, enabling `EnsRainbowApiClient.config()` to be called immediately. ENSRainbow's `/ready` endpoint returns 200 only after internal initialization completes and label healing (`/v1/heal`) is available. ENSIndexer needs `EnsRainbowPublicConfig` as early as possible — it's a component of `EnsIndexerStackInfo`, which must be upserted into ENSDb before ENSApi can read it (ENSApi reads `EnsIndexerStackInfo` from the `IndexingMetadataContext` record to build its own `EnsNodeStackInfo`). After the config is upserted, ENSIndexer blocks on the long-running readiness check (which may take 30+ minutes on ENSRainbow cold start) before processing any onchain events.

- **Why switch ENSIndexer HTTP endpoints from ENSDb reads to in-memory builders?** Within the ENSIndexer app, `PublicConfigBuilder` and `IndexingStatusBuilder` are the ultimate source of truth for `EnsIndexerPublicConfig` and `IndexingStatusSnapshot` respectively. The ENSDb was functioning as a cache populated by these same builders via `EnsDbWriterWorker` — but with a lag equal to the worker's update interval. Reading from the builders directly eliminates this unnecessary indirection and provides real-time data.

- Alternative considered: Keep separate metadata records and batch-write them atomically. Rejected because it doesn't solve the read side — consumers (ENSApi, ENSIndexer HTTP handlers) still need to assemble composite information from multiple records, and a batch write doesn't help a consumer that reads between writes.

</details>

- Planning artifacts: Issue [#1884](https://github.com/namehash/ensnode/issues/1884)
- Reviewed / approved by: N/A

---

## Self-Review

<details>

</details>

- Bugs caught: Ponder build failure — static imports of modules that transitively depend on `ponder:api` are prohibited outside `ponder/src/api`. Fixed by switching to `await import()` inside `eventHandlerPreconditions()` in `ponder.ts`.

- Logic simplified:
  - `initializeIndexingSetup()` and `initializeIndexingActivation()` removed from `ponder.ts` — their logic consolidated into the new `initIndexingOnchainEvents()` function. The old split was unreliable because setup handlers may be skipped entirely (no setup handler in `ensv2`, checkpoints persist across restarts).
  - `EnsDbWriterWorker` reduced from three separate fetch-and-write tasks to one recurring task: delegate to `IndexingMetadataContextBuilder.getIndexingMetadataContext()` and upsert. The builder handles all data sourcing and validation internally.
  - Extracted `upsertIndexingMetadataContextRecord()` helper from `initIndexingOnchainEvents` to separate the upsert orchestration from the startup sequencing.
  - `EnsNodeMetadata` union and `EnsNodeMetadataKeys` — removed the three old record types and keys entirely rather than deprecating them, since they were internal-only.

- Naming / terminology improved:
  - `EnsNodeStackInfo` → `EnsIndexerStackInfo`: the type represents the stack of an ENSIndexer instance (ensDb + ensIndexer + ensRainbow). `EnsNodeStackInfo` = `EnsIndexerStackInfo` + `EnsApiPublicConfig` is the ENSApi-level composite.
  - `EventTypeIds.Onchain` → `EventTypeIds.OnchainEvent` in `ponder.ts`: more explicit about referring to onchain events specifically, not indexing activation.
  - Status code values `"Uninitialized"` → `"uninitialized"`, `"Initialized"` → `"initialized"`: lowercase for consistency with other status code conventions in the codebase.

- Dead or unnecessary code removed:
  - `initializeIndexingSetup()` and `initializeIndexingActivation()` — inline functions removed from `ponder.ts`. Logic consolidated into the new `initIndexingOnchainEvents()` function in its own module.
  - `getValidatedEnsIndexerPublicConfig()` and `getValidatedIndexingStatusSnapshot()` from `EnsDbWriterWorker` — responsibilities moved to `IndexingMetadataContextBuilder` and `initIndexingOnchainEvents`.
  - `upsertEnsDbVersion()`, `upsertEnsIndexerPublicConfig()`, `upsertIndexingStatusSnapshot()` from `EnsDbWriter` — replaced by `upsertIndexingMetadataContext()`.
  - `getEnsDbVersion()`, `getEnsIndexerPublicConfig()`, `getIndexingStatusSnapshot()` from `EnsDbReader` — replaced by `getIndexingMetadataContext()`.
  - `EnsNodeMetadataEnsDbVersion`, `EnsNodeMetadataEnsIndexerPublicConfig`, `EnsNodeMetadataEnsIndexerIndexingStatus` — removed from `ensnode-metadata.ts` union and serialized types.
  - `EnsNodeMetadataKeys.EnsDbVersion`, `.EnsIndexerPublicConfig`, `.EnsIndexerIndexingStatus` — removed.
  - `migrateEnsNodeSchema()` call and `startEnsDbWriterWorker()` call from `apps/ensindexer/ponder/src/api/index.ts` — moved to `initIndexingOnchainEvents`.

---

## Cross-Codebase Alignment

<details>
<summary>Related code you checked</summary>

</details>

- Search terms used: `EnsNodeStackInfo`, `EnsNodeMetadata`, `EnsDbWriterWorker`, `EnsIndexerPublicConfig`, `stackInfo`, `indexingStatus`, `ensIndexerPublicConfig`, `ensNodeMetadata`, `migrateEnsNodeSchema`, `getIndexingStatusSnapshot`, `getEnsIndexerPublicConfig`
- Reviewed but unchanged: `packages/ensnode-sdk/src/indexing-status/` (snapshot building), `packages/ensdb-sdk/src/lib/drizzle.ts` (schema construction), `apps/ensindexer/src/lib/ensdb/migrate-ensnode-schema.ts` (migration runner), `apps/ensindexer/src/lib/indexing-status-builder/` (omnichain status builder)
- Deferred alignment (with rationale): Dead `SerializedEnsNodeMetadata*` types are no longer referenced by production code. They will be cleaned up in a follow-up to keep this PR's diff focused on the behavioral change.

---

<details>

- **ENSIndexer HTTP handlers** (`/config`, `/indexing-status`) — now read from in-memory builders (`PublicConfigBuilder`, `IndexingStatusBuilder`) instead of ENSDb. Previously they could return 500 at startup if `EnsDbWriterWorker` hadn't yet populated ENSDb; now they always serve current data regardless of ENSDb state.
- **ENSApi caches** (`indexing-status.cache.ts`, `stack-info.cache.ts`) — updated to read the single `IndexingMetadataContext` record from ENSDb (under key `"indexing_metadata_context"`) instead of individual `ensindexer_public_config` and `ensindexer_indexing_status` records. The data they consume is structurally the same (`CrossChainIndexingStatusSnapshot` and `EnsIndexerStackInfo` respectively), but the ENSDb key and record format changed.
- **ENSApi config** (`config.schema.ts`) — `buildConfigFromEnvironment()` fetches `EnsIndexerPublicConfig` from `EnsNodeStackInfo.ensIndexer` (built from `IndexingMetadataContext.stackInfo.ensIndexer` + `EnsApiPublicConfig`) instead of from a standalone metadata record.
- **Integration test orchestrator** — `pollIndexingStatus()` now reads `IndexingMetadataContext` from ENSDb and checks `statusCode === "initialized"` instead of calling the removed `getIndexingStatusSnapshot()`.
- **Operators** — startup failure modes are more explicit. If ENSDb is unhealthy or ENSRainbow is unhealthy, `initIndexingOnchainEvents` logs the error and terminates with a non-zero exit code. Previously, the `EnsDbWriterWorker` would attempt to start and silently fail retries.

</details>

- Public APIs affected:
  - **New**: `EnsDbReader.getIndexingMetadataContext()`, `EnsDbReader.isHealthy()`, `EnsDbReader.isReady()`, `EnsDbWriter.upsertIndexingMetadataContext()`.
  - **Removed**: `EnsDbReader.getEnsIndexerPublicConfig()`, `EnsDbReader.getIndexingStatusSnapshot()`, `EnsDbReader.getEnsDbVersion()`, `EnsDbWriter.upsertEnsDbVersion()`, `EnsDbWriter.upsertEnsIndexerPublicConfig()`, `EnsDbWriter.upsertIndexingStatusSnapshot()`. All were public on their respective classes but only consumed internally within the ENSNode codebase.
- Docs updated: Inline JSDoc on `IndexingMetadataContext`, `IndexingMetadataContextBuilder`, `StackInfoBuilder`, `EnsDbWriterWorker`, `EnsDbReader.isHealthy()`, `EnsDbReader.isReady()`. Expanded module-level documentation in `ponder.ts` describing Ponder's full startup sequence.
- Naming decisions worth calling out: `EnsIndexerStackInfo` (stored in ENSDb as part of `IndexingMetadataContextInitialized.stackInfo`) vs. `EnsNodeStackInfo` (built by ENSApi = `EnsIndexerStackInfo` + `EnsApiPublicConfig`). This distinction prevents confusion about which type belongs to which layer.

---

## Testing Evidence

<details>

</details>

- Testing performed:
  - `IndexingMetadataContextBuilder` tests (7): fresh start (`IndexingMetadataContextUninitialized` + `omnichainStatus === "unstarted"`), restart with `EnsIndexerPublicConfig` compatibility validation (pass and fail), dev mode skips validation, dev mode tolerates incompatible config, invariant violation (non-unstarted omnichain status on fresh start), parallel fetch of all three data sources.
  - `StackInfoBuilder` tests (6): successful `EnsIndexerStackInfo` build from three public config sources, caching (second call returns cached value without re-fetching), error propagation from `EnsDbReader`, `EnsRainbowApiClient`, and `PublicConfigBuilder`.
  - `EnsDbWriterWorker` tests (8): periodic upsert via `IndexingMetadataContextBuilder`, error recovery between ticks, `process.exitCode` set on fatal error, `stop()` called on update failure.
  - `EnsDbReader` and `EnsDbWriter` tests updated for new `IndexingMetadataContext` methods.
  - `config.schema.test.ts` updated to use mock data from `config.schema.mock.ts` (new file).
  - `initIndexingOnchainEvents` coverage in `ponder.test.ts` updated.

- Known gaps: No integration test covering the full ENSIndexer startup sequence end-to-end. Each component is tested in isolation. A test that verifies the ordering `isHealthy() → migrateEnsNodeSchema() → waitForEnsRainbowToBeHealthy() → upsertIndexingMetadataContextRecord() → isReady() → waitForEnsRainbowToBeReady() → startEnsDbWriterWorker()` would increase confidence but requires ENSDb and ENSRainbow test infrastructure.
- What reviewers have to reason about manually (and why):
  - The ordering of the seven startup steps in `initIndexingOnchainEvents`. Each step has a dependency on the previous: ENSDb must be healthy before migrations run; migrations must complete before `isReady()` passes; `EnsIndexerStackInfo` must be upserted before ENSApi can read it; ENSRainbow must be ready before onchain events fire.
  - Whether `process.exitCode = 1; throw error` is the correct failure mode for every step. Some failures are definitely fatal (ENSDb health, migration failure, ENSRainbow readiness never reached); confirm that a transient ENSRainbow health check failure warrants a full process exit.

---

## Scope Reductions

<details>

</details>

- Follow-ups:
  - Store `EnsIndexerPublicConfig.ensIndexerBuildId` in ENSDb so `IndexingMetadataContextInitialized.stackInfo.ensIndexer` can be updated when the build ID changes across deployments.
  - Add `EnsDbReader.waitForEnsDbToBeHealthy()` with retries before migrations to match the `waitForEnsRainbowToBeHealthy()` pattern.
  - Transfer `EnsIndexerPublicConfig` fetching in ENSApi to a middleware layer (issue [#1806](https://github.com/namehash/ensnode/issues/1806)). This will require some updates on how we manage dependency injection on the ENSApi side. More details in [this Slack thread](https://namehash.slack.com/archives/C086Z6FNBHN/p1776841729795179).
- Why they were deferred: Each is independently valuable but would expand an already large PR. The core change — introducing `IndexingMetadataContext` and restructuring the startup flow — is self-contained.

---

## Risk Analysis

<details>

</details>

- Risk areas:
  1. **Schema migration moved from Ponder API entry point to `initIndexingOnchainEvents`.** Ponder dispatches no onchain events until the precondition promise returned by `eventHandlerPreconditions()` resolves. Since `initIndexingOnchainEvents` is called within that promise, migrations will complete before any handler fires. However, Ponder's HTTP API is created before onchain events start — ENSIndexer HTTP endpoints could receive requests before migrations complete. This is mitigated by the `/indexing-status` and `/config` endpoints now using in-memory builders (not ENSDb reads), so they don't depend on migration completion.
  2. **`EnsDbWriterWorker` error handling now terminates the process.** Previously, errors in the update loop were logged but the worker kept running. Now `stop()` is called and `process.exitCode = 1` is set before rethrowing. If the `IndexingMetadataContextBuilder` or `upsertIndexingMetadataContext()` intermittently fails, this will crash the ENSIndexer instance. Mitigation: the builder fetches data from in-memory sources that are already validated; the only failure point is the ENSDb upsert, which should succeed if ENSDb is healthy (verified by `isReady()` before the worker starts).
  3. **Removal of individual metadata record writes from ENSDb.** No external system consumed these records. If a rollback is needed, the `IndexingMetadataContextInitialized` type contains all the same data (`CrossChainIndexingStatusSnapshot` + `EnsIndexerStackInfo` which includes `EnsDbPublicConfig` and `EnsIndexerPublicConfig`).

- Mitigations or rollback options: The `upsertEnsNodeMetadata()` private method on `EnsDbWriter` remains available as a general-purpose metadata write primitive. The old individual record types can be restored if needed.
- Named owner if this causes problems: @tkopacki

---


apps/ensindexer/src/lib/indexing-engines/ponder.ts
```diff
@@ -192,25 +143,47 @@ let indexingActivationPromise: Promise<void> | null = null;
...
+            // Mark the preconditions as fully executed after the first time we execute
+            // the preconditions for onchain events, since that's the "hot path" and we want to
+            // minimize the overhead of this function in the long run.
+            eventHandlerPreconditionsFullyExecuted = true;
+          });
       }
 
-      return await indexingActivationPromise;
+      return await indexingOnchainEventsPromise;
     }
```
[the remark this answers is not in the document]
tk-o: It's not possible that ENSIndexer produces no onchain events.

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -43,86 +30,39 @@ export class EnsDbWriterWorker {
...
-      module: "EnsDbWriterWorker",
-    });
-
-    // Task 3: recurring upsert of Indexing Status Snapshot into ENSDb.
+    // Recurring update of the Indexing Metadata Context record in ENSDb.
     this.indexingStatusInterval = setInterval(
-      () => this.upsertIndexingStatusSnapshot(),
+      () => this.updateIndexingMetadataContext(),
       secondsToMilliseconds(INDEXING_STATUS_RECORD_UPDATE_INTERVAL),
     );
```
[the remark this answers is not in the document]
tk-o: The operations in this task are very quick and orders of magnitude belowe the interval lenght.

apps/ensindexer/src/lib/indexing-metadata-context-builder/indexing-metadata-context-builder.ts
```diff
@@ -0,0 +1,115 @@
...
+    );
+
+    if (
+      storedIndexingMetadataContext.statusCode === IndexingMetadataContextStatusCodes.Uninitialized
+    ) {
+      logger.info({ msg: `Indexing Metadata Context is "uninitialized"` });
+
+      invariant_indexingStatusIsUnstartedForIndexingMetadataContextUninitialized(
+        inMemoryIndexingStatusSnapshot,
+      );
```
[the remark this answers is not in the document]
tk-o: Indexing onchain events can start if, and only if, the "stored" `IndexingMetadataContext` is initialized. In other words, it's not possible for the "in-memory" indexing status snapshot to be other "unstarted" if the "stored" `IndexingMetadataContext` was "uninitialized".

apps/ensindexer/src/lib/ensrainbow/singleton.ts
```diff
@@ -24,6 +24,69 @@ export const ensRainbowClient = new EnsRainbowApiClient({
...
+  waitForEnsRainbowToBeHealthyPromise = pRetry(async () => ensRainbowClient.health(), {
+    retries: 3,
+    onFailedAttempt: ({ attemptNumber, retriesLeft }) => {
+      logger.warn({
+        msg: `ENSRainbow health check failed`,
+        attempt: attemptNumber,
+        retriesLeft,
+        ensRainbowInstance: ensRainbowUrl.href,
+        advice: `This might be a transient issue after ENSNode deployment. If this persists, it might indicate an issue with the ENSRainbow instance or connectivity to it.`,
+      });
```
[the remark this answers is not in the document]
tk-o: No need for that, ENSRainbow instance should be healthy right away after its process started, while the ENSRainbow readiness might take a long time to be reached (30+ minutes).

tk-o:
@greptile review

apps/ensindexer/src/lib/ensrainbow/singleton.ts
```diff
@@ -24,6 +24,69 @@ export const ensRainbowClient = new EnsRainbowApiClient({
...
+
+      // Throw the error to terminate the ENSIndexer process due to the failed health check of a critical dependency
+      throw error;
+    });
+
+  return waitForEnsRainbowToBeHealthyPromise;
+}
+
 /**
```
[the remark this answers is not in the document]
tk-o: ENSRainbow instance is expected to be healthy even before the ENSIndexer instance starts.

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -140,162 +88,23 @@ export class EnsDbWriterWorker {
...
-   * stored one. Validation is skipped if the local Ponder app is in dev mode,
-   * allowing to override the stored config in ENSDb with the current in-memory
-   * config, without having to keep them compatible.
-   *
-   * @returns The in-memory config when validation passes or no stored config
-   *          exists.
-   * @throws Error if either fetch fails, or if the in-memory config is
-   *         incompatible with the stored config.
+   * @throws Error if the update operation fails.
    */
```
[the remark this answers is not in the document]
tk-o: Will update PR description.

apps/ensindexer/src/lib/indexing-metadata-context-builder/indexing-metadata-context-builder.ts
```diff
@@ -0,0 +1,118 @@
...
+  async getIndexingMetadataContext(): Promise<IndexingMetadataContextInitialized> {
+    const [
+      inMemoryIndexingStatusSnapshot,
+      inMemoryEnsIndexerStackInfo,
+      storedIndexingMetadataContext,
+    ] = await Promise.all([
+      this.indexingStatusBuilder.getOmnichainIndexingStatusSnapshot(),
+      this.stackInfoBuilder.getStackInfo(),
+      this.ensDbClient.getIndexingMetadataContext(),
+    ]);
```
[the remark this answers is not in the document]
tk-o: This is minor

tk-o:
@greptile review

apps/ensapi/src/cache/indexing-status.cache.ts
```diff
@@ -1,50 +1,67 @@
...
+          // IndexingMetadataContext was uninitialized in ENSDb.
+          // Therefore, throw an error so that this current invocation of `readCache` will:
+          // - Reject the newly fetched response (if any) such that it won't be cached.
+          // - Return the most recently cached value from prior invocations, or `null` if no prior invocation successfully cached a value.
+          logger.error(
+            error,
+            `Error occurred while loading Indexing Metadata Context record from ENSNode Metadata table in ENSDb. ` +
+              `Where clause applied: ("ensIndexerSchemaName" = "${ensDbClient.ensIndexerSchemaName}", "key" = "${EnsNodeMetadataKeys.IndexingMetadataContext}"). ` +
+              `The cached indexing status snapshot (if any) will not be updated.`,
+          );
```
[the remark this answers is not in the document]
tk-o: That's ok

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+export async function initIndexingOnchainEvents(): Promise<void> {
+  try {
+    // Ensure ENSDb instance is healthy before trying to run any queries against it.
+    const isEnsDbHealthy = await ensDbClient.isHealthy();
+
+    // Invariant: ENSDb instance must be healthy by now.
+    if (!isEnsDbHealthy) {
+      throw new Error("ENSDb instance must be healthy");
+    }
```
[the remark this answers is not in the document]
tk-o: If ENSDb cannot perform a simple query, it's evident that ENSDb connection is not working.

packages/ensdb-sdk/src/client/ensdb-reader.ts
```diff
@@ -130,33 +128,33 @@ export class EnsDbReader<
...
-    const record = await this.getEnsNodeMetadata<SerializedEnsNodeMetadataEnsDbVersion>({
-      key: EnsNodeMetadataKeys.EnsDbVersion,
-    });
-
-    return record;
+  async isHealthy(): Promise<boolean> {
+    try {
+      await this.ensDb.execute("SELECT 1;");
+      return true;
+    } catch {
```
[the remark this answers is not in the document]
tk-o: The only error possible here is the connection issue.

apps/ensindexer/src/lib/ensrainbow/singleton.ts
```diff
@@ -60,12 +123,11 @@ export function waitForEnsRainbowToBeReady(): Promise<void> {
     retries: 60, // This allows for a total of over 1 hour of retries with 1 minute between attempts.
     minTimeout: secondsToMilliseconds(60),
     maxTimeout: secondsToMilliseconds(60),
-    onFailedAttempt: ({ error, attemptNumber, retriesLeft }) => {
+    onFailedAttempt: ({ attemptNumber, retriesLeft }) => {
       logger.warn({
         msg: `ENSRainbow health check failed`,
```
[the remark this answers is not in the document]
tk-o: ENSRainbow client does not implement the `ready()` method yet.

apps/ensindexer/src/lib/stack-info-builder/stack-info-builder.ts
```diff
@@ -0,0 +1,46 @@
...
+      const ensRainbowPublicConfig = await this.ensRainbowClient.config();
+
+      this.immutableStackInfo = buildEnsIndexerStackInfo(
+        ensDbPublicConfig,
+        ensIndexerPublicConfig,
+        ensRainbowPublicConfig,
+      );
+    }
+
+    return this.immutableStackInfo;
```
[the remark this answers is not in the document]
tk-o: Prefer code simplicity, the concurrent read case is very unlikely.

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -1,137 +1,85 @@
...
 import { logger } from "@/lib/logger";
-import type { PublicConfigBuilder } from "@/lib/public-config-builder/public-config-builder";
 
 /**
  * Interval in seconds between two consecutive attempts to upsert
- * the Indexing Status Snapshot record into ENSDb.
+ * the IndexingMetadataContext record into ENSDb.
  */
-const INDEXING_STATUS_RECORD_UPDATE_INTERVAL: Duration = 1;
+const INDEXING_METADATA_CONTEXT_RECORD_UPDATE_INTERVAL: Duration = 1;
```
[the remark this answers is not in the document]
tk-o: Ack

apps/ensindexer/src/lib/ensrainbow/singleton.ts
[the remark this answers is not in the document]
tk-o: Will use `count()` response to see if records count is `> 0`.

apps/ensindexer/src/lib/indexing-engines/ponder.ts
```diff
@@ -224,23 +214,31 @@ async function eventHandlerPreconditions(eventType: EventTypeId): Promise<void>
...
-        // Initialize the indexing setup just once.
-        indexingSetupPromise = initializeIndexingSetup();
-      }
-
-      return await indexingSetupPromise;
+      // For some ENSIndexer instances, the setup handlers are not defined at all,
+      // for example, if the ENSIndexer instance has only the `ensv2` plugin activated.
+      // In this case, some important logic, such as running migrations for ENSNode Schema
+      // in ENSDb, would not be executed at all, which would cause the ENSIndexer instance
+      // to not work properly. Therefore, all logic required to be executed before
```
shrugs, line 221: Maybe this is the wrong time to do this, but i do think that we should just remove the distinction between a setup event and an onchain event and just have two different integrations, one that runs once (upon any event) like this init function, and then another that runs for _every_ event like my performance tracking helper, and that's it. I don't think we need to disambiguate between setup and onchain events.

apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
```diff
@@ -22,12 +20,7 @@ export function startEnsDbWriterWorker() {
     throw new Error("EnsDbWriterWorker has already been initialized");
   }
 
-  ensDbWriterWorker = new EnsDbWriterWorker(
-    ensDbClient,
-    publicConfigBuilder,
-    indexingStatusBuilder,
-    localPonderClient,
-  );
+  ensDbWriterWorker = new EnsDbWriterWorker(ensDbClient, indexingMetadataContextBuilder);
```
lightwalker-eth, line 23: Would it make sense to rename the `EnsDbWriterWorker` idea to something like `EnsIndexerMetadataWriterWorker` or something like that?

As I understand, it's not just a writer worker for ENSDb (in general). It specifically is the writer worker for writing the ENSIndexerMetadata to the Metadata table inside the ENSNode schema in ENSDb? Is that right?

Appreciate your advice on how we might optimize the name for this idea.

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -140,162 +88,23 @@ export class EnsDbWriterWorker {
...
-    const omnichainSnapshot = await this.indexingStatusBuilder.getOmnichainIndexingStatusSnapshot();
-
-    // It only makes sense to write Indexing Status Snapshots into ENSDb once
-    // the indexing process has started, as before that there is no meaningful
-    // status to record.
-    // Invariant: the Omnichain Status must indicate that indexing has started already.
-    if (omnichainSnapshot.omnichainStatus === OmnichainIndexingStatusIds.Unstarted) {
-      throw new Error("Omnichain Status must not be 'Unstarted'.");
-    }
+  private async updateIndexingMetadataContext(): Promise<void> {
```
lightwalker-eth, line 104:
```suggestion
  private async upsertIndexingMetadataContext(): Promise<void> {
```

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -140,162 +88,23 @@ export class EnsDbWriterWorker {
...
-      this.indexingStatusInterval = null;
+    if (this.indexingMetadataContextUpdateInterval) {
+      clearInterval(this.indexingMetadataContextUpdateInterval);
+      this.indexingMetadataContextUpdateInterval = null;
     }
   }
 
   /**
-   * Get validated ENSIndexer Public Config object for the ENSDb Writer Worker.
+   * Update the current IndexingMetadataContext into ENSDb.
```
lightwalker-eth, line 98:
```suggestion
   * Upsert the current IndexingMetadataContext into ENSDb.
```

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -140,162 +88,23 @@ export class EnsDbWriterWorker {
...
-   * mode, the in-memory config is validated for compatibility against the
-   * stored one. Validation is skipped if the local Ponder app is in dev mode,
-   * allowing to override the stored config in ENSDb with the current in-memory
-   * config, without having to keep them compatible.
-   *
-   * @returns The in-memory config when validation passes or no stored config
-   *          exists.
-   * @throws Error if either fetch fails, or if the in-memory config is
-   *         incompatible with the stored config.
+   * @throws Error if the update operation fails.
```
lightwalker-eth, line 102:
```suggestion
   * @throws Error if the upsert operation fails.
```

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -1,137 +1,85 @@
...
+ * A worker responsible for writing the current {@link CrossChainIndexingStatusSnapshot} into
+ * the {@link IndexingMetadataContext} record in ENSDb.
  */
 export class EnsDbWriterWorker {
   /**
-   * Interval for recurring upserts of Indexing Status Snapshots into ENSDb.
+   * Interval for recurring updates of IndexingMetadataContext record in ENSDb.
    */
-  private indexingStatusInterval: ReturnType<typeof setInterval> | null = null;
+  private indexingMetadataContextUpdateInterval: ReturnType<typeof setInterval> | null = null;
```
lightwalker-eth, line 25:
```suggestion
  private indexingMetadataContextUpsertInterval: ReturnType<typeof setInterval> | null = null;
```

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -1,137 +1,85 @@
...
- * - ENSDb version
- * - ENSIndexer Public Config
- * - ENSIndexer Indexing Status Snapshots
+ * A worker responsible for writing the current {@link CrossChainIndexingStatusSnapshot} into
+ * the {@link IndexingMetadataContext} record in ENSDb.
  */
 export class EnsDbWriterWorker {
   /**
-   * Interval for recurring upserts of Indexing Status Snapshots into ENSDb.
+   * Interval for recurring updates of IndexingMetadataContext record in ENSDb.
```
lightwalker-eth, line 23:
```suggestion
   * Interval for recurring upserts of IndexingMetadataContext in ENSDb.
```

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
```diff
@@ -1,137 +1,85 @@
...
+const INDEXING_METADATA_CONTEXT_RECORD_UPDATE_INTERVAL: Duration = 1;
 
 /**
  * ENSDb Writer Worker
  *
- * A worker responsible for writing ENSIndexer-related metadata into ENSDb, including:
- * - ENSDb version
- * - ENSIndexer Public Config
- * - ENSIndexer Indexing Status Snapshots
+ * A worker responsible for writing the current {@link CrossChainIndexingStatusSnapshot} into
```
lightwalker-eth, line 18: I think there's an opportunity to refine the docs in relevant places to be more specific / explicit where this is writing to in ENSDb. Ex: the ENSNode schema, etc.

apps/ensindexer/src/lib/ensdb-writer-worker/ensdb-writer-worker.ts
lightwalker-eth, line 13:
```suggestion
const INDEXING_METADATA_CONTEXT_RECORD_UPSERT_INTERVAL: Duration = 1;
```

lightwalker-eth, line 13: Suggest to just make this the default, and then have an optional param in the constructor for `EnsDbWriterWorker` to set a custom interval. If no custom interval is defined then this is the default value that should be used.

Goal: Take a step toward optimizing for this logic to move into `ensdb-sdk`.

apps/ensindexer/ponder/src/api/index.ts
```diff
@@ -5,27 +5,10 @@ import { cors } from "hono/cors";
 
 import type { ErrorResponse } from "@ensnode/ensnode-sdk";
 
-import { migrateEnsNodeSchema } from "@/lib/ensdb/migrate-ensnode-schema";
-import { startEnsDbWriterWorker } from "@/lib/ensdb-writer-worker/singleton";
 import { logger } from "@/lib/logger";
 
 import ensNodeApi from "./handlers/ensnode-api";
```
lightwalker-eth, line 10: Hmm. Should we rename this? It seems strange to call it an "ENSNode API".

apps/ensindexer/ponder/src/api/index.ts
```diff
@@ -5,27 +5,10 @@ import { cors } from "hono/cors";
...
-    module: "ponder-api",
-  });
-  process.exitCode = 1;
-  throw error;
-});
-
-// The entry point for the ENSDb Writer Worker.
-startEnsDbWriterWorker();
-
 const app = new Hono();
```
lightwalker-eth, line 13: Several lines below I see:

```
// use ENSNode HTTP API at /api
app.route("/api", ensNodeApi);
```

Please see related feedback above on the terminology in the comment

apps/ensindexer/src/lib/indexing-engines/ponder.ts
shrugs: but let's queue that up for a future pr

shrugs (approved):
it's quite structural, so frankly the bots will be better than my eyes for catching errors. but sound architecturally. i did not manually check the tests/mocks, and trust that to you + bots

apps/ensapi/src/cache/indexing-status.cache.ts
```diff
@@ -1,50 +1,67 @@
...
-            );
-            throw error;
-          }),
+          // The CrossChainIndexingStatusSnapshot has been successfully loaded for caching.
+          // Therefore, return it so that this current invocation of `readCache` will:
+          // - Replace the currently cached value (if any) with this new value.
+          // - Return this non-null value.
+          return indexingMetadataContext.indexingStatus;
+        } catch (error) {
+          // IndexingMetadataContext was uninitialized in ENSDb.
```
lightwalker-eth, line 52:
```suggestion
          // IndexingMetadataContext was uninitialized in ENSDb (or some error occurred while reading from ENSDb)
```

apps/ensapi/src/cache/indexing-status.cache.ts
```diff
@@ -1,50 +1,67 @@
...
+            `Error occurred while loading Indexing Metadata Context record from ENSNode Metadata table in ENSDb. ` +
+              `Where clause applied: ("ensIndexerSchemaName" = "${ensDbClient.ensIndexerSchemaName}", "key" = "${EnsNodeMetadataKeys.IndexingMetadataContext}"). ` +
+              `The cached indexing status snapshot (if any) will not be updated.`,
+          );
+          throw error;
+        }
+      },
       // We need to refresh the indexing status cache very frequently.
       // ENSDb won't have issues handling this frequency of queries.
       ttl: 1, // 1 second
```
lightwalker-eth, line 67: Instead of hardcoding to 1-second, it would be nice to make it configurable with perhaps 1-second just be the default.

Goal: Take steps towards moving this into `ensdb-sdk`.

apps/ensapi/src/cache/stack-info.cache.ts
```diff
@@ -2,76 +2,91 @@ import config from "@/config";
...
-  const ensRainbowPublicConfig = ensIndexerPublicConfig.ensRainbowPublicConfig;
-
-  return buildEnsNodeStackInfo(
-    ensApiPublicConfig,
-    ensDbPublicConfig,
-    ensIndexerPublicConfig,
-    ensRainbowPublicConfig,
-  );
-}
+export type EnsNodeStackInfoCache = SWRCache<EnsNodeStackInfo>;
```
lightwalker-eth, line 19: Hmm. Not sure I understand. I thought the indexing status cache would now hold both the stack info AND the indexing status? Therefore, why have two separate caches? Why are they not combined?

apps/ensapi/src/config/config.schema.mock.ts
```diff
@@ -0,0 +1,108 @@
...
+  namespace: "mainnet",
+  ensIndexerSchemaName: "ensindexer_0",
+  ensRainbowPublicConfig: {
+    serverLabelSet: { labelSetId: "subgraph", highestLabelSetVersion: 0 },
+    versionInfo: {
+      ensRainbow: packageJson.version,
+    },
+  },
+  indexedChainIds: [1],
+  isSubgraphCompatible: false,
```
lightwalker-eth, line 40: I believe we'll want to work towards removing this `isSubgraphCompatible` field from this data model. Not in this PR, but could you please create a follow-up issue for us to track this opportunity?

Goal: We want to begin optimizing more for ENSv2 and the ENS Omnigraph and preparing to leave behind the ENS Subgraph. This idea is also related to the goal of us pushing to get this data model to a place where it is super stable and we don't need to keep breaking it.

apps/ensapi/src/config/config.schema.mock.ts
```diff
@@ -0,0 +1,108 @@
...
+  clientLabelSet: { labelSetId: "subgraph", labelSetVersion: 0 },
```
lightwalker-eth, line 41: I believe we'll want to work towards making it optional to run ENSRainbow. Not in this PR, but could you please create a follow-up issue for us to track this opportunity.

Goal: We should make the stack info data model support the possibility that there's no ENSRainbow, and therefore that there also is no `clientLabelSet` for ENSIndexer.

apps/ensapi/src/config/config.schema.ts
```diff
@@ -70,13 +74,17 @@ export async function buildConfigFromEnvironment(env: EnsApiEnvironment): Promis
...
       async () => {
-        const config = await ensDbClient.getEnsIndexerPublicConfig();
+        const indexingMetadataContext = await ensDbClient.getIndexingMetadataContext();
 
-        if (!config) {
-          throw new Error("ENSIndexer Public Config not yet available in ENSDb.");
+        if (
+          indexingMetadataContext.statusCode === IndexingMetadataContextStatusCodes.Uninitialized
+        ) {
+          throw new Error(
```
lightwalker-eth, line 82: I assume it would be helpful here to include info on the ENSIndexerSchemaName?

apps/ensapi/src/config/config.schema.ts
```diff
@@ -70,13 +74,17 @@ export async function buildConfigFromEnvironment(env: EnsApiEnvironment): Promis
...
+            "EnsIndexerPublicConfig could not be fetched, the IndexingMetadataContext record has not been initialized in ENSDb yet.",
+          );
         }
 
-        return config;
+        return indexingMetadataContext.stackInfo.ensIndexer;
       },
       {
         retries: 13, // This allows for a total of over 1 hour of retries with the exponential backoff strategy
```
lightwalker-eth, line 90: Should we keep such a long retry period still? Appreciate your advice!

apps/ensindexer/ponder/src/api/handlers/ensnode-api.ts
```diff
@@ -10,45 +11,31 @@ import {
...
-    // an error response.
-    if (typeof crossChainSnapshot === "undefined") {
-      throw new Error("ENSDb does not contain an Indexing Status Snapshot");
-    }
-
-    const projectedAt = getUnixTime(new Date());
-    const realtimeProjection = createRealtimeIndexingStatusProjection(
-      crossChainSnapshot,
-      projectedAt,
+    const now = getUnixTime(new Date());
```
lightwalker-eth, line 31: Hmm. Should we move the call to `now` before we build `omnichainSnapshot`? Appreciate your advice.

apps/ensindexer/src/lib/ensrainbow/singleton.ts
```diff
@@ -24,6 +24,69 @@ export const ensRainbowClient = new EnsRainbowApiClient({
...
+
+  waitForEnsRainbowToBeHealthyPromise = pRetry(async () => ensRainbowClient.health(), {
+    retries: 3,
+    onFailedAttempt: ({ attemptNumber, retriesLeft }) => {
+      logger.warn({
+        msg: `ENSRainbow health check failed`,
+        attempt: attemptNumber,
+        retriesLeft,
+        ensRainbowInstance: ensRainbowUrl.href,
+        advice: `This might be a transient issue after ENSNode deployment. If this persists, it might indicate an issue with the ENSRainbow instance or connectivity to it.`,
```
lightwalker-eth, line 66:
```suggestion
        advice: `This might be a transient issue after a new ENSNode deployment. If this persists, it might indicate an issue with the ENSRainbow instance or connectivity to it.`,
```

apps/ensindexer/src/lib/ensrainbow/singleton.ts
```diff
@@ -56,40 +119,45 @@ export function waitForEnsRainbowToBeReady(): Promise<void> {
...
-      logger.warn({
-        msg: `ENSRainbow health check failed`,
-        attempt: attemptNumber,
-        retriesLeft,
-        error: retriesLeft === 0 ? error : undefined,
-        ensRainbowInstance: ensRainbowUrl.href,
-        advice: `This might be due to ENSRainbow having a cold start, which can take 30+ minutes.`,
-      });
+  waitForEnsRainbowToBeReadyPromise = pRetry(
+    // TODO: replace this count check with an explicit `ready()` method in ENSRainbow Client.
```
lightwalker-eth, line 123: Could you please create a follow-up issue to help us remember this and then link to that issue in a comment here?

apps/ensindexer/src/lib/ensrainbow/singleton.ts
```diff
@@ -56,40 +119,45 @@ export function waitForEnsRainbowToBeReady(): Promise<void> {
...
-      const errorMessage = error instanceof Error ? error.message : "Unknown error";
-
       logger.error({
-        msg: `ENSRainbow health check failed after multiple attempts`,
+        msg: `ENSRainbow readiness check failed after multiple attempts`,
         error,
         ensRainbowInstance: ensRainbowUrl.href,
       });
 
       // Throw the error to terminate the ENSIndexer process due to the failed health check of a critical dependency
```
lightwalker-eth, line 159:
```suggestion
      // Throw the error to terminate the ENSIndexer process due to the failed readiness check of a critical dependency
```

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // Ensure the ENSNode Schema in ENSDb is up to date by running any pending migrations.
+    await migrateEnsNodeSchema();
+
+    // Before calling `ensRainbowClient.config()`, we want to make sure that
+    // the ENSRainbow instance is healthy and ready to serve requests.
+    // This is a quick check, as we expect the ENSRainbow instance to be healthy
+    // by the time ENSIndexer instance executes `initIndexingOnchainEvents`.
+    await waitForEnsRainbowToBeHealthy();
```
lightwalker-eth, line 67: Shouldn't this be checking ready, not healthy?

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
```diff
@@ -0,0 +1,106 @@
...
+    // not completed as expected, the ENSDb instance might not be ready, and
+    // we want to catch this issue before we start processing onchain events.
+    const isEnsDbReady = await ensDbClient.isReady();
+    if (!isEnsDbReady) {
+      throw new Error("ENSDb instance must be ready before onchain events can be indexed.");
+    }
+
+    // Before starting to process onchain events, we want to make sure that
+    // ENSRainbow is ready to serve the "heal" requests.
+    await waitForEnsRainbowToBeReady();
```
lightwalker-eth, line 87: Ah I see the ready check here. Hmm. Why are we checking both healthy and ready?

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // TODO: start Indexing Status Sync worker
```
lightwalker-eth, line 89: Could you please add a follow-up issue to our GitHub for this and add a link to it in a comment here? Thanks

apps/ensindexer/src/lib/indexing-engines/ponder.ts
```diff
@@ -1,9 +1,48 @@
 /**
- * This module is an abstraction layer for the Indexing Engine of ENSIndexer.
- * It decouples core indexing logic from Ponder-specific implementation details.
- * Benefits of this decoupling include:
- * - Building a custom context data model.
- * - Implementing shared logic before or after event handlers, if needed.
+ * Ponder Indexing Engine
+ *
+ * This module provides an abstraction layer over the Ponder Indexing Engine
+ * to decouple the core indexing logic of the ENSIndexer from Ponder-specific
```
lightwalker-eth, line 5:
```suggestion
 * to decouple the core indexing logic of ENSIndexer from Ponder-specific
```

apps/ensindexer/src/lib/public-config-builder/public-config-builder.ts
```diff
@@ -37,16 +37,19 @@ export class PublicConfigBuilder {
...
    *
-   * @throws if the built ENSIndexer Public Config does not conform to
+   * @throws if the built {@link EnsIndexerPublicConfig} does not conform to
    *         the expected schema
    */
   async getPublicConfig(): Promise<EnsIndexerPublicConfig> {
     if (typeof this.immutablePublicConfig === "undefined") {
       const [versionInfo, ensRainbowPublicConfig] = await Promise.all([
         this.getEnsIndexerVersionInfo(),
+        // TODO: remove dependency on ENSRainbow by dropping `ensRainbowPublicConfig` from `EnsIndexerPublicConfig`.
```
lightwalker-eth, line 52: Could you add a comment here with a link to a related GitHub Issue? Thanks

packages/ensdb-sdk/src/client/ensdb-reader.ts
```diff
@@ -130,33 +135,33 @@ export class EnsDbReader<
...
+   * Check if the ENSDb instance is ready by verifying that it is
+   * healthy and the {@link IndexingMetadataContext} has been initialized for
+   * the ENSIndexer Schema used by this ENSDbReader instance.
    */
-  async getEnsIndexerPublicConfig(): Promise<EnsIndexerPublicConfig | undefined> {
-    const record = await this.getEnsNodeMetadata<SerializedEnsNodeMetadataEnsIndexerPublicConfig>({
-      key: EnsNodeMetadataKeys.EnsIndexerPublicConfig,
-    });
+  async isReady(): Promise<boolean> {
+    const isHealthy = await this.isHealthy();
```
lightwalker-eth, line 156: Maybe no need to also call isHealthy here since `getIndexingMetadataContext` performs an even more complex select?

packages/ensdb-sdk/src/client/ensdb-reader.ts
```diff
@@ -130,33 +135,33 @@ export class EnsDbReader<
...
-   * @returns the existing record, or `undefined`.
+   * Check if the ENSDb instance is ready by verifying that it is
+   * healthy and the {@link IndexingMetadataContext} has been initialized for
+   * the ENSIndexer Schema used by this ENSDbReader instance.
    */
-  async getEnsIndexerPublicConfig(): Promise<EnsIndexerPublicConfig | undefined> {
-    const record = await this.getEnsNodeMetadata<SerializedEnsNodeMetadataEnsIndexerPublicConfig>({
-      key: EnsNodeMetadataKeys.EnsIndexerPublicConfig,
-    });
+  async isReady(): Promise<boolean> {
```
lightwalker-eth, line 155:
```suggestion
  async isReadyForReads(): Promise<boolean> {
```

wdyt? I'm not sure if I remember right but I think the ENSDbWriter extends the reader? If not then please ignore this comment. My worry here was if it does extend then the writer might say it's not ready yet (for a write) when it actually is ready for the write to initialize!

Appreciate your advice.

packages/ensdb-sdk/src/client/ensdb-reader.ts
```diff
@@ -171,22 +176,20 @@ export class EnsDbReader<
   }
 
   /**
-   * Get Indexing Status Snapshot
+   * Get Indexing Metadata Context
    *
-   * @returns the existing record, or `undefined`.
+   * @returns the initialized record, or a default uninitialized one if no record exists in ENSDb.
```
lightwalker-eth, line 181:
```suggestion
   * @returns the initialized `IndexingMetadataContext`, or an uninitialized `IndexingMetadataContext` if no initialized `IndexingMetadataContext` exists in ENSDb.
```

packages/ensdb-sdk/src/client/ensnode-metadata.ts
```diff
@@ -1,40 +1,31 @@
-import type {
-  CrossChainIndexingStatusSnapshot,
-  EnsIndexerPublicConfig,
-} from "@ensnode/ensnode-sdk";
+import type { IndexingMetadataContextInitialized } from "@ensnode/ensnode-sdk";
 
 /**
  * Keys used to distinguish records in `ensnode_metadata` table in the ENSDb.
```
lightwalker-eth, line 4:
```suggestion
 * Keys used to distinguish records in the `ensnode_metadata` table in ENSDb.
```

packages/ensdb-sdk/src/client/ensnode-metadata.ts
```diff
@@ -1,40 +1,31 @@
...
-  value: EnsIndexerPublicConfig;
-}
-
-export interface EnsNodeMetadataEnsIndexerIndexingStatus {
-  key: typeof EnsNodeMetadataKeys.EnsIndexerIndexingStatus;
-  value: CrossChainIndexingStatusSnapshot;
+/**
+ * ENSNode Metadata record for Indexing Metadata Context
+ *
+ * This record is used to store the Indexing Metadata Context in
```
lightwalker-eth, line 15:
```suggestion
 * This record is used to store the Indexing Metadata Context in the
```

packages/integration-test-env/src/orchestrator.ts
```diff
@@ -199,9 +202,14 @@ async function pollIndexingStatus(
     while (Date.now() - start < timeoutMs) {
       checkAborted();
       try {
-        const snapshot = await ensDbClient.getIndexingStatusSnapshot();
-        if (snapshot !== undefined) {
-          const omnichainStatus = snapshot.omnichainSnapshot.omnichainStatus;
+        const indexingMetadataContext = await ensDbClient.getIndexingMetadataContext();
```
lightwalker-eth, line 205: For the catch block associated with this statement, maybe we should log something?

lightwalker-eth (approved):
@tk-o Great work here! Shared feedback and suggestions. Could you take the lead to action these in a new PR? Thanks!

apps/ensapi/src/cache/stack-info.cache.ts
tk-o: @lightwalker-eth we can have a single cache for indexing metadata context. It'd include both, the indexing status snapshot, and the stack info. Would that work?

apps/ensindexer/src/lib/indexing-engines/ponder.ts
tk-o: Created issue here:
- https://github.com/namehash/ensnode/issues/2027

apps/ensindexer/src/lib/ensdb-writer-worker/singleton.ts
tk-o: We could call it `IndexingStatusSyncWorker` as it synchronises the indexing status snapshot held in ENSDb.

apps/ensindexer/src/lib/ensrainbow/singleton.ts
tk-o: It's been already sorted out 👍 
https://github.com/namehash/ensnode/blob/d7bdabc8d0702cbf80769f3ef06bb3a204fd346c/apps/ensindexer/src/lib/ensrainbow/singleton.ts#L149

apps/ensapi/src/config/config.schema.mock.ts
tk-o: Issue logged:
- https://github.com/namehash/ensnode/issues/2031

apps/ensapi/src/config/config.schema.mock.ts
tk-o: Issue logged:
- https://github.com/namehash/ensnode/issues/2032

apps/ensapi/src/config/config.schema.ts
tk-o: We should be able to remove this whole async call once #1806 is resolved.

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
tk-o: This is a defensive check ensuring that ENSRainbow instance is there and that we can fetch its public config.

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
tk-o: Health check is needed for ENSIndexer to fetch public config from ENSRainbow instance. The readiness check is needed for ENSIndexer instance to begin indexing onchain events.

apps/ensindexer/src/lib/indexing-engines/init-indexing-onchain-events.ts
tk-o: Issue logged here:
- https://github.com/namehash/ensnode/issues/2033

packages/ensdb-sdk/src/client/ensdb-reader.ts
tk-o: Maybe we should rename to `isFullyPopulated()` or is `hasCompletedInitialization()`?

PR #1999, opened 2026-04-27
docs(example): ensnode-react

notrab:
# Todo

- [ ] Indexing Status

# Lite PR

## Summary

- What changed (1-3 bullets, no essays).

---

## Why

- Why this change exists. Link to related GitHub issues where relevant.

---

## Testing

- How this was tested.

---

---

- [ ] Relevant changesets are included (or are not required)


packages/ensnode-react/example/README.md
```diff
@@ -0,0 +1,20 @@
+# ensnode-react Example
```
lightwalker-eth: It's super logical to put the `ensnode-react-example` inside `examples/ensnode-react-example/` within the monorepo.

The context we have right now though is an unusual one. This `ensnode-react-example` app is important for us to ship right now for the purpose of internally refining how `ensnode-react` works so that we can make a plan for how best to integrate it into `enskit`. But at this time I don't think we want to be promoting the `ensnode-react` package. Therefore we also don't want to be promoting any example apps for `ensnode-react`.

I don't feel good about putting this app inside the root `examples` directory at this time. Could you please share your advice on a temporary alternative? For example, I'm thinking we stick it under `docs` temporarily.

These are all temporary steps, but everything we're doing here right now is very important. We need to have a *maximally clear* narrative to promote for how to integrate with ENSNode and that answer is `enskit` not `ensnode-react`. We therefore should take some special actions here now.

We also need to promote `enskit` integrations *now* and not later after we figure out how we want to integrate `ensnode-react` into `enskit`.

packages/ensnode-react/example/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "vite build",
+    "preview": "vite preview",
+    "typecheck": "tsgo --noEmit"
+  },
+  "dependencies": {
+    "@ensnode/ensnode-react": "workspace:*",
+    "@ensnode/ensnode-sdk": "workspace:*",
```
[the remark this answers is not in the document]
lightwalker-eth: @notrab Yes I imagine this app may not need `ensnode-sdk`.

packages/ensnode-react/example/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@ensnode/ensnode-react-example",
```
lightwalker-eth, line 2: @notrab Can you please take the lead to wire up Vercel so that this example app can automatically get all the Vercel stuff working?

examples/ensnode-react-example/src/PrimaryNameView.tsx
```diff
@@ -0,0 +1,51 @@
+import type { Address, ChainId } from "enssdk";
+import { useState } from "react";
+
+import { usePrimaryName } from "@ensnode/ensnode-react";
+
+const DEFAULT_ADDRESS: Address = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"; // vitalik.eth
```
lightwalker-eth: All address values we use in this app should use `NormalizedAddress` as imported from `enssdk`. You also need to call a utility function such as `asNormalizedAddress` on this hardcoded address string.

examples/ensnode-react-example/src/PrimaryNameView.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <div>
+      <h2>Primary Name</h2>
+      <p>
+        Resolves the ENSIP-19 Mainnet Primary Name for an address using <code>usePrimaryName</code>.
+      </p>
+
+      <form
+        onSubmit={(event) => {
+          event.preventDefault();
+          setAddress(input as Address);
```
lightwalker-eth: @notrab I need to stop seeing code like this. Ever. It completely goes against our culture as a team and company. I'm going completely crazy at how many times I have to repeat the related ideas. There's so many problems in this 1 line of code I'm not going to spend the time to explain them all over and over again.

examples/ensnode-react-example/src/App.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+export function App() {
+  return (
+    <StrictMode>
+      <EnsNodeProvider options={options}>
+        <h1>
+          <code>ensnode-react</code> Example App
+        </h1>
+        <p>
+          Connected to <code>{ENSNODE_URL}</code>
```
lightwalker-eth: For this example app to achieve its goals, we need it to implement more critical ideas:

1. How is it guaranteeing that it was able to successfully connect to the provided ENSNode?
2. If for some reason it was unable to connect to the provided ENSNode, how is this case being gracefully handled? All the meaningfully distinct edge cases should have explicit handling. I don't want to see just a single "error connecting". I want more detailed error recognition such that you can disambiguate an error connecting (network-level error) from an unsupported ENSNode error (application-level error) where the config returned by the connected ENSNode is either not fetched or deserialized successfully.
3. How is this app managing the latest indexing status projection / snapshot asynchronously in the background? I want to see a UI component for this in the example app similar to what's implemented for this inside the ENSAdmin UI (the little info icon).
4. How is this app navigating the vital question of ENS Namespace? There's 2 options for this:
   1. Option 1: The example app hardcodes its own ENS Namespace configuration (ex: mainnet) and then upon connecting to an ENSNode instance, verifies that its config matches the expected namespace. If it doesn't, the example app should refuse the connection.
   2. Option 2: The example app doesn't hardcode any ENS Namespace configuration. Instead, it waits to connect to an ENSNode instance and then makes use of the whatever namespace that ENSNode instance's config is using.

I believe that Option 1 is the direction that 99% of our customers will want to use as there are benefits to knowing the ENS namespace even before a successful connection to ENSNode is established. Therefore this example app should implement Option 1.

That means this example app should also read an ENS namespace as an environment variable that defaults to mainnet.

examples/ensnode-react-example/src/PrimaryNameView.tsx
```diff
@@ -0,0 +1,51 @@
+import type { Address, ChainId } from "enssdk";
+import { useState } from "react";
+
+import { usePrimaryName } from "@ensnode/ensnode-react";
+
+const DEFAULT_ADDRESS: Address = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045"; // vitalik.eth
+const MAINNET: ChainId = 1;
```
lightwalker-eth: This form shouldn't hardcode the chainId this way and also make reference to ENSIP-19 (which is multichain).

We should reference what we've already built for this idea before in ENSAdmin.

CleanShot 2026-05-04 at 14 20 16

packages/ensnode-react/example/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+A minimal React app demonstrating how to use `@ensnode/ensnode-react` to resolve
```
lightwalker-eth: Please see my other comments with key ideas that should be included here for the key goals of this example app:

1. The bigger comment I shared several lines above in this file.
2. The bigger comment I shared in `examples/ensnode-react-example/src/App.tsx`

I want to see this example app being an absolute A+ flawless bulletproof example of how an example app should manage the connection details with an ENSNode instance. This includes negotiating the connection itself, and the related validation and graceful / specialized error handling for that. It also includes maintaining an ongoing indexing status projection.

lightwalker-eth (commented):
@notrab Thanks for getting started on this PR. Noted how this was only a draft. Shared some suggestions with feedback.

packages/ensnode-react/example/README.md
notrab: @lightwalker-eth I'd suggest we move it from the root `examples` directory, but place it inside the `ensnode-react` folder. Having it inside a `docs` folder just feels weird.

Once merged, and we work to remove ensnode-react, we can move the example to the correct folder once later when we come to delete the package.

wdyt

packages/ensnode-react/example/src/PrimaryNameView.tsx
```diff
@@ -0,0 +1,147 @@
...
+    accelerate: true,
+  });
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    try {
+      setAddress(toNormalizedAddress(input.trim()));
+      setInputError(null);
+    } catch (err) {
+      setInputError(err instanceof Error ? err.message : "Invalid EVM address.");
```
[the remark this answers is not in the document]
notrab: To investiate; `FormEvent` deprecated?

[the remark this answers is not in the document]
notrab: To investiate; `FormEvent` deprecated?

pnpm-workspace.yaml
```diff
@@ -3,6 +3,8 @@ packages:
   - docs/*
   - examples/*
   - packages/*
+  # TODO: remove once we can move into `examples` folder enskit-example
+  - packages/ensnode-react/example
```
notrab, line 7: No way around this unless we changed `packages/*` to `packages/**`, but I preferred keeping the blast radius here small.

packages/ensnode-react/example/README.md
```diff
@@ -0,0 +1,49 @@
...
+A bulletproof reference for managing a frontend's connection to an ENSNode instance.
+Resolving a Mainnet Primary Name via `usePrimaryName` is just the payoff. The
+interesting part is everything that gates it:
+
+1. **Connection negotiation.** Wait for a healthy `useIndexingStatus` response before
+   mounting feature UI.
+2. **Disambiguated error handling.** Connection failures are classified as `network`
+   (fetch / DNS / CORS), `application` (bad response, or `responseCode === "error"`),
+   or `unsupported-namespace`. See
+   [`classify-connection-error.ts`](./src/lib/classify-connection-error.ts) and
```
notrab, line 19: need to better format these, leave it with me.

packages/ensnode-react/example/README.md
```diff
@@ -0,0 +1,49 @@
...
+> This app is internaldocumentation for refining `ensnode-react` before its functionality
+> is folded into `enskit`.
+
+## What it demonstrates
+
+A bulletproof reference for managing a frontend's connection to an ENSNode instance.
+Resolving a Mainnet Primary Name via `usePrimaryName` is just the payoff. The
+interesting part is everything that gates it:
+
+1. **Connection negotiation.** Wait for a healthy `useIndexingStatus` response before
```
notrab:
```suggestion
1. **Connection negotiation.** Wait for a healthy `useIndexingStatus` response
```

packages/ensnode-react/example/README.md
```diff
@@ -0,0 +1,49 @@
...
+> is folded into `enskit`.
+
+## What it demonstrates
+
+A bulletproof reference for managing a frontend's connection to an ENSNode instance.
+Resolving a Mainnet Primary Name via `usePrimaryName` is just the payoff. The
+interesting part is everything that gates it:
+
+1. **Connection negotiation.** Wait for a healthy `useIndexingStatus` response
+   mounting feature UI.
```
notrab:
```suggestion
```

PR #2000, opened 2026-04-27
cicd: add override for changeset version command

sevenzing:
## Summary

- Configure `changesets/action` in `.github/workflows/release.yml` to use `version: pnpm changeset:version`.

---

## Why

- Fix release CI error where `OpenAPI Spec Sync Check` fails after merging “Release New Version” because `ensapi-openapi.json` was not regenerated after package version bumps.

---

## Testing

- Not run locally in this change (workflow/script wiring only).

---


package.json
```diff
@@ -12,6 +12,7 @@
     "audit:osv": "osv-scanner scan source --lockfile pnpm-lock.yaml",
     "typecheck": "pnpm -r --parallel --aggregate-output typecheck",
     "changeset": "changeset",
+    "changeset:version": "changeset version && pnpm -F @docs/ensnode generate:openapi",
```
tk-o: What do you think about using the generic `generate:openapi` script defined in this file? The idea is to support future use cases where more apps in the monorepo can have their own OpenAPI spec.
```suggestion
    "changeset:version": "changeset version && pnpm generate:openapi",
```

tk-o (approved):
Nice use of `changesets/action` 🚀 Left minor comment, otherwise, LGTM!

package.json
sevenzing: agree

package.json
```diff
@@ -12,6 +12,7 @@
...
     "changeset:next": "changeset version --snapshot next",
```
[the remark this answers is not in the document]
sevenzing: okay. I dont really know if we actually use `release_snapshot.yml`, but I see no reason why not to include it. applied

PR #1965, opened 2026-04-20
feat: add docker directory with several docker composes

sevenzing:
## Summary

* Reorganized Docker Compose setup into `docker/` directory with purpose-specific compose files:
  * docker-compose.yml
  * docker-compose.devnet.yml
  * docker-compose.orchestrator.yml
* Extracted individual service definitions into `docker/services/*.yml` fragments using extends to eliminate duplication
* aslo updated `orchestrator.ts`  to use new docker compose, `package.json` with build scripts,  added `docker/README.md` with usage docs for developers

## Why

Resolves #1923 

---

Testing

* Manually tested docker-compose.devnet.yml stack 
* Manually tested docker-compose.yml stack
* `docker compose config` used to verify resolved output after restructure
* orchestrator.ts updated to point to new file path; CI will validate on merge

---

Notes for Reviewer (Optional)

* Maybe rename `postgres` service to `ensdb` ? not sure, probably ensdb is word for database schema, not for postgres service

* env_file is not defined in services/*.yml — each compose file declares its own env_file list explicitly to control  order and `required` param
* volumes: must be re-declared in every compose file
* Initially wanted to add docker-compose.yml to root using symlink, but dropped it. it broke `extends` path resolution when Docker Compose resolved paths relative to the symlink location rather than the real file :( 

---

Pre-Review Checklist (Blocking)

[x] This PR does not introduce significant changes and is low-risk to review quickly.
[x] Relevant changesets are included (or are not required)

PR #2004, opened 2026-04-28
chore: more docker improvements

sevenzing:
## Summary

- Updates Docker compose stack structure by renaming `postgres` service config to `ensdb` and refining service/environment wiring across compose files.
- Updates `docker/envs/.env.docker.devnet` and removes `environment` key from docker-composes
- Introduces `scripts/sync-docker-services-tags.mjs` (**AI GENERATED!!!**) and updates package/workflow wiring to keep Docker service tags synchronized for release flows.

---

## Why

Resolving comments after previous PR was merged:
https://github.com/namehash/ensnode/pull/1965#pullrequestreview-4178988172

---

## Testing

- `ENSNODE_TAG=123 docker compose -f docker/docker-compose.devnet.yml config` shows replaced data

---


scripts/sync-docker-services-tags.mjs
```diff
@@ -0,0 +1,63 @@
...
+  "docker/services/ensadmin.yml",
+  "docker/services/ensapi.yml",
+  "docker/services/ensindexer.yml",
+  "docker/services/ensrainbow.yml",
+];
+
+const semverRegex = /^\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
+
+async function getVersionFromEnsapi() {
+  const ensapiPackageJsonPath = resolve(rootDir, "apps/ensapi/package.json");
```
shrugs: add a note that because all of the package versions are fixed, this version applies to all of our services

scripts/sync-docker-services-tags.mjs
```diff
@@ -0,0 +1,63 @@
...
+  const packageJsonContent = await readFile(ensapiPackageJsonPath, "utf8");
```
shrugs: we might be able to just `import "../../apps/ensapi/package.json" as {type: 'json'};` or something

.github/workflows/release.yml
```diff
@@ -67,8 +67,8 @@ jobs:
         with:
           commit: "chore(release): version apps"
           title: "Release New Version"
-          version: pnpm changeset:version
-          publish: pnpm changeset-publish
+          version: pnpm changeset:version && pnpm release:postversion
```
shrugs: agreed with bots, move this `&& postversion` into the `changeset:version` pnpm script so that the workflow is a single script call; easier to reason about

docker/services/ensadmin.yml
```diff
@@ -1,7 +1,7 @@
 services:
   ensadmin:
     container_name: ensadmin
-    image: ghcr.io/namehash/ensnode/ensadmin:latest
+    image: ghcr.io/namehash/ensnode/ensadmin:${ENSNODE_TAG:-1.10.1}
```
shrugs: hmm, what if these were `${ENSNODE_TAG:-latest}` and the versioning script just updated the `ENSNODE_TAG` variable instead of each of these docker files?

scripts/sync-docker-services-tags.mjs
```diff
@@ -0,0 +1,63 @@
...
+
+    const updated = content.replace(replacePattern, replacement);
+    await writeFile(absolutePath, updated, "utf8");
+    console.log(`Updated ${relativePath} -> ${version}`);
+  }
+}
+
+async function main() {
+  const versionFromArg = process.argv[2];
+  const version = versionFromArg ?? (await getVersionFromEnsapi());
```
shrugs: Why don't we make this whole file simpler by only accepting the version in the argument and then as part of the pnpm script we can pass it like so?

```
cat ./apps/ensapi/package.json | jq '.version' | node scripts/sync-docker-services-tags.mjs
```

also switch this file to `.ts`, ensure there's a tsconfig extending the base config in this package, and execute with tsx. then introduce a script in _this_ package called `sync-docker-services-version` or `docker:version:sync` and call _that_ script from the monorepo package.json instead of leaking internal details about this package at the monorepo root

sevenzing: hm, we might do that but I'm almost sure that `jq` isn't installed by default in workflow container, but I see your direction, let me see what I can do here

docker/services/ensadmin.yml
sevenzing: yes, I also wanted to do this way, but it complicates things a lot:

docker compose need to know about file with `ENSNODE_TAG` and bad news are that we **cannot** specify path to this file in `docker-compose.yml`, only in command `docker compose --env-file <path>`. So if we will store it inside `docker/envs/.env.docker.common` all our commands will be like


```bash
docker compose --env-file docker/envs/.env.docker.common -f docker/docker-compose.yml up -d
```

It's already quite a lot, so I dont want ask developers to put `--env-file` argument

That leaves us with 2 ways: 
1. EITHER store version in `docker/.env` and push it to repo. `.env` file is automatically attached to `docker compose` command. 
2. OR use approach I suggested


Personally I dont like first approach, so decided to keep it complicated on our side and easiler for developers

.github/workflows/release.yml
sevenzing: yeah, I was like at a crossroads between these two decisions, honestly I think my approach is slightly better but I dont know best practice here, so I will just do as you suggested ;)

scripts/sync-docker-services-tags.mjs
sevenzing: decided to do it this way:

root package.json:
```json
"docker:tag:sync": "node ./scripts/sync-docker-services-tags.mjs \"$(pnpm -F ensapi -s version:current)\"",
```

ensapi package.json:

```json
"version:current": "node -p \"require('./package.json').version\""
```

docker/envs/.env.docker.common
```diff
@@ -1,16 +1,15 @@
 # Shared Docker Compose environment variables
 # Used by docker/docker-compose.yml and its variants
 
-
 # Postgres
```
lightwalker-eth, line 4: An idea but please attack this as I'm not confident about what is right.

What do you think about changing the defaults for these Postgres environment variables?

```
POSTGRES_DB=ensdb
POSTGRES_USER=ensdbuser
POSTGRES_PASSWORD=ensdbpass
```

I assume I'm missing something about making a change like this and that it would create some problem / complexity?

My only goal in making this comment is further strengthening the branding and narrative for ENSDb. Please ignore if it has problems.

docker/envs/.env.docker.common
```diff
@@ -1,16 +1,15 @@
...
 POSTGRES_PASSWORD=password
 
 # Internal service URLs (container-to-container)
-ENSDB_URL=postgresql://postgres:password@postgres:5432/postgres
+ENSDB_URL=postgresql://postgres:password@ensdb:5432/postgres
 ENSRAINBOW_URL=http://ensrainbow:3223
 
-# ENS Admin
+# ENSAdmin
 ENSADMIN_PUBLIC_URL=http://localhost:4173
```
lightwalker-eth, line 14: For these ENSAdmin environment variables, does it make sense that they are explicitly localhost or would it be nicer to use the special hostnames we're defining inside of docker such as `ensapi` or `ensadmin`.

Please ignore if some issue with this comment 👍

docker/envs/.env.docker.devnet
```diff
@@ -1,6 +1,22 @@
 # Default configuration for devnet docker-compose stack.
 # These values work out of the box — override by creating .env.docker.local.
 
+# ENSIndexer
 PLUGINS=subgraph,ensv2
+# ENSIndexer and ENSRainbow
```
lightwalker-eth: `NAMESPACE` should only be for ENSIndexer as I understand it

docker/envs/.env.docker.devnet
```diff
@@ -1,6 +1,22 @@
...
+# ENSIndexer and ENSApi
+ENSINDEXER_SCHEMA_NAME=docker_devnet_v1
+# ENSIndexer and ENSApi
+RPC_URL_1=http://devnet:8545
+# ENSIndexer and ENSRainbow
 LABEL_SET_VERSION=0
+# ENSIndexer and ENSRainbow
+LABEL_SET_ID=ens-test-env
+# ENSRainbow
 DB_SCHEMA_VERSION=3
```
lightwalker-eth, line 15: There should almost never be a need to customize this environment variable for ENSRainbow.

Not sure I'm understanding this file correctly. Is it intended customize non-default environment variables? If so, then I would suggest moving the configuration of `DB_SCHEMA_VERSION` just into the base template for ENSRainbow and not including it anywhere else.

Please ignore if not relevant.

docker/envs/.env.docker.devnet
```diff
@@ -1,6 +1,22 @@
...
+
+
+# TODO: in future we will migrate devnet to chain_id=1
```
lightwalker-eth: Maybe we can update this comment based on this Slack discussion? https://namehash.slack.com/archives/C08BMPFUUDD/p1777413276847409

docker/services/ensadmin.yml
lightwalker-eth: Happy to follow the approach here that you guys think will be best. This is outside my expertise. Some more general feedback:

1. If we decide to use this `ENSNODE_TAG` strategy I'm concerned about the terminology. "tag" has so many meanings in so many contexts. "tag" for what? Suggest to make this name more specific and precise and self-documenting.
2. How can we better document what's going on here?

docker/services/ensrainbow.yml
```diff
@@ -1,7 +1,7 @@
 services:
   ensrainbow:
     container_name: ensrainbow
-    image: ghcr.io/namehash/ensnode/ensrainbow:latest
+    image: ghcr.io/namehash/ensnode/ensrainbow:${ENSNODE_TAG:-1.10.1}
     build:
```
lightwalker-eth, line 5: Several lines below I note the line which reads:

`restart: unless-stopped`

Should we consider similar policies for any of the other services?

package.json
```diff
@@ -12,13 +12,15 @@
...
     "changeset:next": "changeset version --snapshot next && pnpm generate:openapi",
-    "changeset-publish": "changeset publish",
-    "changeset-publish:next": "changeset publish --no-git-tag --snapshot --tag next",
+    "changeset:publish": "changeset publish",
+    "changeset:publish:next": "changeset publish --no-git-tag --snapshot --tag next",
+    "release:postversion": "pnpm docker:tag:sync && pnpm generate:openapi",
     "packages:prepublish": "pnpm -r prepublish",
     "devnet": "docker compose -f docker/docker-compose.devnet.yml up devnet",
     "docker:build:ensnode": "pnpm run -w --parallel \"/^docker:build:.*/\"",
+    "docker:tag:sync": "node ./scripts/sync-docker-services-tags.mjs",
```
lightwalker-eth: Please see my related comment about refining the use of the word "tag" to something more explicit and precise. Let's not worry about making this command so concise. It's better to make it more clear and self-documenting 👍 

Ex: Do we change "tag" here to something like "version-tag"? Advice appreciated.

scripts/sync-docker-services-tags.mjs
```diff
@@ -0,0 +1,63 @@
...
+const serviceFiles = [
+  "docker/services/ensadmin.yml",
+  "docker/services/ensapi.yml",
+  "docker/services/ensindexer.yml",
+  "docker/services/ensrainbow.yml",
+];
+
+const semverRegex = /^\d+\.\d+\.\d+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
+
+async function getVersionFromEnsapi() {
```
lightwalker-eth:
```suggestion
async function getVersionFromEnsApi() {
```

scripts/sync-docker-services-tags.mjs
```diff
@@ -0,0 +1,63 @@
+import { readFile, writeFile } from "node:fs/promises";
+import { resolve } from "node:path";
+
```
lightwalker-eth, line 11: Can we add some comments up at the top of the file explaining what the goal and responsibility of this file is? If there's some relevant background info that may also be interesting.

Appreciate your advice 👍

scripts/sync-docker-services-tags.mjs
```diff
@@ -0,0 +1,63 @@
...
+  return packageJson.version;
+}
+
+function validateVersion(version) {
+  if (!semverRegex.test(version)) {
+    throw new Error(`Invalid version "${version}". Expected SemVer-like value such as 1.10.1`);
+  }
+}
+
+async function updateServiceDefaultTag(version) {
```
lightwalker-eth: Please see related feedback suggesting we refine the "tag" terminology to be more explicit about exactly what type of tag we're talking about.

lightwalker-eth (approved):
@sevenzing Looks nice! Shared a few suggestions please feel welcome to take the lead to merge when you feel good 👍

docker/envs/.env.docker.common
sevenzing: tricky part that this is public url, it's used by user in ensadmin and user might only access it with localhost

docker/envs/.env.docker.devnet
sevenzing: thanks!

docker/envs/.env.docker.devnet
sevenzing: this is base template of devnet and this file should NOT be editable. `DB_SCHEMA_VERSION` is required variable. Also we have requirement to start devnet with `.env.docker.local` file, that's why we need this variable

docker/envs/.env.docker.common
sevenzing: hm, I think it wont create any complexity since this is in docker env.
it might break existing users who uses ensnode via docker compose but I believe it's okay for now

sevenzing: oops, tried it locally and it might create complexity to migrate to different password. for now, I would prefer to keep `postgres/password/postgres` defaults for local testing. It’s a common practice in local/dev setups

docker/services/ensrainbow.yml
sevenzing: hm, this is legacy from prev docker compose.

I think best strategy here is to specify `restart` only for default docker compose and DONT do it for devnet. for local development it's better to fail service rather than constantly restart it

docker/services/ensadmin.yml
sevenzing: decided to use `ENSNODE_VERSION`. I introduce documentation for that in next PR #2005

PR #2014, opened 2026-04-29
feat(ensv2): HCA-aware Event.sender + filter + docs

### LINKED ISSUE #1813: Omnigraph: Smart Account Handling
ENSv2 has first-party support for smart accounts, which affects usage of the native `msg.sender`.

the `sender` (and ERC1155 `operator`) arguments used in ENSv2 events _is_ already smart-account-adjusted (thank god) so we DON'T need to do anything special to support smart accounts.

we do use the transaction sender within the `Event` model, which requires additional thought, and depends on how the frontends wish to use the `Account.events` or `*.events(where: { from: Address })` filter — the indexed value, currently, is the actual sender of the transaction (i.e. could be anything), NOT the smart account's address.

So we may need to provide an optional `sender` override in the `ensure*Event` helpers such that events created in the context of an ENSv2 event that has a smart account are appropriately attributed.

>   One nuance: the NameRegistered event's sender (line 179) is explicitly commented as "the registrar, not the registrant" -- meaning when registration goes through ETHRegistrar, the sender will be the registrar contract's resolved _msgSender(), not the end user. The end user appears in the `owner` parameter instead.

shrugs:
closes #1813

primary review surface:

1. **`Account.events` semantics flip** (`apps/ensapi/src/omnigraph-api/schema/account.ts`): now filters by `sender` (HCA-aware) instead of `from` (`tx.from`). this is a quiet behavior change — anyone relying on the old "all txs this EOA submitted" semantics will see different results once HCA traffic exists.
2. **handler sender plumbing** (`apps/ensindexer/src/plugins/ensv2/handlers/ensv2/*`, `apps/ensindexer/src/lib/ensv2/event-db-helpers.ts`): the override path is what makes `Event.sender` HCA-aware. confirm each ENSv2 ensureEvent call site passes the right HCA-candidate arg (registrant / unregistrant / sender / operator) and not, e.g., the wrong field.
3. **schema column + index**: new `events.sender` column (NOT NULL) and `events_sender_index`. requires reindex.

---

## Problem & Motivation

ENSv2 contracts route writes through HCAs (Hybrid Custody Accounts / smart accounts) via `IHCAFactoryBasic`. on-chain events emit an explicit `sender`/`owner`/`account` argument that is HCA-aware — the HCA address rather than the controlling EOA. previously the indexed `events` table only stored `tx.from`, which is the EOA/relayer that submitted the transaction and is never HCA-aware. omnigraph consumers asking "which events did this account trigger" had no way to query by the HCA actor.

a `sender` column already existed on the events table and `ensureEvent` already accepted an optional override, but the handlers weren't passing it through, the GraphQL layer didn't expose it, no filter existed, and the schema didn't document HCA semantics anywhere.

---

## What Changed (Concrete)

1. `packages/ensdb-sdk/src/ensindexer-abstract/ensv2.schema.ts`: documented HCA-awareness on `event.sender`, `event.from` (explicit "never HCA-aware"), `domain.ownerId`, `registration.registrantId`/`unregistrantId`, `permissionsUser.user`. added `bySender` index on the events table.
2. `apps/ensindexer/src/lib/ensv2/event-db-helpers.ts`: `ensureEvent(context, event, sender?)` already accepted an HCA-aware override; verified handler audit covered all call sites.
3. ENSv2 handlers (`ETHRegistrar.ts`, `ENSv2Registry.ts`): pass HCA-candidate arg as the sender override on `NameRegistered` (`owner`), `LabelRegistered`/`Reserved` (`sender`), `LabelUnregistered` (`sender`), `ExpiryUpdated` (`sender`), `SubregistryUpdated` (`sender`), `TransferSingle`/`Batch` (`operator`). gaps (no actor in args, falls back to `tx.from`): `EACRolesChanged` granter, `NameRenewed`, `TokenRegenerated`, all shared `Resolver:*` events. accepted.
4. `apps/ensapi/src/omnigraph-api/schema/event.ts`: exposed `Event.sender` field; updated `Event.from` description to call out non-HCA semantics; added `sender` filter to `EventsWhereInput`; updated `from` filter description; updated `AccountEventsWhereInput` to mark `sender` as implied.
5. `apps/ensapi/src/omnigraph-api/lib/find-events/find-events-resolver.ts`: extended `EventsWhere` interface and `eventsWhereConditions` to apply the new `sender` filter.
6. `apps/ensapi/src/omnigraph-api/schema/account.ts`: `Account.events` now resolves with `sender = parent.id` (HCA-aware) instead of `from = parent.id`.
7. omnigraph schema doc updates: `Domain.owner` (polymorphic ENSv1/v2 wording), `Registration.registrant`/`unregistrant` (HCA-aware in v2), `PermissionsUser.user`, `RegistryPermissionsUser.user`, `ResolverPermissionsUser.user`.

---

## Design & Planning

spec'd inline with @shrugs over two passes (initial proposal + re-proposal after the user clarified that `event.sender` already existed as an override mechanism and that they wanted it surfaced via filter). no separate design doc.

alternatives considered:
- exposing only the schema column and skipping the GraphQL filter — rejected because consumers need to query by HCA actor for this to be useful.
- keeping `Account.events` on `from` and adding a parallel `Account.txs` for the `tx.from` view — rejected as YAGNI; the HCA-aware `sender` is the more useful default and `from` filter remains available via `EventsWhereInput`.
- reverse: rename `event.sender` → `event.actor` to avoid confusion with `tx.from` colloquially being "the sender" — rejected to minimize churn; docs clarify the distinction.

- Planning artifacts: conversation transcript only.
- Reviewed / approved by: @shrugs (live).

---

## Self-Review

reviewed every changed file end-to-end after implementation.

- Bugs caught: none new in this pass; the override plumbing was already in place from prior commits on this branch.
- Logic simplified: docstring template unified to `..., the HCA account address, if used.` for non-polymorphic fields and the explicit ENSv1/ENSv2 two-clause wording for `Domain.owner` and `Registration.registrant`/`unregistrant`.
- Naming / terminology improved: `Event.from` description explicitly disclaims HCA-awareness; `Event.sender` describes fallback semantics; filter docs cross-reference each other.
- Dead or unnecessary code removed: none — this PR is additive.

---

## Cross-Codebase Alignment

- Search terms used: `HCA`, `IHCAFactoryBasic`, `ensureEvent`, `event.sender`, `event.from`, `_msgSender`, `sender:`.
- Reviewed but unchanged: ENSv1 handlers in `apps/ensindexer/src/plugins/ensv2/handlers/ensv1/` (intentionally not HCA-aware per project decision); `ENSv1Domain.rootRegistryOwner`; `Event.address` and `Event.to`; `AccountId.address`; registry/registrar contract address columns.
- Deferred alignment: no audit of EACRolesChanged granter (gap accepted; would require an upstream contract change to emit the granter).

---

- Public APIs affected: GraphQL schema gains `Event.sender: Address!` field and a `sender` filter on `EventsWhereInput`. `Account.events` semantics changes from `tx.from`-based to `sender`-based filtering.
- Docs updated: schema doc comments + GraphQL field descriptions throughout. no separate docs site changes (descriptions ship via the GraphQL schema).
- Naming decisions worth calling out: `sender` is overloaded — `tx.from` is colloquially "the sender" but in this schema "sender" is reserved for the HCA-aware actor. descriptions on both `Event.from` and `Event.sender` explicitly cross-link to disambiguate.

---

## Testing Evidence

- Testing performed:
  - `pnpm typecheck` clean across all 24 workspace projects.
  - devnet (`feat/hca-aware-events` branch) reindexed against the running anvil; verified the `events.sender` column is populated for all 762 indexed events.
  - 24 events have `sender != from` — exactly the Registry events routed through ETHRegistrar (`LabelRegistered`, `ExpiryUpdated`, `TransferSingle`). spot-checked tx `0x1c8a1a5c…b0` via `cast tx`: `tx.from = 0x7099…79C8` (user EOA), `tx.to = 0x4C4a…5584` (ETHRegistrar), event emitted from registry `0x8f86…fe4cf`, indexed `event.sender = 0x4C4a…5584` (ETHRegistrar — the registry's `_msgSender()`).
  - confirmed `events_sender_index` exists on the table after reindex.
- Known gaps:
  - devnet test fixtures don't actually deploy + use an HCA, so we can't differentiate "HCA actor" from "intermediary contract caller". the override mechanism is identical for both paths, so the user→HCA→Registry case will produce the HCA address by the same code path; not a separate code path under test.
  - no integration test covers `Account.events` returning HCA-aware results.
- What reviewers have to reason about manually: that each handler's chosen override arg is the right HCA-candidate (e.g., `TransferSingle` uses `operator` rather than `from`/`to`).

---

## Scope Reductions

- Follow-ups:
  - extend devnet fixtures to deploy an HCA via `HCAFactory` and route at least one register/transfer through it for a true HCA-vs-EOA differentiation test.
  - integration test for `Account.events` HCA-aware filtering.
  - upstream contract change to emit the granter address on `EACRolesChanged`.
- Why deferred: each is a stand-alone improvement; this PR's surface is already the doc + plumbing + filter triad and shouldn't grow.

---

## Risk Analysis

- Risk areas:
  - `Account.events` behavior change is silent. callers expecting old semantics will see different results once HCA traffic exists. today the devnet has no HCAs so the practical delta is zero, but staging/prod consumers should be informed.
  - new NOT NULL `events.sender` column requires reindex. existing deployments will need to drop and rebuild the events table (or run a backfill that defaults `sender` to `from`).
  - new index `events_sender_index` is a write-path cost; negligible relative to existing indexes on the same table.
- Mitigations or rollback options: revert is clean — schema column and index can be dropped, `Account.events` resolver reverts to `from = parent.id`, GraphQL schema removes one field and one filter argument.
- Named owner if this causes problems: @shrugs.

---


shrugs:
@greptile review

shrugs:
@greptile review

packages/ensdb-sdk/src/ensindexer-abstract/ensv2.schema.ts
```diff
@@ -96,6 +96,9 @@ export const event = onchainTable(
     // Ponder's event.id
     id: t.text().primaryKey(),
 
+    // The HCA account address if used, otherwise Transaction.from.
```
lightwalker-eth, line 99: @shrugs Docs for this db schema have been added to https://ensnode.io/ensdb/concepts/database-schemas#ensv2

Could you apply the changes from this PR into these docs? Thanks!

Later we'll find a more elegant strategy to keep the docs in sync. For now we were just looking for a quick fix so I asked Tomasz to just ship the ENSDb docs and we can optimize processes later.

lightwalker-eth (approved):
@shrugs Nice 🚀 1 small comment please merge when ready 👍

PR #2038, opened 2026-05-01
Transitioning to use `AccountId` as the referrer identifier

### LINKED ISSUE #1956: Enhance data model for referrers
Currently, our data models in "ENSAnalytics" / our referral program logic uses only a `NormalizedAddress` to identify a referrer.

It will be more mature for us to update every instance of this approach so that instead we use `AccountId` from the `enssdk` package.

I imagine there may be cases in our existing logic where you're currently using a `NormalizedAddress` as some form of a key in a data structure to uniquely identify each referrer. Based on how TypeScript works, it won't be appropriate to use an `AccountId` like this because it will determine equality by object reference rather than object value. To fix this, if you need to generate a key, you can do it via the `stringifyAccountId` helper function which will convert an `AccountId` to a `string` which is then safe to use as a key which will be compared by value.

You can also make use of `parseAccountId` in the `ensnode-sdk` package to attempt to convert back from a `string` to a valid `AccountId`.

The `accountIdEqual` helper function in the `ensnode-sdk` package may also be useful for you.

These updates should also be applied to our API routes. For example, the API route for requesting details about a specific referrer should be enhanced so that instead of taking as input only an address it takes as input a query param identifying the full `AccountId` of the referrer through a string that is formatted as a CAIP-10 account Id as can be produced through our `stringifyAccountId` helper function.

These updates should also be applied to the client helper functions in `packages/ens-referrals/src/client.ts` so that wherever the client helper functions are currently taking just an address as a param they should instead take an `AccountId`. The logic to convert from the provided `AccountId` into it's serialized representation via `stringifyAccountId` should happen internally within the client's logic.

Goader:
# Transitioning to use `AccountId` as the referrer identifier

closes: #1956 

## Summary

- Transitioned to use `AccountId` or `AccountIdString` for keys
- Fixed the bug in the Zod schema in `ensnode-sdk`

---

## Why

- See: #1956 

---

## Testing

- New tests

---

## Notes for Reviewer

- The bug in `ensnode-sdk`
- ***Referrer*** column in the accounting CSV is now stringified `AccountId` (e.g., `eip155:11155111:0xcfa4f8192ad39d1ee09f473e88e79d267e09ddca`)

---


Goader:
@greptile

Goader:
@coderabbit do the review

PR #2039, opened 2026-05-01
fix: use entrypoint in orchestrator instead of re-implementing

shrugs:
simple refactor to DRY up some code


shrugs:
merging as minor, clearly good improvement

PR #2041, opened 2026-05-01
ci: extract promote_vercel_deployment action, run for ensadmin + example app

shrugs:
## Summary

- extracted `promote_ensadmin.sh` into a reusable composite action at `.github/actions/promote_vercel_deployment/`, parameterized by `vercel_project_id` and `vercel_team_slug`
- `deploy_switch_ensnode_environment` now invokes the action twice: ensadmin (`prj_nKcHTO12hq9kcgascQMq4xokRhwp`) and example app (`prj_Ux80vjjNyDfXMNvPEQo4zlx6Eq6D`)
- deleted the now-unused `.github/scripts/promote_ensadmin.sh`

---

## Why

- the example app's vercel deployment should be version-locked to the active ensnode environment, same as ensadmin
- pulled the script out of `scripts/` and into a composite action so we don't fork per project

---

## Testing

- not tested live — workflow only runs via `workflow_dispatch` against real green/blue environments
- validated by inspection: the script is unchanged apart from reading `VERCEL_PROJECT_ID` / `VERCEL_TEAM_SLUG` from env instead of hardcoding, and the action wires the same env vars through that the workflow already exports

---

## Notes for Reviewer

- the example app step runs after the ensadmin step in the same job. if the example app has no vercel deployment matching the active ensindexer commit sha, the workflow step will fail (and the slack notification won't fire). ensadmin already has this exact behavior — we're just extending the same coupling to a second project.
- the script now lives inside the action directory (`.github/actions/promote_vercel_deployment/promote_vercel_deployment.sh`) instead of `.github/scripts/`, so the action is self-contained.

---

## Checklist

### DIFF
--- a/.github/actions/promote_vercel_deployment/action.yml
+++ b/.github/actions/promote_vercel_deployment/action.yml
@@ -1,0 +1,40 @@
+name: Promote Vercel Deployment
+description: Promotes the Vercel production deployment whose commit SHA matches
+  the active ENSIndexer image in the given Railway environment. Ensures exact
+  version matching between the active ENSNode and the Vercel project.
+
+inputs:
+  vercel_team_slug:
+    description: "Vercel team slug"
+    required: true
+
+  vercel_project_id:
+    description: "Vercel project ID to promote"
+    required: true
+
+  vercel_token:
+    description: "Vercel API token"
+    required: true
+
+  railway_token:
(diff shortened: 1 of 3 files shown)


shrugs:
merging to prioritize ensv2 demo velocity

PR #2054, opened 2026-05-05
feat: initialize ENSNameHealer contract

sevenzing:
# Lite PR

## Summary

- Add **ENSNameHealer** — minimal UUPS contract that emits `LabelPublished` for a single onchain publisher (ENSRainbowBeam); indexers heal labels from events.
- Add Foundry project under `contracts/ens-name-healer/` with deploy scripts (standard + CREATE2 vanity via `mine-create2-salt.sh`), tests, and docs
- Add **contracts CI** workflow (discover Foundry projects → fmt / build / test / gas snapshot) and repo hygiene (submodules, Biome `!!` ignores for Foundry artifacts, Prettier ignore for `lib/`)

---

## Testing

- `forge test` — 14 tests (unit + CREATE2 deploy)
- `forge snapshot --check` in contracts CI
- `forge fmt --check`, `forge build --sizes`
- Sepolia: CREATE2 vanity deploy (`0x1abe15…` prefix) + `publishLabel` smoke test

---


contracts/ens-label-healer/.env.example
```diff
@@ -0,0 +1,15 @@
...
+# If omitted, grant it manually after deployment via grantRole.
+SUBMITTER_ADDRESS=
+
+# Private key of the deployer account (required for deployment).
+# Must hold enough ETH to cover gas on the target network.
+DEPLOYER_PRIVATE_KEY=
+
+# RPC URLs for each network.
+SEPOLIA_RPC_URL=
+MAINNET_RPC_URL=
```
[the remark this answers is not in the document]
sevenzing: yes, forgot to remove `--etherscan-api-key $ETHERSCAN_API_KEY` from docs. etherscan is shit and verification doesnt work here

contracts/ens-name-healer/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+# Read the emitted event
+cast logs --rpc-url http://localhost:8545 \
+  --address $PROXY \
+  "NameHealed(bytes32,string,address)"
+
+# Check if a name is already healed
+cast call $PROXY "healed(bytes32)(bool)" \
+  $(cast keccak $(cast abi-encode "f(string)" "vitalik.eth")) \
+  --rpc-url http://localhost:8545
```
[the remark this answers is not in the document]
sevenzing: yeah, forgot to remove it

contracts/ens-name-healer/README.md
```diff
@@ -0,0 +1,161 @@
...
+```bash
+source .env && forge script script/Deploy.s.sol \
+  --rpc-url $SEPOLIA_RPC_URL \
+  --private-key $DEPLOYER_PRIVATE_KEY \
+  --broadcast \
+  --verify \
+  --verifier sourcify
+```
+
+`--verify` uploads the source to Etherscan so you can interact with the contract via the web UI.
```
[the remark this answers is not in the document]
sevenzing: thanks

.github/workflows/contracts_ci.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Find Foundry projects
+        id: find
+        run: |
+          dirs=$(find contracts -name "foundry.toml" -not -path "*/lib/*" -exec dirname {} \; | sort | jq -R -s -c 'split("\n") | map(select(length > 0))')
+          echo "contracts=$dirs" >> $GITHUB_OUTPUT
+
+  ci:
+    name: "CI (${{ matrix.contract }})"
+    needs: discover
+    runs-on: ubuntu-latest
```
shrugs: we use blacksmith runners elsewhere, perhaps preferable?

contracts/ens-name-healer/script/Deploy.s.sol
```diff
@@ -0,0 +1,43 @@
...
+        address admin = vm.envAddress("ADMIN_ADDRESS");
+        address submitter = vm.envOr("SUBMITTER_ADDRESS", address(0));
+
+        vm.startBroadcast();
+
+        ENSNameHealer impl = new ENSNameHealer();
+        bytes memory initData = abi.encodeCall(ENSNameHealer.initialize, (admin));
+        ENSNameHealer proxy = ENSNameHealer(address(new ERC1967Proxy(address(impl), initData)));
+
+        if (submitter != address(0)) {
```
shrugs: i would just split this into two scripts, one for Deploy one for grant

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
+// SPDX-License-Identifier: MIT
```
shrugs, line 1: noting that i'm not sure about the purpose of storing the namehashes on-chain, seems expensive, and then adding the override ability seems to give us a more fragmented system, more complex and less reliable than either storing every namehash or not storing any at all

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Compute the ENS namehash for a dot-separated name string.
+    ///
+    /// Implements EIP-137: iterate labels right-to-left, computing
+    ///   node = keccak256(abi.encodePacked(node, keccak256(label)))
+    /// starting from node = bytes32(0) (the namehash of the empty string).
+    ///
+    /// @param name Dot-separated ENS name (e.g. "vitalik.eth").
+    /// @return node The resulting bytes32 namehash.
+    function _namehash(string memory name) internal pure returns (bytes32 node) {
```
shrugs: surely we should use the NameCoder from ens-contracts instead of re-implementing

contracts/ens-name-healer/test/ENSNameHealer.fuzz.t.sol
```diff
@@ -0,0 +1,233 @@
...
+    function namehash(string memory name) external pure returns (bytes32) {
+        return _namehash(name);
+    }
+}
+
+/// @dev Independent reference implementation of ENS namehash used to validate
+///      the contract's on-chain implementation against a second code path.
+///      Uses a recursive approach vs the contract's iterative approach.
+library ReferenceNamehash {
+    function compute(string memory name) internal pure returns (bytes32 node) {
```
shrugs: this reference should _definitely_ be NameCoder, not a custom implementation

contracts/ens-label-healer/.env.example
```diff
@@ -0,0 +1,15 @@
...
+# Address that receives SUBMITTER_ROLE during deployment (optional).
+# If omitted, grant it manually after deployment via grantRole.
+SUBMITTER_ADDRESS=
+
+# Private key of the deployer account (required for deployment).
+# Must hold enough ETH to cover gas on the target network.
+DEPLOYER_PRIVATE_KEY=
+
+# RPC URLs for each network.
+SEPOLIA_RPC_URL=
```
shrugs: ensure that we're using the foundry networks config, also include an etherscan api key for contract verification

contracts/ens-name-healer/src/ENSNameHealer.sol
lightwalker-eth: Agreed. I don't understand the goal of storing namehashes? In my mind that should be cut.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+
+import "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/utils/PausableUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+
+/// @title ENSNameHealer
+/// @notice On-chain oracle for healing unresolvable ENS names.
+///
+/// ENSv1's old registry contract emits events containing only a namehash — no
```
lightwalker-eth: No. They emit events containing only a labelhash. Please review the related contract for details.

Additionally, this isn't just the "old registry contract". This is true for both the original ENS Registry and the updated ENS Registry With Fallback.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+/// label string. This contract allows trusted submitters to publish the
+/// human-readable name that corresponds to a given namehash, emitting a
+/// `NameHealed` event that any indexer (such as ENSNode) can consume to
+/// recover the readable form of an otherwise unresolvable name.
```
lightwalker-eth: The words "name" and "namehash" should not appear anywhere in this PR. Nothing we are working on here has anything to do with names or namehashes. Everything is about labels and labelhashes.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Role for addresses permitted to submit names for healing.
+    /// Intended for the `ens-rainbow-beam` off-chain service hot key.
+    bytes32 public constant SUBMITTER_ROLE = keccak256("SUBMITTER_ROLE");
+
+    // ── State ──────────────────────────────────────────────────────────────
+
+    /// Maps ENS namehashes to whether they have been healed.
+    /// Writes are skipped when `healedStorageDisabled` is true.
+    mapping(bytes32 => bool) public healed;
```
lightwalker-eth: What's the goal for storing this info onchain? In my mind this should not be done. The responsibility is on the permissioned submitters not to submit duplicate labels. We want to optimize the gas cost for each label healing that is performed. Therefore in my mind all that should happen is to emit an event and nothing else.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    /// When true the deduplication check and storage writes in `submit` are
+    /// both bypassed. Controlled by `DEFAULT_ADMIN_ROLE`.
+    ///
+    /// Intended as an operational escape hatch: if on-chain storage costs
+    /// become prohibitive at scale, an admin can disable the mapping without
+    /// upgrading the contract. Downstream consumers (e.g. the ENSNode
+    /// indexer) must be prepared to handle duplicate `NameHealed` events.
+    bool public healedStorageDisabled;
```
lightwalker-eth: We should remove this. Duplicate events are fine. The permissioned submitters will work to avoid duplications but if one happens its no big deal.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+    /// @param submitter The address that submitted the name.
+    event NameHealed(bytes32 indexed namehash, string name, address indexed submitter);
+
+    /// Emitted when `healedStorageDisabled` is toggled.
+    event HealedStorageDisabledSet(bool disabled);
+
+    // ── Errors ─────────────────────────────────────────────────────────────
+
+    /// Thrown when an empty string is submitted.
+    error InvalidName();
```
lightwalker-eth: There should be no error for an empty label. The empty label is still a label, it is just not a valid ENS normalized label.

All label inputs to this contract should be a `LiteralLabel` based on our definition of this idea in `enssdk`. They are not `InterpretedLabels`. A `LiteralLabel` is any possible string, including the empty string, strings with null bytes inside, etc. etc.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Thrown when `submit` is called for a name whose namehash is already
+    /// recorded in `healed` and `healedStorageDisabled` is false.
+    error AlreadyHealed(bytes32 namehash);
```
lightwalker-eth: We should remove this.

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Submit a name for healing, bypassing the already-healed deduplication
+    /// check.
+    ///
+    /// Use during indexer recovery, after toggling `healedStorageDisabled`, or
+    /// whenever an event must be re-emitted regardless of prior state.
+    ///
+    /// @param name The full dot-separated ENS name (e.g. "vitalik.eth").
+    function forceResubmit(string calldata name) external onlyRole(SUBMITTER_ROLE) whenNotPaused {
```
lightwalker-eth: We can remove this

contracts/ens-name-healer/src/ENSNameHealer.sol
```diff
@@ -0,0 +1,187 @@
...
+        for (uint256 i = 0; i < names.length; i++) {
+            _processName(names[i], false);
+        }
+    }
+
+    /// Submit multiple names for healing, bypassing the already-healed
+    /// deduplication check for all names in the batch.
+    ///
+    /// @param names Array of full dot-separated ENS names.
+    function forceResubmitBatch(string[] calldata names) external onlyRole(SUBMITTER_ROLE) whenNotPaused {
```
lightwalker-eth: We can remove this

contracts/ens-name-healer/src/ENSNameHealer.sol
lightwalker-eth: There should be zero namehash operations in this PR.

Likewise, there should be zero labelhash implementations in this PR too.

This contract needs to be a lot more simple. It receives a label, it emits a label, done. Indexers are perfectly capable of calculating the labelhash of the emitted label on their own.

contracts/README.md
```diff
@@ -0,0 +1,33 @@
+# ENSNode Contracts
+
+Solidity smart contracts used in the ENSNode ecosystem. Each contract lives in its own subdirectory as a standalone [Foundry](https://book.getfoundry.sh/) project.
+
+## Contracts
+
+| Directory | Contract | Purpose |
+|---|---|---|
+| `ens-name-healer/` | `ENSNameHealer` | On-chain oracle for healing unresolvable ENS names |
```
lightwalker-eth: Suggest we change the name because in my mind this contract does not heal names and therefore isn't a name healer.

The role of this contract is to "emit rainbow beams" to ENS indexers.

Each "emitted rainbow beam" symbolically represents a single label that indexers would then add to a dynamically generated rainbow table based on the set of all events indexed from this contract.

Each time an indexer indexes one of these events, it should check its indexed data for any labels formatted as encoded labelhashes of the labelhash generated from the indexed labels. For any found, the indexer should then transform those labels into the healed label. A special detail here is if the healed label is not a normalized label then the indexer must store the healed label as an encoded labelhash of the healed label (which is unnormalized). Therefore I found a problem in our current logic while writing this out. Do you see it? ENSRainbowBeam needs a smarter strategy for how it determines if it should heal a label again or not. It could get in a loop of resubmitting an unnormalized label.. Ah.. wait! Here's an easy fix for now: The ENSRainbowBeam app should validate that all labels submitted to it are normalized. Any that are not normalized should be discarded for now -- later we can expand our indexed data model to track unnormalized labels too, such that we could tell the difference between an unknown label vs an unnormalized label. Today we can't. Therefore this problem. Can you please take the lead to notify @djstrong of this important change needed to the ENSRainbowBeam app?

Each permissioned submitter has the responsibility to verify that each label submitted would result in 1+ already indexed labels that are currently unknown transitioning to becoming healed (known). This contract is not responsible for that verification. The permissioned submitters are. Instead, this contract is optimized for minimum gas usage per emitted label, over time we may be submitting a lot, and we are the ones paying for all the gas.

The actual name healing process happens inside the indexer when it heals an unknown label into a known label.

Therefore suggest to call this contract something like `ens-rainbow-beam-emitter`

Appreciate your help to transfer these ideas into nice comments where relevant.

contracts/ens-name-healer/.env.example
```diff
@@ -0,0 +1,15 @@
+# Address that receives DEFAULT_ADMIN_ROLE on the proxy (required).
+# Use a Gnosis Safe multisig on Sepolia/mainnet.
```
lightwalker-eth: Why? And why only Sepolia/mainnet?

lightwalker-eth (commented):
@sevenzing Appreciate your updates here. We should take a more simplified strategy for this contract. Please see comments for details 👍

contracts/README.md
sevenzing: `ens-rainbow-beam-emitter` 😄 

decided to keep `ENSLabelHealer`

contracts/ens-label-healer/src/ENSLabelHealer.sol
```diff
@@ -0,0 +1,49 @@
...
+
+    /// @param admin Address to assign `DEFAULT_ADMIN_ROLE`.
+    function initialize(address admin) external initializer {
+        __AccessControl_init();
+        __Pausable_init();
+        __UUPSUpgradeable_init();
+        _grantRole(DEFAULT_ADMIN_ROLE, admin);
+    }
+
+    function pause() external onlyRole(DEFAULT_ADMIN_ROLE) {
```
shrugs: I like the pausability in the abstract, but I can't think of a concrete reason we'd need it in practice; wdyt?

shrugs (approved):
looks great to me!

contracts/ens-label-healer/src/ENSLabelHealer.sol
sevenzing: for example we could face an error in ENS-Rainbow-Emitter. for example it could send duplicate events. in theory we dont want to turn off ens-rainbow-emiiter to accept labels from users, but want to pause emitting events.

this is just one idea, something else could happend. adding this function would only increase gas during **deploy**, no runtime (emitting LabelHealed) gas increase

contracts/README.md
```diff
@@ -0,0 +1,9 @@
+# ENSNode Contracts
+
+Solidity smart contracts used in the ENSNode ecosystem. Each contract lives in its own subdirectory as a standalone [Foundry](https://book.getfoundry.sh/) project.
```
lightwalker-eth:
```suggestion
Solidity smart contracts created as an explicit part of the ENSNode platform. Each contract lives in its own subdirectory as a standalone [Foundry](https://book.getfoundry.sh/) project.

Note that this is not a directory of all ENS contracts. For general directories of ENS contracts see:
1. The [datasources package in the ENSNode monorepo](https://github.com/namehash/ensnode/tree/main/packages/datasources).
2. The [ens-contracts repo](https://github.com/ensdomains/ens-contracts).
3. The [contracts-v2 repo](https://github.com/ensdomains/contracts-v2).
```

contracts/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Contracts
+
+| Directory           | Contract         | Purpose                                                       |
+| ------------------- | ---------------- | ------------------------------------------------------------- |
+| `ens-label-healer/` | `ENSLabelHealer` | Permissioned on-chain label emitter for unresolved ENS labels |
```
lightwalker-eth:
```suggestion
| `ens-name-healer/` | `ENSNameHealer` | Enable healed labels in ENS names to automatically propagate to all decentralized ENSNode instances |
```

Feedback:

1. ENSNameHealer is better branding than ENSLabelHealer. As discussed we can name this according to it's ultimately value proposition, rather than as a technical description of how it is implemented. Please apply all the related renamings.
2. Updated purpose to focus more on the value proposition.

contracts/ens-label-healer/src/ENSLabelHealer.sol
```diff
@@ -0,0 +1,64 @@
...
+///
+/// @dev Upgradeable via UUPS. Storage layout must be preserved across upgrades.
+contract ENSLabelHealer is
+    Initializable,
+    OwnableUpgradeable,
+    AccessControlUpgradeable,
+    PausableUpgradeable,
+    UUPSUpgradeable
+{
+    bytes32 public constant SUBMITTER_ROLE = keccak256("SUBMITTER_ROLE");
```
lightwalker-eth: I note how the current approach here defines the submitter as a role that any number of accounts can be granted.

We don't ever want there to be more than 1 submitter though. That increases the probability of race conditions and coordination problems across submitters. Allowing more than 1 submitter is a source of problems.

The submitter should not be a role. It should be a single address.

Likewise, we should remove the idea of this contract being pausable. If we want to pause the contract for some reason, we will do this by changing the address of the sole-submitter to the null address.

We should therefore also change `grantSubmitter` / `revokeSubmitter` (and their associated scripts). The operation here should be to `setSubmitter` and which is what we would want scripts to help with.

contracts/ens-label-healer/src/ENSLabelHealer.sol
```diff
@@ -0,0 +1,64 @@
...
+import "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/utils/PausableUpgradeable.sol";
+
+/// @title ENSLabelHealer
+/// @notice Permissioned on-chain emitter for unresolved ENS labels.
+///
+/// Some ENS registry contracts emit events containing only a labelhash.
```
lightwalker-eth: We've previously invested in writing up docs with related info. There's an opportunity to better align what's written here with our existing docs.

1. https://ensnode.io/ensrainbow#the-problem-unknown-labels
2. https://ensnode.io/ensrainbow/concepts/unknown-labels

contracts/ens-label-healer/src/ENSLabelHealer.sol
lightwalker-eth: We should document the *why* behind our key design choices and make them explicit.

For example, with regard to the single submitter design:

A single submitter may be designed at a time. This supports the goal of minimizing the number of duplicate `LabelHealed` events indexers need to consume. Supporting multiple submitters concurrently would create coordination complexity (ex: mitigating race conditions). Only a single instance of ENSRainbowBeam, fulfilling the role of the submitter, is needed across the whole ENS ecosystem.

contracts/ens-label-healer/src/ENSLabelHealer.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    function grantSubmitter(address submitter) external onlyOwner {
+        _grantRole(SUBMITTER_ROLE, submitter);
+    }
+
+    function revokeSubmitter(address submitter) external onlyOwner {
+        _revokeRole(SUBMITTER_ROLE, submitter);
+    }
+
+    function submit(string calldata label) external onlyRole(SUBMITTER_ROLE) whenNotPaused {
```
lightwalker-eth: We should rename / refine some of our terminology to avoid confusion.

The terminology of "submitter" / "submit" will cause confusion based on how the overall solution here will work. We need to think about this contract not only in isolation by itself, but as part of a larger solution of systems and people working together.

Note how other people / other apps will be "submitting" labels to ENSRainbowBeam through pathways such as:

1. An updated version of https://ensrainbow.io/ that will add a multi-line text field where anyone can go and dump a bunch of labels into and press "Heal".
2. Anyone who might implement the new `EnsRainbowBeamClient` that I understand @djstrong is building in his ENSRainbowBeam PR which is a lightweight wrapper around the ENSRainbowBeam REST API. For example, the official ENS apps may start using this when they discover a label based on a user's input.
3. Anyone who is directly sending HTTP requests to our ENSRainbowBeam instance that we will publicly host for the whole ENS ecosystem to use.

Therefore, who is really "the submitter" here? See how it will be confusing?

To solve for this, here's some suggested terminology:

1. `discoverer` - this is the address of who *discovers* a label and submits it to ENSRainbowBeam. Each time a label submission is made to ENSRainbowBeam (of a single label, or many labels in bulk) that submission may optionally have a `discoverer` who would then receive attribution for any of the submitted labels that are ultimately emitted onchain.
   1. ... part of the goal here is that we will eventually build a leaderboard for who the best discoverers are. We want to recognize and reward contributions!
   2. Note how the `discoverer` is optional. Sometimes labels will be discovered but the `discoverer` will be undefined. That's ok.
2. `publisher` - this is the sole and exclusive address assigned the authority to publish labels that have been discovered to the ENSNameHealer contract. In other words, I'm suggesting to rename the idea of the "submitter" as is currently implemented into this contract to be the "publisher" instead.

Another idea that we need to give special attention to here is the word `label` or `labels`. This is another key source of potential confusion. We must add formalism to explicitly document what these label values can be, because it's not what many developers might initially assume: all of the label values are what we refer to as a `LiteralLabel`.

A LiteralLabel is any possible string and is not constrained to the set of strings that are valid labels in ENSIP-15 normalized ENS names. Therefore, all of the following are examples of possible LiteralLabel values:
- The empty string
- A string containing label separator characters '.'
- A string containing null byte characters.
- A string containing character sequences that are not ENSIP-15 normalized or normalizable.
- Etc.

We can note how this enables even unnormalized ENS names to become healed, and how this has a benefit of enabling better classification of label between "unknown" vs "unnormalized".

Based on all of the ideas above, suggest to change the following functions:

1. `function submit(string calldata label) external onlyRole(SUBMITTER_ROLE) whenNotPaused`
2. `function submitBatch(string[] calldata labels) external onlyRole(SUBMITTER_ROLE) whenNotPaused`

To the following instead:

1. `function publishLabel(string calldata literalLabel) external onlyPublisher`
2. `function publishLabel(string calldata literalLabel, address discoverer) external onlyPublisher`
3. `function publishLabels(string[] calldata literalLabels) external onlyPublisher`
4. `function publishLabels(string[] calldata literalLabels, address discoverer) external onlyPublisher`

(please note: in the above suggestion I'm not focused on anything technical, which may need further refinement. In the suggestions above I'm exclusively focused on narrative)

We should also rename `LabelHealed` to `LabelPublished` -- this contract doesn't heal anything. When we discussed calling it "ENSNameHealer" we agreed to do that in the context of talking about it's value proposition, not its actual implementation. This contract publishes labels to all distributed ENSNode instances so that their ENSIndexer service may use published labels that are indexed to heal unknown labels (where possible).

We should also explicitly be documenting key that shape the fundamental design of how everything here works together, including this contract, the publisher, the community of discoverers, and indexers.

1. The open community of unknown label discoverers use any available techniques to discover labels in indexed ENS name data that are unknown. These unknown labels are represented as encoded labelhashes in indexed ENS name data. Therefore, a discovered label represents the preimage of the encoded labelhash of an unknown label.
2. When anyone in the open community of unknown label discoverers believes they discovered the preimage of the encoded labelhash of an unknown label in 1 or more indexed ENS names, that discoverer should submit the discovered labels (labelhash preimages) to the ENSRainbowBeam instance operating as the sole "publisher" of labels through the ENSNameHealer contract. Each time a discoverer submits labels to ENSRainbowBeam they may optionally attribute the discovery of those labels to their address for recognition of the discovery.
3. ENSRainbowBeam has the responsibility to minimize the volume of labels published through the ENSNameHealer contract that would not result in indexers healing 1 or more already-indexed labels. In other words, to minimize unproductive work for all ENS indexers.
4. The ENSNameHealer contract has the responsibility to enforce that only the designated publisher can publish labels and that all published labels are emitted to indexers.
5. Indexers have the responsibility to correctly interpret each LiteralLabel emitted by the ENSNameHealer contract and to appropriately update indexed data. Indexers must gracefully handle the possibility that a published label has already been published or results in no incremental labels being healed. Indexers are suggested to store all published labels indefinitely such that a published label indexed at time T could also be used by the indexer to heal another label at time T+N.

contracts/ens-label-healer/src/ENSLabelHealer.sol
```diff
@@ -0,0 +1,64 @@
...
+pragma solidity ^0.8.24;
+
+import "@openzeppelin/contracts-upgradeable/access/AccessControlUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/utils/PausableUpgradeable.sol";
+
+/// @title ENSLabelHealer
+/// @notice Permissioned on-chain emitter for unresolved ENS labels.
```
lightwalker-eth: Please see related feedback on how we should introduce this contract. This includes how we explicitly document its role as part of a larger set of systems, etc.

contracts/ens-label-healer/README.md
```diff
@@ -0,0 +1,177 @@
+# ENSLabelHealer
+
+Permissioned on-chain label emitter for unresolved ENS labels.
```
lightwalker-eth: Please see related feedback for how we document this contract, introduce its purpose, and document it's role and responsibilities within a larger set of systems.

lightwalker-eth (commented):
@sevenzing Hey thanks for the updates here. Shared feedback on next steps 👍

biome.jsonc
```diff
@@ -1,5 +1,8 @@
 {
   "$schema": "https://biomejs.dev/schemas/2.3.2/schema.json",
+  "files": {
+    "includes": ["**", "!!contracts/**/lib", "!!contracts/**/out", "!!contracts/**/cache"]
+  },
```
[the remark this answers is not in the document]
sevenzing: `!!` is intentional — Biome 2.3+ force-ignore syntax (see https://biomejs.dev/blog/biome-v2-3/). Single `!` still indexes files for the type system; Foundry `lib/`, `out/`, and `cache/` should be fully excluded. `packages/enssdk` uses `!` for generated code that may still be needed for types; that does not apply here.

contracts/ens-name-healer/script/DeployCreate2.s.sol
```diff
@@ -0,0 +1,46 @@
...
+/// @notice CREATE2 vanity deploy via Nick's factory. Mine salt with `script/mine-create2-salt.sh` first.
+///
+/// Required: OWNER_ADDRESS, CREATE2_SALT
+/// Optional: PUBLISHER_ADDRESS, CREATE2_IMPL_SALT, CREATE2_PREFIX (logged for verification only)
+contract DeployCreate2 is Script {
+    function run() external {
+        address owner = vm.envAddress("OWNER_ADDRESS");
+        address publisher = vm.envOr("PUBLISHER_ADDRESS", address(0));
+        bytes32 implSalt = vm.envOr("CREATE2_IMPL_SALT", bytes32(0));
+        bytes32 proxySalt = vm.envBytes32("CREATE2_SALT");
```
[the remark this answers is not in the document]
sevenzing: Added `require(owner != address(0), "OWNER_ADDRESS must not be zero")` to match `Deploy.s.sol`.

contracts/ens-name-healer/script/ComputeCreate2.s.sol
```diff
@@ -0,0 +1,45 @@
...
+
+/// @notice Writes proxy CREATE2 init code to `cache/create2-proxy-init-code.hex` for `cast create2`.
+///
+/// Required: OWNER_ADDRESS
+/// Optional: PUBLISHER_ADDRESS, CREATE2_IMPL_SALT (defaults to zero)
+contract ComputeCreate2 is Script {
+    function run() external {
+        address owner = vm.envAddress("OWNER_ADDRESS");
+        address publisher = vm.envOr("PUBLISHER_ADDRESS", address(0));
+        bytes32 implSalt = vm.envOr("CREATE2_IMPL_SALT", bytes32(0));
```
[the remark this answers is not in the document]
sevenzing: Added the same zero-address guard as `Deploy.s.sol` so mining fails fast before producing unusable init code.

contracts/ens-label-healer/src/ENSLabelHealer.sol
sevenzing: Addressed in the refactor: contract is now `ENSNameHealer` without `Pausable`. Publishing is paused by setting `publisher` to `address(0)` via `setPublisher`.

sevenzing:
@greptile review

shrugs (commented):
LGTM

PR #2073, opened 2026-05-08
fix(ensindexer): handle non-UTF-8 label args in RegistrarController invariants

shrugs:
## Summary

- The \`Invariant\` check on \`RegistrarController:NameRegistered\` and \`:NameRenewed\` required the emitted \`string label\` arg to round-trip to the emitted \`labelHash\`. Some contracts emit raw bytes in \`string\` args; viem replaces non-UTF-8 sequences with U+FFFD during ABI decoding, breaking the round-trip and crashing the indexer with a fatal \`IndexingFunctionError\`.
- Replace the throwing invariant with a defensive fall-through: if the decoded label doesn't round-trip, treat it as unemitted and use the heal path. The \`labelHash\` is canonical \`bytes32\` so the registration still indexes correctly under an unknown label.

## Repro

Discovered on a mainnet backfill that crashed at **block 24,026,216** after ~8h of progress. The crashing tx was \`0x405f2c36...2cfff8\` (from \`0xe747...6e1c\` to \`0x5815...24c0\`); the \`string label\` arg decoded to ~580 U+FFFD characters, indicating the contract emitted non-UTF-8 bytes.

## Scope

Two call sites: \`handleNameRegisteredByController\` and \`handleNameRenewedByController\` in \`apps/ensindexer/src/plugins/ensv2/handlers/ensv1/RegistrarController.ts\`.

## Related, not in this PR

There's a structurally similar \`Sanity Check\` in \`apps/ensindexer/src/plugins/ensv2/handlers/ensv2/ENSv2Registry.ts:77-81\` that throws on the same labelhash-mismatch condition for ENSv2-native \`LabelRegistered\`/\`LabelReserved\` events. Same anti-pattern, but the v2 contracts are new and well-controlled; the bug is far less likely to fire there in practice. Worth a follow-up cleanup but kept out of this PR to keep the change tight.

## Test plan

- [x] \`pnpm -F ensindexer typecheck\` — clean
- [x] \`pnpm lint\` — clean (biome reformatted one ternary)
- [x] \`pnpm test --project ensindexer\` — 217/217 passed
- [ ] Verify on a fresh mainnet backfill that crosses block 24,026,216 without aborting

### DIFF
--- a/.changeset/registrar-controller-non-utf8-label.md
+++ b/.changeset/registrar-controller-non-utf8-label.md
@@ -1,0 +1,5 @@
+---
+"ensindexer": patch
+---
+
+**ENSIndexer**: handle non-UTF-8 `string label` args in `RegistrarController:NameRegistered` and `:NameRenewed` events without crashing the indexer. ABI-decoding replaces non-UTF-8 byte sequences with U+FFFD, which then fails the labelhash round-trip. Previously this threw a fatal `Invariant(RegistrarController:NameRegistered)` and aborted the run; now the label is treated as unemitted and the heal path indexes the registration under the canonical `labelHash`.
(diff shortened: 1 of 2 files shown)


shrugs:
@greptile review

PR #2089, opened 2026-05-11
Prevent the "Release" workflow from running for any commit merged into `main` branch

### LINKED ISSUE #2087: Make the Release workflow to only run for the Release PRs from the Changesets bot
Currently, the "Release" workflow is executed for any commit merged into `main`:
https://github.com/namehash/ensnode/actions/workflows/release.yml

This makes the version tags used in ghcr.io to be overridden and breaks the guarantee that the given version tag is immutable and always point to the same docker image.

Image

We need the workflow to only run for commits created from Release PRs from the Changesets bot. This will result in the workflow only running when needed.

tk-o:
# Lite PR

## Summary

- Made the `publish` input to `changesets/action@v1.7.0` conditional: only set to `pnpm changeset:publish` when the commit message starts with `Release New Version` (i.e., a merged Release PR). On regular pushes to `main`, the input is empty so the action skips publish mode.
- Preserved commit-message guards on downstream jobs (`build-and-push-ensnode`, `build-lambdas`, `create-github-release`) so they only run for actual release PR merges.

---

## Why

- The previous workflow triggered `changeset publish` on every push to `main` that had no pending changesets. A regular docs commit could cause the action to enter publish mode and attempt to publish packages, which in turn set `published: 'true'` and caused downstream Docker/lambda builds to run.
- By conditionally providing the `publish` input, the `changesets` job can still run on every push to keep the Release PR up to date, but `changeset publish` and downstream builds are restricted to actual release merges.

Related run: https://github.com/namehash/ensnode/actions/runs/25669535621/job/75350813075

---

## Testing

- Not tested end-to-end in CI (requires a merge to `main`). The logic was verified by reviewing the `changesets/action` source code to confirm it skips `runPublish` when the `publish` input is empty.

---

- The `changesets` job must remain unguarded at the job level so that regular commits with new changesets still update the Release PR. The guard is applied only to the `publish` input.
- Resolves #2087 

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,7 +131,7 @@ jobs:
     name: ${{ matrix.apps.name }} ${{ matrix.apps.version }}
     # if changesets published our npm packages, also build docker images
     needs: changesets
-    if: startsWith(github.event.head_commit.message, 'Release New Version') && needs.changesets.outputs.published == 'true' && needs.changesets.outputs.publishedApps != '[]'
+    if: needs.changesets.outputs.published == 'true' && needs.changesets.outputs.publishedApps != '[]'
     runs-on: blacksmith-4vcpu-ubuntu-2204
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -68,7 +68,8 @@ jobs:
           commit: "chore(release): version apps"
           title: "Release New Version"
           version: pnpm changeset:version
-          publish: pnpm changeset:publish
+          # Only publish when this is a merged Release PR (squash merge uses PR title as commit subject)
+          publish: ${{ startsWith(github.event.head_commit.message, 'Release New Version') && 'pnpm changeset:publish' || '' }}
```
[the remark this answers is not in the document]
shrugs: don't care

shrugs, line 72: let's keep this conditional, but remove the others — my gut says that if there's no publish then `needs.changesets.outputs.published` will be false and the other `if:` blocks stay as they are

tk-o: We can try that simple approach 👍

shrugs (approved):

PR #2092, opened 2026-05-12
chore(docs/ensnode.io): Update OG images

Y3drk:
# Lite PR &rarr; Update OG images

## Summary

- Swapped OG images in `docs/ensnode.io` to use the new ones provided by @sskvts 

---

## Why

- Changing our narrative for ENSNode

---

## Testing

- Due to Twitter-related limitations, we will only be able to test this once the PR is merged into main. However, my experience suggests that it will all be fine (we only swap images, paths stay exactly the same, the change is minimal)

---

- It might take some time for the Twitter bot and the other ones to re-crawl the page for the new image, so the change in the previews might not be immediate.
- I think this change doesn't require a changeset since it's not even a content change (just alignment with the previous changes). Let me know if I should still add one.

---


lightwalker-eth (approved):
👍

PR #2101, opened 2026-05-12
feat: materialize Domain.canonical{Name,LabelHashPath,Node}

shrugs:

honestly, just trust it — generated by claude + @shrugs together, integration tests carry the weight here. if you want to spot-check, the highest-leverage areas are the recursive-CTE walk in `cascadeCanonicality` (now carries `parent_path` / `parent_name` accumulators down the tree and re-materializes Domain rows on parent-identity changes, not just flag flips) and the heal propagation path (`ensureLabel` → `cascadeLabelHeal` GIN-indexed bulk UPDATE).

## Problem & Motivation

`Domain.name` and `Domain.path` previously required a dataloader walk up the canonical tree on every read, joining Label rows per step. fine for one-off lookups, awful for any query that returns more than a handful of Domains. the canonical subgraph is the part of the namegraph we actually care about materializing; everything else (effective resolver, full record resolution) stays on-demand.

this PR materializes the canonical-tree fields directly on `Domain` so name / path / node become indexed column reads.

## What Changed (Concrete)

1. add `canonicalName: InterpretedName`, `canonicalLabelHashPath: LabelHashPath` (head-first, root → leaf), `canonicalNode: Node` columns on `Domain`. all three are NULL iff `canonical = false`.
2. add four indexes on Domain: hash on `canonicalName` (exact lookup, avoids the btree 8191-byte row-size hazard for spam names), GIN trigram on `canonicalName` (substring / fuzzy), GIN on `canonicalLabelHashPath` (heal cascade containment lookup), hash on `canonicalNode` (resolver-record joins).
3. `ensureDomainInRegistry` now takes `labelHash` and materializes name/path/node inline when the parent Registry is canonical. callers (`ENSv1Registry`, `ENSv2Registry`) updated; the ENSv1 label-heal block moved above `ensureDomainInRegistry` so the Label row exists when materialization reads it.
4. `reconcileRegistryCanonicality` now takes `prevCanonicalDomainId` and cascades through `cascadeCanonicality` when (a) the flag flipped, or (b) the flag stays canonical but the parent Domain identity changed (descendants' materialized paths would otherwise stay rooted at the prior parent).
5. `cascadeCanonicality` rewritten as a single recursive CTE carrying `parent_path` / `parent_name` accumulators, plus a JS-side chunked pass (10k batch) to namehash each row's path into `canonicalNode`. one-statement Registry + Domain batch UPDATE inside the CTE.
6. `cascadeLabelHeal` (new): when `ensureLabel` upgrades an existing Label row's `interpreted`, a GIN-indexed bulk UPDATE re-renders `canonicalName` across every Domain containing that labelHash in its path. `canonicalLabelHashPath` / `canonicalNode` are untouched (heals don't change labelHashes).
7. `namehashLabelHashPath` helper + unit tests. simple `toReversed → encodeLabelHash → namehashInterpretedName` implementation; TODO left to namehash directly over labelHashes later.
8. omnigraph schema (breaking): remove top-level `Domain.canonical: Boolean!`, `Domain.name`, `Domain.path`. add `Domain.canonical: DomainCanonical` (nullable object with `{ name: InterpretedName!, path: [Domain!]!, node: Node! }`). null when the Domain is not canonical.
9. omnigraph: move `Domain.registry` and `Domain.subregistry` onto the `Domain` interface (previously isolated to `ENSv2Domain`). `Domain.parent` now does a single unidirectional walk via `getRegistryParentDomain(registryId)` instead of consulting the canonical path dataloader.
10. split `apps/ensapi/src/omnigraph-api/schema/domain.ts` — inputs into `domain-inputs.ts`, the canonical object into `domain-canonical.ts`.
11. update example yoga health probe, `enskit-react-example` (`AccountView`, `DomainView`, `SearchView`), and the omnigraph example queries to the new `canonical { name }` shape.
12. 3 changesets: ensapi minor (breaking schema), ensindexer + ensdb-sdk minor (materialized columns + indexes), ensapi patch (registry/subregistry interface move).

## Design & Planning

no upfront doc. built incrementally on top of the existing `canonicality-db-helpers.ts` cascade machinery. the JS-side chunked namehash pass was the main design call — keeping `canonicalNode` derivation out of postgres avoids reimplementing the namehash recursion in SQL and keeps the CTE single-statement.

- Planning artifacts: none
- Reviewed / approved by: n/a (solo + claude)

## Self-Review

- Bugs caught: reconciliation no-op bug (b97403d28) — fast-path skipped writes when `__hasChildren = false` even if the flag flipped; correct order of `LabelHashPath` (c9e968bed) — head-first vs leaf-first ambiguity caught by the namehash unit test.
- Logic simplified: `Domain.parent` now does a single registry lookup instead of loading the entire canonical path then taking `[1]`; `Domain.canonical` resolver collapsed to `(domain) => (domain.canonical ? domain : null)`.
- Naming / terminology improved: `Domain.canonical` is now a structured object, so the API surface mirrors the materialized column set rather than three loosely-related top-level fields.
- Dead or unnecessary code removed: removed `rejectAnyErrors` import from `domain.ts` (path/name resolvers no longer dataloader-walk), removed `interpretedLabelsToInterpretedName` import.

## Cross-Codebase Alignment

- Search terms used: `canonicalPath`, `Domain.name`, `Domain.path`, `interpretedLabelsToInterpretedName`, `ensureDomainInRegistry`, `cascadeCanonicality`.
- Reviewed but unchanged: ENSDb SDK consumers of `canonical` boolean (no callers read the old top-level fields outside the omnigraph layer); `find-domains` resolver layer — only import paths shifted to `domain-inputs.ts`.
- Deferred alignment: `migrated_nodes` and `labels` prefetch opportunities tracked separately; not in scope here.

- Public APIs affected: omnigraph `Domain.name`, `Domain.path`, `Domain.canonical: Boolean` removed in favor of `Domain.canonical: DomainCanonical`. any consumer reading those fields breaks. example app + example queries + yoga health probe migrated as part of this PR.
- Docs updated: changesets describe the breakage; schema description on `Domain.canonical` updated; `ensv2.schema.ts` top-of-file commentary updated to reflect that the canonical subgraph is materialized while resolution stays on-demand.
- Naming decisions worth calling out: `DomainCanonical` (the new object type) holds `name`, `path`, `node` — keeping the field names short under a clear container reads better than `canonicalName` / `canonicalPath` / `canonicalNode` at the API level.

## Testing Evidence

- Testing performed: `domain.integration.test.ts` updated for the new `canonical { name, path, node }` shape (177 lines reworked, ~115 added); new `query.integration.test.ts` and `account.integration.test.ts` adjustments; new `namehash-label-hash-path.test.ts` unit tests covering single-label, multi-label, and unknown-label fallback cases; updated `module.integration.test.ts` + `test-domain-pagination` in enssdk.
- Known gaps: no full re-index validation yet. no perf comparison against the `ensv2-only mainnet baseline (15h 36m / 2,820 EPS @ main c38ad750b)`. GIN indexes are realtime-only in ponder, so backfill-time heal cascades fall back to seq scan — flagged in code comment, not measured.
- What reviewers have to reason about manually: the cascade CTE's `parent_path` / `parent_name` accumulator correctness on parent-identity-changed rewrites; the chunk-size choice (10k) for the JS-side namehash pass; index selectivity on real-world data.

## Scope Reductions

- Follow-ups: perf-test the re-index against the ensv2-only baseline; consider a non-ponder-managed index on `canonicalLabelHashPath` so backfill-time heal cascades don't seq-scan; namehash directly over labelHashes (avoid the `encodeLabelHash` round-trip in `namehashLabelHashPath`); DRY `labelExists` vs the `ensDb.find` re-implementation in `ensureLabel`.
- Why they were deferred: keep this PR focused on the materialization + schema change; perf measurement is a separate workstream.

## Risk Analysis

- Risk areas: (a) materialization writes added to every canonical Domain creation — measurable hot-path cost; (b) parent-identity-changed cascade is new behavior, previously the reconciler short-circuited when the flag matched; (c) heal cascade triggers a non-cache UPDATE outside ponder's flush boundaries; (d) GIN trigram index build cost on large `canonicalName` populations.
- Mitigations or rollback options: schema-only change on the indexer side, so a re-index reverts cleanly; the ENSv1 fast-path (`__hasChildren = false`) still no-ops the CTE for the dominant case; the bulk UPDATE in `cascadeLabelHeal` only runs when an existing label's interpreted value actually changes.


packages/enssdk/src/omnigraph/generated/schema.graphql
```diff
@@ -213,56 +213,72 @@ scalar ChainId
 scalar CoinType
 
 """
-A Domain represents an individual Label within the ENS namegraph. It may or may not be Canonical. It may be an ENSv1Domain or an ENSv2Domain.
+A Domain represents an on-chain Domain, i.e. an individual Label within the ENS namegraph. It may or may not be Canonical. It may be an ENSv1Domain or an ENSv2Domain.
```
lightwalker-eth: Do we need to explicitly say "on-chain"? That's true for now, but ideally we expand to support offchain domains in the future?

What if instead we wrote this as:

```suggestion
Represents a Domain, i.e. an individual Label within the ENS NameGraph. It may or may not be Canonical. It may be an ENSv1Domain or an ENSv2Domain.
```

packages/enssdk/src/omnigraph/generated/schema.graphql
```diff
@@ -213,56 +213,72 @@ scalar ChainId
...
 """
 interface Domain {
-  """Whether the Domain is Canonical."""
-  canonical: Boolean!
+  """
+  Metadata (name, path, and node) related to the Domain's canonicality, if known. Null when the Domain is not Canonical.
```
lightwalker-eth:
```suggestion
  Metadata (name, path, and node) related to the Domain's canonicality, if known. Null when the Domain is not in the canonical NameTree.
```

Wdyt?

packages/enssdk/src/omnigraph/generated/schema.graphql
```diff
@@ -213,56 +213,72 @@ scalar ChainId
...
 
   """All Events associated with this Domain."""
   events(after: String, before: String, first: Int, last: Int, where: EventsWhereInput): DomainEventsConnection
 
-  """A unique reference to this Domain."""
+  """A unique and stable reference to this Domain."""
   id: DomainId!
 
-  """The Label this Domain represents in the ENS Namegraph"""
+  """The Label this Domain represents in the ENS Namegraph."""
```
lightwalker-eth: Should we say something like the following?

```suggestion
  """The Label associated with this Domain in the ENS Namegraph."""
```

packages/enssdk/src/omnigraph/generated/schema.graphql
```diff
@@ -213,56 +213,72 @@ scalar ChainId
...
   registrations(after: String, before: String, first: Int, last: Int): DomainRegistrationsConnection
 
+  """The Registry under which this Domain exists."""
+  registry: Registry!
+
   """
-  The Resolver that this Domain has assigned, if any. NOTE that this is the Domain's _assigned_ Resolver, _not_ its _effective_ Resolver, which can only be determined by following ENS Forward Resolution and ENSIP-10.
+  The Resolver that this Domain has assigned, if any. NOTE that this is the Domain's _assigned_ Resolver, _not_ its _effective_ Resolver, which can only be determined by following ENS Forward Resolution and ENSIP-10. Do NOT use this Domain-Resolver relationship in isolation to resolve records, that operation is NOT ENS Forward Resolution.
   """
   resolver: Resolver
```
lightwalker-eth:
```suggestion
  assignedResolver: Resolver
```

Wdyt? And then later we could add `effectiveResolver` in some future PR.

packages/ensdb-sdk/src/ensindexer-abstract/ensv2.schema.ts
```diff
@@ -24,8 +26,9 @@ import type { EncodedReferrer } from "@ensnode/ensnode-sdk";
  *
  * While the initial approach was a highly materialized view of the ENS protocol, abstracting away
  * as many on-chain details as possible, in practice—due to the sheer complexity of the protocol at
- * resolution-time—it becomes more or less impossible to appropriately materialize the canonical
- * namegraph.
+ * resolution-time—full materialization of resolution behavior is impractical. The canonical
+ * subgraph, however, is materialized inline via synchronous handler-side cascades; see
```
lightwalker-eth:
```suggestion
 * nametree, however, is materialized inline via synchronous handler-side cascades; see
```

packages/ensdb-sdk/src/ensindexer-abstract/ensv2.schema.ts
```diff
@@ -50,10 +53,11 @@ import type { EncodedReferrer } from "@ensnode/ensnode-sdk";
  * guarantees (for example, ENSv1 BaseRegistrar Registrations may have a gracePeriod, but ENSv2
```
lightwalker-eth, line 53: As I understand the idea to remove the grace period in ENSv2 is now changed and that grace periods will remain. Not sure how that aligns with the current state of the contracts at this moment. We should follow whatever is in the contracts, not what the latest public announcement of intention is 👍 Therefore, perhaps no need to change anything here at this time.

apps/ensindexer/src/lib/ensv2/namehash-label-hash-path.ts
```diff
@@ -0,0 +1,25 @@
...
+  namehashInterpretedName,
+} from "enssdk";
+
+/**
+ * Namehash a LabelHashPath.
+ *
+ * `LabelHashPath` is head-first (root → leaf), but ENS name strings are leaf-first
+ * ("vitalik.eth"), so we reverse before encoding each labelHash as `[<hash>]` and joining.
+ *
+ * TODO: this could more accurately perform the namehash algorithm over the LabelHashes directly
```
lightwalker-eth:
```suggestion
 * TODO: optimize by performing the namehash algorithm over the LabelHashes directly
```

Right? The current comment sounds like this is a question of accuracy, but as I understand it's a question of performance optimization.

apps/ensindexer/src/lib/ensv2/label-db-helpers.ts
```diff
@@ -20,15 +21,23 @@ export async function labelExists(context: IndexingEngineContext, labelHash: Lab
 
 /**
  * Ensures that the LiteralLabel `label` is interpreted and upserted into the Label rainbow table.
+ * When this upgrades an existing row's `interpreted` value (a heal), propagates the new value to
```
lightwalker-eth, line 24: Nice 👍

apps/ensapi/src/omnigraph-api/schema/domain.ts
```diff
@@ -82,13 +86,13 @@ export const ENSv2DomainRef = builder.objectRef<ENSv2Domain>("ENSv2Domain");
 //////////////////////////////////
 DomainInterfaceRef.implement({
   description:
-    "A Domain represents an individual Label within the ENS namegraph. It may or may not be Canonical. It may be an ENSv1Domain or an ENSv2Domain.",
+    "A Domain represents an on-chain Domain, i.e. an individual Label within the ENS namegraph. It may or may not be Canonical. It may be an ENSv1Domain or an ENSv2Domain.",
```
lightwalker-eth: Oops. I left a related comment on this sentence in some other file.

apps/ensapi/src/omnigraph-api/schema/domain.ts
```diff
@@ -296,7 +263,7 @@ DomainInterfaceRef.implement({
 // ENSv1Domain Implementation
 //////////////////////////////
 ENSv1DomainRef.implement({
-  description: "An ENSv1Domain represents an ENSv1 Domain.",
+  description: "An ENSv1Domain represents an on-chain ENSv1 Domain.",
```
lightwalker-eth: Please see related feedback about "on-chain" qualifier.

apps/ensapi/src/omnigraph-api/schema/domain.ts
```diff
@@ -327,7 +294,7 @@ ENSv1DomainRef.implement({
 // ENSv2Domain Implementation
 //////////////////////////////
 ENSv2DomainRef.implement({
-  description: "An ENSv2Domain represents an ENSv2 Domain.",
+  description: "An ENSv2Domain represents an on-chain ENSv2 Domain.",
```
lightwalker-eth: Please see related feedback about "on-chain" qualifier.

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -15,7 +24,9 @@ import { ensIndexerSchema, type IndexingEngineContext } from "@/lib/indexing-eng
  * upon canonical edges. A canonical edge between Registry R and parent Domain D requires both
  * unidirectional pointers to agree (`R.canonicalDomainId = D.id` ↔ `D.subregistryId = R.id`)
  * AND for D itself to be canonical. Root Registries (ENSv1 root, ENSv2 root) are canonical by
- * definition and seeded as Canonical at `ensureRegistry`-time.
+ * definition and seeded as Canonical at `ensureRegistry`-time. This definition of canonicality is
+ * taken from ENSv2 and then applied to ENSv1, so naturally some ENSv1 Domains will not be considered
+ * canonical, since ENSv1 didn't necessitate the same concept.
  *
  * Concretely, this means a v1 Domain with a Bridged Resolver leaves its v1 children as non-canonical.
```
lightwalker-eth, line 31: Really appreciate the docs here! Very good!

My understanding of the example cited here of `bridge.linea.eth` is the inverse though.

If `bridge.linea.eth` explicitly exists both on mainnet as an ENSv1 domain and on Linea as an ENSv1 domain, my understanding is that:
- `bridge.linea.eth` on mainnet is canonical according to ENS resolution, ENSIP-10 / wildcard resolvers don't apply.
- `bridge.linea.eth` on Linea is non-canonical according to ENS resolution.

It's quite late here, so if I just failed to interpret the comments here correctly then please ignore.

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -25,26 +36,34 @@ import { ensIndexerSchema, type IndexingEngineContext } from "@/lib/indexing-eng
...
  * The unidirectional pointers `Registry.canonicalDomainId` and `Domain.subregistryId` are written
- * blindly when their onchain events fire, to store the on-chain state as-is. Edge authentication
+ * directly when their onchain events fire, to store the on-chain state as-is. Edge authentication
  * is done on-demand at query time. The boolean flags `Registry.canonical` and `Domain.canonical`
  * are materialized from membership in the canonical nametree, and are kept up-to-date by reconciling
  * during updates to the uni-directional pointers.
  *
- * When `reconcileRegistryCanonicality` determines a Registry's flag actually flips, we cascade
- * the change through the canonical subgraph beneath it. Two paths:
+ * `reconcileRegistryCanonicality` cascades through the canonical subgraph beneath the Registry
```
lightwalker-eth:
```suggestion
 * `reconcileRegistryCanonicality` cascades through the canonical nametree beneath the Registry
```

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +344,147 @@ async function reconcileRegistryCanonicality(
...
+    // the Ponder cache (ENSv1 fast-path). A parent-only change with no descendants has nothing
+    // to re-materialize, so we skip the write entirely.
     await context.ensDb
       .update(ensIndexerSchema.registry, { id: registryId })
       .set({ canonical: nextCanonical });
   }
 }
 
+/**
+ * Propagate a Label heal to every canonical Domain whose `canonicalLabelHashPath` contains
```
lightwalker-eth, line 389: Shouldn't we also propagate the Label heal even to non-canonical Domain who's label has a matching labelhash?

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +344,147 @@ async function reconcileRegistryCanonicality(
...
+        FROM unnest(d.canonical_label_hash_path) WITH ORDINALITY AS p(lh, ord)
+        JOIN ${ensIndexerSchema.label} l ON l.label_hash = p.lh
+      )
+      WHERE d.canonical = true
+        AND d.canonical_label_hash_path @> ARRAY[${labelHash}]::text[];
+  `);
+}
+
 /**
  * Walk the canonical subgraph rooted at `registryId` and set `canonical = nextCanonical` on
```
lightwalker-eth:
```suggestion
 * Walk the canonical nametree rooted at `registryId` and set `canonical = nextCanonical` on
```

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +344,147 @@ async function reconcileRegistryCanonicality(
...
- * every Registry and Domain it visits.
+ * every Registry and Domain it visits, additionally materializing canonical-tree fields on
+ * every affected Domain.
+ *
+ * Two phases:
+ *   - Phase A is a single statement: one recursive CTE that enumerates the canonical subgraph
```
lightwalker-eth:
```suggestion
 *   - Phase A is a single statement: one recursive CTE that enumerates the canonical nametree
```

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -326,9 +493,44 @@ async function cascadeCanonicality(
...
+        )
+      RETURNING d.id, d.canonical_label_hash_path;
   `);
+
+  // Phase B: when flipping to canonical, compute and write `canonical_node` per affected Domain.
+  // When flipping to non-canonical, Phase A already nulled `canonical_node` — nothing to do.
+  // NOTE: this is necessary because there's no labelhash-aware namehash fn in postgres
+  // TODO: perhaps we could add a namehash fn in eventHandlerPreconditions and collapse this into
+  // a single update?
+  // NOTE: this step could be avoided if we didn't need to materialize the Canonical Node — if not
```
lightwalker-eth, line 534: My first thought is that this is useful to keep, but feedback welcome 👍

lightwalker-eth (approved):
@shrugs Nice work 🚀 This is big! Shared a few suggestions. Please feel welcome to merge when ready 👍

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +344,147 @@ async function reconcileRegistryCanonicality(
...
+ * cascades degenerate to a sequential scan; re-asses whether a lookup table would help, or perhaps
+ * introduce a non-ponder-managed index on this column via eventHandlerPreconditions.
+ */
+export async function cascadeLabelHeal(
+  context: IndexingEngineContext,
+  labelHash: LabelHash,
+): Promise<void> {
+  await context.ensDb.sql.execute(sql`
+    UPDATE ${ensIndexerSchema.domain} AS d
+      SET canonical_name = (
```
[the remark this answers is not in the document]
shrugs: Acknowledged but intentional. `cascadeLabelHeal` only runs *after* `ensureLabel(label)` writes the heal target, and `canonical_label_hash_path` entries are sourced from the same `label` table — every labelHash in a canonical path is guaranteed to have a row by the same invariant that lets `ensureDomainInRegistry` throw if a path label is missing. A silent `COALESCE` fallback would mask the canonical-tree-corruption signal we *want* loud; if we ever break that invariant, the `DomainCanonical.name` resolver throwing at query time is the correct alarm. Resolving.

apps/ensapi/src/omnigraph-api/lib/get-registry-parent-domain.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Returns the Registry's parent DomainId, if known.
+ */
+export async function getRegistryParentDomain(registryId: RegistryId): Promise<DomainId | null> {
+  const registry = await ensDb.query.registry.findFirst({
+    where: (t, { eq }) => eq(t.id, registryId),
+    columns: { canonicalDomainId: true },
+  });
+  return registry?.canonicalDomainId ?? null;
```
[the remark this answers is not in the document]
shrugs: Real concern. Adding a DataLoader is the right fix but is its own change — added a TODO inline (`get-registry-parent-domain.ts`) and will follow up in a dedicated PR rather than scope-creep this one. Resolving.

apps/ensindexer/src/plugins/ensv2/handlers/ensv1/ENSv1Registry.ts
```diff
@@ -74,6 +74,34 @@ export default function () {
...
+        // avoid addr.reverse healing for that namespace so indexing progresses smoothly
+        // TODO: remove this once Sepolia V2 is decomissioned
+        config.namespace !== ENSNamespaceIds.SepoliaV2
+      ) {
+        const label = await healAddrReverseSubnameLabel(context, event, labelHash);
+        await ensureLabel(context, label);
+      } else {
+        await ensureUnknownLabel(context, labelHash);
+      }
+    }
```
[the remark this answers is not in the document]
shrugs: Not applicable to this flow. The `addr.reverse` heal path runs from `ENSv1Registry:NewOwner`, which fires once on Domain creation; there is no separate later event that re-enters this block for an existing unknown label. If a row already exists, label discovery has already happened for it via this same handler. Resolving.

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
shrugs: Only canonical Domains carry a materialized `canonicalName` (it's null when `canonical = false`), so there is no `canonical_name` to update on non-canonical rows — the `WHERE canonical = true` clause is exactly the set of rows where a label heal can change anything. Resolving.

shrugs:
@greptile review

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +352,147 @@ async function reconcileRegistryCanonicality(
...
+        SELECT string_agg(l.interpreted, '.' ORDER BY p.ord DESC)
+        FROM unnest(d.canonical_label_hash_path) WITH ORDINALITY AS p(lh, ord)
+        JOIN ${ensIndexerSchema.label} l ON l.label_hash = p.lh
```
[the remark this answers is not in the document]
shrugs: Same concern as the resolved Greptile thread on this file. Intentional: `cascadeLabelHeal` only runs after `ensureLabel(label)` writes the heal target, and every labelHash in `canonical_label_hash_path` is sourced from the same `label` table. A silent `WHERE l.interpreted IS NOT NULL` would mask the canonical-tree-corruption signal we want loud — if that invariant ever breaks, the `DomainCanonical.name` resolver throwing at query time is the correct alarm. Resolving.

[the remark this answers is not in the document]
shrugs: Duplicate of the sibling VADE thread on the same line; see rationale there. Resolving.

shrugs:
@greptile review

apps/ensapi/src/omnigraph-api/schema/domain-canonical.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (canonicalPath instanceof Error) throw canonicalPath;
+        if (canonicalPath === null) {
+          throw new Error(
+            `Invariant(DomainCanonical.path): canonical Domain '${domain.id}' produced null canonical path.`,
+          );
+        }
+
+        return canonicalPath;
+      },
+    }),
```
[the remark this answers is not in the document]
shrugs: Real follow-up. Added a TODO inline on the resolver — the materialized `canonicalLabelHashPath` column lets us derive each ancestor DomainId and batch them through `DomainInterfaceRef.getDataloader` in a future PR. Out of scope here to keep the diff bounded. Resolving.

apps/ensapi/src/omnigraph-api/schema/domain.ts
```diff
@@ -108,79 +111,23 @@ DomainInterfaceRef.implement({
...
       type: DomainInterfaceRef,
       nullable: true,
-      resolve: async (domain, _args, context) => {
-        const path = await context.loaders.canonicalPath.load(domain.id);
-        if (path instanceof Error) throw path;
-        return path?.[1] ?? null;
-      },
+      resolve: async (domain, _args, context) =>
+        context.loaders.registryParentDomain.load(domain.registryId),
     }),
```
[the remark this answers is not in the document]
shrugs: Good catch — keeping the new semantics (unidirectional walk over `Registry.canonicalDomainId`, no agreement check) and documenting them. Updated the field description on both `Domain.parent` references and called out the asymmetry in `.changeset/ensapi-canonical-name.md` so consumers can adjust. Resolving.

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -326,9 +501,44 @@ async function cascadeCanonicality(
...
+    const ids = batch.map((r) => r.id);
+    const nodes = batch.map((r) => namehashLabelHashPath(r.canonical_label_hash_path));
+
+    await context.ensDb.sql.execute(sql`
+      UPDATE ${ensIndexerSchema.domain} AS d
+        SET canonical_node = upd.canonical_node
+        FROM (
+          SELECT unnest(${ids}::text[]) AS id,
+                 unnest(${nodes}::text[]) AS canonical_node
+        ) upd
```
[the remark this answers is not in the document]
shrugs: Fixed. Switched Phase B to `unnest(${ids}::text[], ${nodes}::text[]) AS upd(id, canonical_node)` — multi-argument unnest is the documented, contractually-aligned form. Resolving.

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +352,147 @@ async function reconcileRegistryCanonicality(
...
+): Promise<void> {
+  await context.ensDb.sql.execute(sql`
+    UPDATE ${ensIndexerSchema.domain} AS d
+      SET canonical_name = (
+        SELECT string_agg(l.interpreted, '.' ORDER BY p.ord DESC)
+        FROM unnest(d.canonical_label_hash_path) WITH ORDINALITY AS p(lh, ord)
+        JOIN ${ensIndexerSchema.label} l ON l.label_hash = p.lh
+      )
+      WHERE d.canonical = true
+        AND d.canonical_label_hash_path @> ARRAY[${labelHash}]::text[];
```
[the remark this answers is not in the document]
shrugs: Two issues — addressing separately:

1) `string_agg` NULL: same rationale as the resolved Greptile/VADE threads on this file. `cascadeLabelHeal` only runs after `ensureLabel(label)` writes the heal target, and unknown labels still get a row via `ensureUnknownLabel` (with `interpreted` set to `[<hash>]`). Every label in `canonical_label_hash_path` is guaranteed a `label` row by construction, so `string_agg` cannot return NULL. Loud query-time failure on invariant break is preferred over silent COALESCE.

2) `bytea[]` cast: false positive. Ponder maps `t.hex()` to `text` storage (hex strings with a leading `0x`, not `bytea`). The `::text[]` cast is correct and exercised by the existing integration tests. The `byCanonicalLabelHashPath` index uses default `array_ops`, which is what we need for `@>` containment.

Resolving.

apps/ensindexer/src/lib/ensv2/canonicality-db-helpers.ts
```diff
@@ -270,54 +352,147 @@ async function reconcileRegistryCanonicality(
...
+      SELECT
+        d.id AS domain_id,
+        w.parent_path || ARRAY[d.label_hash] AS new_path,
+        COALESCE(l.interpreted || '.' || w.parent_name, l.interpreted) AS new_name
+      FROM walk w
+      JOIN ${ensIndexerSchema.domain} d
+        ON d.registry_id = w.registry_id
+      JOIN ${ensIndexerSchema.label} l
+        ON l.label_hash = d.label_hash
     ),
```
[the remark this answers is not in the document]
shrugs: Added a clarifying comment to `domain_targets` explaining why the agreement filter is intentionally omitted: every Domain under a canonical Registry inherits that Registry's canonical flag (no separate canonical subtree), and the bidirectional agreement check on `walk` already guarantees the strict-tree property. Resolving.

apps/ensindexer/src/lib/ensv2/label-db-helpers.ts
```diff
@@ -20,15 +21,23 @@ export async function labelExists(context: IndexingEngineContext, labelHash: Lab
...
+  // TODO: this re-implements labelExists, can likely DRY it up
+  const prev = await context.ensDb.find(ensIndexerSchema.label, { labelHash });
+
   await context.ensDb
     .insert(ensIndexerSchema.label)
     .values({ labelHash, interpreted })
     .onConflictDoUpdate({ interpreted });
+
+  // if this was a heal (was previous seen, now updated), cascade the update to the materialized names
+  if (prev && prev.interpreted !== interpreted) await cascadeLabelHeal(context, labelHash);
```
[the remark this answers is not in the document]
shrugs: Intended envelope. `cascadeLabelHeal` runs once per heal, where a heal is `prev.interpreted !== interpreted` — i.e. either an unknown→known upgrade (the addr.reverse case) or a re-interpretation rule change. The single GIN-indexed UPDATE is bounded to the affected canonical Domains. We accept the cache-flush cost because heals are rare in steady state and the alternative (deferring the cascade) would let the materialized `canonicalName` drift from `Label.interpreted`. Will revisit if measurement shows it as a hot path. Resolving.

packages/ensnode-sdk/src/omnigraph-api/example-queries.ts
```diff
@@ -104,18 +104,13 @@ query DomainByName($name: InterpretedName!) {
...
     id
     label { interpreted hash }
-    name
+    canonical { name node path { id } }
     owner { address }
+    subregistry { contract { chainId address } }
 
     ... on ENSv1Domain {
       rootRegistryOwner { address }
     }
```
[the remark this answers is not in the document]
shrugs: No v1/v2-specific comments needed to update in this file. The `subregistry { ... }` selection is correctly at the top level (since `subregistry` is now on the `Domain` interface), and the remaining v1-specific arm (`... on ENSv1Domain { rootRegistryOwner }`) is still correct. Resolving.

shrugs:
@greptile review

PR #2104, opened 2026-05-13
hotfix: find-resolver correctly construct new domain ids

shrugs:
caused during id refactor (unified datamodel) that made ensv1domainids `(registry, node)` instead of just `node`. the type-unsafe cast in forward-resolution wasn't updated, so the active resolver for a name couldn't be identified, leading to null results

 the ENSv1RegistryOld change need not be included but it's safe; the indexer writes all records to ENSv1Registry now, (appropriately ignoring the Old ones after migration) so the DRRs attached to ENSv1Registry are guaranteed to implement the fallback correctly

### DIFF
--- a/.changeset/eager-yaks-shout.md
+++ b/.changeset/eager-yaks-shout.md
@@ -1,0 +1,5 @@
+---
+"ensapi": patch
+---
+
+Hotfixes an issue where Accelerated records would always be null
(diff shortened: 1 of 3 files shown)


lightwalker-eth (approved):
@shrugs Nice, thanks!

sevenzing (approved):

tk-o (approved):
LGTM

PR #2181, opened 2026-05-22
fix(ensindexer): reject ENSRainbow heals that don't hash back to the requested labelHash

shrugs:
## what

`EnsRainbowApiClient.heal()` now verifies a healed label hashes back to the requested labelHash (compared against the client's normalized labelHash). A mismatch — a malformed rainbow record — is returned as `NotFound`, so consumers fall back gracefully via the existing unhealable path instead of keying data under the wrong labelHash.

## why

A v1.14.0-alpha reindex crashed at ~78.6% backfill on `unigraph/ENSv1Registry:NewOwner` (chain 1, block 15546441):

```
Invariant(ensureDomainInRegistry): Label '0x00677002…' must exist before linking Domain '…'.
```

Root cause — a **malformed ENSRainbow record**:

- The event registers `"007".eth` — the on-chain label is the literal 5-char string `"007"` (quotes included), and `keccak("007"-with-quotes) == 0x00677002…`.
- The alpha's ENSRainbow label set heals `0x00677002…` to `007` (quotes stripped — `"` is the CSV quote char, so a CSV/SQL ingest step mangled the label while preserving the original labelHash key). ENSRainbow re-validates neither at ingest nor at heal time.
- `ensureUnknownLabel` → `ensureLabel` **re-keys on the healed label**, writing the row under `keccak("007") = 0xadafce…` instead of `0x00677002…`. The requested labelHash is never persisted.
- `ensureDomainInRegistry` sees the `.eth` virtual registry is canonical, does `find(label, { labelHash: 0x00677002… })`, gets `null`, and throws — a fail-fast that kills the indexer.

## notes

- The upstream label set should also be rebuilt with CSV-safe label handling so `"007"` and similar quoted/special-char labels are recoverable rather than silently dropped to Encoded LabelHashes.

### DIFF
--- a/.changeset/reject-malformed-rainbow-heals.md
+++ b/.changeset/reject-malformed-rainbow-heals.md
@@ -1,5 +1,5 @@
 ---
-"ensindexer": patch
+"@ensnode/ensrainbow-sdk": patch
 ---
 
 ENSIndexer no longer crashes on malformed ENSRainbow records and verifies that any ENSRainbow-healed label labelhashes to the expected `labelHash`.
--- a/.changeset/reject-malformed-rainbow-heals.md
+++ b/.changeset/reject-malformed-rainbow-heals.md
@@ -2,4 +2,4 @@
 "@ensnode/ensrainbow-sdk": patch
 ---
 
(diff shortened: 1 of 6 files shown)


shrugs:
@greptile review

shrugs:
@greptile review

shrugs:
@greptile review

PR #2232, opened 2026-06-01
fix(ensapi): small omnigraph api updates

sevenzing:
# Lite PR

## Summary

- What changed (1-3 bullets, no essays).

---

## Why

- Why this change exists. Link to related GitHub issues where relevant.

---

## Testing

- How this was tested.

---

---

### DIFF
--- a/.changeset/neat-dingos-attend.md
+++ b/.changeset/neat-dingos-attend.md
@@ -1,0 +1,7 @@
+---
+"@ensnode/ensnode-sdk": patch
+"enssdk": patch
+"ensapi": patch
+---
+
+omnigraph: rename chain to chainName, remove DEFAULT chain, add Hex type to address.
--- a/.changeset/neat-dingos-attend.md
+++ b/.changeset/neat-dingos-attend.md
@@ -1,7 +1,0 @@
----
-"@ensnode/ensnode-sdk": patch
-"enssdk": patch
-"ensapi": patch
----
-
-omnigraph: rename chain to chainName, remove DEFAULT chain, add Hex type to address.
--- a/apps/ensapi/src/lib/resolution/accelerate-known-onchain-static-resolver.ts
+++ b/apps/ensapi/src/lib/resolution/accelerate-known-onchain-static-resolver.ts
@@ -1,4 +1,4 @@
-import type { AccountId, Node } from "enssdk";
+import type { AccountId, Hex, Node } from "enssdk";
 
 import type { ResolverRecordsSelection } from "@ensnode/ensnode-sdk";
 
@@ -50,7 +50,7 @@ function resolveOperationWithIndex(op: Operation, records: IndexedRecords): Oper
     case "addr": {
       const coinType = op.args[1];
       const found = records?.addressRecords.find((r) => r.coinType === coinType);
-      return { ...op, result: found?.value ?? null };
+      return { ...op, result: (found?.value ?? null) as Hex | null };
     }
     case "text": {
       const key = op.args[1];
--- a/apps/ensapi/src/omnigraph-api/schema/records.ts
+++ b/apps/ensapi/src/omnigraph-api/schema/records.ts
@@ -31,7 +31,7 @@ ResolvedRawTextRecordRef.implement({
 ///////////////////////////
 export const ResolvedAddressRecordRef = builder.objectRef<{
   coinType: CoinType;
-  address: string | null;
+  address: Hex | null;
 }>("ResolvedAddressRecord");
 
 ResolvedAddressRecordRef.implement({
@@ -43,9 +43,10 @@ ResolvedAddressRecordRef.implement({
       nullable: false,
       resolve: (r) => r.coinType,
     }),
-    address: t.exposeString("address", {
+    address: t.expose("address", {
+      type: "Hex",
       description:
-        "The raw address record value, or null if not set. May be any address format for the associated coin type and may require validation or preprocessing before use. Only EVM addresses are guaranteed to be NormalizedAddress values; see ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder).",
+        "The on-chain address record bytes as hex, or null if not set. Decode with ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder) for the associated coin type.",
       nullable: true,
     }),
   }),
--- a/apps/ensapi/src/omnigraph-api/schema/records.ts
+++ b/apps/ensapi/src/omnigraph-api/schema/records.ts
@@ -46,7 +46,7 @@ ResolvedAddressRecordRef.implement({
     address: t.expose("address", {
       type: "Hex",
       description:
-        "The on-chain address record bytes as hex, or null if not set. Decode with ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder) for the associated coin type.",
+        'The "raw" resolved address record as hex, or null if not set. Decode with ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder) for the associated coin type. May be a hex value representing 0 or more bytes. There is no guarantee that an EVM coinType returns an address value of any particular byte length.',
       nullable: true,
     }),
   }),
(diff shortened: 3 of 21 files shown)


apps/ensapi/src/lib/resolution/accelerate-known-onchain-static-resolver.ts
```diff
@@ -50,7 +50,7 @@ function resolveOperationWithIndex(op: Operation, records: IndexedRecords): Oper
     case "addr": {
       const coinType = op.args[1];
       const found = records?.addressRecords.find((r) => r.coinType === coinType);
-      return { ...op, result: found?.value ?? null };
+      return { ...op, result: (found?.value ?? null) as Hex | null };
```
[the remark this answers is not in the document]
sevenzing: I added this only for typecheck

.changeset/neat-dingos-attend.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ensnode/ensnode-sdk": patch
+"enssdk": patch
+"ensapi": patch
+---
+
+omnigraph: rename chain to chainName, remove DEFAULT chain, add Hex type to address.
```
tk-o: @sevenzing I wonder if we need to create a new changesets file for the feature that has not yet been part of any official release. Perhaps we should just update [the changests from PR 1974](https://github.com/namehash/ensnode/pull/1974/changes#diff-fe478cc41d5d7264a0bb3ca297ada2fb7fa94f2c7bee8b781271be6978e90ca4) to describe the current data model. What do you think?

lightwalker-eth: @tk-o Good suggestion! @sevenzing suggest to remove this changeset and instead only make updates to the earlier created changeset if any changes are needed to it based on this PR 😄

packages/enssdk/src/omnigraph/generated/schema.graphql
```diff
@@ -254,13 +254,12 @@ type CanonicalName {
 scalar ChainId
 
 """
-Primary-name chains supported by the Omnigraph API. Use `DEFAULT` for the default EVM chain.
+Primary-name chains supported by the Omnigraph API.
```
lightwalker-eth:
```suggestion
The names of chains that the Omnigraph API supports identifying by name as a syntactic convenience. The Omnigraph API supports identification of additional chains beyond this list, but those chains must be identified through other mechanisms such as `coinType` or `chainId`.
```

Goals:
1. In theory these aren't only for Primary name lookups, right? These could be used in other contexts too.
2. It's possible to make a primary name lookup on the "default" chain when querying by coinType (and also later when we add support to query by chainId). Therefore we don't want to say these are the only chains where you can make a primary name lookup.
3. The big problem we are solving by removing `DEFAULT` from this list of ChainName values all comes from the word "default" and how that will cause endless big mistakes and misunderstandings because there's a big difference between what the word "default" normally means and what it would represent if used in this context as a ChainName.

Could you please also verify that all will be ok if someone makes a query using coinType X and then asks to fetch the chainName for coinType X and we don't have any defined chainName for that coinType. Assuming this is all ok then no need to message me back about it. Just want to be sure 🫡

packages/enssdk/src/omnigraph/generated/schema.graphql
```diff
@@ -1487,9 +1486,9 @@ type ResolvedAbiRecord {
 """A resolved address record for an ENS name."""
 type ResolvedAddressRecord {
   """
-  The raw address record value, or null if not set. May be any address format for the associated coin type and may require validation or preprocessing before use. Only EVM addresses are guaranteed to be NormalizedAddress values; see ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder).
+  The on-chain address record bytes as hex, or null if not set. Decode with ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder) for the associated coin type.
```
lightwalker-eth:
```suggestion
  The "raw" resolved address record as hex, or null if not set. Decode with ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder) for the associated coin type. May be a hex value representing 0 or more bytes. There is no guarantee that an EVM coinType returns an address value of any particular byte length.
```

Goals:
1. We shouldn't constrain this to say it only comes from on-chain data sources. The value here may come from offchain datasources too via CCIP-read from non-accelerated resolution requests.
2. Adding more explicit warnings about how the raw value of this field may not be what someone might at first expect.

packages/ensnode-sdk/src/shared/interpretation/interpret-record-values.ts
```diff
@@ -54,14 +53,17 @@ export function interpretAddressRecordValue(value: string): string | null {
   // interpret empty bytes as deletion of address record
   if (value === "0x") return null;
 
-  // if it's not an EVM address, return as-is
-  if (!isAddress(value, { strict: false })) return value;
+  if (!isHex(value)) return null;
```
lightwalker-eth, line 56: I just thought of something else we should explicitly warn about when it comes to the "raw" variation of address records: there's no guarantee that there's an even number of hex digits.

Can you please add an explicit warning about this in the relevant place(s)?

packages/ensnode-sdk/src/shared/interpretation/interpret-record-values.ts
```diff
@@ -54,14 +53,17 @@ export function interpretAddressRecordValue(value: string): string | null {
...
+
+  const hex = value.toLowerCase() as Hex;
 
   // interpret zeroAddress as deletion
-  if (isAddressEqual(value, zeroAddress)) return null;
+  if (isAddress(hex, { strict: false }) && isAddressEqual(hex, zeroAddress)) return null;
+
+  // normalize EVM addresses to lowercase NormalizedAddress hex
+  if (isAddress(hex, { strict: false })) return toNormalizedAddress(hex);
```
lightwalker-eth: Several lines above you are already calling:

```
const hex = value.toLowerCase() as Hex;
```

Therefore this entire line of code can be removed.

packages/ensnode-sdk/src/resolution/resolver-records-response.ts
```diff
@@ -18,11 +18,9 @@ export type ResolverRecordsResponseBase = {
    * Address records, keyed by CoinType.
    * Value is null if no record for the specified CoinType is set.
    *
-   * NOTE: ENS resolver address records can store arbitrary string values,
-   * including non-EVM addresses — always validate the record value against
-   * the format your application expects.
+   * Values are the on-chain record bytes as hex.
```
lightwalker-eth:
```suggestion
   * Values are the raw record bytes as hex.
```

Could you please also review the special warnings we wrote about this raw field in other places and include that here or elsewhere that is can be relevant? Thanks!

apps/ensapi/src/omnigraph-api/schema/records.ts
```diff
@@ -43,9 +43,10 @@ ResolvedAddressRecordRef.implement({
       nullable: false,
       resolve: (r) => r.coinType,
     }),
-    address: t.exposeString("address", {
+    address: t.expose("address", {
+      type: "Hex",
       description:
-        "The raw address record value, or null if not set. May be any address format for the associated coin type and may require validation or preprocessing before use. Only EVM addresses are guaranteed to be NormalizedAddress values; see ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder).",
+        "The on-chain address record bytes as hex, or null if not set. Decode with ENSIP-9 (https://docs.ens.domains/ensip/9) and address-encoder (https://github.com/ensdomains/address-encoder) for the associated coin type.",
```
lightwalker-eth: Please see my related feedback in other comments 😄

lightwalker-eth (approved):
@sevenzing Nice thank you! Shared a few suggestions, please take the lead to merge this once you're ready 👍

.changeset/neat-dingos-attend.md
sevenzing: ok

packages/ensnode-sdk/src/shared/interpretation/interpret-record-values.ts
sevenzing: it's not possible, viem always returns string in hex format

packages/enssdk/src/omnigraph/generated/schema.graphql
sevenzing: agree.

yeah, it will be `chainName: null`

sevenzing: we dont have `chainId` yet I dont like to write it, but probably I will forget about that later so, okay, let me add it now

PR #2296, opened 2026-06-12
feat: index the Ethereum Follow Protocol (EFP) and expose it via the Omnigraph API

shrugs:
Brings the EFP feature (originally PR #2224, `Quantumlyy/efp-plugin`) onto a `namehash/ensnode` branch, with the Omnigraph API surface refactored.

## What it does

- **ENSIndexer EFP plugin** — indexes EFP list NFTs, records, tags, and account metadata across Base/Optimism/Ethereum into ENSDb's `efp_*` tables (enable via `efp` in `PLUGINS`).
- **Omnigraph API** — exposes EFP under a single nullable `efp` root field:
  - `Query.efp` (protocol-rooted): `list(by:)`, `lists(where:)`, `listRecords(where:)` (each record exposing its owning `list`), cursor-paginated with owner/user/manager and recordData filters.
  - `Account.efp` (account-rooted): validated `primaryList` (EFP two-step user-role check), the social graph `following` / `followers`, the `lists` it is the `user` of, and account `metadata(key:)` / `metadatas`.
  - `EfpListRecord.account` links a record to the `Account` it points at.
  - Both `efp` fields are **null** when the indexer lacks the `efp` plugin.

### Social graph (`Account.efp.following` / `Account.efp.followers`)

- `following` — the address records in the account's **validated primary list**, excluding `block`/`mute`-tagged records; empty when the account has no validated primary list.
- `followers` — the distinct `user`s whose validated primary list holds this account as a non-`block`/`mute` record.
  - Implemented as a single three-table SQL join (`efp_list_records` → `efp_lists` → `efp_account_metadata`) that encodes the EFP two-step Primary List validation directly in SQL.

## Surface refactor (vs. the original #2224 branch)

- `TokenId` scalar (uint256 → decimal string); `tokenId` is a real `bigint` end-to-end, dropping the `idx_tokenId_numeric` workaround index and its numeric-text pagination.
- `by{}` / `where{}` argument pattern across the EFP queries.
- Account-rooted queries moved from `Query.efp` onto `Account.efp` (`primaryList`, `following`, `followers`, `metadata`/`metadatas`).
- Shared EFP id helpers extracted into the `enssdk/efp` submodule (consumed by both ENSIndexer and ENSApi).
- EFP libs relocated to `apps/ensindexer/src/lib/efp/`; `efp.ts` → `query-efp.ts`.

## Areas that would benefit from EFP-expert review

These are deliberate decisions where ENSNode conventions and EFP/`api-v2` behaviour diverge — worth a second opinion from someone who knows EFP semantics:

- **NULL-byte handling is split by field, on purpose.** Account-metadata **keys** are *rejected* when they contain a NULL byte (`interpretMetadataKey` → `hasNullByte`, the standard ENSNode pattern), not stripped: a Postgres `text` column can't store a NULL byte, and stripping would silently collapse distinct on-chain keys onto one stored key (see `AccountMetadata.ts`, `enssdk/efp/ids.ts`). **Tags**, by contrast, still *strip* embedded NULL bytes to match `api-v2` behaviour (`parse-list-op.ts`). Is the tag-strip `api-v2` parity the right call, or should tags also reject? Are there other EFP key/value surfaces where a NULL byte should reject vs. strip?
- **Two-step Primary List validation as a SQL join.** `followersJoins()` / `resolveValidatedPrimaryListTokenId` express "metadata's `primaryListTokenId` = the list's `id` AND the metadata owner = the list's `user`" as join predicates. Confirm this exactly matches the EFP Primary List validation spec (no missing leg, correct handling of re-points and stale metadata).
- **Follow-graph exclusion set.** `EFP_NON_FOLLOW_TAGS = ["block", "mute"]` is the full set of tags that exclude a record from `following`/`followers`. Is that complete and correct per EFP, and should the address record type filter (`EFP_ADDRESS_RECORD_TYPE = 1`) be broadened?
- **Canonical record keying.** Records are keyed on the canonical 22-byte `version | type | address` prefix with trailing junk truncated — confirm this is the right canonicalization for EFP list ops.

## Validation

- `pnpm typecheck`, `pnpm lint`, unit tests, `pnpm generate` (idempotent, SDL/introspection/SKILL regenerated).
- `pnpm test:integration:ci` green with the EFP docker stack.


.changeset/efp-omnigraph.md
```diff
@@ -0,0 +1,6 @@
+---
+"ensapi": minor
+"enssdk": minor
+---
+
+Expose EFP (Ethereum Follow Protocol) data through the Omnigraph API. Protocol-rooted queries live under a single nullable `efp` root field (`Query.efp`, null when the indexer lacks the `efp` plugin): the `EfpQuery` namespace provides `list(by:)`, `lists(where:)`, and `listRecords(where:)` (each record exposing its owning `list`), with cursor-paginated connections and where-filters (owner/user/manager, recordData). Account-rooted EFP data hangs off the ENS graph via the nullable `Account.efp`: an account's validated `primaryList` (applying the EFP two-step user-role check), the `lists` it is the `user` of, and its account `metadata(key:)` / `metadatas`. `EfpListRecord.account` links a record to the `Account` it points at. List NFT token ids are surfaced through a dedicated `TokenId` scalar (a uint256, serialized as a decimal string), and the shared EFP id helpers are published from `enssdk/efp`. Requires the `efp` plugin to be enabled on the indexer.
```
shrugs:
```
Exposes EFP (Ethereum Follow Protocol) data through the Omnigraph API.

`Query.efp` (null if `efp` plugin is not enabled) exposes the EFP-protocol-level queries `Query.efp.list(by:)`, `Query.efp.lists(where:)`, and `Query.efp.listRecords(where:)` (each record exposing its owning `list`), with cursor-paginated connections and where-filters (owner/user/manager, recordData).

`Account.efp` (null if `efp` plugin is not enabled) exposes the Account-specific queries `Account.queries.primaryList` , `Account.efp.lists` it is the `user` of, and its account `Account.efp.metadata(key:)` / `Account.efp.metadatas`.
```

.changeset/efp-plugin.md
```diff
@@ -0,0 +1,8 @@
+---
+"@ensnode/datasources": minor
+"@ensnode/ensdb-sdk": minor
+"@ensnode/ensnode-sdk": minor
+"ensindexer": minor
+---
+
+Add an EFP (Ethereum Follow Protocol) indexer plugin. Enable it by including `efp` in the `PLUGINS` environment variable (on the `mainnet` ENS namespace, or the `ens-test-env` devnet) to index EFP list NFTs, records, tags, and account metadata into ENSDb's `efp_*` tables.
```
shrugs: scope to datasources with comment

```
Introduces the `EFP*` Datasources on the `mainnet` and `ens-test-env` namespaces.
```

scope to ensdb-sdk with comment

```
Introduces the `efp.schema.ts` EFP plugin tables to the Abstract ENSIndexer Schema.
```

scope to ensindexer with comment

```
Introduces the EFP (Ethereum Follow Protocol) plugin (`efp`). Enabling it will index EFP list NFTs, records, tags, and account metadata into ENSDb's `efp_*` tables and, when combined with the `unigraph` plugin will enable the `Query.efp` and `Account.efp` fields in the Omnigraph API.
```

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
```diff
@@ -0,0 +1,127 @@
...
+
+AccountEfpRef.implement({
+  description: "An account's Ethereum Follow Protocol (EFP) presence.",
+  fields: (t) => ({
+    /////////////////////////////
+    // AccountEfp.primaryList
+    /////////////////////////////
+    primaryList: t.field({
+      description:
+        "The account's validated primary EFP list: the list named by its `primary-list` metadata, returned only if that list's `user` role matches the account (the EFP two-step Primary List validation). Null if unset, not indexed, or unvalidated.",
```
shrugs: now that `efp` is nullable due to indexability, this field need not over-specify if "not indexed"

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
```diff
@@ -0,0 +1,127 @@
...
+        const { ensDb, ensIndexerSchema } = di.context;
+        // (address, key) identifies the row; the PK also includes the AccountMetadata contract's
+        // chainId, which the API doesn't carry, so query the indexed columns. Return the full row so
+        // the loadable ref resolves it directly, with no second fetch by id.
+        const [row] = await ensDb
+          .select()
+          .from(ensIndexerSchema.efpAccountMetadata)
+          .where(
+            and(
+              eq(ensIndexerSchema.efpAccountMetadata.address, address as Hex),
```
shrugs: `as Hex` no longer necessary, address is already `NormalizedAddress`

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
```diff
@@ -0,0 +1,127 @@
...
+        'Get one of this account\'s EFP account-metadata values by key (e.g. "primary-list").',
+      type: EfpAccountMetadataRef,
+      nullable: true,
+      args: { key: t.arg({ type: "String", required: true }) },
+      resolve: async (address, args) => {
+        const { ensDb, ensIndexerSchema } = di.context;
+        // (address, key) identifies the row; the PK also includes the AccountMetadata contract's
+        // chainId, which the API doesn't carry, so query the indexed columns. Return the full row so
+        // the loadable ref resolves it directly, with no second fetch by id.
+        const [row] = await ensDb
```
shrugs: perhaps this would benefit from 

```ts
ensIndexerSchema.efpAccountMetadata.query.findFirst(...)
```

or whatever the correct syntax is for drizzle query api

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    ////////////////////////
+    // AccountEfp.metadatas
+    ////////////////////////
+    metadatas: t.connection({
+      description: "All of this account's EFP account-metadata entries.",
+      type: EfpAccountMetadataRef,
+      resolve: (address, args) => {
+        const { ensDb, ensIndexerSchema } = di.context;
+        const scope = eq(ensIndexerSchema.efpAccountMetadata.address, address as Hex);
```
shrugs: `as Hex` not necessary

apps/ensapi/src/omnigraph-api/schema/efp-plugin.ts
```diff
@@ -0,0 +1,11 @@
+import { PluginName } from "@ensnode/ensnode-sdk";
+
+import di from "@/di";
+
+/**
+ * Whether the connected ENSIndexer has the EFP plugin active. The EFP query surface (`Query.efp`,
+ * `Account.efp`) resolves to `null` when it does not, since no `efp_*` data is indexed.
+ */
+export function isEfpPluginEnabled(): boolean {
```
shrugs: extract this helper to an `isPluginEnabled(pluginName: PluginName)` helper and put it in `omnigraph-api/lib/is-plugin-enabled.ts` and find other callsites that could use it

apps/ensapi/src/omnigraph-api/schema/efp-account-metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+//////////////////////
+// EfpAccountMetadata
+//////////////////////
+EfpAccountMetadataRef.implement({
+  description: 'An EFP `(address, key) -> value` account-metadata entry (e.g. "primary-list").',
+  fields: (t) => ({
+    //////////////////////////
+    // EfpAccountMetadata.id
+    //////////////////////////
+    id: t.field({ type: "String", nullable: false, resolve: (metadata) => metadata.id }),
```
shrugs: use `ID` scalar for these IDs

apps/ensapi/src/omnigraph-api/schema/efp-account-metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    ///////////////////////////////
+    // EfpAccountMetadata.chainId
```
shrugs: instead of `EfpAccountMetadata.chainId/address` expose `EfpAccountMetadata.contract: AccountId`

apps/ensapi/src/omnigraph-api/schema/efp-account-metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+    }),
+
+    ///////////////////////////////
+    // EfpAccountMetadata.address
+    ///////////////////////////////
+    address: t.field({
+      description: "The account this metadata belongs to.",
+      type: "Address",
+      nullable: false,
+      resolve: (metadata) => metadata.address as NormalizedAddress,
```
shrugs: as NormalizedAddress likely not necessary

apps/ensapi/src/omnigraph-api/schema/efp-account-metadata.ts
shrugs: and the `as` cast likely not necessary

apps/ensapi/src/omnigraph-api/schema/efp-account-metadata.ts
shrugs: and expand to multiline

apps/ensapi/src/omnigraph-api/schema/efp-account-metadata.ts
```diff
@@ -0,0 +1,94 @@
...
+      description: "The metadata value (raw bytes).",
+      type: "Hex",
+      nullable: false,
+      resolve: (metadata) => metadata.value,
+    }),
+
+    /////////////////////////////////
+    // EfpAccountMetadata.createdAt
+    /////////////////////////////////
+    createdAt: t.field({ type: "BigInt", nullable: false, resolve: (m) => m.createdAt }),
```
shrugs: add desc that it's in Unix Timestamp Seconds

shrugs: likewise for all timestamp fields, description like "When this __ was created, in Unix Timestamp Seconds" etc etc

apps/ensapi/src/omnigraph-api/schema/efp-list-record.ts
```diff
@@ -0,0 +1,163 @@
...
+/////////////////
+// EfpListRecord
+/////////////////
+EfpListRecordRef.implement({
+  description: "A single record within an EFP list (an address it follows), with its tags.",
+  fields: (t) => ({
+    /////////////////////
+    // EfpListRecord.id
+    /////////////////////
+    id: t.field({ type: "String", nullable: false, resolve: (record) => record.id }),
```
shrugs: multiline

apps/ensapi/src/omnigraph-api/schema/efp-list-record.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    //////////////////////////
+    // EfpListRecord.chainId
```
shrugs: likewise here, use `EfpListRecord.contract: AccountId` instead of individual fields

apps/ensapi/src/omnigraph-api/schema/efp-list-record.ts
```diff
@@ -0,0 +1,163 @@
...
+    ///////////////////////
+    list: t.loadable({
+      description: "The EFP list this record belongs to.",
+      type: EfpListRef,
+      nullable: true,
+      // Resolve each record to its storage-location id, then batch the location -> list lookup in
+      // `load` so a page of records resolves all `list` back-refs in two queries rather than one per
+      // node (avoids an N+1 on `efp.listRecords { node { list } }`).
+      resolve: (record) => storageLocationId(record.chainId, record.contractAddress, record.slot),
+      load: async (locationIds: string[]) => {
```
shrugs, line 124: perhaps this logic is better suited for a dataloader?

apps/ensapi/src/omnigraph-api/schema/efp-list-record.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    ///////////////////////////
+    // EfpListRecord.account
+    ///////////////////////////
+    account: t.field({
+      description:
+        "The Account this record points to (its `recordData`). Always resolvable: an Account exists for any address (see `query.account`), so a record's target can always be walked into its ENS names and own EFP presence.",
+      type: AccountRef,
+      nullable: false,
+      resolve: (record) => record.recordData as NormalizedAddress,
```
shrugs: `as` cast not necessary

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,199 @@
...
+
+export const EfpListRef = builder.loadableObjectRef("EfpList", {
+  load: (tokenIds: TokenId[]) => {
+    const { ensDb, ensIndexerSchema } = di.context;
+    return ensDb
+      .select()
+      .from(ensIndexerSchema.efpLists)
+      .where(inArray(ensIndexerSchema.efpLists.tokenId, tokenIds));
+  },
+  // efp_lists is keyed by `tokenId`, not a synthetic `id`.
```
shrugs: drop comment

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,199 @@
...
+  toKey: (list) => list.tokenId,
+  cacheResolved: true,
+  sort: true,
+});
+
+export type EfpList = Exclude<typeof EfpListRef.$inferType, string>;
+
+/**
+ * Connection args paginated by a list NFT's `tokenId` (efp_lists has no synthetic `id` column).
```
shrugs: this would all be simpler if the efp_lists had an `id` column that was TokenId, yes? if so just make that refactor

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,199 @@
...
+    }),
+
+    ////////////////
+    // EfpList.owner
+    ////////////////
+    owner: t.field({
+      description: "The current ERC-721 owner of the list NFT.",
+      type: "Address",
+      nullable: false,
+      resolve: (list) => list.owner as NormalizedAddress,
```
shrugs: all of these `as` casts likely unnecessary

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,199 @@
...
+      description: "The address allowed to administer this list.",
+      type: "Address",
+      nullable: true,
+      resolve: (list) => (list.manager ?? null) as NormalizedAddress | null,
+    }),
+
+    /////////////////////
+    // EfpList.nftChainId
+    /////////////////////
+    nftChainId: t.field({
```
shrugs: `EfpList.contract: AccountId` or `EfpList.nft: AccountId`

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,199 @@
...
+    /////////////////////////////
+    nftContractAddress: t.field({
+      description: "The ListRegistry contract address.",
+      type: "Address",
+      nullable: false,
+      resolve: (list) => list.nftContractAddress as NormalizedAddress,
+    }),
+
+    //////////////////////////////////////
+    // EfpList.listStorageLocationChainId
```
shrugs: likewise, `EfpList.listStorageLocation: AccountId`

apps/ensapi/src/omnigraph-api/schema/efp-primary-list.ts
```diff
@@ -0,0 +1,65 @@
...
+const PRIMARY_LIST_VALUE_HEX_LENGTH = 2 + 32 * 2;
+
+/**
+ * Decode a `primary-list` account-metadata value (`abi.encodePacked(uint256 tokenId)`) into a
+ * token id, or `null` if it isn't well-formed. EFP defines the value as exactly a 32-byte uint256,
+ * so reject any other length rather than let `BigInt` coerce a malformed value (e.g. `0x01`) into a
+ * real token id.
+ */
+export function decodePrimaryListTokenId(value: Hex): TokenId | null {
+  if (value.length !== PRIMARY_LIST_VALUE_HEX_LENGTH) return null;
```
shrugs: use `viem#size(value) === PRIMARY_LIST_VALUE_SIZE` and convert `PRIMARY_LIST_VALUE_HEX_LENGTH` to `PRIMARY_LIST_VALUE_SIZE = 32`

apps/ensapi/src/omnigraph-api/schema/efp-primary-list.ts
```diff
@@ -0,0 +1,65 @@
...
+  try {
+    return BigInt(value) as TokenId;
```
shrugs: `viem#hexToBigint as TokenId`

apps/ensapi/src/omnigraph-api/schema/efp-primary-list.ts
```diff
@@ -0,0 +1,65 @@
...
+ * two-step Primary List validation). `null` if unset, not indexed, or unvalidated.
+ *
+ * Used by `Account.efp.primaryList`.
+ */
+export async function resolveValidatedPrimaryListTokenId(
+  address: NormalizedAddress,
+): Promise<TokenId | null> {
+  const { ensDb, ensIndexerSchema } = di.context;
+
+  const [metadata] = await ensDb
```
shrugs: likewise also benefits from drizzle query api using `.findFirst`

apps/ensapi/src/omnigraph-api/schema/efp-primary-list.ts
```diff
@@ -0,0 +1,65 @@
...
+    )
+    .limit(1);
+  if (!metadata) return null;
+
+  const tokenId = decodePrimaryListTokenId(metadata.value);
+  if (tokenId === null) return null;
+
+  // EFP "Primary List" is only valid when the named list's `user` role matches the account.
+  // Compare case-insensitively so validation is independent of input casing.
+  const [list] = await ensDb
```
shrugs: `findFirst`

apps/ensapi/src/omnigraph-api/schema/efp-primary-list.ts
```diff
@@ -0,0 +1,65 @@
...
+    .select({ user: ensIndexerSchema.efpLists.user })
+    .from(ensIndexerSchema.efpLists)
+    .where(eq(ensIndexerSchema.efpLists.tokenId, tokenId))
+    .limit(1);
+  if (!list?.user || list.user.toLowerCase() !== address.toLowerCase()) return null;
```
shrugs: no need to lowercase compare, trust the NormalizedAddress type

apps/ensapi/src/omnigraph-api/schema/query-efp.ts
```diff
@@ -0,0 +1,167 @@
...
+    ///////////////
+    lists: t.connection({
+      description: "Find EFP lists, optionally filtered by owner / user / manager.",
+      type: EfpListRef,
+      args: { where: t.arg({ type: EfpListsWhereInput }) },
+      resolve: (_parent, args) => {
+        const { ensDb, ensIndexerSchema } = di.context;
+        const where = args.where;
+        const scope = and(
+          where?.owner ? eq(ensIndexerSchema.efpLists.owner, where.owner as Hex) : undefined,
```
shrugs: `as` casts not necessary

apps/ensapi/src/omnigraph-api/schema/query-efp.ts
```diff
@@ -0,0 +1,167 @@
...
+      description:
+        "Find EFP list records. Filter by `recordData` to answer 'which lists follow this address?'.",
+      type: EfpListRecordRef,
+      args: { where: t.arg({ type: EfpListRecordsWhereInput }) },
+      resolve: (_parent, args) => {
+        const { ensDb, ensIndexerSchema } = di.context;
+        const where = args.where;
+        const scope = and(
+          where?.recordData
+            ? eq(ensIndexerSchema.efpListRecords.recordData, where.recordData as Hex)
```
shrugs: `as` cast not necessary

apps/ensapi/src/omnigraph-api/schema/scalars.ts
```diff
@@ -167,6 +168,13 @@ builder.scalarType("ChainId", {
   parseValue: (value) => makeChainIdSchema("ChainId").parse(value),
 });
 
+builder.scalarType("TokenId", {
+  description:
+    "TokenId represents an enssdk#TokenId: an ERC-721 token id (a uint256), serialized as a decimal string.",
```
shrugs: drop the ERC721 ref, just say `TokenId represents an enssdk#TokenId` and leave it at that

apps/ensindexer/src/lib/efp/constants.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
shrugs, line 1: perhaps these constants want to move to the `enssdk/efp` submodule

apps/ensindexer/src/lib/efp/list-metadata.ts
```diff
@@ -0,0 +1,16 @@
+import { type NormalizedAddress, toNormalizedAddress } from "enssdk";
+import type { Hex } from "viem";
+
+/**
+ * Interpret an EFP list-metadata `value` as a role address. The well-known `user` / `manager` keys
+ * carry exactly a 20-byte address; the generic metadata setter can emit arbitrary bytes, so any
+ * other length is malformed and returns `null` to clear the role rather than store a truncated or
+ * empty address (which would later surface through a GraphQL `Address`).
+ */
+export function metadataValueToAddress(value: Hex): NormalizedAddress | null {
```
shrugs: rename to `interpretMetadataValue`

apps/ensindexer/src/lib/efp/parse-list-storage-location.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  it("decodes a well-formed 86-byte onchain (locationType=1) payload", () => {
+    const chainIdHex = (8453).toString(16).padStart(64, "0");
+    const addressHex = "ab".repeat(20);
+    const slotHex = "cd".repeat(32);
+    const payload = `0x0101${chainIdHex}${addressHex}${slotHex}` as `0x${string}`;
+
+    expect(parseListStorageLocation(payload)).toEqual({
+      version: 1,
+      chainId: 8453,
+      contractAddress: `0x${addressHex}` as `0x${string}`,
```
shrugs: use `viem#Address` and `viem#Hex` semantic types throughout this file

apps/ensindexer/src/plugins/efp/handlers/AccountMetadata.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Registers the EFP `AccountMetadata` event handler (UpdateAccountMetadata).
+ */
+export default function () {
+  // UpdateAccountMetadata — writes a single (key, value) pair for an account (today: `primary-list`).
+  addOnchainEventListener(
+    namespaceContract(pluginName, "AccountMetadata:UpdateAccountMetadata"),
+    async ({ context, event }) => {
+      const ts = event.block.timestamp;
+      const address = event.args.addr.toLowerCase() as Hex;
```
shrugs: just `const {addr: address, key} = event.args` and type the event as EventWithArgs<{...}> correctly, follow this pattern for all indexer handlers

apps/ensindexer/src/plugins/efp/handlers/AccountMetadata.ts
```diff
@@ -0,0 +1,41 @@
...
+      // `key` is a free-form on-chain string used as a primary-key component. Strip NUL bytes: a
+      // Postgres text column rejects them (a NUL key would crash the insert), and the tag path
+      // strips them too for api-v2 parity.
```
shrugs: replace with an `interpretMetadataKey` helper in efp-account-metadata.ts and that function should do `if hasNullBytes(...) return null` (using existing has null bytes helper)

apps/ensindexer/src/plugins/efp/handlers/AccountMetadata.ts
```diff
@@ -0,0 +1,41 @@
...
+      const key = event.args.key.replace(/\0/g, "");
+
+      await context.ensDb
+        .insert(ensIndexerSchema.efpAccountMetadata)
+        .values({
+          id: accountMetadataId(context.chain.id, address, key),
+          chainId: context.chain.id,
+          contractAddress: event.log.address.toLowerCase() as Hex,
```
shrugs: STOP lowercasing and `as` casting, trust the types

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: Done — renamed the `efp_lists` primary key to `id` (kept `bigint`/`TokenId`), so `EfpListRef` now uses the standard `getModelId` toKey and the special-case comment is gone. Pagination stays `paginateByBigInt` since `id` is still a bigint, preserving the numeric ordering the >9-lists test guards.

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: A bare `AccountId` would drop the slot, so I added an `EfpListStorageLocation { chainId, address, slot }` type and exposed it as the nullable `EfpList.storageLocation` (replacing the three flat `listStorageLocation*` fields).

apps/ensapi/src/omnigraph-api/schema/efp-list-record.ts
shrugs: This already is a batched dataloader: `t.loadable` with a `load` that resolves a page of records' `list` back-refs in two queries (location→tokenId, then tokenId→list), so there is no N+1. Left as-is.

apps/ensindexer/src/lib/efp/constants.ts
shrugs: These are indexer-internal parse constants (op/record/LSL versions, opcodes, well-known list-metadata keys), not part of the writer/reader wire contract that `enssdk/efp` owns — keeping them local to the EFP plugin.

apps/ensindexer/src/plugins/efp/handlers/AccountMetadata.ts
shrugs: Destructured the args. Note the event is already strongly typed via `addOnchainEventListener<EventName>` (it infers `event.args` from the ABI), so an explicit `EventWithArgs<{…}>` annotation isn't needed on these EFP handlers.

apps/ensapi/src/omnigraph-api/schema/query-efp.ts
```diff
@@ -0,0 +1,167 @@
...
+
+        return lazyConnection({
+          totalCount: () => ensDb.$count(ensIndexerSchema.efpListRecords, scope),
+          connection: () =>
+            resolveCursorConnection(
+              { ...ID_PAGINATED_CONNECTION_ARGS, args },
+              ({ before, after, limit, inverted }: ResolveCursorConnectionArgs) =>
+                ensDb
+                  .select()
+                  .from(ensIndexerSchema.efpListRecords)
```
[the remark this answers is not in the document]
shrugs: `totalCount` is lazy (only computed when the field is selected, via `lazyConnection`), so the unfiltered `COUNT(*)` is not paid by default. Leaving `where` optional is intentional for now; will revisit a required-filter / omit-totalCount guard if mainnet-scale profiling shows it is needed.

apps/ensapi/src/omnigraph-api/schema/efp.integration.test.ts
```diff
@@ -0,0 +1,312 @@
...
+describe("EFP handler edge cases (seeded)", () => {
+  type SeededRecordsResult = {
+    efp: {
+      listRecords: GraphQLConnection<{
+        recordData: string;
+        tags: string[];
+        account: { id: string };
+        list: { user: string | null } | null;
+      }>;
+    };
```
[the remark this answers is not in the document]
shrugs: Removed the magic constants — the deep-walk now asserts `tokenId` matches `/^\d+$/` plus the structural walk (it reaches bob and his own validated primary list), so it no longer depends on demoGraph mint order.

apps/ensapi/src/omnigraph-api/schema/scalars.ts
```diff
@@ -167,6 +168,13 @@ builder.scalarType("ChainId", {
...
+  serialize: (value: TokenId) => value.toString(),
+  parseValue: (value) => z.coerce.bigint().parse(value) as TokenId,
+});
```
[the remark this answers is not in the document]
shrugs: Tightened `TokenId.parseValue`: it now coerces to a string and requires `/^\d+$/` (non-negative decimal), so hex / signed / fractional inputs are rejected before reaching a resolver or DB query.

apps/ensindexer/src/lib/efp/parse-list-op.ts
```diff
@@ -0,0 +1,146 @@
...
+  // type checked, lowercased), so the tag keys into the same row the record was stored under.
+  const parsed = parseRecord(data.slice(0, RECORD_PREFIX_WITH_0X_LENGTH) as Hex);
+  if (!parsed) return null;
+
+  const tagHex = data.slice(RECORD_PREFIX_WITH_0X_LENGTH);
+  // hex must contain whole bytes
+  if (tagHex.length % 2 !== 0) return null;
+
+  // Match api-v2: decode as UTF-8 and strip embedded NULs.
+  const tag = hexToString(`0x${tagHex}`).replace(/\0/g, "");
```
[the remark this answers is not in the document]
shrugs: Added a generous 512-byte cap in `parseTagOp` — a tag over the limit is skipped (returns null) rather than indexed, bounding row size against an arbitrary on-chain payload.

apps/ensapi/src/omnigraph-api/schema/efp-follows.ts
```diff
@@ -0,0 +1,231 @@
...
+      // A list holds `target` when `target` is a record at the list's decoded storage location.
+      .innerJoin(
+        efpLists,
+        and(
+          eq(efpLists.listStorageLocationChainId, efpListRecords.chainId),
+          eq(efpLists.listStorageLocationContractAddress, efpListRecords.contractAddress),
+          eq(efpLists.listStorageLocationSlot, efpListRecords.slot),
+        ),
+      )
+      .leftJoin(
```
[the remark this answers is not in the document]
shrugs: False positive here: this join keys on the constant `EFP_PRIMARY_LIST_KEY` ("primary-list"), not user input, so there is no NUL-byte exposure. The user-facing `Account.efp.metadata(key:)` resolver does reject NUL-byte keys via `interpretMetadataKey`, matching the indexer which now rejects them on write too.

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
```diff
@@ -0,0 +1,196 @@
...
+                    ),
+                  )
+                  .orderBy(orderPaginationBy(ensIndexerSchema.efpListRecords.recordData, inverted))
+                  .limit(limit)
+                  .then((rows) => rows.map((row) => row.recordData as NormalizedAddress)),
+            );
+          },
+        });
+      },
+    }),
```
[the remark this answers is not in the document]
shrugs: Done — `followingScope` now caches the resolved `SQL` (async wrapper) instead of the promise.

apps/ensapi/src/omnigraph-api/schema/efp-follows.ts
```diff
@@ -0,0 +1,231 @@
...
+          isNotNull(efpLists.user),
+          cursor.after ? gt(efpLists.user, cursor.after as Hex) : undefined,
+          cursor.before ? lt(efpLists.user, cursor.before as Hex) : undefined,
+        ),
+      )
+      .groupBy(efpLists.user, efpAccountMetadata.value)
+      .orderBy(inverted ? desc(efpLists.user) : asc(efpLists.user))
+      .limit(limit) as Promise<FollowerCandidate[]>
+  );
+}
```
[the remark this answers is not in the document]
shrugs: Done — removed the `as Hex` casts on `cursor.after`/`cursor.before` (and on `target`); they are already `NormalizedAddress`, matching the columns.

apps/ensindexer/src/plugins/efp/handlers/ListRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        const address = metadataValueToAddress(meta.value);
+        await context.ensDb
+          .update(ensIndexerSchema.efpLists, { tokenId })
+          .set(
+            key === EFP_LIST_METADATA_KEYS.USER
+              ? { user: address, updatedAt: ts }
+              : { manager: address, updatedAt: ts },
+          );
+      }
```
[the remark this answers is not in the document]
shrugs: Leaving as-is: only two keys (USER/MANAGER), and Ponder serializes handlers per chain inside a transaction, so the extra round-trip is negligible and batching buys no consistency benefit.

packages/datasources/src/ens-test-env.ts
```diff
@@ -157,4 +161,62 @@ export default {
...
+  [DatasourceNames.EFPEthereum]: {
+    chain: ensTestEnvChain,
+    contracts: {
+      ListRecords: {
+        abi: efp_ListRecords,
+        address: efpContracts.EFPListRecords,
+        startBlock: 0,
+      },
+    },
+  },
```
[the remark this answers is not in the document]
shrugs: Leaving as-is: all three devnet `ListRecords` datasources reference the same `efpContracts.EFPListRecords` constant, so they cannot diverge without an obvious source edit — a runtime assertion would guard a case the shared reference already prevents.

packages/datasources/src/mainnet.ts
```diff
@@ -502,4 +506,74 @@ export default {
...
+      // multiple chains for different contracts. Confirmed as ListRecords on Ethereum mainnet at
+      // https://docs.efp.app/production/deployments/ (and ethereumfollowprotocol/api-v2); it is NOT
+      // a copy-paste of the Base AccountMetadata address.
+      ListRecords: {
+        abi: efp_ListRecords,
+        address: "0x5289fe5dabc021d02fddf23d4a4df96f4e0f17ef",
+        startBlock: 20820743,
+      },
+    },
+  },
```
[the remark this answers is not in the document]
shrugs: False positive — verified against a Base RPC via `eth_getCode`: the EFPListRegistry (0x0e688f…) has no code at block 10721682 or 11000000 and is first deployed at exactly block 20197231 (empty at 20197230). The PR's `startBlock: 20197231` is correct; the cited BaseScan block looks like a web-search hallucination.

apps/ensindexer/src/lib/efp/parse-list-op.ts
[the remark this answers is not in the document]
shrugs: Already addressed in 5798ffe03 — `parseTagOp` now caps the tag at 512 bytes and skips anything larger.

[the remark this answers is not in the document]
shrugs: Already addressed in 5798ffe03 — `parseTagOp` now caps the tag at 512 bytes and skips anything larger.

apps/ensapi/src/omnigraph-api/schema/scalars.ts
```diff
@@ -167,6 +168,12 @@ builder.scalarType("ChainId", {
   parseValue: (value) => makeChainIdSchema("ChainId").parse(value),
 });
 
+builder.scalarType("TokenId", {
+  description: "TokenId represents an enssdk#TokenId.",
+  serialize: (value: TokenId) => value.toString(),
+  parseValue: (value) => z.coerce.bigint().parse(value) as TokenId,
```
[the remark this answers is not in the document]
shrugs: Already addressed in 5798ffe03 — `TokenId.parseValue` now requires a non-negative decimal string (`/^\d+$/`), rejecting hex and negatives (equivalent to this suggestion).

shrugs:
@greptile review

apps/ensindexer/src/plugins/efp/handlers/ListRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+      });
+
+      // Claim this slot's reverse mapping for the list. The slot is attacker-settable, so a
+      // different list may already own this mapping; `onConflictDoNothing` keeps first-writer-wins —
+      // an existing mapping is never overwritten — matching `deleteStorageLocationIfOwnedBy`.
+      await context.ensDb
+        .insert(ensIndexerSchema.efpListStorageLocations)
+        .values({ id: newLocationId, chainId, contractAddress, slot, tokenId, updatedAt: ts })
+        .onConflictDoNothing();
+
```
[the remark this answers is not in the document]
shrugs: By design and documented: EFP `user`/`manager` roles are per-storage-location, so a list pointing at a slot correctly reflects that slot's current roles at repoint time. The reverse-map (first-writer-wins) only governs which list future `UpdateListMetadata` events re-route to — the schema documents this ("when lists share a slot, only the owner's role routing tracks that slot"), and the role-durability integration test covers it. Shared slots are an attacker-settable edge case; keeping as-is.

packages/enssdk/src/efp/ids.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/** `efp_account_metadata` key: a `(chainId, address, key)` tuple (NUL bytes stripped from the key). */
+export function accountMetadataId(
+  chainId: ChainId,
+  address: NormalizedAddress,
+  key: string,
+): string {
+  return [chainId, address, key.replace(/\0/g, "")].join("-");
+}
```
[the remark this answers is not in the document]
shrugs: Fixed in 3923c6c8e — removed the NUL strip from `accountMetadataId`. Keys are now rejected upstream via `interpretMetadataKey` (reject, not strip), so the helper trusts its already-validated input.

packages/enssdk/src/lib/coin-type.ts
```diff
@@ -66,3 +66,19 @@ export const bigintToCoinType = (value: bigint): CoinType => {
...
+ *
+ * @throws if `value` is too large to fit in Number.MAX_SAFE_INTEGER
+ */
+export const bigintToChainId = (value: bigint): ChainId => {
+  if (value > BigInt(Number.MAX_SAFE_INTEGER)) {
+    throw new Error(`'${value}' cannot represent as ChainId, it is too large.`);
+  }
+
+  return Number(value) as ChainId;
+};
```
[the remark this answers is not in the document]
shrugs: Fixed in 3923c6c8e — `bigintToChainId` now throws for non-positive values (`<= 0n` guard), enforcing the positive-integer ChainId contract, with test coverage for `0n`/`-1n`.

shrugs:
@greptile review

shrugs:
@greptile review

apps/ensindexer/src/plugins/efp/handlers/ListRegistry.ts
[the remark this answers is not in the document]
shrugs: Same as the Copilot thread on this handler: by design and documented. EFP `user`/`manager` roles are per-storage-location, so a list pointing at a slot correctly reflects that slot's current roles at repoint time. The reverse-map (first-writer-wins) only governs which list future `UpdateListMetadata` events re-route to; the schema documents this ("when lists share a slot, only the owner's role routing tracks that slot") and the role-durability integration test covers it. Shared slots are an attacker-settable edge case. Keeping as-is.

apps/ensindexer/src/plugins/efp/handlers/ListRegistry.ts
```diff
@@ -0,0 +1,202 @@
...
+      // (Re-)apply this storage location's durable user/manager metadata to the list. Keyed by
+      // location, so it restores roles whenever a list points at (or re-points to) a slot whose
+      // metadata was already recorded; it is not deleted, as it stays valid for the slot. Combined
+      // with the role clear on a move above, this rederives the list's roles from its location.
+      for (const key of [EFP_LIST_METADATA_KEYS.USER, EFP_LIST_METADATA_KEYS.MANAGER] as const) {
+        const meta = await context.ensDb.find(ensIndexerSchema.efpListMetadata, {
```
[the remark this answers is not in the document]
shrugs: Duplicate of the thread above / the Copilot thread on this handler — this is the intended, documented per-storage-location role behavior (the reverse-map first-writer-wins only governs future `UpdateListMetadata` routing). Keeping as-is.

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
```diff
@@ -0,0 +1,189 @@
...
+    following: t.connection({
+      description:
+        "The accounts this account follows: the address records in its validated primary EFP list, excluding `block`/`mute`-tagged records. Empty when the account has no validated primary list.",
+      type: AccountRef,
+      resolve: (address, args) => {
+        const { ensDb, ensIndexerSchema } = di.context;
+        // `following` derives from the account's primary list; resolve that scope once and share it
+        // between `totalCount` and the page query.
+        let scope: SQL | null = null;
+        const followingScope = async () => {
```
shrugs: no need to make this lazy, if someone's selecting this connection they're choosing at least one of the subfields both of which depend on this scope, just await it inline

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
shrugs: can we not just return the account metadata ref id and let the loadable pull the row?

apps/ensapi/src/omnigraph-api/schema/efp-follows.ts
```diff
@@ -0,0 +1,252 @@
...
+export const EFP_ADDRESS_RECORD_TYPE = 1;
+
+/**
+ * EFP record tags that exclude a record from the social follow graph: a `block`ed or `mute`d address
+ * is present in a list but is not a "follow". `following` / `followers` omit any record carrying one.
+ */
+export const EFP_NON_FOLLOW_TAGS = ["block", "mute"] as const;
+
+/** Relay connection args for a connection of account addresses, cursored by the address itself. */
+export const ADDRESS_PAGINATED_CONNECTION_ARGS = {
```
shrugs: wants to live in `apps/ensapi/src/omnigraph-api/schema/constants.ts`?

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,217 @@
...
+  },
+  toKey: getModelId,
+  cacheResolved: true,
+  sort: true,
+});
+
+export type EfpList = Exclude<typeof EfpListRef.$inferType, string>;
+
+/** Connection args paginated by a list NFT's token id (the `efp_lists` primary key). */
+export const TOKEN_ID_PAGINATED_CONNECTION_ARGS = {
```
shrugs: likewise maybe lives in `apps/ensapi/src/omnigraph-api/schema/constants.ts`

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,217 @@
...
+// EfpList
+///////////
+EfpListRef.implement({
+  description: "An EFP list NFT (a ListRegistry token) and the records it holds.",
+  fields: (t) => ({
+    //////////////////
+    // EfpList.tokenId
+    //////////////////
+    tokenId: t.field({
+      description: "The ERC-721 token id of the list NFT.",
```
shrugs: "TokenId" not "token id"

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,217 @@
...
+      nullable: true,
+      resolve: (list) => list.manager,
+    }),
+
+    //////////////
+    // EfpList.nft
+    //////////////
+    nft: t.field({
+      description:
+        "The list NFT's CAIP-10 account id: the ListRegistry contract and the chain it lives on (Base / 8453 on the mainnet namespace; otherwise the active namespace's EFP deployment chain, e.g. 31337 on the ens-test-env devnet).",
```
shrugs: just `The list NFT's AccountId`

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,217 @@
...
+      nullable: false,
+      resolve: (list) => ({ chainId: list.nftChainId, address: list.nftContractAddress }),
+    }),
+
+    /////////////////////////
+    // EfpList.storageLocation
+    /////////////////////////
+    storageLocation: t.field({
+      description:
+        "The decoded onchain location of this list's records. Null until the list emits an `UpdateListStorageLocation` (or when its payload is undecodable).",
```
shrugs: just `The decoded onchain location of this list's records, or null if not known.`

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
```diff
@@ -0,0 +1,217 @@
...
+    records: t.connection({
+      description: "The records currently in this list (the addresses it follows).",
+      type: EfpListRecordRef,
+      resolve: (list, args) => {
+        const { ensDb, ensIndexerSchema } = di.context;
+
+        // A list's records live at its decoded onchain storage location. If the list has not yet
+        // emitted an UpdateListStorageLocation, it has no records to resolve.
+        if (
+          list.listStorageLocationChainId === null ||
```
shrugs: we do this triple null check at least 3 times already, DRY it up

apps/ensapi/src/omnigraph-api/schema/account-efp.ts
shrugs: Done — `following`'s resolver is now `async` and awaits `buildFollowingScope(address)` once up front; both `totalCount` and the page query share the resolved scope (no lazy closure).

apps/ensapi/src/omnigraph-api/schema/efp-follows.ts
shrugs: Moved to `apps/ensapi/src/omnigraph-api/schema/constants.ts`.

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: Moved to `apps/ensapi/src/omnigraph-api/schema/constants.ts` (alongside `ADDRESS_PAGINATED_CONNECTION_ARGS`).

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: Done — "TokenId".

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: Done — `nft` is now "The list NFT's AccountId."

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: Done — "The decoded onchain location of this list's records, or null if not known."

apps/ensapi/src/omnigraph-api/schema/efp-list.ts
shrugs: Done — extracted `decodedStorageLocation(list)` (returns the `{chainId, contractAddress, slot}` tuple or null), now used by `EfpList.storageLocation`, `EfpList.records`, and `buildFollowingScope`. Note this whole area was rewritten per your other thread: `primary-list` is decoded to a numeric `efp_account_metadata.primaryListTokenId` at index time, so `followers`/`following` are now a single SQL join with no in-app enumeration.

.changeset/efp-ensindexer.md
```diff
@@ -0,0 +1,5 @@
+---
+"ensindexer": minor
+---
+
+Introduces the EFP (Ethereum Follow Protocol) plugin (`efp`). Enabling it will index EFP list NFTs, records, tags, and account metadata into ENSDb's `efp_*` tables and, when combined with the `unigraph` plugin, will enable the `Query.efp` and `Account.efp` fields in the Omnigraph API.
```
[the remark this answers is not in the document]
shrugs: Fixed — `efp` alone enables `Query.efp`/`Account.efp` (`hasOmnigraphApiConfigSupport` treats `efp` as sufficient), so the "when combined with unigraph" clause is removed.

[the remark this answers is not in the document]
shrugs: Fixed — removed the incorrect "when combined with unigraph" clause; the `efp` plugin alone enables these fields.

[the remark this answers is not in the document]
shrugs: Fixed — removed the incorrect "when combined with unigraph" clause; the `efp` plugin alone enables these fields.

apps/ensindexer/src/lib/efp/parse-list-op.ts
```diff
@@ -0,0 +1,156 @@
...
+  /** Record type (1 = 20-byte address). */
+  recordType: number;
+  /**
+   * Canonical record bytes `recordVersion | recordType | address` (0x-prefixed, exactly 22 bytes),
+   * with any trailing junk after the 20-byte address truncated. Records are keyed by this so tag
+   * and remove ops, which carry the same 22-byte prefix, resolve to the same row.
+   */
+  record: Hex;
+  /** Record payload — for `recordType === 1`, the 20-byte followed address. */
+  recordData: NormalizedAddress;
```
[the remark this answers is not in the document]
shrugs: Done — "the 20-byte address".

apps/ensindexer/src/lib/efp/parse-list-op.ts
```diff
@@ -0,0 +1,153 @@
...
+  version: number;
+  /** Record type (1 = 20-byte address). */
+  recordType: number;
+  /**
+   * Canonical record bytes `recordVersion | recordType | address` (0x-prefixed, exactly 22 bytes),
+   * with any trailing junk after the 20-byte address truncated. Records are keyed by this so tag
+   * and remove ops, which carry the same 22-byte prefix, resolve to the same row.
+   */
+  record: Hex;
+  /** Record payload — for `recordType === 1`, the 20-byte followed address. */
```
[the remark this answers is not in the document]
shrugs: Done — "the 20-byte address".

shrugs:
@greptile review

Quantumlyy:
Starting review.
One thought I had was being able to have it run with just 3/6 chains (Ethereum Mainnet, Base, Optimism).

apps/ensapi/src/omnigraph-api/schema/scalars.ts
```diff
@@ -167,6 +168,27 @@ builder.scalarType("ChainId", {
...
+        if (!/^\d+$/.test(ctx.value)) {
+          ctx.issues.push({
+            code: "custom",
+            message: "TokenId must be a non-negative uint256 (decimal string)",
+            input: ctx.value,
+          });
+        }
+      })
+      .transform((val) => BigInt(val) as TokenId)
+      .parse(value),
```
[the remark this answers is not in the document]
shrugs: Good catch — added a uint256 upper-bound check (reject `> 2^256 - 1`) alongside the existing digit-only check so the scalar contract is honest. Fixed in 1a5b1e82a.

shrugs:
@greptile review

shrugs:
@greptile review

packages/enssdk/src/lib/coin-type.ts
```diff
@@ -66,3 +66,22 @@ export const bigintToCoinType = (value: bigint): CoinType => {
...
+ *
+ * @throws if `value` is not a positive integer, or is too large to fit in Number.MAX_SAFE_INTEGER
+ */
+export const bigintToChainId = (value: bigint): ChainId => {
+  if (value <= 0n) {
+    throw new Error(`'${value}' cannot represent a ChainId; it must be a positive integer.`);
+  }
+  if (value > BigInt(Number.MAX_SAFE_INTEGER)) {
+    throw new Error(`'${value}' cannot represent as ChainId, it is too large.`);
+  }
```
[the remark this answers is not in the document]
shrugs: Fixed in 7ba8ed5a6 — reworded both the ChainId and the sibling CoinType too-large errors to "cannot be represented as a {CoinType,ChainId}; it is too large." for consistent, correct phrasing.

apps/ensapi/src/omnigraph-api/schema/account.ts
```diff
@@ -2,15 +2,19 @@ import { type ResolveCursorConnectionArgs, resolveCursorConnection } from "@poth
...
 import { builder } from "@/omnigraph-api/builder";
 import { orderPaginationBy, paginateBy } from "@/omnigraph-api/lib/connection-helpers";
 import { resolveFindDomains } from "@/omnigraph-api/lib/find-domains/find-domains-resolver";
 import { resolveFindEvents } from "@/omnigraph-api/lib/find-events/find-events-resolver";
 import { resolveAccountNameReferences } from "@/omnigraph-api/lib/find-name-references/find-name-references-resolver";
+import { isPluginEnabled } from "@/omnigraph-api/lib/is-plugin-enabled";
 import { lazyConnection } from "@/omnigraph-api/lib/lazy-connection";
 import { buildAccountPrimaryNamesSelection } from "@/omnigraph-api/lib/resolution/account-primary-names-selection";
 import { resolvePrimaryNameRecords } from "@/omnigraph-api/lib/resolution/resolve-primary-name-records";
+import { AccountEfpRef } from "@/omnigraph-api/schema/account-efp";
```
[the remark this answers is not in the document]
shrugs: Resolving as a non-issue: this is the standard Pothos objectRef pattern. AccountRef and AccountEfpRef are each created eagerly via builder.objectRef(...) with no cross-reference, and are only dereferenced inside their lazy fields: (t) => ({...}) callbacks, which Pothos evaluates at schema-build time after all modules have loaded — so there is no partially-initialized-export hazard. This bidirectional import is already pervasive across the schema/ directory (domain.ts, permissions.ts, profile.ts, registration.ts, etc. all import AccountRef while account.ts imports them back).

packages/ensdb-sdk/src/ensindexer-abstract/efp.schema.ts
```diff
@@ -0,0 +1,185 @@
...
+     * `(chainId, contract, slot)`; this records the FIRST list to claim it (first writer wins), and
+     * the `UpdateListStorageLocation` handler gates every write/delete of this row on that ownership.
+     * A consequence: when lists share a slot, only the owner's `EfpListRecord.list` back-ref and
+     * `user`/`manager` role routing track that slot.
+     */
+    tokenId: t.bigint().notNull().$type<TokenId>(),
+    updatedAt: t.bigint().notNull().$type<DurationBigInt>(),
+  }),
+  (t) => ({
+    idx_tokenId: index().on(t.tokenId),
```
Quantumlyy: nothing queries by `tokenId` here as far as I can tell, it's only ever looked up by `id`. looks like a dead index.

apps/ensindexer/src/lib/efp/parse-list-op.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+/**
+ * Decode an ADD_TAG / REMOVE_TAG payload, where the data layout is
+ * `record (22 bytes) | tag (UTF-8 bytes)`. Returns `null` when the record prefix is missing or is
+ * not a valid address record.
+ */
+export function parseTagOp(data: Hex | string | null | undefined): ParsedTagOp | null {
+  if (!data || typeof data !== "string" || !isHex(data)) return null;
+  if (data.length < RECORD_PREFIX_WITH_0X_LENGTH) return null;
```
Quantumlyy, line 128: < lets a prefix-with-no-tag op through and stores an empty tag. <=, or reject when `tagHex` is empty.

apps/ensapi/src/omnigraph-api/schema/account.ts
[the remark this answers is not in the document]
shrugs: False positive. Both AccountRef (in account-efp.ts) and AccountEfpRef (here) are referenced only inside lazy Pothos `fields: (t) => ...` callbacks, which Pothos evaluates at schema-build time, not module-load time — so the circular import resolves fine. Confirmed: the generated SDL contains `type AccountEfp { ... }` and the EFP integration tests pass. Resolving.

packages/ensdb-sdk/src/ensindexer-abstract/efp.schema.ts
shrugs: Good catch — confirmed tokenId is only ever SELECTed, every lookup/delete/join uses the id PK. Removed idx_tokenId in 64f981626.

apps/ensindexer/src/lib/efp/parse-list-op.ts
shrugs: Fixed in 64f981626. Rejected the empty decoded tag (`if (tag === "") return null;`) rather than flipping to `<=` — that way it also covers the all-NULL payload case (which decodes to empty after NULL-stripping), closing Greptiles separate observation in the same guard. Added a test.

shrugs:
@greptile review

apps/ensapi/src/omnigraph-api/schema/scalars.ts
```diff
@@ -167,6 +168,37 @@ builder.scalarType("ChainId", {
...
+        }
+
+        // Enforce the uint256 upper bound so the scalar contract stays honest.
+        if (BigInt(ctx.value) > 2n ** 256n - 1n) {
+          ctx.issues.push({
+            code: "custom",
+            message: "TokenId must not exceed the uint256 maximum (2^256 - 1)",
+            input: ctx.value,
+          });
+        }
```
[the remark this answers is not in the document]
shrugs: Fixed in eee3e8e1b. Added a 78-digit length guard (uint256 max is 78 decimal digits) that short-circuits before the BigInt() conversion, so an over-long input is rejected without an expensive parse.

.changeset/efp-omnigraph.md
```diff
@@ -0,0 +1,10 @@
+---
+"ensapi": minor
+"enssdk": minor
+---
```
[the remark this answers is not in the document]
shrugs: Added changesets in eee3e8e1b for the two published packages with real source changes here: enskit (EFP urql cache resolvers) and @ensnode/ensnode-sdk (PluginName.EFP + EFP Omnigraph support). Skipped ensskills — its only diff is an AUTOGEN:SCHEMA region regenerated by `pnpm generate`, a byproduct of the schema change already described in efp-omnigraph.md — and @ensnode/integration-test-env, which is private (the fixed group bumps it automatically, no release note needed).

apps/ensapi/src/omnigraph-api/schema/scalars.ts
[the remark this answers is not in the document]
shrugs: Fixed in eee3e8e1b — TokenId now rejects strings longer than 78 digits (uint256 max) before the BigInt() conversion, so an extremely large input never triggers an expensive parse.

shrugs:
@greptile review

PR #2310, opened 2026-06-17
Fix published module metadata

Pugsin:
## Summary

- point the published `module` entries for four packages at `./dist/index.js`
- keep `main`, `types`, and exports aligned with the files emitted in published tarballs
- add a changeset for the affected public packages

## Verification

- `npm pack @ensnode/ensrainbow-sdk@1.15.2`, `@ensnode/datasources@1.15.2`, `@namehash/namehash-ui@1.15.2`, and `@ensnode/ponder-subgraph@1.15.2`
- confirmed each tarball contains `package/dist/index.js` and `package/dist/index.d.ts`, but not `package/dist/index.mjs`
- node assertion that all four `publishConfig.module` values are now `./dist/index.js`
- `git diff --check`

### DIFF
--- a/.changeset/quiet-maps-join.md
+++ b/.changeset/quiet-maps-join.md
@@ -1,0 +1,8 @@
+---
+"@ensnode/datasources": patch
+"@ensnode/ensrainbow-sdk": patch
+"@ensnode/ponder-subgraph": patch
+"@namehash/namehash-ui": patch
+---
+
+Fix published package `module` entries to point at emitted `dist/index.js` files.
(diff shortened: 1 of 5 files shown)


shrugs:
Thanks — the underlying issue is real (`publishConfig.module` pointed at `./dist/index.mjs` while these `type: module` packages emit `./dist/index.js` via tsup). We've implemented the fix ourselves in #2331, which also aligns the four packages with the rest of the repo's published manifests. Closing in favor of that.

PR #2329, opened 2026-06-20
feat(checkpoint): remote-checkpoint indexing → load → deploy pipeline

shrugs:
# Remote-checkpoint indexing → load → deploy

## Summary

- Productionizes indexing a release candidate on disposable **Cherry Servers** hardware and loading the resulting schemas into a target ENSDb — cutting a full index from **~1 week to ~6 hours**, with the deployed alpha/mainnet indexers **resuming** the warm schema instead of re-indexing.
- Adds two private packages — `@ensnode/ensdb-cli` (dump/load ENSIndexer schemas + their ENSNode metadata) and `@ensnode/ensindexer-checkpoint` (Cherry/R2 box orchestration) — plus four `[next]`-prefixed workflows.
- Fixes the stale `alpha` env defaults (searchlight/1 + `unigraph`) and introduces committed `apps/ensindexer/configs/*.env` as the single-source-of-truth **build_id parity contract**.

---

## Why

Indexing ENS from scratch on Railway/Render takes ~1 week, so it's never a viable release step. This builds the fast path proven by an earlier prototype into a one-trigger pipeline:

1. Resolve the release-candidate commit (head of the changesets "Release New Version" PR), build sha-tagged images, and index the two heavy mainnet-namespace configs (`alpha`, `mainnet`) on two Cherry boxes from the R2 `ponder_sync` seed.
2. Load the output schemas into the target env's Postgres as warm `<ns>Schema<VERSION>`.
3. Dispatch the existing blue/green deploy — alpha/mainnet resume in minutes; sepolia/etc index from scratch as normal.

**Resume mechanism:** the box runs the exact production identity (sourced from `configs/<config>.env`, no `END_BLOCK_*`), so Ponder's `build_id` matches the deployed service and it resumes rather than crashing. No build_id patching.

**Identifiers:** docker image = `sha-<short>`; schema name = `<ns>Schema<VERSION>` (so the RC pre-seed and the eventual release deploy share the schema); checkpoint dump = sha-keyed in R2.

---

## What's here

- **`packages/ensdb-cli`** — `dump`/`load` (schema via `pg_dump`/`pg_restore`, ENSNode metadata via ensdb-sdk, auto schema-rename, `--skip-if-exists`).
- **`packages/ensindexer-checkpoint`** — bash: Cherry lifecycle (+ self-destruct watchdog), write-tuned Postgres, full-backfill + end-block modes, sha-keyed R2 lock/idempotency, alpha seed-refresh.
- **`apps/ensindexer/configs/{alpha,mainnet}.env`** + drift-guard test — the build_id parity contract.
- **`environment-defaults.ts`** — alpha → searchlight/1 + unigraph (EFP excluded from the shared default: its datasources are mainnet-only and would break alpha-sepolia; enabled explicitly in `alpha.env`).
- **Workflows** (`[next]`): `index_checkpoint_and_deploy.yml`, `checkpoint.yml` (dev point-in-time), `deploy_ensapi_hotfix.yml`, `checkpoint_reaper.yml`.
- **`deploy_ensnode_blue_green.yml`** — backward-compatible optional `image_tag` input (decouples image tag from schema version).

---

## Testing

- ✅ `pnpm test` (ensindexer + ensdb-cli): **256/256 pass**; typecheck clean; biome clean.
- ✅ All box scripts **shellcheck-clean**; all workflow YAML validated.
- ✅ R2 seed bootstrapped into the `ensindexer-checkpoints` bucket (server-side copy, byte-verified).
- ⏳ **Not yet exercised against live Cherry/R2/Railway** — workflows are `workflow_dispatch`-only and need to be on `main` to run. The first run (planned: `checkpoint.yml`, `config=mainnet`, Jan 2019) is the real integration test. The one thing only a live run confirms is **config parity** (box identity == deployed service env).

---

## Notes for Reviewer

- Nothing auto-runs (all manual `workflow_dispatch`); the only scheduled job (reaper) only ever terminates `ensindexer-checkpoint-*` Cherry boxes — never production.
- The single existing-workflow change (`deploy_ensnode_blue_green.yml`) is additive and backward-compatible.
- **Yellow** (Render/Terraform) is intentionally out of scope for the index/deploy workflow — documented as a fast-follow.
- New GitHub secrets the workflows expect: `CHERRY_*`, `CF_R2_ACCESS_KEY_ID`/`CF_R2_SECRET_ACCESS_KEY`, `<TARGET>_ENSDB_ADMIN_URL` (existing `ALCHEMY_API_KEY` reused).

Changeset: included for `environment-defaults` (behavior change for alpha instances); the two new packages are private (no changeset required).


.github/workflows/checkpoint.yml
```diff
@@ -0,0 +1,100 @@
...
+        required: true
+        type: choice
+        options:
+          - alpha
+          - mainnet
+      timestamp:
+        description: "Chain-time to index up to (unix seconds)"
+        required: true
+        type: string
+      chain_ids:
```
shrugs: remove this chain_ids input variable concept entirely

.github/workflows/checkpoint.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Write SSH key + resolve indexed chain ids
+        run: |
+          install -m600 /dev/null /tmp/cherry_key
+          printf '%s\n' "$CHERRY_SSH_KEY" > /tmp/cherry_key
+          echo "SSH_KEY=/tmp/cherry_key" >> "$GITHUB_ENV"
+          CHAIN_IDS="${{ inputs.chain_ids }}"
+          if [ -z "$CHAIN_IDS" ]; then
+            case "${{ inputs.config }}" in
+              alpha) CHAIN_IDS="1,8453,10,42161,59144,534352" ;;
```
shrugs: derive these chain ids somehow, don't hardcode. maybe a simple script that imports the config and outputs indexedChainIds?

.github/workflows/checkpoint_reaper.yml
```diff
@@ -0,0 +1,49 @@
+name: "[next] Checkpoint: reap stale Cherry boxes"
```
shrugs: call this a garbage collector not a reaper

.github/workflows/deploy_ensnode_blue_green.yml
```diff
@@ -11,9 +11,14 @@ on:
           - green
           - blue
       tag:
-        description: "Docker Image Tag (also used for schema name generation)"
+        description: "Version used for ENSINDEXER schema names (e.g. 1.16.0)"
```
shrugs: rename this `tag` to `version` and image_tag becomes `tag`

.github/workflows/deploy_ensapi_hotfix.yml
```diff
@@ -0,0 +1,129 @@
...
+              ;;
+            *)
+              echo "Environment not recognized, skipping workflow"
+              exit 1
+              ;;
+          esac
+
+      - name: Update ENSApi service images
+        run: |
+          update_service_image() {
```
shrugs: extract this logic for deploying a docker version to a specific instance into a re-usable action and use it all of these workflows. i.e. matrix over the SVC_IDs or whatever to update all of them. ensure that ensrainbow gets updated first (feel free to do them all in series, not parallel)

shrugs: Bundle the `updateServiceImage` function and the `redeployService` function to be service-specific and then call each in series.

.github/workflows/index_checkpoint_and_deploy.yml
```diff
@@ -0,0 +1,159 @@
...
+          registry_user: ${{ github.actor }}
+          registry_token: ${{ secrets.GITHUB_TOKEN }}
+          tags: |
+            type=raw,value=${{ needs.resolve.outputs.image_tag }}
+            type=sha
+
+  index-and-load:
+    needs: resolve
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    timeout-minutes: 600 # GitHub-hosted's 6h cap doesn't apply to Blacksmith; default is 360, raise it
```
shrugs: change comment to "ensure timeout is large enough for full index" and actually set that to 960 minutes

.github/workflows/index_checkpoint_and_deploy.yml
```diff
@@ -0,0 +1,159 @@
...
+    env:
+      # Cherry + R2 + RPC come from secrets. Confirm the R2_* secret names match your existing ones.
```
shrugs, line 242: drop this comment

.github/workflows/index_checkpoint_and_deploy.yml
```diff
@@ -0,0 +1,159 @@
...
+          VERSION: ${{ needs.resolve.outputs.version }}
+          DO_LOAD: "1" # load both alpha + mainnet checkpoints into the target ENSDb
+          DO_SEED: "1" # refresh the canonical R2 seed from the enriched (all-chain) ponder_sync
+        run: bash packages/ensindexer-checkpoint/scripts/index-both.sh
+
+      - name: Tear down the Cherry box (always)
+        if: always()
+        run: bash packages/ensindexer-checkpoint/scripts/cherry-down.sh || true
+
+  deploy:
```
shrugs: Using the abstract deploy action mentioned above, update the other services ahead of the long alpha/mainnet ones so that they can all index in parallel. 

so:
1. deploy commit to sepolia,alphaSepolia,v2Sepolia on railway
2. index alpha/mainnet, load alpha/mainnet
3. deploy alpha/mainnet

apps/ensindexer/configs/README.md
```diff
@@ -0,0 +1,21 @@
+# ENSIndexer named configs
+
+Canonical, importable `.env` files for the named ENSIndexer "indexing identity" configurations used
```
shrugs, line 3: call these "Hosted Instance Configurations"

shrugs: and go ahead and write the others as well for completeness. sepolia (mainnet-style but pointed at sepollia), alphaSepolia (alpha, no efp, sepolia namespace), and v2Sepolia (alpha, no efp, sepolia-v2 namespace)

apps/ensindexer/configs/README.md
```diff
@@ -0,0 +1,21 @@
...
+by the remote-checkpoint pipeline (`alpha`, `mainnet`).
+
+Each file holds only the env vars that determine **indexing behavior** and therefore feed Ponder's
```
shrugs: drop this discussion paragraph

apps/ensindexer/configs/README.md
```diff
@@ -0,0 +1,21 @@
...
+must run the exact same identity as the deployed Railway service, or the warm-loaded schema will fail
+to resume (Ponder throws on a Build ID mismatch).
+
+Deployment-specific vars are **not** here — they are supplied per-environment:
+
+- `ENSDB_URL`, `ENSINDEXER_SCHEMA_NAME`
+- `ENSRAINBOW_URL`
+- an `RPC_URL_<chainId>` (or provider key) for every chain the active plugins index
+
+Consumed by sourcing in bash (`set -a; . configs/<config>.env; set +a`) or any dotenv reader.
```
shrugs: remove this line

apps/ensindexer/configs/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+Note: `EFP` is enabled only in `alpha.env`, not in the shared `EnvironmentDefaults.alpha` default,
```
shrugs: remove this discussion

apps/ensindexer/src/config/named-configs.test.ts
```diff
@@ -0,0 +1,47 @@
+import { readFileSync } from "node:fs";
```
shrugs: delete this test

packages/ensdb-cli/src/commands/dump.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { defineCommand } from "citty";
+
+import { dumpMetadata } from "../lib/ensdb";
+import { dumpSchema } from "../lib/pgtools";
+
+export const dump = defineCommand({
+  meta: {
+    name: "dump",
+    description: "Dump an ENSIndexer schema (custom-format) and, optionally, its ENSNode metadata.",
```
shrugs: all checkpoints should include metadata, no option for without

packages/ensdb-cli/src/commands/dump.ts
```diff
@@ -0,0 +1,52 @@
...
+  },
+  args: {
+    schema: {
```
shrugs: call this ensIndexerSchemaName for 1:1 terminology, use throughout this pr

packages/ensdb-cli/src/commands/load.ts
```diff
@@ -0,0 +1,88 @@
...
+    schema: {
+      type: "string",
+      description: "Target schema name (renamed to this; defaults to the dump's own schema name)",
+    },
+    metadata: {
+      type: "string",
+      description:
+        "ENSNode metadata JSON (from `dump --metadata-out`) to upsert under the target schema",
+    },
+    "skip-if-exists": {
```
shrugs: replace this logic with a `[--force]` argument; always fail fast if the destination schema or metadata entry exists and tell the user why. If force is passed, allow overwriting.

packages/ensdb-cli/src/lib/ensdb.ts
```diff
@@ -0,0 +1,78 @@
+import { createRequire } from "node:module";
+import { join } from "node:path";
+
+import { eq, sql } from "drizzle-orm";
+
+import { EnsDbReader, type EnsDbWriter } from "@ensnode/ensdb-sdk";
+
+/** One ENSNode metadata record, keyed by `key` within an ENSIndexer schema's namespace. */
+export interface MetadataRow {
```
shrugs: Surely this type is available from the ensdb sdk.

packages/ensdb-cli/src/lib/ensdb.ts
```diff
@@ -0,0 +1,78 @@
...
+  key: string;
+  value: unknown;
+}
+
+// The ENSDb SDK ships the ENSNode Schema migrations alongside its build (`files: ["dist","migrations"]`);
+// resolve the directory at runtime the same way ENSIndexer does.
+function ensnodeMigrationsDir(): string {
```
shrugs: Why?

packages/ensdb-cli/src/lib/ensdb.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export async function schemaExists(db: DrizzleClient, name: string): Promise<boolean> {
+  const result = (await db.execute(
+    sql`select 1 from information_schema.schemata where schema_name = ${name} limit 1`,
+  )) as { rows?: unknown[] };
+  return (result.rows?.length ?? 0) > 0;
+}
+
+export async function dropSchema(db: DrizzleClient, name: string): Promise<void> {
+  await db.execute(sql`drop schema if exists ${sql.identifier(name)} cascade`);
```
shrugs: Consider moving this exists, drop, and rename logic into the ensdb-sdk itself

packages/ensdb-cli/src/lib/ensdb.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export async function renameSchema(db: DrizzleClient, from: string, to: string): Promise<void> {
+  await db.execute(sql`alter schema ${sql.identifier(from)} rename to ${sql.identifier(to)}`);
+}
+
+/** Read the ENSNode metadata rows belonging to `schema` from the source ENSDb. */
+export async function dumpMetadata(srcUrl: string, schema: string): Promise<MetadataRow[]> {
+  const reader = new EnsDbReader(srcUrl, schema);
```
shrugs: Surely there's a get metadata function here, right?

packages/ensdb-cli/src/lib/ensdb.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Ensure the shared `ensnode` schema exists, then upsert `rows` under `targetSchema`'s namespace.
+ * Re-keys the records to `targetSchema` so a checkpoint loaded under a renamed schema keeps its
+ * metadata.
+ */
+export async function loadMetadata(
+  writer: EnsDbWriter,
+  targetSchema: string,
+  rows: MetadataRow[],
+): Promise<void> {
+  await writer.migrateEnsNodeSchema(ensnodeMigrationsDir());
```
shrugs: A dump should only ever have a single metadata row to insert, so only support that. Also, the ensdb-sdk should surely include a writemetadata helper, yes? does it not?

packages/ensdb-cli/src/lib/pgtools.ts
```diff
@@ -0,0 +1,57 @@
...
+ * `alphaSchema1.16.0`), which `pg_dump -n` treats as a schema.table pattern unless quoted.
+ */
+export function quoteIdent(name: string): string {
+  return `"${name.replace(/"/g, '""')}"`;
+}
+
+/** `pg_dump` args for a single-schema, custom-format, owner/ACL-stripped dump. */
+export function pgDumpArgs(schema: string, url: string, outFile: string): string[] {
+  return [
+    "-Fc",
```
shrugs, line 16: Add comments to each one of these arguments describing why it exists and what it does.

packages/ensdb-cli/tsup.config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default defineConfig({
+  entry: { cli: "src/cli.ts" },
+  format: ["esm"],
+  platform: "node",
+  target: "es2022",
+  bundle: true,
+  // Bundle the workspace packages (their dev `exports` point at TS source, which plain `node` can't
+  // resolve), but keep the heavy/real node_modules deps external — they're valid ESM packages and
+  // resolve from node_modules at runtime (this is a private, in-repo tool, so node_modules is always
```
shrugs, line 25: This is intended to be publicly available, although I'm not sure that we will be publishing it soon. Make sure that this module access works if this CLI is itself bundled with tsup

packages/ensdb-cli/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@ensnode/ensdb-cli",
+  "version": "0.0.0",
```
shrugs: Use the current word version from other packages 1.16.0

packages/ensindexer-checkpoint/scripts/config.example.sh
```diff
@@ -0,0 +1,57 @@
...
+export CHERRY_IMAGE="${CHERRY_IMAGE:-ubuntu_22_04}" # OS image slug
+export CHERRY_SSH_KEY_ID="${CHERRY_SSH_KEY_ID:-}"   # numeric id of the SSH key registered with Cherry
+export SSH_KEY="${SSH_KEY:-$HOME/.ssh/id_ed25519}"  # matching private key for ssh/scp to the box
+export BOX_USER="${BOX_USER:-root}"
+export BOX_HOSTNAME="${BOX_HOSTNAME:-ensindexer-checkpoint}"  # used as the Cherry hostname + reaper tag
+# postgres refuses to run as root; on a root box run the cluster as the `postgres` OS user.
+export PG_RUN_USER="${PG_RUN_USER:-postgres}"
+# Safety net: the box terminates ITSELF this many hours after provisioning, even if the orchestrating
+# runner dies and never tears it down. Keep comfortably above a full index (~6h).
+export SELF_DESTRUCT_HOURS="${SELF_DESTRUCT_HOURS:-12}"
```
shrugs: set to 20 for comfort

shrugs (commented):
I feel like there's some duplication going on between these two approaches (index and load, checkpoint). but actually they're the same thing, just with `END_BLOCK` configuration in the checkpoint case and no load, just dump/upload?

shrugs:
@greptile review

.github/workflows/checkpoint_reaper.yml
shrugs: Done — renamed to `checkpoint_gc.yml` ("Garbage Collector"), job `collect`, + concurrency group & timeout. (77d8982e9)

.github/workflows/checkpoint.yml
shrugs: Done — chains derived from the config now (no hardcoding): `remote-index-one.sh` runs `apps/ensindexer/scripts/print-indexed-chain-ids.ts` (builds the config, prints `indexedChainIds`). (77d8982e9)

.github/workflows/checkpoint.yml
shrugs: Done — `chain_ids` input removed entirely; derived from config. (77d8982e9)

.github/workflows/deploy_ensapi_hotfix.yml
shrugs: Done — extracted `.github/actions/deploy_railway_services` (loops service ids in series, ENSRainbow first). Used by hotfix, blue/green, and the new index workflow. Note: introduces new `*_SVC_ID` secrets for the per-service lists. (77d8982e9)

.github/workflows/deploy_ensnode_blue_green.yml
shrugs: Done — `tag`→`version` (schema names), `image_tag`→`tag` (images); all refs + dispatchers updated. (77d8982e9)

.github/workflows/index_checkpoint_and_deploy.yml
shrugs: Done — comment dropped. (77d8982e9)

.github/workflows/index_checkpoint_and_deploy.yml
shrugs: Done — `timeout-minutes: 960` with comment "ensure timeout is large enough for full index". (77d8982e9)

.github/workflows/index_checkpoint_and_deploy.yml
shrugs: Done — reworked into 3 phases via the reusable action: (1) deploy light sepolia/alphaSepolia/v2Sepolia, (2) index+load alpha/mainnet on Cherry, (3) deploy heavy alpha/mainnet. (77d8982e9)

apps/ensindexer/configs/README.md
shrugs: Done — retitled "Hosted Instance Configurations". (77d8982e9)

apps/ensindexer/configs/README.md
shrugs: Done — discussion paragraph removed. (77d8982e9)

apps/ensindexer/configs/README.md
shrugs: Done — line removed. (77d8982e9)

apps/ensindexer/configs/README.md
shrugs: Done — discussion removed. (77d8982e9)

apps/ensindexer/src/config/named-configs.test.ts
shrugs: Done — test deleted. (77d8982e9)

packages/ensdb-cli/package.json
shrugs: Done — version set to 1.16.0. (77d8982e9)

packages/ensdb-cli/src/commands/dump.ts
shrugs: Done — renamed to `ensIndexerSchemaName` throughout. (77d8982e9)

packages/ensdb-cli/src/commands/dump.ts
shrugs: Done — metadata always written to the `<dump>.metadata.json` sidecar; `--metadata-out` removed. (77d8982e9)

packages/ensdb-cli/src/commands/load.ts
shrugs: Done — `--skip-if-exists` replaced with `--force`; default fails fast if the target schema or metadata entry exists (naming which), `--force` overwrites. (77d8982e9)

packages/ensdb-cli/src/lib/ensdb.ts
shrugs: Done — `schemaExists`/`dropSchema`/`renameSchema` moved into ensdb-sdk (EnsDbReader/Writer); CLI consumes them. (77d8982e9)

packages/ensdb-cli/src/lib/ensdb.ts
shrugs: Replaced — the createRequire hack is gone; tsup now copies the SDK migrations into `dist/migrations` and they're resolved relative to the bundle (works standalone). (77d8982e9)

packages/ensdb-cli/src/lib/ensdb.ts
shrugs: Done — now uses the SDK's `SerializedEnsNodeMetadata`; the CLI's `MetadataRow` is gone. (77d8982e9)

packages/ensdb-cli/src/lib/ensdb.ts
shrugs: Done — public `getEnsNodeMetadata` on `EnsDbReader`. (77d8982e9)

packages/ensdb-cli/src/lib/ensdb.ts
shrugs: Done — modeled as a single metadata row; public `writeEnsNodeMetadata` helper added to `EnsDbWriter`. (77d8982e9)

packages/ensdb-cli/src/lib/pgtools.ts
shrugs: Done — each `pg_dump`/`pg_restore` arg now has an inline comment. (77d8982e9)

packages/ensdb-cli/tsup.config.ts
shrugs: Done — tsup `onSuccess` copies the SDK `migrations/` into `dist/migrations`; resolved via `import.meta.url` so it works when bundled standalone. (77d8982e9)

packages/ensindexer-checkpoint/scripts/config.example.sh
shrugs: Done — `SELF_DESTRUCT_HOURS` set to 20. (77d8982e9)

packages/ensindexer-checkpoint/scripts/index-both.sh
```diff
@@ -0,0 +1,52 @@
...
+  on_box "cd ~/$REMOTE_DIR && bash remote-provision.sh"
+fi
+
+log "running remote-checkpoint-prod.sh on the box (the long part)"
+on_box "cd ~/$REMOTE_DIR && \
+  SHA='$SHA' VERSION='$VERSION' \
+  ALCHEMY_API_KEY='${ALCHEMY_API_KEY:-}' \
+  DO_LOAD='${DO_LOAD:-0}' TARGET_URL='${TARGET_URL:-}' \
+  DO_SEED='$DO_SEED' \
+  bash remote-checkpoint-prod.sh"
```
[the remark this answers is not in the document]
shrugs: Fixed — `index-both.sh` is deleted; the unified `checkpoint.sh` now ships `TARGET_URL` (and the Cherry token) to the box via a 600 `.run-env` file sourced on the box, not interpolated into the SSH command argv. (77d8982e9)

packages/ensindexer-checkpoint/scripts/cherry-up.sh
```diff
@@ -0,0 +1,61 @@
...
+    break
+  fi
+  sleep 10
+done
+[ "$up" = "1" ] || die "timed out waiting for SSH to $ip"
+
+# Arm the self-destruct watchdog: a detached process on the box that, after SELF_DESTRUCT_HOURS,
+# DELETEs this server via the Cherry API. Survives runner death; cherry-down.sh is the normal path.
+log "arming self-destruct watchdog (${SELF_DESTRUCT_HOURS}h) on the box"
+on_box "CHERRY_API_TOKEN='$CHERRY_API_TOKEN' setsid bash -c '
```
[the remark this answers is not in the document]
shrugs: Fixed — the watchdog token is shipped via a 600 `~/.cherry-token` file and read at DELETE time; no longer embedded in the SSH command argv. (77d8982e9)

packages/ensdb-cli/src/commands/load.ts
```diff
@@ -0,0 +1,88 @@
...
+        return;
+      }
+
+      // Make the restore idempotent: clear the dump's own schema name, restore, then rename to the
+      // target (replacing any prior target). The Postgres schema name does not affect Ponder's
+      // build_id, so renaming is safe for resume.
+      if (dumpSchemaName) await dropSchema(writer.ensDb, dumpSchemaName);
+      await restoreDump(args.into, args.dump);
+
+      if (dumpSchemaName && dumpSchemaName !== targetSchema) {
```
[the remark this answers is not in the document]
shrugs: Fixed — `load` now fails fast if the dump's schema name is unresolvable (and `--schema` is required), so it can never restore into the wrong/zero schema. (77d8982e9)

packages/ensindexer-checkpoint/scripts/remote-run.sh
```diff
@@ -0,0 +1,150 @@
...
+if [ "$MODE" = "end-block" ]; then
+  : "${TIMESTAMP:?}" "${CHAIN_IDS:?}"
+  END_BLOCK_LINES="$(MANIFEST_OUT="/tmp/${SCHEMA}.blocks.json" bash "$LIB_DIR/remote-resolve-end-blocks.sh")"
+fi
+
+# ── .env.local: identity + runtime ───────────────────────────────────────────
+ENV_FILE="$REPO_DIR/apps/ensindexer/.env.local"
+log "writing $ENV_FILE (mode=$MODE, config=$CONFIG, schema=$SCHEMA)"
+{
+  echo "NAMESPACE=$NAMESPACE"
```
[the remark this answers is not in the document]
shrugs: Addressed — `remote-run.sh` is gone; the ENSRainbow gate (now in `remote-index-one.sh`) requires BOTH the DB marker AND a live `/health` response, and dies otherwise. (77d8982e9)

packages/ensindexer-checkpoint/scripts/remote-run.sh
```diff
@@ -0,0 +1,150 @@
...
+pg -c "drop schema if exists \"$SCHEMA\" cascade;"
+
+# ── indexer (detached) + ensapi (status polling) ─────────────────────────────
+INDEXER_LOG="/tmp/${SCHEMA}-indexer.log"
+log "starting indexer -> $INDEXER_LOG"
+(cd "$REPO_DIR/apps/ensindexer" &&
+  setsid bash -c "ENSINDEXER_SCHEMA_NAME='$SCHEMA' pnpm start > '$INDEXER_LOG' 2>&1" </dev/null &)
+
+log "starting ensapi on :$ENSAPI_PORT (retry until indexer metadata ready)"
+setsid bash -c "cd '$REPO_DIR/apps/ensapi'; while true; do
```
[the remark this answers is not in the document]
shrugs: Addressed — crash detection (now in `remote-index-one.sh`) is process-liveness-first: if the indexer's process group dies before `is_ready`, that's a crash regardless of log text. The pattern list is now just a fast-path. (77d8982e9)

.github/workflows/checkpoint_gc.yml
```diff
@@ -0,0 +1,49 @@
...
+    - cron: "17 * * * *" # hourly
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  reap:
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+    env:
```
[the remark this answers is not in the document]
shrugs: Fixed — workflow renamed to `checkpoint_gc.yml` with an explicit `timeout-minutes` and a concurrency group. (77d8982e9)

packages/ensindexer-checkpoint/scripts/remote-checkpoint.sh
```diff
@@ -0,0 +1,74 @@
...
+    TIMESTAMP="${TIMESTAMP:-}" CHAIN_IDS="${CHAIN_IDS:-}" \
+    bash "$LIB_DIR/remote-run.sh"
+
+  log "building ensdb-cli"
+  pnpm -C "$REPO_DIR" -F @ensnode/ensdb-cli build >/dev/null
+  log "dumping schema $SCHEMA -> $LOCAL_DUMP (+ metadata)"
+  ensdb_cli dump "$SCHEMA" --from "$ENSDB_URL" -f "$LOCAL_DUMP" --metadata-out "$LOCAL_META"
+
+  log "uploading checkpoint + metadata to R2"
+  rclone copyto "$LOCAL_DUMP" "$(r2_checkpoint "$CKPT_NAME")"
```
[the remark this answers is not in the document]
shrugs: Fixed — the unified `remote-checkpoint.sh` always runs `remote-checkout.sh` (which builds ensdb-cli) before any load, so a `DO_LOAD=1` run on a fresh box / reuse path has the CLI available. (77d8982e9)

.github/workflows/checkpoint.yml
```diff
@@ -0,0 +1,88 @@
...
+            echo "### Checkpoint ready"
+            echo ""
+            echo "- R2: \`${R2_CHECKPOINTS_BUCKET}/checkpoints/${OBJ}\`"
+            echo "- metadata: \`${R2_CHECKPOINTS_BUCKET}/checkpoints/${OBJ%.dump}.metadata.json\`"
+            echo ""
+            echo "Restore locally:"
+            echo '```'
+            echo "rclone copy r2:${R2_CHECKPOINTS_BUCKET}/checkpoints/${OBJ} ."
+            echo "ensdb-cli load ${OBJ} --into \$LOCAL_ENSDB_URL --metadata ${OBJ%.dump}.metadata.json"
+            echo '```'
```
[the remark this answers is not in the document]
shrugs: Good catch — fixed: the snippet now uses `--schema <ensIndexerSchemaName>`, drops `--metadata`, copies the `<dump>.metadata.json` sidecar (read automatically by load), and points at `${OBJ}.metadata.json`.

packages/ensindexer-checkpoint/scripts/remote-seed-export.sh
```diff
@@ -0,0 +1,29 @@
...
+# chain-1-only (a subset).
+source "$(dirname "$0")/lib.sh"
+require psql
+require pg_dump
+require rclone
+
+OUT="/tmp/${R2_SEED_OBJECT}"
+log "ponder_sync size: $(pg -tAc "SELECT pg_size_pretty(sum(pg_total_relation_size(c.oid))) FROM pg_class c JOIN pg_namespace n ON n.oid=c.relnamespace WHERE n.nspname='ponder_sync' AND c.relkind='r'" | tr -d '[:space:]')"
+log "pg_dump -Fc -Z3 ponder_sync -> $OUT (compressed; long)"
+pg_dump -Fc -Z3 --no-owner --no-privileges -n ponder_sync "$PG_CONN" -f "$OUT"
```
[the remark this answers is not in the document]
shrugs: Fixed — the seed dump now writes to `$DATA_MOUNT` (the large NVMe) instead of `/tmp`.

packages/ensindexer-checkpoint/scripts/remote-index-one.sh
```diff
@@ -0,0 +1,155 @@
...
+  log "[$CONFIG] deriving indexed chain ids from config"
+  CHAIN_IDS="$(cd "$REPO_DIR/apps/ensindexer" &&
+    ENSRAINBOW_URL="http://localhost:$RAINBOW_PORT" ENSINDEXER_SCHEMA_NAME="$SCHEMA" \
+      pnpm exec tsx scripts/print-indexed-chain-ids.ts | paste -sd, -)"
+  [ -n "$CHAIN_IDS" ] || die "[$CONFIG] could not derive indexed chain ids from config"
+  log "[$CONFIG] indexed chains: $CHAIN_IDS"
+  while IFS= read -r line; do
+    [ -n "$line" ] && END_BLOCK_ENV+=("$line")
+  done < <(CHAIN_IDS="$CHAIN_IDS" MANIFEST_OUT="/tmp/${SCHEMA}.blocks.json" bash "$LIB_DIR/remote-resolve-end-blocks.sh")
+fi
```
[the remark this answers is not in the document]
shrugs: Good catch (P1) — fixed: end-block resolution now writes to a file with an explicit `|| die` on non-zero exit, instead of a process substitution that swallowed the resolver's failure. (fd869cb0e+1)

shrugs:
@greptile review

packages/ensindexer-checkpoint/scripts/remote-checkpoint.sh
```diff
@@ -0,0 +1,135 @@
...
+
+# ── load each config into the target ENSDb as <config>Schema<VERSION> ─────────
+if [ "$DO_LOAD" = "1" ]; then
+  mkdir -p "$DATA_MOUNT"
+  for c in "${CONFIG_LIST[@]}"; do
+    cn="$(ckpt_name "$c")"
+    ld="$DATA_MOUNT/$cn"
+    if [ ! -f "$ld" ]; then
+      log "[$c] downloading $cn (+ metadata) from R2 for load"
+      rclone copyto "$(r2_checkpoint "$cn")" "$ld"
```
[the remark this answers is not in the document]
shrugs: Intentional — fail-fast is the chosen policy (it must never silently overwrite a `<config>Schema<VERSION>` that could already be live). The re-run-after-partial-failure case is handled operationally: drop the target schema, then re-run (indexing is skipped via the sha-keyed R2 checkpoint, so recovery is cheap). `--force` exists on the CLI for an operator who explicitly wants to overwrite. Not changing the default.

packages/ensdb-cli/src/commands/load.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (conflicts.length > 0) {
+          throw new Error(`${conflicts.join(" and ")} already exist(s); pass --force to overwrite`);
+        }
+      }
+
+      // Make the restore idempotent: clear the dump's own schema name, restore, then rename to the
+      // target (dropping any prior target). The Postgres schema name does not affect Ponder's
+      // build_id, so renaming is safe for resume.
+      await writer.dropSchema(dumpSchemaName);
+      await restoreDump(args.into, args.dump);
```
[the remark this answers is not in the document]
shrugs: Fixed (c43b9ade6) — the dump's staging schema is now part of the fail-fast conflict check: it's only dropped when it doesn't pre-exist, or with `--force`. No silent data loss.

packages/ensindexer-checkpoint/README.md
```diff
@@ -0,0 +1,109 @@
...
+since two configs run from the same repo dir). The deployed Railway service must use the same identity.
+
+**`build_id` is environment-specific — it hashes the indexed code, not just the env.** Ponder gates
+crash-recovery resume on `_ponder_meta.app.build_id` matching (`ponder/dist/.../database/index.js:498`
+throws and crashes on mismatch). The Build ID is derived from the indexing _code/contracts_ plus the
+identity above. So a checkpoint **resumes a stock deployed indexer only when it was produced from the
+exact same commit/image that the indexer runs.** The production workflow guarantees this by
+construction: `index_checkpoint_and_deploy.yml` builds the `sha-<short>` images **from the indexed
+commit** and the deploy step dispatches `deploy_ensnode_blue_green.yml` with that same `image_tag` — so
+the box, the checkpoint, and the deployed image are all the same SHA.
```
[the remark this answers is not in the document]
shrugs: Fixed (c43b9ade6) — README corrected: the 3-phase workflow deploys the same sha images inline (deploy-light/deploy-heavy via the deploy_railway_services action), not by dispatching blue/green.

.github/workflows/deploy_ensnode_blue_green.yml
```diff
@@ -10,10 +10,15 @@ on:
         options:
           - green
           - blue
-      tag:
-        description: "Docker Image Tag (also used for schema name generation)"
+      version:
+        description: "Version used for ENSINDEXER schema names (e.g. 1.16.0)"
         required: true
         type: string
+      tag:
```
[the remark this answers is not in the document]
shrugs: Intentional — this is the maintainer-directed rename (`tag`→`version` for schemas, `image_tag`→`tag` for images); `version` is deliberately the required primary input. In-repo dispatchers are updated in this PR; manual runbooks should pass `version`. Not adding a version→tag fallback, as that would re-muddy the split the rename is meant to create.

shrugs:
@greptile review

.github/actions/deploy_railway_services/action.yml
```diff
@@ -0,0 +1,81 @@
...
+          local service_id=$2
+          echo "Redeploying $service_id"
+          curl --request POST \
+            --silent \
+            --url https://backboard.railway.app/graphql/v2 \
+            --header 'Authorization: Bearer '"$RAILWAY_TOKEN" \
+            --header 'Content-Type: application/json' \
+            --data "{\"query\":\"mutation serviceInstanceDeploy(\$serviceId: String!, \$environmentId: String!) { serviceInstanceDeploy(serviceId: \$serviceId, environmentId: \$environmentId) }\",\"variables\":{\"environmentId\":\"${environment_id}\",\"serviceId\":\"${service_id}\"}}"
+          echo "Finished redeploying $service_id"
+        }
```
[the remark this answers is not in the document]
shrugs: Fixed — `redeploy_service` now uses `--fail --show-error` (matching `update_service_image`), so a Railway API error aborts the step instead of being swallowed. Centralized in the action, so hotfix + blue/green both get it.

shrugs:
@greptile review

shrugs:
@greptile review

.github/workflows/index_checkpoint_and_deploy.yml
```diff
@@ -0,0 +1,347 @@
...
+              --header 'Authorization: Bearer '${{ env.RAILWAY_TOKEN }} \
+              --header 'Content-Type: application/json' \
+              --data '{"query": "mutation variableUpsert { variableUpsert(input: { projectId: \"'${{ env.RAILWAY_PROJECT_ID }}'\", environmentId: \"'${environment_id}'\", name: \"'${variable_name}'\", value: \"'${variable_value}'\" }) }"}'
+            echo "Finished setting $variable_name"
+          }
+
+          set_shared_variable ${RAILWAY_ENVIRONMENT_ID} "SEPOLIA_ENSINDEXER_SCHEMA_NAME" "sepoliaSchema${VERSION}"
+          set_shared_variable ${RAILWAY_ENVIRONMENT_ID} "ALPHA-SEPOLIA_ENSINDEXER_SCHEMA_NAME" "alphaSepoliaSchema${VERSION}"
+          set_shared_variable ${RAILWAY_ENVIRONMENT_ID} "V2-SEPOLIA_ENSINDEXER_SCHEMA_NAME" "v2SepoliaSchema${VERSION}"
+
```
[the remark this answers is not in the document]
shrugs: Already hardened — `set_shared_variable` captures the response and fails on a body-level `errors[]` (this thread is anchored to the pre-hardening line). (50d25c5b1)

packages/ensindexer-checkpoint/scripts/remote-checkpoint.sh
```diff
@@ -0,0 +1,135 @@
...
+SUFFIX=""
+[ "$MODE" = "end-block" ] && SUFFIX="-t${TIMESTAMP}"
+
+box_schema() { echo "ckpt_${1}_${SHA}${SUFFIX}"; }
+ckpt_name() { checkpoint_object "$1" "$SHA" "$SUFFIX"; } # <config>-<sha>[-t<ts>].dump
+
+LOCK_KEY="${CONFIGS// /+}-${SHA}${SUFFIX}"
+acquire_lock "$LOCK_KEY"
+trap 'release_lock "$LOCK_KEY"' EXIT
+
```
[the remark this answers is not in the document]
shrugs: Fixed — the on-box staging schema now uses a 12-char sha (`ckpt_<config>_<short>[-t<ts>]`), well under Postgres' 63-byte limit. It's box-local only, so it doesn't affect build_id or the R2 object name (those keep the full sha). (50d25c5b1)

packages/ensindexer-checkpoint/scripts/remote-checkout.sh
```diff
@@ -0,0 +1,48 @@
...
+cd "$REPO_DIR" || die "cannot cd to $REPO_DIR"
+log "checking out $SHA"
+git fetch --all --quiet
+git checkout --quiet "$SHA"
+log "pnpm install"
+pnpm install --frozen-lockfile
+
+log "building ensdb-cli"
+pnpm -F @ensnode/ensdb-cli build >/dev/null
+
```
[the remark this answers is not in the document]
shrugs: Acknowledged as a known limitation: production indexes release-candidate commits, which always contain `packages/ensdb-cli`. Dev-checkpointing a commit that predates ensdb-cli is out of scope (the config-from-orchestrating-checkout support is for identity env, not the toolchain). Not fixing here.

packages/ensindexer-checkpoint/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Secrets and per-run state — never commit. Configure via env (CI) or a local copy of config.example.sh.
+scripts/config.sh
+scripts/.box-id
+scripts/.box-host
```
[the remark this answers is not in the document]
shrugs: Fixed — added `scripts/.known_hosts` to the package .gitignore. (50d25c5b1)

packages/ensdb-cli/src/commands/dump.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    schemaExistsMock.mockResolvedValueOnce(true);
+    getEnsNodeMetadataMock.mockResolvedValueOnce({ key: "indexing_metadata_context", value: 1 });
+
+    await dump.run!({
+      args: {
+        ensIndexerSchemaName: "alphaSchema1.16.0",
+        from: "postgres://x/y",
+        out: "/tmp/o.dump",
+      },
+    } as any);
```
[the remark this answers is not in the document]
shrugs: Left as-is: these success cases assert on side-effects / emitted output rather than a resolved value, so `await`-then-assert reads more clearly than `.resolves`. The tests are correct and pass.

packages/ensdb-cli/src/commands/load.test.ts
```diff
@@ -0,0 +1,162 @@
...
+  });
+
+  it("migrates the ensnode schema before restoring", async () => {
+    await load.run!({ args: { ...baseArgs } } as any);
+
+    expect(migrateEnsNodeSchemaMock).toHaveBeenCalledWith("/migrations");
+    const migrateOrder = migrateEnsNodeSchemaMock.mock.invocationCallOrder[0]!;
+    const restoreOrder = restoreDumpMock.mock.invocationCallOrder[0]!;
+    expect(migrateOrder).toBeLessThan(restoreOrder);
+  });
```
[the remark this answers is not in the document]
shrugs: Same as the dump.test note — these cases assert on side-effects/output, not a resolved value; await-then-assert is intentional here.

packages/ensdb-cli/src/commands/load.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (!force) {
+        const conflicts: string[] = [];
+        if (targetSchemaExists) conflicts.push(`schema "${ensIndexerSchemaName}"`);
+        if (stagingSchemaCollides) conflicts.push(`staging schema "${dumpSchemaName}"`);
+        if (existingMetadata) {
+          conflicts.push(`ENSNode metadata entry for "${ensIndexerSchemaName}"`);
+        }
+        if (conflicts.length > 0) {
+          throw new Error(`${conflicts.join(" and ")} already exist(s); pass --force to overwrite`);
+        }
```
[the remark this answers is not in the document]
shrugs: Intentional — a `null` metadata sidecar means the source schema had no IndexingMetadataContext row yet; skipping the upsert is correct (the resuming indexer re-writes it via EnsDbWriterWorker). The dump always writes the sidecar; `null` is a valid value, not a missing file.

packages/ensindexer-checkpoint/README.md
```diff
@@ -0,0 +1,110 @@
...
+  production and dev via `CONFIGS`/`MODE`/`DO_LOAD`/`DO_SEED`).
+
+## R2 layout
+
+```
+<R2_CHECKPOINTS_BUCKET>/
+  seed/ponder_sync.dump                       # canonical zero-RPC cache (refreshed by alpha runs)
+  checkpoints/<config>-<sha>[-t<ts>].dump      # sha-keyed checkpoint + .metadata.json sidecar
+  locks/<configs>-<sha>[-t<ts>]               # best-effort run lock (GH concurrency is the hard one)
+```
```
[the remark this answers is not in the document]
shrugs: Fixed — added a `text` language tag to the R2 layout fence. (50d25c5b1)

packages/ensindexer-checkpoint/scripts/lib.sh
```diff
@@ -0,0 +1,153 @@
...
+  if r2_exists "$p"; then
+    held="$(rclone cat "$p" 2>/dev/null | head -1)"
+    held_ts="${held%% *}"
+    held_owner="${held#* }"
+    if [[ "$held_ts" =~ ^[0-9]+$ ]] && [ $((now - held_ts)) -lt "$LOCK_TTL_SECONDS" ] &&
+      [ "$held_owner" != "$LOCK_OWNER" ]; then
+      die "R2 lock held for '$key' (age $((now - held_ts))s < TTL, owner $held_owner). Another checkpoint run in progress?"
+    fi
+    [ "$held_owner" = "$LOCK_OWNER" ] || warn "breaking stale R2 lock for '$key' (age $((now - held_ts))s)"
+  fi
```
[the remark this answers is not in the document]
shrugs: Fixed — lock age is now computed only for a numeric payload (`age=-1` otherwise, treated as stale); a malformed lock no longer crashes `acquire_lock` under set -e. (50d25c5b1)

packages/ensindexer-checkpoint/scripts/remote-index-one.sh
```diff
@@ -0,0 +1,162 @@
...
+INDEXER_ENV=(
+  "NAMESPACE=$NAMESPACE"
+  "PLUGINS=$PLUGINS"
+  "LABEL_SET_ID=$LABEL_SET_ID"
+  "LABEL_SET_VERSION=$LABEL_SET_VERSION"
+  "ENSINDEXER_SCHEMA_NAME=$SCHEMA"
+  "ENSDB_URL=$ENSDB_URL"
+  "ENSRAINBOW_URL=http://localhost:$RAINBOW_PORT"
+  "ALCHEMY_API_KEY=${ALCHEMY_API_KEY:-}"
+)
```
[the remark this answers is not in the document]
shrugs: Fixed — added `: "${ENSDB_URL:?}"` so it fails fast if unset (it's defaulted by config.example.sh, but now asserted). (50d25c5b1)

shrugs:
@greptile review

shrugs:
@greptile review

packages/ensdb-cli/src/lib/migrations.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Absolute path to the ENSNode Schema migrations directory.
+ *
+ * tsup copies the ENSDb SDK's `migrations/` folder next to the bundled `dist/cli.js` at build time
+ * (see `tsup.config.ts`), so we resolve it relative to this module's URL. This works whether the CLI
+ * runs from source (tsx) or as the standalone bundle, without depending on `node_modules` resolution.
+ */
+export function ensnodeMigrationsDir(): string {
+  return fileURLToPath(new URL("./migrations", import.meta.url));
+}
```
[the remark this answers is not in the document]
shrugs: Fixed — corrected the comment: the bundled `dist/migrations` resolution is for the built CLI (`node dist/cli.js`), which is how it's always run; it's explicitly not meant for tsx-from-source. (fd8…)

.github/workflows/deploy_ensnode_blue_green.yml
```diff
@@ -10,10 +10,15 @@ on:
...
+        description: "Docker image tag to deploy (defaults to `version`; set to a sha- tag for release-candidate images)"
+        required: false
+        type: string
+        default: ""
```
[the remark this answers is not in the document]
shrugs: Fixed — reversed my earlier call: `version` is now optional and falls back to `tag` (and `tag` falls back to `version`), with a guard requiring at least one. Old `tag`-only callers keep working; the version(schemas)/tag(images) split still works for RC deploys.

.github/workflows/deploy_ensnode_blue_green.yml
```diff
@@ -24,11 +29,16 @@ jobs:
...
+      # `version` names the ENSINDEXER schemas; `tag` selects the Docker images and defaults to
+      # `version`. They differ when deploying release-candidate images, which are sha-tagged while
+      # the schemas keep the release version (so a pre-seeded checkpoint and the eventual release
+      # deploy share the schema).
+      VERSION: ${{ inputs.version }}
+      TAG: ${{ inputs.tag != '' && inputs.tag || inputs.version }}
+      ENSINDEXER_DOCKER_IMAGE: "ghcr.io/namehash/ensnode/ensindexer:${{ inputs.tag != '' && inputs.tag || inputs.version }}"
+      ENSAPI_DOCKER_IMAGE: "ghcr.io/namehash/ensnode/ensapi:${{ inputs.tag != '' && inputs.tag || inputs.version }}"
+      ENSRAINBOW_DOCKER_IMAGE: "ghcr.io/namehash/ensnode/ensrainbow:${{ inputs.tag != '' && inputs.tag || inputs.version }}"
+      ENSADMIN_DOCKER_IMAGE: "ghcr.io/namehash/ensnode/ensadmin:${{ inputs.tag != '' && inputs.tag || inputs.version }}"
```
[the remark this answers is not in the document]
shrugs: Fixed — `VERSION` is now `inputs.version != '' && inputs.version || inputs.tag`, so schema naming works for tag-only runs; images already used `tag || version`.

shrugs:
@greptile review

packages/ensindexer-checkpoint/scripts/lib.sh
```diff
@@ -0,0 +1,156 @@
...
+  cat >"$dest" <<EOF
+[${RCLONE_REMOTE}]
+type = s3
+provider = Cloudflare
+access_key_id = ${R2_ACCESS_KEY_ID}
+secret_access_key = ${R2_SECRET_ACCESS_KEY}
+endpoint = ${R2_ENDPOINT}
+no_check_bucket = true
+EOF
+}
```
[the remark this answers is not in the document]
shrugs: Fixed — write_rclone_conf now creates the file mode 600 via a subshell-scoped `umask 077`, so the R2 creds aren't world-readable (and the umask change doesn't leak).

packages/ensindexer-checkpoint/scripts/remote-rehydrate.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+# BOX: prepare ephemeral storage — mount the data NVMe, init a write-tuned postgres cluster on it, and
+# restore ponder_sync from the canonical R2 seed. Idempotent within a box's lifetime (skips work
+# already done). Local NVMe can't be snapshotted, so this runs on each fresh box.
+source "$(dirname "$0")/lib.sh"
+require rclone
+require psql
+
```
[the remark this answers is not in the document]
shrugs: Fixed — added `require pg_restore` to remote-rehydrate.sh for a clear early failure.

shrugs:
@greptile review

.github/workflows/deploy_ensnode_blue_green.yml
```diff
@@ -24,17 +30,29 @@ jobs:
...
       RAILWAY_ENVIRONMENT_ID: ${{ secrets.RAILWAY_ENVIRONMENT_ID}}
       RAILWAY_TEAM_TOKEN: ${{ secrets.RAILWAY_TEAM_TOKEN }}
 
     steps:
+      - name: Require version or tag
+        run: |
+          if [ -z "${{ inputs.version }}" ] && [ -z "${{ inputs.tag }}" ]; then
+            echo "Provide at least one of: version (schema names) or tag (image tag)." >&2
+            exit 1
+          fi
```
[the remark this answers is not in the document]
shrugs: Fixed — routed `inputs.version`/`inputs.tag` through step-level `env:` (`INPUT_VERSION`/`INPUT_TAG`) and reference the shell vars in the guard, eliminating the template-injection surface.

PR #2335, opened 2026-06-23
fix: ENSApi initialization logic

### LINKED ISSUE #2272: ENSApi: DI init crashes instead of waiting when ENSIndexer metadata isn't published yet
## Summary

ENSApi crashes permanently (`process.exit(1)`) at startup if it initializes before ENSIndexer has published its `ensnode.metadata`, instead of waiting/retrying. This is a startup-ordering race: any deployment where ENSApi can come up before the indexer's first metadata write (fresh DB, co-located stacks, restarts) hard-fails the API.

## Symptom

```
ERROR: Error initializing DI container
  DI container initialization failed: could not connect to ENS Root Chain RPC
  due to relation "ensnode.metadata" does not exist
 ELIFECYCLE  Command failed with exit code 1.
```

ENSApi exits and does not recover on its own; only a restart *after* the indexer has written metadata succeeds.

## Root cause

- `apps/ensapi/src/index.ts` runs `di.init().catch(() => process.exit(1))` — no retry.
- `di.init()` (`apps/ensapi/src/di.ts`) eagerly reads the indexer's published `ensnode.metadata` via `stackInfoCache.read()` / `indexingStatusCache.read()` and the root-chain RPC config derived from it. When `ensnode.metadata` doesn't exist yet, these throw and init rejects.

So a transient, expected startup-ordering condition (indexer hasn't published metadata yet) is treated as a fatal error.

## Proposed fix

Treat "indexer metadata not published yet" as a retryable condition during DI init:

- Add a classifier `isIndexerMetadataNotReadyError(err)` matching the missing-relation / empty-metadata case.
- In `di.init()`, wrap the metadata-dependent initialization (the three cache reads + root-chain RPC config/getBlockNumber) in a bounded retry-with-backoff loop. On a "not ready" error: log and retry with backoff; on any other error: fail fast; give up after a configurable timeout (env, default ~10 min) so a genuine misconfig still surfaces. Ensure caches re-query on retry rather than caching the failure.

This makes ENSApi tolerant of being started before/alongside the indexer, which is the common case for fresh deployments and co-located dev/checkpoint stacks.

## Notes

- Keep genuine config/connectivity failures fail-fast — only the metadata-not-ready class should retry.
- Found while running co-located indexer+ENSApi checkpoint stacks, where ENSApi reliably loses the race against the indexer's first metadata write and dies.

Part of #1360 (tracking).

tk-o:
# Lite PR

## Summary

- ENSApi now waits for ENSDb to become ready before initializing its dependency graph, preventing startup crashes when ENSApi begins before the relevant ENSNode Metadata record is available in the ENSDb instance.
- Added `waitForEnsDbToBeReady` helper in `apps/ensapi/src/lib/ensdb.ts` with a 15-minute retry window (15 retries × 60 seconds).
- Added unit tests for the retry helper and updated the DI bootstrap test to cover the retry path.
- Minor logging improvements in `apps/ensapi/src/index.ts` to ensure error objects are logged under the `err` key.
- Added a top-level middleware (`serviceStatusMiddleware`) that provides `isHealthy` and `isReady` values to downstream middleware and route handlers.
- Added a middleware (`assertApiReadinessMiddleware`) that short-circuit HTTP requests to APIs that require the ENSApi instance readiness.

---

## Why

- Fixes #2272: ENSApi was crashing with `process.exit(1)` at startup if it initialized before the ENSDb metadata was ready.
- ENSApi readiness is based on ENSDb readiness because it guarantees that the relevant ENSNode Metadata record is available for relevant APIs that need this record.

---

## Testing

- `pnpm -F ensapi test` — **297 passed**.
- `pnpm -F ensapi typecheck` — **passed**.
- `pnpm typecheck` (all workspaces) — **passed**.
- `pnpm lint` on changed files — **no issues**.

---

- The retry loop runs **after** `ensDbClient.isHealthy()` has already verified connectivity, so the loop only waits for metadata readiness.
- The timeout is intentionally hardcoded to 15 minutes; the interval was not made configurable per the implementation plan.
- The `waitForEnsDbToBeReady` helper does not cache its promise because `di.init()` runs once per process.
- `serviceStatusMiddleware` is added as a top-level middleware as all downstream middleware and route handlers must be able to see if the ENSApi instance is healthy/ready.
- `assertApiReadinessMiddleware` is added as a first middleware to all routers handling APIs which require ENSApi readiness.

---

### DIFF
--- a/.changeset/ensapi-ensdb-ready.md
+++ b/.changeset/ensapi-ensdb-ready.md
@@ -1,0 +1,5 @@
+---
+"ensapi": patch
+---
+
+ENSApi now waits for ENSDb to become ready before initializing its dependency graph. This prevents startup crashes when ENSApi starts before the relevant ENSNode Metadata record is available in ENSDb instance.
--- a/.changeset/fancy-buses-wear.md
+++ b/.changeset/fancy-buses-wear.md
@@ -1,0 +1,5 @@
+---
+"@ensnode/ensdb-sdk": minor
+---
+
+Updated `isReady` method in `EnsDbReader` class.
--- a/apps/ensapi/src/app.ts
+++ b/apps/ensapi/src/app.ts
@@ -7,6 +7,7 @@ import { html } from "hono/html";
 import { errorResponse } from "@/lib/handlers/error-response";
 import { createApp } from "@/lib/hono-factory";
 import logger from "@/lib/logger";
+import { serviceStatusMiddleware } from "@/middleware/service-status.middleware";
 import { generateOpenApi31Document } from "@/openapi-document";
 
 import realtimeApi from "./handlers/api/meta/realtime-api";
@@ -15,7 +16,7 @@ import ensanalyticsApi from "./handlers/ensanalytics/ensanalytics-api";
 import ensApiProbesApi from "./handlers/ensapi-probes/ensapi-probes-api";
 import subgraphApi from "./handlers/subgraph/subgraph-api";
 
-const app = createApp();
+const app = createApp({ middlewares: [serviceStatusMiddleware] });
 
 // set the X-ENSNode-Version response header to the current version
 app.use(async (ctx, next) => {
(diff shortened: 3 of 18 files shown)


tk-o:
@greptile review

tk-o:
Replaced by #2346

PR #2339, opened 2026-06-29
fix(ensindexer): label heal cascade never fired due to find/upsert aliasing

shrugs:
## Problem

A `Domain` whose label is healed **after** its `canonicalName` has already been materialized keeps a stale `[labelHash]` `canonicalName`, even though its `Label` row is healed correctly — leaving the indexed DB internally inconsistent (healed `Label`, stale `Domain.canonicalName`).

On `ens-test-env` this affects e.g. the ENSv1 `zero` domain and `legacy-v1-unwrapped.eth`: their `[labelHash]` leaf label is later healed via an ENSv2 subname that shares the same leaf label (`zero.contenthash.eth`), but the materialized name never updates.

## Root cause

`ensureLabel` (`src/lib/ensv2/label-db-helpers.ts`):

```ts
const prev = await context.ensDb.find(ensIndexerSchema.label, { labelHash }); // live buffer reference
await context.ensDb.insert(...).values({ labelHash, interpreted }).onConflictDoUpdate({ interpreted }); // mutates prev.interpreted in place
if (prev && prev.interpreted !== interpreted) await cascadeLabelHeal(context, labelHash); // always false
```

Ponder's Store API returns a **live reference** to the in-memory buffered row, and `onConflictDoUpdate` mutates that same object's `interpreted` field in place. So by the time the heal check runs, `prev.interpreted === interpreted` and `cascadeLabelHeal` is **never invoked**. `cascadeLabelHeal` itself is correct — it just never gets called for these heals.

## Fix

Snapshot the heal decision **before** the mutating upsert. One-line change; `cascadeLabelHeal` is unchanged.

## Verification

Reindexed `NAMESPACE=ens-test-env PLUGINS=unigraph,protocol-acceleration` before/after:

- Exactly **2 domains** changed — `[29319238…]` → `zero`, `[86563c40…].eth` → `legacy-v1-unwrapped.eth` — both now consistent with their healed `Label` rows.
- **Zero regressions**: every other domain/label is byte-identical; counts unchanged (328 domains / 321 labels). The remaining `[hash]` domains are genuinely-unknown labels (no preimage), correct in both runs.

Adds `label-db-helpers.test.ts` — a regression test that reproduces Ponder's in-place find/upsert aliasing with a fake store and asserts the cascade fires on a heal (fails on the old code, passes on the fix).


.changeset/fix-label-heal-cascade.md
```diff
@@ -0,0 +1,5 @@
+---
+"ensindexer": patch
+---
+
+Fix the Label heal cascade silently never firing. `ensureLabel` read `prev` via `context.ensDb.find` (a live reference to Ponder's in-memory buffered row) and then evaluated the heal check `prev.interpreted !== interpreted` _after_ the `onConflictDoUpdate`, which mutates that same object in place — so the check always read the just-written value and was never true. As a result, a Domain whose label was healed _after_ the Domain's `canonicalName` had already been materialized (e.g. an ENSv1 Domain whose `[labelHash]` leaf label is later healed via an ENSv2 subname sharing that label) kept a stale `[labelHash]` `canonicalName` even though its `Label` row was correctly healed. The heal decision is now snapshotted before the upsert.
```
shrugs: Just 

```
Domains composed of Labels that are healed after discovery now have their Canonical Names correctly updated.
```

apps/ensindexer/src/lib/ensv2/label-db-helpers.test.ts
```diff
@@ -0,0 +1,76 @@
+import { type LiteralLabel, labelhashLiteralLabel, literalLabelToInterpretedLabel } from "enssdk";
```
shrugs: eh, just drop this test entirely

.changeset/fix-label-heal-cascade.md
shrugs: Done — changeset reworded to the one-liner.

apps/ensindexer/src/lib/ensv2/label-db-helpers.test.ts
shrugs: Done — test file removed entirely.

apps/ensindexer/src/lib/ensv2/label-db-helpers.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+// Mock the cascade so we can assert whether a heal triggers it.
+vi.mock("./canonicality-db-helpers", () => ({ cascadeLabelHeal: vi.fn() }));
+// Stub transitive imports that would otherwise load `@/config` (which `process.exit`s without env).
+vi.mock("@/lib/graphnode-helpers", () => ({ labelByLabelHash: vi.fn() }));
+vi.mock("@/lib/indexing-engines/ponder", () => ({ ensIndexerSchema: { label: {} } }));
+
+import { cascadeLabelHeal } from "./canonicality-db-helpers";
+import { ensureLabel } from "./label-db-helpers";
+
```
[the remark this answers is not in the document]
shrugs: Moot — the test file has been removed entirely, so the mock-path mismatch no longer applies.

shrugs:
@greptile review

PR #2346, opened 2026-07-06
Make ENSApi to wait for ENSDb to be ready

### LINKED ISSUE #2272: ENSApi: DI init crashes instead of waiting when ENSIndexer metadata isn't published yet

tk-o:
# Lite PR

## Summary

- Updated `isReady()` method in ENSDb SDK to check for relevant database schemas (ENSNode Schema, ENSDb Writer Schema) existence.
- Updated the ENSApi dependency injection container to wait for ENSDb instance to be ready before initializing caches.

---

## Why

- Fixes #2272 

---

## Testing

- Ran all static code checks and testing suites
- Ran an ENSApi instance locally and confirmed that it waits for ENSDb instance to be ready before proceeding to ENSApi caches initialization.

---

- [The idea implemented in this PR was discussed on Slack](https://namehash.slack.com/archives/C086Z6FNBHN/p1782725305079179)
- Related PR #2335

---

### DIFF
--- a/.changeset/odd-keys-like.md
+++ b/.changeset/odd-keys-like.md
@@ -1,0 +1,5 @@
+---
+"@ensnode/ensdb-sdk": minor
+---
+
+Updated `isReady()` method in `EnsDbReader` to check for relevant database schemas existence.
--- a/apps/ensapi/src/config/config.singleton.test.ts
+++ b/apps/ensapi/src/config/config.singleton.test.ts
@@ -6,10 +6,12 @@ vi.mock("@/lib/logger", () => ({
   default: {
     error: vi.fn(),
     info: vi.fn(),
+    warn: vi.fn(),
   },
   makeLogger: vi.fn(() => ({
     error: vi.fn(),
     info: vi.fn(),
+    warn: vi.fn(),
   })),
 }));
 
@@ -23,6 +25,9 @@ vi.mock("@ensnode/ensdb-sdk", async (importOriginal) => {
     ensIndexerSchema = {};
     ensIndexerSchemaName = "ensindexer_test";
     async isHealthy() {
+      return true;
+    }
+    async isReady() {
       return true;
     }
     async destroy() {}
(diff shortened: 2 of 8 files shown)


tk-o:
@greptile review

shrugs (approved):
LGTM

